PR #1, opened 2020-11-12
Add schema for the models in the indices that will be shared between Fleet Server and Kibana

blakerouse:
## Description

Defines 4 different schema's each representing the data that will be placed into different indexes. This will be the data contract between Kibana and Fleet Server.

### Mapping to Indices

* `definitions/server` -> `.fleet-servers`
  * This is 1 document per server. The same document is always updated when something is updated for the server.
* `definitions/agent` -> `.fleet-agents`
  * This is 1 document per agent. The same document is always updated when something is updated for the agent.
* `definitions/policy` -> `.fleet-policies`
  * This is an append only index. A new document will be inserted per new revision per policy.
* `definitions/checkin` -> `logs-fleet.checkin-default`
  * This is a data stream that will get checkin information about all agents, every 30 seconds.


model/agent.json
```diff
@@ -0,0 +1,86 @@
...
+    },
+    "_version": {
+      "description": "The version of the document in the index",
+      "type": "integer"
+    },
+    "agent_version": {
+      "description": "The version of the Elastic Agent",
+      "type": "string"
+    },
+    "local_metadata": {
```
scunningham, line 20: How do you model the second level fields with this structure?

model/checkin.json
```diff
@@ -0,0 +1,136 @@
...
+      "description": "Date/time the checkin was created",
+      "type": "string",
+      "format": "date-time"
+    },
+    "agent": {
+      "title": "Checkin Agent",
+      "description": "The current metadata information for the Elastic Agent",
+      "type": "object",
+      "properties": {
+        "id": {
```
ruflin, line 22: Lets use the same structure for agent in all docs. I like this one here ;-)

model/checkin.json
```diff
@@ -0,0 +1,136 @@
...
+          "format": "uuid"
+        },
+        "revision": {
+          "description": "The revision of the policy",
+          "type": "integer"
+        },
+        "inputs": {
+          "title": "Checkin Policy Input",
+          "description": "The current input status per policy",
+          "type": "array",
```
ruflin, line 87: One thing we should be careful here is if we index this or not. If we do, each array entry becomes its own document in Lucene. So if we only need the blob for retrieval that would be best. It still applies that we should have it in the schema as it is here.

model/checkin.json
```diff
@@ -0,0 +1,136 @@
...
+                "type": "string",
+                "format": "uuid"
+              },
+              "status": {
+                "description": "The current status of the input",
+                "type": "string",
+                "enum": ["healthy", "error", "degraded"]
+              },
+              "message": {
+                "description": "The current status message of the intput",
```
ruflin, line 107:
```suggestion
                "description": "The current status message of the input",
```

model/agent.json
blakerouse: @nchaulet Are these modeled or just free-form fields?

We could change to model them if they don't change.

model/agent.json
```diff
@@ -0,0 +1,86 @@
...
+  "title": "Agent",
+  "description": "An Elastic Agent that has enrolled into Fleet",
+  "type": "object",
+  "properties": {
+    "_id": {
+      "description": "The unique identifier for the Elastic Agent",
+      "type": "string",
+      "format": "uuid"
+    },
+    "_version": {
```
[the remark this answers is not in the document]
blakerouse: Yes, this is just to ensure the generated golang structure will decode that property.

model/agent.json
```diff
@@ -0,0 +1,86 @@
...
+      "description": "The version of the document in the index",
+      "type": "integer"
+    },
+    "agent_version": {
```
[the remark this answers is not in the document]
blakerouse: I have not hard feelings about it. If we go with the dot then its probably better to change to:

```
"agent": {
  "id": "id",
  "version": "version",
}
```

We just ensure that `agent.id` matches `_id` on the document. Indexing might be useful as well.

model/agent.json
```diff
@@ -0,0 +1,86 @@
...
+    "local_metadata": {
+      "description": "Local metadata information for the Elastic Agent",
+      "type": "object",
+      "additionalProperties": true
+    },
+    "access_api_key_id": {
+      "description": "ID of the API key the Elastic Agent must used to contact Fleet Server",
+      "type": "integer"
+    },
+    "default_api_key": {
```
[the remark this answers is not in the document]
blakerouse: Yes.

model/checkin.json
blakerouse: Will do.

model/checkin.json
blakerouse: At the moment, I don't know if we can keep it as a blob. When we do work on the coordinator piece this might need to be indexed.

Or with a transform that updates the latest status, then only the result of that would need to be indexed.

model/agent.json
nchaulet: These models were changing a lot during the first iterations of Fleet, but we can probably have a fixed model now.

model/checkin.json
nchaulet: We probably want to be able to search for a particular input status, but maybe not in the logs, more in a document that will be created with a transform

blakerouse:
I have updated the schema from the feedback. Every object now has `agent.*` so it's uniform. I have updated the files to be in one filed `schema.json` because of how the golang generator works. I have added the `.fleet-servers` index that each Fleet Server will write to. Also went ahead and did the generation of the schema.

I think we need to add in the actions schema to this data model, then it's probably enough to start with on the implementation. I am sure we will need to make more adjustments once we get into it.

scunningham (approved):
Looks good for a first rev.  Generator tool seems to do a reasonable job.  I do wonder how performant the generated code is.  On inspection seems ok.

ruflin (approved):
LGTM. I assume we will have a few more iterations on it when we put actual data into it.

nchaulet (approved):
Looks good 👍

PR #2, opened 2020-11-16
Add initial configuration for fleet-server

blakerouse:
## Overview

This adds the initial configuration for the Fleet Server.

Adds 3 sections to the configuration:

- `elasticsearch`: configuration for communication with Elasticsearch (requires username/password, API key errors)
- `server`: configuration for the HTTP server that fleet-server runs for the Elastic Agents to connect to (missing server-side TLS options)
- `logging` configuration for the logging

At the moment only `logging` is actually used, the others need to be plumbed in.

Adds `-c` and `--config` to specify the configuration of Fleet Server as well as `--help` using `cobra`.

Adds default configuration that pulls settings from ENV so running it in container can be configured. But this really should not be needed as it should only really be executable from Elastic Agent, where it passes those settings down.


scunningham (approved):
👍   Thanks for keeping the logger in place.

internal/pkg/config/server.go
```diff
@@ -0,0 +1,43 @@
...
+// InitDefaults initializes the defaults for the configuration.
+func (c *ServerTimeouts) InitDefaults() {
+	c.Read = 5        // 5 seconds
+	c.Write = 60 * 10 // 10 minutes (long poll)
+}
+
+// Server is the configuration for the server
+type Server struct {
+	Host     string         `config:"host"`
+	Port     uint16         `config:"host"`
```
ruflin, line 35: `config:"host"` should be `config:"port"`

internal/pkg/config/server.go
```diff
@@ -0,0 +1,43 @@
...
+// software distributed under the License is distributed on an
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+package config
+
+// ServerTimeouts is the configuration for the server timeouts
+type ServerTimeouts struct {
```
ruflin, line 21: Any chance to add these also to the fleet-server.yml file so we have all the configs in a single place "documented"?

internal/pkg/config/logging.go
```diff
@@ -0,0 +1,104 @@
...
+	"fmt"
+	"io"
+	"os"
+	"strings"
+
+	"github.com/rs/zerolog"
+)
+
+// Logging is the logging configuration
+type Logging struct {
```
ruflin, line 30: Same for these config options, please add them to the fleet-server config file.

PR #5, opened 2020-11-18
Fix docker image build

aleksmaus:
* Put back the build/ directory that was lost during the repo move.
* Adjusted the Dockerfile to fix the image for the config file and changes project structure (main.go) in ./
* Updated RPM spec and systemd file to include fleet-server.yml

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,6 +2,4 @@
 
 bin/
 *.rpm
-build/
 fleet-server
-fleet_server
--- a/build/package/Dockerfile
+++ b/build/package/Dockerfile
@@ -1,0 +1,101 @@
+# syntax=docker/dockerfile:experimental
+
+############################
+# STEP 0 build environment
+############################
+# golang:1.15.2-alpine3.12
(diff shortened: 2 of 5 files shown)


scunningham (approved):

ph (approved):
LGTM, sorry for removing this, I think this is OK for dev, but we will need more work to fit into the existing Elastic distribution. 

@michalpristas known a bit about how it works in Beats and it might be able to help.

ruflin:
The way we use the `build` directory in most other projects is that it only contains generated files and can be thrown away at any time. For the files we have here we sometimes use `dev-tools` / `deploy`. No need to align right now but I think long term it will be confusing when jumping between repos and not be consistent.

aleksmaus:
yeah, can move however we need, just let me know when

ruflin:
@ph @aleksmaus Any preference on a name which is not `build` for these files?

ph:
I will mimic beats' I think _dev-tools_ was an appropriate choice.

ruflin:
Lets go with `dev-tools` for now if there are no objections. @aleksmaus Can you follow up?

aleksmaus:
yep, will move to dev-tools

PR #7, opened 2020-11-19
Add to the schema for coordinator

blakerouse:
Adds to the `schema.json` for the policy coordinator.

Adds `coordinator_idx` to `.fleet-policies` document. Renames `revision` to `revision_idx` so it matches the coordinator index, and make it clear it's an index.

Add `policy-leader` that will be in `.fleet-policies-leader` index to mark which Fleet Server is the current leader of a policy.

### DIFF
--- a/model/schema.json
+++ b/model/schema.json
@@ -21,7 +21,7 @@
         "version"
       ]
     },
-    "host": {
+    "host-metadata": {
       "title": "Host Metadata",
       "description": "The host metadata for the Elastic Agent",
       "type": "object",
@@ -87,8 +87,8 @@
           "type": "integer"
         },
         "agent": { "$ref": "#/definitions/agent-metadata" },
-        "host": { "$ref": "#/definitions/host" },
-        "server": { "ref":  "#definitions/server" }
+        "host": { "$ref": "#/definitions/host-metadata" },
+        "server": { "ref":  "#/definitions/server-metadata" }
       },
       "required": [
         "_id",
@@ -163,7 +163,7 @@
           "type": "string",
           "format": "date-time"
         },
-        "server": { "ref":  "#definitions/server" }
+        "server": { "ref":  "#/definitions/server-metadata" }
       },
       "required": [
         "_id",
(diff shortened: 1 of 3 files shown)


model/schema.json
```diff
@@ -21,7 +21,7 @@
         "version"
       ]
     },
-    "host": {
+    "host-metadata": {
```
ruflin, line 24: Any reason you renamed this? Before it matched ECS so nicely :-)

model/schema.json
```diff
@@ -52,6 +52,26 @@
         "name"
       ]
     },
+    "server-metadata": {
```
ruflin, line 55: Lets use `_` to be consistent with other fields in Beats / ECS.

ruflin: Why not pack this into server.metadata?

model/schema.json
```diff
@@ -62,9 +82,21 @@
...
         },
+        "_version": {
+          "description": "The version of the document in the index",
+          "type": "integer"
+        },
         "agent": { "$ref": "#/definitions/agent-metadata" },
-        "host": { "$ref": "#/definitions/host" }
-      }
+        "host": { "$ref": "#/definitions/host-metadata" },
+        "server": { "ref":  "#/definitions/server-metadata" }
```
ruflin, line 91: Can the server metadata be used in two places? I think I need to read up on how this works. What is definition and hwat is actual usuage?

model/schema.json
blakerouse: Think of this name as just a variable. I updated it so it matched `agent-metadata`, and `server-metadata`. But these are just variables that are referenced by other objects in the schema.

So `host-metadata` is used like:

```
"host": { "$ref": "#/definitions/host-metadata" }
```

So the resulting document in ES will be:

```
host: {
  ...fields from host-metadata...
}
```

model/schema.json
blakerouse: This is the same as my comment above. This is just a reference.

model/schema.json
blakerouse: The reason I named them `*-metadata` is because all the metadata objects are shared references from multiple objects in the definition.

model/schema.json
ruflin: Got it.

ruflin (approved):
LGTM. I think as soon as we have some example documents we should have a look again an iterate on it.

PR #12, opened 2020-11-30
Agent actions

aleksmaus:
The first working POC for the new agent actions with the ack token exchange on check-in.
The corresponding agent changes are in another draft:
https://github.com/elastic/beats/pull/22782

What works:
0. The Initial indices/datastreams bootrapping is done in the fleet-server for now for development purposes
1. New actions are detected and forwarded to the agent
2. The ack token exchange between agent and the fleet is implemented as specified in https://docs.google.com/document/d/15sF6IZdH5KfbJmkDP97qIpmbNOUBWUGbBOEFS1ZigvQ/edit?ts=5fb69468

What doesn't work:
1. The action is not forwarded to the beat on the agent side (see the agent PR with the empty handler).
2. The results and the data are not posted back to ES yet.

For development and compatibility purposes the old saved objects based storage is working side-by-side with the new .fleet-* indices. So for now the we have double number of elasticsearch requests, until we fully migrate to the new schema (including kibana) 

There are still few TODO items left in the code, some refactoring/clean up and utz would be nice to have. 

Just wanted to get the draft PR out, so you can estimate the areas of changes involved.
Depending on the external expose and rick tolerance of fleet-server project we could merge these changes to avoid blocking anybody (big PRs could be difficult to review too) and continue iterate on this. You might need some of these changes in the work you are planning to do.

Meanwhile I'll continue working in this area and adding more code to this PR.


aleksmaus:
Did quick and dirty modifications to the current generator that we use to generate Go structs from schema to make it generate elasticsearch mapping in my fork: https://github.com/aleksmaus/generate
Bootstrapped all the indices currently defined in the schema on fleet start for now @nchaulet in case if you need it.

While hacking the generator for ES mapping found few typos in the schema ("ref" instead of "$ref") and fixed in the last commit 
 @blakerouse

aleksmaus:
Screen Shot 2020-11-30 at 6 56 30 PM

aleksmaus:
Screen Shot 2020-11-30 at 6 56 30 PM
Screen Shot 2020-11-30 at 7 08 52 PM

internal/pkg/esboot/ilm.go
```diff
@@ -0,0 +1,226 @@
...
+	{
+		"policy":{
+		   "phases":{
+		   }
+		}
+	}`
+
+	ilmRolloverSize = 300
+	ilmRolloverAge  = 30
+	ilmDeleteAge    = 90
```
ruflin, line 30: Lets not delete data by default, at least not now. I think this could bite us later if we forget to remove it.

ruflin (commented):
Great progress. Good to see we already have most of the code for setup in place. I hope we can get rid of it long term but this will get us moving now.

Lets also figure out good ways on how to test all this code before it gets merged. I guess some of the tests will have to involve Agent and Elasticsearch.

cmd/fleet/actionDispatcher.go
```diff
@@ -0,0 +1,120 @@
+package fleet
+
+import (
+	"context"
+	"fleet/internal/pkg/action"
+	"fleet/internal/pkg/esutil"
+	"fleet/internal/pkg/seqno"
```
blakerouse, line 7: Should move this down to be like:

```
import (
    "context"
    "sync"

    ..fleet imports..

    ..external imports..
)

File: cmd/fleet/actionDispatcher.go [hunk 3]
```diff
@@ -0,0 +1,120 @@
...
+		case <-ctx.Done():
+			return
+		case hits := <-ad.am.Output():
+			ad.process(ctx, hits)
+		}
+	}
+}
+
+func (ad *ActionDispatcher) Subscribe(agentId string, seqNo int64) *ActionSubX {
+	cbCh := make(chan []ActionX, 1)
```
blakerouse, line 51: Being this is a buffered channel of 1, if the Agent doesn't quickly handle the action will this block the sending go routine affecting other Agent's?

cmd/fleet/actionDispatcher.go
```diff
@@ -0,0 +1,120 @@
...
+	agentActions := make(map[string][]ActionX)
+	for _, action := range actions {
+		for _, agentId := range action.Agents {
+			arr := agentActions[agentId]
+			arr = append(arr, action)
+			agentActions[agentId] = arr
+		}
+	}
+
+	// TODO: revisit actions rollout
```
blakerouse, line 97: Whats the TODO here? Would be good to comment on the improvement you think should be made here.

cmd/fleet/actionDispatcher.go
```diff
@@ -0,0 +1,120 @@
...
+	ad.mx.RLock()
+	sub, ok := ad.subs[agentId]
+	ad.mx.RUnlock()
+	return sub, ok
+}
+
+func (ad *ActionDispatcher) dispatch(ctx context.Context, agentId string, actions []ActionX) {
+	sub, ok := ad.getSub(agentId)
+	if !ok {
+		log.Info().Str("agent_id", agentId).Msg("Agent is not currently connected. Not dispatching actions.")
```
blakerouse, line 113: Should this be an info log message? When running multiple Fleet Servers the Agent will only ever be connected to 1.

cmd/fleet/actionDispatcher.go
```diff
@@ -0,0 +1,120 @@
...
+		return
+	}
+	select {
+	case sub.Ch() <- actions:
+	case <-ctx.Done():
```
blakerouse, line 118: After the context is closed it will call `ad.getSub` again, that seems wrong. Shouldn't it return and not do any work?

cmd/fleet/main.go
```diff
@@ -19,10 +19,18 @@ package fleet
...
+	"fleet/internal/pkg/action"
+	"fleet/internal/pkg/agent"
+	"fleet/internal/pkg/bulk"
 	"fleet/internal/pkg/config"
-	"github.com/spf13/cobra"
+	"fleet/internal/pkg/esboot"
+	"fleet/internal/pkg/seqno"
 	"time"
 
+	"github.com/elastic/go-elasticsearch/v8"
```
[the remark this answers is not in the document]
aleksmaus: It depends. If we are not hitting exclusively ES 8.x APIs it should work.
The bootstrapping currently fails on 7.10 with:
400 - illegal_argument_exception: data_stream [.fleet-actions-results] must not start with '.'

...and of course the datastreams would not work on 7.6 for example.

Currently following the main branch of ES.
Can adjust later as needed later.

cmd/fleet/auth.go
```diff
@@ -73,27 +75,44 @@ func authApiKey(r *http.Request, client *elasticsearch.Client) (*apikey.ApiKey,
...
 	}
 
-	return agent, nil
+	if af != nil {
+		seqno, err := af.FetchAgentSeqNo(r.Context(), id)
+
+		agnt.ActionSeqNo = -1
+		if err != nil {
+			if err == agent.ErrNotFound {
+				log.Debug().Str("agent_id", id).Msg("Agent action sequence not found")
```
blakerouse, line 107: Would be good to add `action_seq_no` to the log message even if its `-1`.

aleksmaus:
@ruflin that the plan for this and probably the next week. There are still some things to do, some things to make cleaner and move around and get some tests in place.

cmd/fleet/handleCheckin.go
```diff
@@ -122,37 +169,54 @@ func (ct *CheckinT) _handleCheckin(w http.ResponseWriter, r *http.Request, id st
...
-			break LOOP
-		case <-longPoll.C:
-			log.Trace().Msg("Fire long poll")
-			break LOOP
-		case <-tick.C:
-			ct.bc.CheckIn(agent.Id, nil)
 		}
 	}
 
 	// For now, empty response
```
blakerouse, line 215: Is `empty response` correct? Seems JSON is being returned.

blakerouse (commented):
Looks really good. Some comments.

cmd/fleet/actionDispatcher.go
aleksmaus: it will not, it will exit the func. 
this construct is useful as a breaker on context cancel in case if you are blocked on sent to the full channel

cmd/fleet/actionDispatcher.go
aleksmaus: it should not be. will be doing few more passes to refactor and cleanup

cmd/fleet/actionDispatcher.go
aleksmaus: yes, there is a slim chance that sending to this channel can block because the agent long poll didn't exit and unsubscribed from the action dispatcher before the new action came in. 

That's on my TODO item below.
```
// TODO: revisit actions rollout
``` 

was focusing on getting this work end-to-end first, but actions rollout details are not fully fleshed out yet

aleksmaus: the easiest for now send with select default (drop the actions) for now so even if the channel is full it will not block.
and if the channel is full we should not dispatch anything anyways cause the agent has response at this point.

cmd/fleet/main.go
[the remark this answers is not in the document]
ruflin: 👍 

Side note: We only need to support ES >=7.11

aleksmaus:
Closing this PR draft, since rewriting lots of it right now. Will nave a new PR today or tomorrow.

PR #14, opened 2020-12-01
More work on the configuration and wire in ES connection and server binding

blakerouse:
## What does this PR do?

Adds more configuration to the fleet-server. Uses `inputs` for configuring the fleet-server as it will be done passed through the Elastic Agent. Connects the configuration to starting the elasticsearch connection and the fleet-server HTTP endpoint.

## Why is it important?

So the configuration can be done through the configuration file and the format matches what Elastic Agent will send it with very little modification.

## Checklist


cmd/fleet/handleCheckin.go
```diff
@@ -38,8 +38,8 @@ var (
 	ErrAgentNotFound  = errors.New("agent not found")
 	ErrAgentCorrupted = errors.New("agent record corrupted")
 
-	kCheckinTimeout  = env.CheckinTimeout(30)   // 30s
-	kLongPollTimeout = env.LongPollTimeout(300) // 5m
+	kCheckinTimeout  = 30 * time.Second
+	kLongPollTimeout = 300 * time.Second // 5m
```
aleksmaus, line 42: hardcoded now?

cmd/fleet/main.go
```diff
@@ -34,6 +34,8 @@ import (
 	"fleet/internal/pkg/signal"
 )
 
+const kPolicyThrottle = time.Millisecond * 5
```
aleksmaus, line 37: assuming this will be configurable?

go.mod
```diff
@@ -9,10 +9,12 @@ require (
 	github.com/elastic/go-elasticsearch/v8 v8.0.0-20200728144331-527225d8e836
 	github.com/elastic/go-ucfg v0.8.3
 	github.com/gofrs/uuid v3.3.0+incompatible
+	github.com/google/go-cmp v0.4.0
```
aleksmaus, line 12: 👍

internal/pkg/config/config_test.go
```diff
@@ -0,0 +1,222 @@
...
+				if err == nil {
+					t.Error("no error was reported")
+				} else {
+					cfgErr := err.(ucfg.Error)
+					require.Equal(t, test.err, cfgErr.Reason().Error())
+				}
+			} else {
+				require.NoError(t, err)
+				if !assert.True(t, cmp.Equal(test.cfg, cfg)) {
+					diff := cmp.Diff(test.cfg, cfg)
```
aleksmaus, line 214: 👍

internal/pkg/config/input.go
```diff
@@ -0,0 +1,96 @@
...
+	RateLimitBurst    int            `config:"rate_limit_burst"`
+	RateLimitInterval time.Duration  `config:"rate_limit_interval"`
+	MaxEnrollPending  int64          `config:"max_enroll_pending"`
+	Profile           ServerProfile  `config:"profile"`
+}
+
+// InitDefaults initializes the defaults for the configuration.
+func (c *Server) InitDefaults() {
+	c.Host = "0.0.0.0"
+	c.Port = 8000
```
aleksmaus, line 69: different defaults?  above is: localhost:6000

internal/pkg/config/output.go
```diff
@@ -0,0 +1,199 @@
...
+	for i, host := range c.Hosts {
+		addr, err := makeURL(c.Protocol, c.Path, host, 9200)
+		if err != nil {
+			return elasticsearch.Config{}, err
+		}
+		addrs[i] = addr
+	}
+
+	// build the transport from the config
+	httpTransport := http.DefaultTransport.(*http.Transport)
```
aleksmaus, line 101: nit: better to create the transport rather than cloning from the globally accessible variable.
some reasoning here:
https://github.com/hashicorp/go-cleanhttp

aleksmaus (approved):
few nits, otherwise LGTM

cmd/fleet/handleCheckin.go
blakerouse: Being they are not configurable in the Elastic Agent, I just don't see the need to make it configurable. I think its better to not give so many options that can cause issues if a user configures it incorrectly.

cmd/fleet/main.go
blakerouse: Whats the benefit of it being configurable?

internal/pkg/config/input.go
blakerouse: The one above is for the profiler, not for the HTTP endpoint.

internal/pkg/config/output.go
blakerouse: changed it to create a new transport instead of cloning the default

cmd/fleet/main.go
aleksmaus: 👍

aleksmaus (approved):
thanks for replies and the change 👍

PR #15, opened 2020-12-01
Hook up ES mapping generator

aleksmaus:
## What does this PR do?

Hooked up the modified version of generator, that allows to regenerate the ES mapping from schema as well.

This is an optional PR, but is useful for development, while we changing the schema.
Let me know.


blakerouse (approved):
Looks good.

PR #20, opened 2020-12-03
Fix generator directives, that were lost on previous copyright header bulk update

aleksmaus:
## What does this PR do?

Fix generator directives, that were lost on previous copyright header bulk update

@blakerouse we probably need to tweak whatever was used to update the copyright


blakerouse:
This is the tool that is used by `make check` and the same tool used to update the copyright headers.

https://github.com/elastic/go-licenser

Surprised that it removed those lines, being it is used by lots of projects at Elastic. I would guess that now that you have added them back it will not be an issue.

Seems that `make check` on Jenkins is having issue with running the generation.

blakerouse (approved):
Looks good.

aleksmaus:
Found couple of more deficiencies of generator and updated.
1. Generate int64 fields for integer fields in the schema
2. If there are two object fields like "Data" in our case, it randomly creates the top level struct that could result in overwriting the Data struct with different comments on different generator runs and failure on make check.

PR #21, opened 2020-12-04
Agent actions support

aleksmaus:
## What does this PR do?

Implements the fleet agent actions feature according to the initial proposal.

* The feature is disabled by default, is behind the configurable feature flag. It still creates the indices and everything, but doesn't listen or check for actions. 
So if you run the current agent against the server it will work as is.

* The agent record is created in the .fleet-agents index as well upon enrollment
* The agent record in the .fleet-agents index is utilized to store the last agent action sequence number returned by the Agent upon checkin request
* Rewrote the POC code for data access to use dsl and bulker where possible. All indexes access is encapsulated in the dl package now.
* Leveraging the schema generated structures for data layer.
* The action monitor tracks the index global checkpoint changes and uses that value to cap the actions query in order to avoid missing not fully indexed actions.
* The action token is derived from to the action document id. The token resolver utilizes the LRU cache for resolving tokens to the _seq_no.
* Extended dsl package to support the new queries.
* Fixed few schema fields types.
* Extended bulk result to support aggregation queries.

1. Enable "actions" feature in the config
2. Run the server pointing to you local ES instance
```
ELASTICSEARCH_USERNAME=elastic ELASTICSEARCH_PASSWORD=changeme ELASTICSEARCH_HOSTS=localhost:9200 LOG_LEVEL=DEBUG ./bin/fleet
```
3. The agent from my PR branch https://github.com/elastic/beats/pull/22782 should follow the protocol and receive the action. It doesn't forward the action to the osquerybeat yet.

Example of creating the action for the agent:
```
POST .fleet-actions/_doc/
{
  "id": "bc4cc421-32c5-494c-9868-1b2249cc8eac",
  "@timestamp": "2020-11-20T00:36:30Z",
  "expiration": "2020-12-25T00:36:30Z",
  "type" : "APP_ACTION",
  "input_id": "osquery",
  "agents": [
    "b5be2866-542f-45df-8e30-82b7a3689094"
  ],
  "data" : {
    "commands" : [
      {
        "id": "123-456",
        "type" : "query",
        "sql" : "select * from users"
      }
    ]    
  }
}
```

The action property "data" is opaque and the payload format is not finalized for osquery specifically.

P.S. Going forward will try to piecemeal the changes, to make the changes smaller and easier to review and avoid blocking anybody working on the same code.
This one was split in two PR, and the first part the bootstrapping landed before.


fleet-server.yml
```diff
@@ -3,3 +3,12 @@ output:
...
     password: '${ELASTICSEARCH_PASSWORD:}'
+
+fleet:
+  agent:
+    logging:
+      level: '${LOG_LEVEL:INFO}'
+
+features:
+  actions:
+    enabled: false
```
blakerouse, line 14: We should work to remove this quickly. Good for now.

blakerouse (approved):
This looks really good. Lets get it landed now, but none of this has any testing. We need to go back and add unit tests for the areas that need it.

fleet-server.yml
aleksmaus: I added this in due to the questions raised before on what happens if the existing agent connects to the fleet.
Also the scale testing "horde" harness needs to be changed to work correctly against these changes.
Can remove once the "horde" is ready.

aleksmaus:
Just didn't want to block anybody while working on this piece any longer.
Most of the testing for this should be done against some ES instance, cause most of the pieces for actions are tied to ES pretty closely. It's on my to do list.

scunningham (approved):
:kudos: for figuring out the dsl. you'll have to remind me how that works.

PR #22, opened 2020-12-05
Set up integration tests

aleksmaus:
## What does this PR do?

* Set up integration tests environment. Add an example integration test that uses elasticsearch. 
* Use ```make test``` run the unit and integration tests.
* Use ```make test-unit``` to run the unit tests only.
* Use ```make test-int``` to run the integration tests only. This is slow because it starts up and shuts down elasticsearch. Prefer ```make test-int-set``` for development that relies on you starting and stopping the ES, see the notes below.
* For development of integration tests it's more convenient to start the integration docker environment first, because starting elasticsearch takes time.
```
make int-docker-start
```

Run integration tests while the environment is up and running
```
make test-int-set
```

Stop the environment after you are done.
```
make int-docker-start
```

* Use build tag ```+build integration``` to mark integration tests
* Use build tag ```+build !integration``` to exclude unit tests from integration tests, otherwise it will run the unit tests again (due to Go build tag implementation) when integration tests are ran.

* Moved ES initialization to it's own package so it can be accessed for the integration tests.
* Added ```rnd``` utility for random values, useful for tests


aleksmaus:
@ruflin @blakerouse Not sure what needs to happen with CI to get the docker compose working. 
The ```make test-int``` target works otherwise, tested on Mac and Ubuntu.
Commented that from running with "test" target for now.

ruflin:
@elastic/observablt-robots Can you help out here with getting the docker-compose tests working?

Makefile
```diff
@@ -75,5 +75,50 @@ check-no-changes:
...
+
+# Start ES with docker without waiting
+.PHONY: int-docker-start-async
+int-docker-start-async:
+	@docker-compose -f ./dev-tools/integration/docker-compose.yml --env-file ./dev-tools/integration/.env up  -d --remove-orphans elasticsearch
+
+# Wait for ES to be ready
+.PHONY: int-docker-wait
+int-docker-wait:
+	@./dev-tools/integration/wait-for-elasticsearch.sh ${ELASTICSEARCH_USERNAME}:${ELASTICSEARCH_PASSWORD}@localhost:9200
```
ruflin, line 100: @mtojek It crossed my mind to potentially use elastic-package here to not require our own yaml compose files etc. Would this make sense?

mtojek: I didn't analyze this use case deeper, but it seems that you're using additional options (ulimits, ES bootstrap, ES discovery). Not sure if you need them or they've been copied from another source. Just in case, these are definitions for elastic-package:  https://github.com/elastic/elastic-package/tree/master/internal/install

cachedout:
@ruflin A PR to fix this issue [was submitted here](https://github.com/elastic/infra/pull/25264). Once it is merged, it will be safe to uncomment the line regarding the Docker integration tests. Let me know if you need anything else.

Makefile
aleksmaus: I copied the settings from my compose files that I used for other projects, some for scale testing, nothing specifically mandatory for our use case, just need the snapshot instance of elasticsearch for integration tests.
Works for now.
Can switch once the PR is merged in the following this PRs or let you adjust how it suppose to be done.

blakerouse (approved):
Looks good.

Great to see tests!

cachedout:
@aleksmaus and @ruflin The[ relevant PR to unblock the Docker issue has now been merged](https://github.com/elastic/infra/pull/25264). 

I suggest waiting about a day before submitting a PR to re-enable the Docker tests, however, as there are automated processes which rebuild the images periodically. Please contact me or @elastic/observablt-robots if you're having trouble after that point. Thanks!

Makefile
ruflin: Definitively, no urgency to switch over. Was just thinking about simplifying / unifying our test environments long term.

PR #23, opened 2020-12-07
Add ability to reload configuration

blakerouse:
## What does this PR do?

Adds the ability in the future to call `fleetServer.Reload`, which will reload the server based on the updated configuration.

The usage of the actual reload is not used, because I do not want to currently add the requirement of needed Elastic Agent to run Fleet Server.

## Why is it important?

Needed so once the `elastic-agent-client` is added to the project it is very easy to connect configuration reloading.

## Checklist


cmd/fleet/main.go
```diff
@@ -65,47 +65,54 @@ func runBulkActions(ctx context.Context, sv saved.CRUD) *BulkActions {
...
 	go func() {
-		err := ad.Run(ctx)
-		checkErr(err)
+		for {
+			err := ad.Run(ctx)
+			if err == context.Canceled {
+				break
+			}
+			log.Error().Err(err).Msg("Restart action dispatcher on error")
+		}
```
aleksmaus, line 111: This could cause tight loops on errors on Run of server different subcomponents, difficult to manage state when some of the components started and some are in the tight running loops trying to start.

maybe should consider the fleet server as one restartable unit and if any of its components fail then exit the server and have some kind of restart with backoff logic for the server restart.
I can address in the next PR, did something similar for Endgame event streaming before.

cmd/fleet/main.go
```diff
@@ -184,3 +160,142 @@ func NewCommand(version string) *cobra.Command {
...
+
+// Run runs the fleet server.
+func (f *fleetServer) Run(ctx context.Context) error {
+	// Initial indices bootstrapping, needed for agents actions development
+	// TODO: remove this after the indices bootstrapping logic implemented in ES plugin
+	err := esboot.EnsureESIndices(ctx, f.es)
+	if err != nil {
+		return err
+	}
+
```
aleksmaus, line 198: maybe pefer
```
defer profilerCancel() 
```

cmd/fleet/main.go
```diff
@@ -184,3 +160,142 @@ func NewCommand(version string) *cobra.Command {
...
+	errCh := make(chan error)
+	serverCancel, err := f.runServer(ctx, errCh)
+	if err != nil {
+		profilerCancel()
+		serverCancel()
+		return err
+	}
+	f.lock.Lock()
+	f.profilerCancel = profilerCancel
+	f.serverCancel = serverCancel
```
aleksmaus, line 217: maybe could use one cancellable context for both server and profiler?

aleksmaus (approved):

cmd/fleet/main.go
blakerouse: no don't want to defer that as, it can be replaced with a different cancellable on reload

cmd/fleet/main.go
blakerouse: then it would restart both, when either changed

you might want to turn on profiling while the server is already running, but if you restart the server all the current status of the server will be lost and the information you are trying to get by turning on the profiler might be lost

cmd/fleet/main.go
blakerouse: i think if we handle the restart as a single unit, that would be good

PR #24, opened 2020-12-08
Refactor restart

aleksmaus:
## What does this PR do?

* Refactored restart cancellation and some repetitive code.
* Moved the cancellables from the server struct scope into the "Runnable" function scope. I think this way the cancellation handling is cleaner and doesn't require locking, synchronization.
* Trimmed the server Reload function.

@blakerouse let me know what you think

### DIFF
--- a/cmd/fleet/main.go
+++ b/cmd/fleet/main.go
@@ -10,13 +10,13 @@ import (
 	"time"
 
 	"fleet/internal/pkg/action"
-	"fleet/internal/pkg/bulk"
 	"fleet/internal/pkg/config"
 	"fleet/internal/pkg/env"
 	"fleet/internal/pkg/es"
 	"fleet/internal/pkg/esboot"
 	"fleet/internal/pkg/logger"
 	"fleet/internal/pkg/profile"
+	"fleet/internal/pkg/runner"
 	"fleet/internal/pkg/saved"
 	"fleet/internal/pkg/signal"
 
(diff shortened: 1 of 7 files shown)


blakerouse (approved):
Really like it, looks great.

Just one comment.

cmd/fleet/main.go
```diff
@@ -161,141 +83,160 @@ func NewCommand(version string) *cobra.Command {
...
-		return err
+	var (
+		proCancel, srvCancel context.CancelFunc
+		proWg, srvWg         sync.WaitGroup
+	)
+
+	for {
+		ech := make(chan error, 2)
+
+		if curCfg == nil || curCfg.Inputs[0].Server.Profile.Bind != newCfg.Inputs[0].Server.Profile.Bind {
```
blakerouse, line 133: When will `curCfg` be `nil`? Seems that it should never be nil.

aleksmaus: it's on initial pass when the server is started for the first time.
the newCfg is assigned to f.cfg initially (line 102) and then we get the updates from the Reload channel (line 149)
otherwise following the same logic comparing the config params as it was before in the Reload func,
but doing start/restart in the scope of the func.

blakerouse: Still confused on when it will be `nil` then. Look at it, if `curCfg = f.cfg` on initial `Run` then its always set.

aleksmaus:
```
newCfg := f.cfg
```
curCfg in my interpretation here is the "currently applied/running config" (only after it applied)

maybe can reverse the naming if confusing, let me know

blakerouse: Ah, I see. I was mixing up `curCfg` with `newCfg`. Looks good.

blakerouse (approved):

PR #26, opened 2020-12-08
Add fleet X-elastic-product-origin header

aleksmaus:
## What does this PR do?

Adds X-elastic-product-origin header to all the requests to elasticsearch.
This eliminates the Warning header from Elasticsearch responses while querying against .fleet-* indices, 
where otherwise you would get:
```
299 Elasticsearch-8.0.0-SNAPSHOT-5859c4d6e62606428c2e49a181c7f845cadabecb "this request accesses system indices: [.fleet-actions], but in a future major version, direct access to system indices will be prevented by default"
```

### DIFF
--- a/internal/pkg/config/output.go
+++ b/internal/pkg/config/output.go
@@ -48,6 +48,12 @@ func (c *Elasticsearch) InitDefaults() {
 	c.MaxRetries = 3
 	c.MaxConnPerHost = 128
 	c.BulkFlushInterval = 250 * time.Millisecond
+
+	// Set special header "X-elastic-product-origin" for .fleet-* indices based on the latest conversation with ES team
+	// This eliminates the warning while accessing the system index
+	c.Headers = map[string]string{
+		"X-elastic-product-origin": "fleet",
+	}
 }
 
 // Validate ensures that the configuration is valid.
--- a/internal/pkg/config/output.go
+++ b/internal/pkg/config/output.go
@@ -122,7 +122,7 @@ func (c *Elasticsearch) ToESConfig() (elasticsearch.Config, error) {
 
 	// Set special header "X-elastic-product-origin" for .fleet-* indices based on the latest conversation with ES team
 	// This eliminates the warning while accessing the system index
-	c.Headers["X-elastic-product-origin"] = "fleet"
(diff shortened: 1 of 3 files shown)


internal/pkg/config/output.go
```diff
@@ -48,6 +48,12 @@ func (c *Elasticsearch) InitDefaults() {
 	c.MaxRetries = 3
 	c.MaxConnPerHost = 128
 	c.BulkFlushInterval = 250 * time.Millisecond
+
+	// Set special header "X-elastic-product-origin" for .fleet-* indices based on the latest conversation with ES team
+	// This eliminates the warning while accessing the system index
+	c.Headers = map[string]string{
+		"X-elastic-product-origin": "fleet",
+	}
```
blakerouse, line 56: I think this is the incorrect to place to set it. It should not be in the configuration at the `config.Elasticsearch` level. I think it should just always be injected in the `ToESConfig` function. That function can ensure to add it to the `http.Headers`.

blakerouse (changes requested):
See inline comment, think we should switch this around a little.

internal/pkg/config/output.go
aleksmaus: sounds good, will change

internal/pkg/config/output.go
```diff
@@ -120,6 +120,12 @@ func (c *Elasticsearch) ToESConfig() (elasticsearch.Config, error) {
 		h.Set(key, val)
 	}
 
+	// Set special header "X-elastic-product-origin" for .fleet-* indices based on the latest conversation with ES team
+	// This eliminates the warning while accessing the system index
+	c.Headers = map[string]string{
+		"X-elastic-product-origin": "fleet",
+	}
```
blakerouse, line 127: You need to change this to:

```
h.Set("X-elastic-product-origin", "fleet")
```

Modifying the `c.Headers` is not something `ToESConfig` should do.

blakerouse (changes requested):

internal/pkg/config/output.go
aleksmaus: 🤦

blakerouse (approved):
Looks good.

PR #27, opened 2020-12-08
Add enrollment API key definition into the schema

aleksmaus:
## What does this PR do?
* Add enrollment API key definition into the schema
* Add migration code for development that migrates the keys from .kibana
  index into .fleet-enrollment-api-keys index

This PR is based on my previous PR that is still pending review, so this one contains the commit from that PR. So only commits after the first one are directly related to this change.


blakerouse (commented):
Is this fleet server performing the migration? Is this just for now?

It will be Kibana doing the migration, correct?

aleksmaus:
Migrating just for now just to be able to have the data in the new index and test the actual enrollment. 
I left the comment in the code mentioning that's this is for development only, helpful for us and kibana devs to test this end-to-end. 
Once kibana starts handling the data migration we should remove it.

blakerouse (approved):
Perfect!

PR #29, opened 2020-12-09
Generic index seq_no monitor

aleksmaus:
## What does this PR do?

Generic index seq_no monitor.
Can be reused for the policy work @blakerouse .

To run:
```
mon, err = monitor.New("myindex", es)
....
err = mon.Run()

```

I will make some internal modifications later to optimize how the monitoring is done.


internal/pkg/es/error.go
```diff
@@ -0,0 +1,65 @@
...
+// you may not use this file except in compliance with the Elastic License.
+
+package es
+
+import (
+	"errors"
+	"fmt"
+)
+
+type ErrElastic struct {
```
aleksmaus, line 12: this is moved out from bulk package, since it can be shared for both bulk and non-bulk queries

internal/pkg/es/result.go
```diff
@@ -0,0 +1,65 @@
+// Copyright Elasticsearch B.V. and/or licensed to Elasticsearch B.V. under one
+// or more contributor license agreements. Licensed under the Elastic License;
+// you may not use this file except in compliance with the Elastic License.
+
+package es
+
+import "encoding/json"
+
+// Error
+type ErrorT struct {
```
aleksmaus, line 10: this is moved out from bulk package, since it can be shared for both bulk and non-bulk queries

blakerouse (approved):
Looks like a rather mechanical change, looks good.

aleksmaus:
Yeah, pretty much just moved some things that can be reused around and cleaned up the monitor to make it more generic.

PR #30, opened 2020-12-10
Add policy leader election with coordinator v0

blakerouse:
## What does this PR do?

Adds the `policy.Manager` and `policy.Coordinator` that handles leader election for running coordinators per policy.

Each Fleet Server updates the `.fleet-policies-leader` index every 20 seconds when it is the leader of a policy. When the Fleet Server is the leader it runs the coordinator for that index.

The `v0` coordinator just takes a new policy and sets the `coordinator_idx: 1` and inserts the new policy revision in the `.fleet-policies`.

## Why is it important?

This allows the coordinator that will run in Fleet Server to have all the required plumbing to add on to it. At the moment the `v0` coordinator does nothing with the policy data, but as well add that logic none of this plumbing changes.

## Checklist

- [X] I have made corresponding change to the default configuration files


internal/pkg/coordinator/monitor.go
```diff
@@ -0,0 +1,360 @@
...
+	"fleet/internal/pkg/bulk"
+	"fleet/internal/pkg/config"
+	"fleet/internal/pkg/dl"
+	"fleet/internal/pkg/es"
+	"fleet/internal/pkg/model"
+	"fleet/internal/pkg/monitor"
+)
+
+const (
+	defaultCheckInterval           = 20 * time.Second // check for valid leaders every 20 seconds
```
ruflin, line 28: Making this configurable later on will become useful for integration tests. Then short intervals can be used with multiple fleet-servers to see if things work as expected.

internal/pkg/coordinator/monitor.go
```diff
@@ -0,0 +1,360 @@
...
+	metadataInterval  time.Duration
+	coordRestartDelay time.Duration
+
+	policies map[string]policyT
+}
+
+// NewMonitor creates a new coordinator policy monitor.
+func NewMonitor(fleet config.Fleet, version string, bulker bulk.Bulk, monitor monitor.Monitor, factory Factory) Monitor {
+	return &monitorT{
+		log:               log.With().Str("ctx", "policy leader manager").Logger(),
```
ruflin, line 69: Do we plan to use ECS logger?

internal/pkg/dsl/node.go
```diff
@@ -33,6 +33,14 @@ func (n *Node) MarshalJSON() ([]byte, error) {
 	return []byte(kKeywordNULL), nil
 }
 
+func (n *Node) MustMarshalJSON() []byte {
+	res, err := n.MarshalJSON()
+	if err != nil {
+		panic(err)
```
ruflin, line 39: Panic sounds extreme here?

ruflin (commented):
There is a LOT in this PR 🎉  I need to go through it again to understand all the bits. @aleksmaus Could you have a look?

One thing I tried to find out: Does the server version play a role in the leader election or not?

internal/pkg/coordinator/monitor.go
blakerouse: Yep, it is exposed as a value on the struct allowing it to be adjusted in integration test before starting the monitor.

internal/pkg/coordinator/monitor.go
blakerouse: We are currently using the zerolog, which works nicely.

internal/pkg/dsl/node.go
blakerouse: This is the expected behavior of any `Must` function. It is only used in a global var, which would get caught with testing.

The usage of a `Must` function should always be cautions, but in the case it is used correctly.

internal/pkg/coordinator/monitor.go
ruflin: No problem at the moment but we should make sure that fleet-server has ECS / JSON formatted logs when Elastic Agent (FB) starts to collect it to no transformation is needed.

internal/pkg/coordinator/monitor.go
```diff
@@ -0,0 +1,369 @@
...
+		}
+	}
+
+	// take/keep leadership and start new coordinators
+	res := make(chan policyT)
+	for _, p := range lead {
+		pt, _ := m.policies[p.PolicyId]
+		pt.id = p.PolicyId
+		go func(p model.Policy, pt policyT) {
+			l := m.log.With().Str(dl.FieldPolicyId, pt.id).Logger()
```
aleksmaus, line 222: maybe 
```
defer func() {
    res <- pt
}()
```
so don't forget to return something at every return from the func?

internal/pkg/coordinator/monitor.go
```diff
@@ -0,0 +1,369 @@
...
+		go func(pt policyT) {
+			if pt.cord != nil {
+				pt.canceller()
+			}
+			// uses a background context, because the context for the
+			// monitor will be cancelled at this point in the code
+			err := dl.ReleasePolicyLeadership(context.Background(), m.bulker, pt.id, m.agentMetadata.Id, m.leaderInterval, dl.WithIndexName(m.leadersIndex))
+			if err != nil {
+				l := m.log.With().Str(dl.FieldPolicyId, pt.id).Logger()
+				l.Err(err).Msg("failed to release leadership")
```
aleksmaus, line 278: nit:
what's the convention for the log messages? 
was following with the capital letter start convention before in other places.
Just need to make it consistent one way or another.

internal/pkg/coordinator/monitor.go
```diff
@@ -0,0 +1,369 @@
...
+			}
+			wg.Done()
+		}(pt)
+	}
+	wg.Wait()
```
aleksmaus, line 283: potential for long blocking on Wait here since the release is using background context. Do we expect this operation to finish in how long? worst case maybe use context with timeout?

internal/pkg/coordinator/monitor.go
```diff
@@ -0,0 +1,369 @@
...
+func runCoordinator(ctx context.Context, cord Coordinator, l zerolog.Logger, d time.Duration) {
+	for {
+		l.Info().Str("coordinator", cord.Name()).Msg("starting coordinator for policy")
+		err := cord.Run(ctx)
+		if err != context.Canceled {
+			l.Err(err).Msg("coordinator failed")
+			select {
+			case <-time.After(d):
+				break
+			case <-ctx.Done():
```
aleksmaus, line 347: maybe use timer that can reset here, otherwise, time.After will keep ticking until fired

internal/pkg/dl/actions_integration_test.go
```diff
@@ -89,27 +88,14 @@ func storeRandomActions(ctx context.Context, bulker bulk.Bulk, index string) ([]
...
-	if err != nil {
-		t.Fatal(err)
-	}
-
-	err = esboot.EnsureIndex(ctx, cli, index, es.MappingAction)
-	if err != nil {
-		t.Fatal(err)
-	}
+func setupActions(ctx context.Context, t *testing.T) (string, bulk.Bulk, []model.Action) {
+	index, bulker := ftesting.SetupIndexWithBulk(ctx, t, es.MappingAction)
```
aleksmaus, line 92: thank you!

internal/pkg/dl/policies.go
```diff
@@ -0,0 +1,82 @@
...
+	rSort.SortOrder(FieldRevisionIdx, dsl.SortDescend)
+	rSort.SortOrder(FieldCoordinatorIdx, dsl.SortDescend)
+	return root.MustMarshalJSON()
+}
+
+// QueryLatestPolices gets the latest revision for a policy
+func QueryLatestPolicies(ctx context.Context, bulker bulk.Bulk, opt ...Option) ([]model.Policy, error) {
+	initQueryLatestPoliciesOnce.Do(func() {
+		tmplQueryLatestPolicies = prepareQueryLatestPolicies()
+	})
```
aleksmaus, line 43: nit:
I moved all the queries preparations into package inited vars in the last PRs, cleaner and should fail sooner.
I can take care of it later when it's merged to make it more consistent

internal/pkg/dl/policies.go
```diff
@@ -0,0 +1,82 @@
...
+
+	o := newOption(FleetPolicies, opt...)
+	aggErr := fmt.Errorf("missing expected aggregation result")
```
aleksmaus, line 46: maybe move the error out so it's accessible for the callers

internal/pkg/dl/policies.go
```diff
@@ -0,0 +1,82 @@
...
+		return []model.Policy{}, nil
+	}
+	policies := make([]model.Policy, len(policyId.Buckets))
+	for i, bucket := range policyId.Buckets {
+		revisionIdx, ok := bucket.Aggregations[FieldRevisionIdx]
+		if !ok || len(revisionIdx.Hits) != 1 {
+			return nil, aggErr
+		}
+		hit := revisionIdx.Hits[0]
+		err = json.Unmarshal(hit.Source, &policies[i])
```
aleksmaus, line 66:
```
hit.Unmarshal
```

internal/pkg/dl/policies_leader.go
```diff
@@ -0,0 +1,147 @@
...
+	}
+	res, err := bulker.Search(ctx, []string{o.indexName}, data)
+	if err != nil {
+		return
+	}
+
+	leaders = map[string]model.PolicyLeader{}
+	for _, hit := range res.Hits {
+		var l model.PolicyLeader
+		err = json.Unmarshal(hit.Source, &l)
```
aleksmaus, line 57:
```
hit.Unmarshal
```

internal/pkg/es/result.go
```diff
@@ -54,8 +57,63 @@ type HitsT struct {
 	MaxScore *float64 `json:"max_score"`
 }
 
+type Bucket struct {
+	Key          string           `json:"key"`
+	DocCount     int64            `json:"doc_count"`
+	Aggregations map[string]HitsT `json:"-"`
+}
+
+type _bucket Bucket
```
aleksmaus, line 66: why?

internal/pkg/es/result.go
```diff
@@ -54,8 +57,63 @@ type HitsT struct {
...
+	err := json.Unmarshal(data, &b2)
+	if err != nil {
+		return err
+	}
+	var aggs map[string]interface{}
+	err = json.Unmarshal(data, &aggs)
+	if err != nil {
+		return err
+	}
+	typ := reflect.TypeOf(b2)
```
aleksmaus, line 79: can this be done without reflection?

internal/pkg/testing/retry.go
```diff
@@ -0,0 +1,57 @@
...
+	for _, opt := range opts {
+		opt(&o)
+	}
+	var err error
+	for i := 0; i < o.count; i++ {
+		err = f(ctx)
+		if err == nil {
+			return
+		}
+		time.Sleep(o.sleep)
```
aleksmaus, line 54: understand this is for testing, but still would be better to 
select on context and timer for wait rather than sleep

aleksmaus (commented):
One bug I think, otherwise mostly nits and questions.
Thank you!

internal/pkg/es/result.go
blakerouse: This is needed because otherwise an infinite loop will occur on the unmarshal because `Bucket` as the `UnmarshalJSON` function defined so it implements the interface.

By using a type alias it ensures that it will not recall the same method over and over again. So this must be used this way.

internal/pkg/es/result.go
blakerouse: I will remove the reflection, but we need to be sure that any keys we add to the `Bucket` type, is also updated in the `UnmarshalJSON`. I added some comments in hopes that it is not missed.

internal/pkg/coordinator/monitor.go
blakerouse: i am a big fan of lowercase log messages everywhere, think I only got the failed and start wrong on the monitor

internal/pkg/coordinator/monitor.go
blakerouse: added a timeout context

internal/pkg/coordinator/monitor.go
aleksmaus: ok, will start moving to that standard then on the pieces that I touch

blakerouse:
@aleksmaus Thanks for the review. It's ready for another look.

internal/pkg/testing/retry.go
```diff
@@ -51,7 +51,15 @@ func Retry(t *testing.T, ctx context.Context, f RetryFunc, opts ...RetryOption)
 		if err == nil {
 			return
 		}
-		time.Sleep(o.sleep)
+		t := time.NewTicker(o.sleep)
```
aleksmaus, line 54: time.NewTimer?

aleksmaus: maybe just add something like this as a function and use in both places?

```
func sleepWithContext(ctx context.Context, dur time.Duration) {
	t := time.NewTimer(dur)
	defer t.Stop()
	select {
	case <-ctx.Done():
	case <-t.C:
	}
}
```

aleksmaus (approved):
Couple of more nits, otherwise LGTM. Thank you!

PR #33, opened 2020-12-14
[CI] Enable JUnit

v1v:
## What does this PR do?

Enable JUnit reporting in the CI.

## Why is it important?

Honour the GitHub comments that report any test failures.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -77,16 +77,16 @@ test:  ## - Run all tests
 	@mkdir -p build
 	@$(MAKE) test-unit 
 	# @$(MAKE) test-int
-	@$(MAKE) junit
+	@$(MAKE) junit-report
 
 .PHONY: test-unit 
 test-unit: ## - Run unit tests only
 	@go test -v -race ./... | tee build/test-unit.out
 
-.PHONY: junit
-junit: ## - Run unit tests only
+.PHONY: junit-report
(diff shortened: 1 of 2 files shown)


aleksmaus (approved):

Makefile
```diff
@@ -74,12 +74,19 @@ check-no-changes:
...
 test:  ## - Run all tests
+	@mkdir -p build
 	@$(MAKE) test-unit 
 	# @$(MAKE) test-int
+	@$(MAKE) junit-report
 
 .PHONY: test-unit 
 test-unit: ## - Run unit tests only
-	@go test -v -race ./...
+	@go test -v -race ./... | tee build/test-unit.out
```
kuisathaverat, line 84: you'd need to set pipefail for bash

```
.SHELLFLAGS = -eo pipefail
```

v1v: I was somehow skeptical when I wrote it since I wanted to delegate the junit reporting to be the one driving the build status. But there might be some corner cases that it's worth to use the shellflags

kuisathaverat: the thing is junit will mark the build as unstable, not as failed

Makefile
```diff
@@ -1,3 +1,5 @@
+SHELL = /bin/bash
```
v1v, line 1:
```suggestion
```

kuisathaverat (approved):

PR #34, opened 2020-12-14
Consolidate sleep.WithContext

aleksmaus:
## What does this PR do?

Consolidate sleep.WithContext.
Same pattern is used in couple of places.


blakerouse (approved):
Looks good.

PR #35, opened 2020-12-15
Remove esutil package. Move some of the craft that is still used for bootstrapping to esboot package, since other parts do not use it and should not.

aleksmaus:
## What does this PR do?

Remove esutil package. Move some of the craft that is still used for bootstrapping to esboot package, since other parts do not use it and should not.


internal/pkg/esutil/hit.go
```diff
@@ -1,21 +0,0 @@
-// Copyright Elasticsearch B.V. and/or licensed to Elasticsearch B.V. under one
-// or more contributor license agreements. Licensed under the Elastic License;
-// you may not use this file except in compliance with the Elastic License.
-
-package esutil
-
-import "encoding/json"
-
-type Hit struct {
```
ruflin, line 9: This code was not used at all?

ruflin (approved):

internal/pkg/esboot/esutil.go
```diff
@@ -2,7 +2,7 @@
 // or more contributor license agreements. Licensed under the Elastic License;
 // you may not use this file except in compliance with the Elastic License.
 
-package esutil
+package esboot
```
simitt, line 5: nit: Would it be worth unexporting the functionality since it is only used in the `esboot` package now?

internal/pkg/esutil/hit.go
aleksmaus: yeah, consolidated some of the pieces, removing unused anymore

internal/pkg/esboot/esutil.go
aleksmaus: sure, will do.

PR #36, opened 2020-12-15
Remove runner package, replace with errgroup

aleksmaus:
## What does this PR do?
Remove runner package, replace with golang.org/x/sync/errgroup

Was not quite happy with the custom runnable, replaced with errgroup, I think it's cleaner.


blakerouse (approved):
Looks good.

PR #37, opened 2020-12-15
Add fleet action results support.

aleksmaus:
## What does this PR do?

* Fleet action results support.
* Fleet server integration test, running full server and currently covering unauthorized requests. Planning to incrementally build on top of that to cover the full agent life-cycle, hitting the actual API endpoints of the server.


aleksmaus:
Well, failed integration test starting the server due to missing .kibana index, working on the fix.

aleksmaus:
was cleaning up some old branches, deleted this one by mistake, thought it was merged. reopened.

cmd/fleet/router.go
```diff
@@ -8,34 +8,39 @@ import (
...
+		ct:       ct,
+		et:       et,
+		validate: validator.New(),
 	}
 
 	router := httprouter.New()
 	router.POST(ROUTE_ENROLL, r.handleEnroll)
 	router.POST(ROUTE_CHECKIN, r.handleCheckin)
 	router.POST(ROUTE_ACKS, r.handleAcks)
+	router.POST(ROUTE_RESULT, r.handleActionResult)
```
blakerouse, line 44: Why the new endpoint? Why not just use the `ROUTE_ACKS`? Elastic Agent already handles actions and response with the Ack.

blakerouse (commented):
Confused on the new endpoint? Why doesn't the `acks` endpoint work? Its already implemented to handle acking actions in Elastic Agent, no code change would be needed in the Elastic Agent for this.

cmd/fleet/router.go
aleksmaus: It was a part of the original proposal doc and I didn't get any objections. The results endpoint is not an ack. The result document is posted upon completion of the action, with the details on of the completion an an error if error happened.
Assumed that the expiration/timeout of the action will be stored as the result document with the error code.

aleksmaus:
With the new agents actions flow we are not acking actions as well. do acks endpoint is not used at least for actions.

cmd/fleet/router.go
blakerouse: So agent will both ack and providing results?

aleksmaus: For the agent actions (that I worked on), the agent doesn't need to ack.
The workflow for actions looks something like this:
1. The Agent calls "checkin" API
2. The Fleet Server checked the "last action token"(if provided), and sends the new actions with the "token" in response. Next time the fleet server "checks in" it sends the token to the fleet server, this updates the agent record so we track the last action received by the agent.
3. The Agent dispatches the actions to the corresponding beats/applications
4. The beats/applications execute the action and stream the data to ES directly
5. The beats/applications communicate the result/error of the action execution back to the agent upon the completion of the action
6. The Agent calls the "result" API

blakerouse (approved):
Looks good.

Will take this and integrate it into the /acks

PR #39, opened 2020-12-15
Remove agent saved object from enrollment code completely

aleksmaus:
## What does this PR do?

Remove agent saved object from enrollment code completely. With this change the agent is not going to appear in kibana upon enrollment anymore until kibana switched to the new indexes.

## Why is it important?

This allows to test at scale fairly (without double saving into .kibana as before)


blakerouse (approved):
Looks good.

PR #41, opened 2020-12-17
Add new policy.Monitor for use with checkin

blakerouse:
## What does this PR do?

Adds and uses a new `policy.Monitor` that reads policy changes from the `.fleet-policies` index. Takes into account the new `coordinator_idx`.

Adds the handling of an ack to update the current policy revision of an Agent.

## Why is it important?

Send policy changes are sent to the Agents.

## Checklist


cmd/fleet/handleCheckin.go
```diff
@@ -315,33 +317,32 @@ func parsePolicy(ctx context.Context, bulker bulk.Bulk, agentId string, action A
...
-		Id:        action.Id,
-		Type:      action.Type,
+		Id:        idForPolicy(policy.PolicyId, policy.RevisionIdx, policy.CoordinatorIdx),
+		Type:      TypePolicyChange,
 	}
 
 	return &resp, nil
 }
 
+func idForPolicy(policyId string, revIdx int64, coordIdx int64) string {
```
aleksmaus, line 342: nit:
this and the code above seem like a good candidate for the policyId wrapper that supports Parse() and String()

cmd/fleet/main.go
```diff
@@ -229,10 +230,8 @@ func (f *FleetServer) runServer(ctx context.Context, cfg *config.Config) (err er
 	g.Go(loggedRunFunc(ctx, "Coordinator policy monitor", cord.Run))
 
 	// Policy monitor
-	pm, err := NewPolicyMon(kPolicyThrottle)
-	g.Go(loggedRunFunc(ctx, "Policy monitor", func(ctx context.Context) error {
-		return pm.Monitor(ctx, sv)
-	}))
+	pm := policy.NewMonitor(bulker, pim, kPolicyThrottle)
+	g.Go(loggedRunFunc(ctx, "Policy monitor", pm.Run))
```
aleksmaus, line 234: 👍

internal/pkg/monitor/monitor.go
```diff
@@ -213,10 +262,20 @@ func (m *monitorT) notify(ctx context.Context, hits []es.HitT) {
...
 		m.storeCheckpoint(maxVal)
 
-		select {
-		case m.outCh <- hits:
-		case <-ctx.Done():
+		m.mut.RLock()
+		var wg sync.WaitGroup
+		wg.Add(len(m.subs))
+		for _, s := range m.subs {
+			go func(s *subT) {
```
aleksmaus, line 269: depending on the usage, there is a chance of the race and go routine leak:
1. you start a go routine in the loop to send the message
2. the client unsubscribes, never consumes the message from the channel
3. by the time the go routine fires it will be blocked on sending the hit and leak. the unbuffered channel can make this happen more often 

with this addition the monitor became heavier since we added additional syncing, kicking off the go routines for each notifications and could potentially leak go routines.
could help a bit if using buffered channel, and maybe drop notifications if channel is full to avoid the go routine leak.

aleksmaus: same problem with the Mock

internal/pkg/policy/monitor.go
```diff
@@ -0,0 +1,351 @@
...
+	zlog := m.log.With().Str("policyId", policy.PolicyId).Logger()
+
+	subs := m.updatePolicy(policy)
+	if subs == nil {
+		return nil
+	}
+	if len(subs) == 0 {
+		zlog.Info().Msg("no pending subscriptions to revised policy")
+		return nil
+	}
```
aleksmaus, line 182: can replace both ifs above with one 
```
if len(subs) == 0 {
```

aleksmaus (approved):
Few nits/questions. One concern about the index monitor change.

internal/pkg/monitor/monitor.go
blakerouse: The solution is to close the channel on unsubscribe. Then the sending goroutine will panic for sending on a closed goroutine, the panic can then be recovered and the goroutine will exit. On the next notify the subscription will be gone and that issue will not be hit.

I have added a tests to confirm this behavior.

internal/pkg/policy/monitor.go
blakerouse: It could but they have a different meaning. `nil` means the policy should not be rolled out because the coordinatorIdx is 0, and `len(0)` means that no one is subscribed to it

internal/pkg/policy/action.go
```diff
@@ -0,0 +1,62 @@
...
+	"fleet/internal/pkg/model"
+	"fmt"
+	"strconv"
+	"strings"
+
+	"github.com/gofrs/uuid"
+)
+
+// Action is a policy action that is sent as an action ID to an agent.
+type Action struct {
```
aleksmaus, line 17: maybe different name? we already have model.Action. could be confusing

internal/pkg/monitor/monitor.go
aleksmaus: Still not a big fan of the locking and go routines and now panic recovery, but ok.

aleksmaus (approved):

internal/pkg/policy/action.go
blakerouse: Change it to `policy.Revision`.

PR #43, opened 2020-12-17
Fix actions ID

aleksmaus:
## What does this PR do?

Fix the action ids. The action document id was used for the action id instead of the uuid action id.

Actions results before the change:
Screen Shot 2020-12-17 at 2 52 22 PM

Actions results after the change:
Screen Shot 2020-12-17 at 2 52 55 PM


blakerouse (approved):
Looks good.

PR #46, opened 2020-12-20
Refactor index monitor.

aleksmaus:
## What does this PR do?

* Refactor index monitor. 
* Extract the simple monitor and subclass the subscription monitor from it 
  that preserves the same interfaces and behavior as before.
* Use the simple monitor for agents actions since it is more efficient and avoids some the issues I listed below.

@blakerouse
I think there are few problems with the latest subscription monitor changes:
1. The callbacks firing each call with the new go routine. 
   Depending on the timing of the go scheduler the order in which the go routines are fired is not predictable. This kind of an edge case since the monitor poll interval is 1 sec, and most likely your go routine will be scheduled within that time, but it's still fragile.
2. The tracking of the sequence number was not quite correct. 
   It was advanced before the notifications were sent globally, which could potentially cause the subscriber to miss some events in case of timeouts.
   To avoid this we would have to track that the sequence number per subscriber, which complicates the monitor even further. And once you start tracking per subscription you might end up in the situation where one of the subscribers lags so much behind that will exceed the default size of the hit documents. So again complicates the monitor further.  
3. It is heavier in terms of the go routines and locking.

I refactored that into base SimpleMonitor, that serves better for the agent actions and avoids the issues listed above.
I "subclassed" (built on top of that) the subscription monitor and have put all the subscription code
that you added later in that piece. So from the Monitor point of view the interface and the behavior is still the same, the tests are still the same.

I'm not sure how critical the problems that I listed above for the policy and coordinator that you were working on, so I leave it up to you for now. Let me know if you want me look instead.

Btw, seeing some flakiness in the policy monitor tests, intermittent failures from time to time.
My last PR has built fine but the build failed with CI once it was merged to master: 
[2020-12-18T19:33:45.979Z]     TestMonitor_NewPolicyExists: monitor_test.go:313: never got policy update; timed out after 500ms
[2020-12-18T19:33:45.979Z] --- FAIL: TestMonitor_NewPolicyExists (2.00s)
[2020-12-18T19:33:45.979Z] FAIL

### DIFF
--- a/cmd/fleet/main.go
+++ b/cmd/fleet/main.go
@@ -237,12 +237,12 @@ func (f *FleetServer) runServer(ctx context.Context, cfg *config.Config) (err er
 	g.Go(loggedRunFunc(ctx, "Policy monitor", pm.Run))
 
 	// Actions monitoring
-	var am monitor.Monitor
+	var am monitor.SimpleMonitor
 	var ad *action.Dispatcher
 	var tr *action.TokenResolver
 
 	// Behind the feature flag
-	am, err = monitor.New(dl.FleetActions, es, monitor.WithExpiration(true))
+	am, err = monitor.NewSimple(dl.FleetActions, es, monitor.WithExpiration(true))
 	if err != nil {
 		return err
 	}
--- a/dev-tools/integration/main.go
+++ b/dev-tools/integration/main.go
@@ -6,10 +6,13 @@ package main
 
 import (
 	"context"
+	"errors"
 	"fleet/internal/pkg/config"
 	"fleet/internal/pkg/es"
 	"fleet/internal/pkg/esboot"
 	"fmt"
+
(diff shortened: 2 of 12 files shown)


aleksmaus:
yep, flaky test again with the second commit @blakerouse 
[2020-12-21T21:00:34.360Z] {"level":"debug","ctx":"policy agent manager","time":"2020-12-21T21:00:31Z","message":"kick channel full"}

[2020-12-21T21:00:34.360Z]     TestMonitor_NewPolicyExists: monitor_test.go:313: never got policy update; timed out after 500ms

[2020-12-21T21:00:34.360Z] --- FAIL: TestMonitor_NewPolicyExists (2.00s)

aleksmaus:
Commented the flaky test, opened an issue. Will merge this now since it is green and probably nobody will look at it till January 4th.

PR #56, opened 2021-01-06
Add notes around encryption key usage

ruflin:
This adds a note on how to pass in the encryption key to fleet-server to connect it to Elasticsearch. This is only temporary until changes were made in Kibana.

Additionaly the fleet-server.dev.yml was added to gitignore. This allows to have a local config file which is not checked in.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,3 +6,4 @@ bin/
 build/
 
 fleet-server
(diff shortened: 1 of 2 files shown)


aleksmaus (approved):
👍

PR #57, opened 2021-01-06
Rename module name to fleet-server

ruflin:
I stumbled over this in two parts: I was surprised that the build binary is called `fleet` instead of `fleet-server` and `go get github.com/elastic/fleet-server` does complain with an error:

```
go get github.com/elastic/fleet-server
go: downloading github.com/elastic/fleet-server v0.0.0-20210105132059-b47b94c40513
go get: github.com/elastic/fleet-server@none updating to
	github.com/elastic/fleet-server@v0.0.0-20210105132059-b47b94c40513: parsing go.mod:
	module declares its path as: fleet
	        but was required as: github.com/elastic/fleet-server
```

I think the module should be called `fleet-server` as this is what we are building. Can someone share some background on why we called it `fleet`. Will this also break other things?


aleksmaus:
The module name should match the github path in general, if anybody needs to pull in as a module. 
So it was probably just overlooked, since nobody needed to pull the fleet-server as a module into other projects.
If we are planning to use any packages from the `fleet-server` in any other projects then it needs to be renamed.

The binary name doesn't matter that much, but we can make it consistent as well and rename the fleet to fleet-server.

go.mod
```diff
@@ -1,4 +1,4 @@
-module fleet
```
aleksmaus, line 1: use the full path for the module name and the imports if we really want to this as a module anywhere. 


For example:
https://github.com/nats-io/nats.go/blob/master/examples/nats-pub/main.go

```
github.com/nats-io/nats.go
```

so in our case it will be 
```
module github.com/elastic/fleet-server
```

and the imports:
```
import "github.com/elastic/fleet-server/internal/pkg/policy"
```

When you start major versioning the module there will be a version in the path as well

```
import "github.com/elastic/fleet-server/v3/internal/pkg/policy"
```

ruflin:
Ok, in this case I would suggest we rename it all the `fleet-server` to be consistent. Agree?

aleksmaus (changes requested):

aleksmaus:
> Ok, in this case I would suggest we rename it all the `fleet-server` to be consistent. Agree?

sounds good to me

blakerouse:
I have a PR up that sets to correct paths and uses github.com/elastic/fleet-server/v7 on #65. Lets use that one instead of this one.

aleksmaus:
Nice! Thanks!

PR #59, opened 2021-01-07
Fix the agent ID handling

aleksmaus:
## What does this PR do?

Sets the correct agent id in the local meta upon enrollment if the agent id was provided. Does modify the local meta if the agent was not provided. 
Eventually the agent id in the local meta should not be used and the doc id value on the agent document should be used everywhere instead.

The related horde fix for the agent ID was done here https://github.com/elastic/horde/pull/23

### DIFF
--- a/cmd/fleet/handleEnroll.go
+++ b/cmd/fleet/handleEnroll.go
@@ -191,12 +191,18 @@ func _enroll(ctx context.Context, bulker bulk.Bulk, c cache.Cache, req EnrollReq
 		Str("defaultOutputApiKey.Id", defaultOutputApiKey.Id).
 		Msg("Created api key")
 
+	// Update the local metadata agent id
+	localMeta, err := updateLocalMetaAgentId(req.Meta.Local, agentId)
+	if err != nil {
+		return nil, err
+	}
+
 	agentData := model.Agent{
 		Active:          true,
 		PolicyId:        erec.PolicyId,
 		Type:            req.Type,
 		EnrolledAt:      nowStr,
-		LocalMetadata:   req.Meta.Local,
+		LocalMetadata:   localMeta,
 		AccessApiKeyId:  accessApiKey.Id,
 		DefaultApiKeyId: defaultOutputApiKey.Id,
 		DefaultApiKey:   defaultOutputApiKey.Token(),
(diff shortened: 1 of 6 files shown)


cmd/fleet/handleEnroll.go
```diff
@@ -229,6 +235,62 @@ func _enroll(ctx context.Context, bulker bulk.Bulk, c cache.Cache, req EnrollReq
...
+		return data, nil
+	}
+
+	var m map[string]interface{}
+	err := json.Unmarshal(data, &m)
+	if err != nil {
+		return nil, err
+	}
+
+	if v, ok := m["elastic"]; ok {
```
ruflin, line 275: Are we using `common.MapStr` anywhere in fleet-server? If yes, this could be used here with `Get`.

dev-tools/rpm/nfpm.yaml
```diff
@@ -10,9 +10,15 @@ maintainer: "devops@endgame.com"
 description: Elastic Fleet
 vendor: "Elastic NV"
 homepage: "http://www.elastic.co/"
-files:
-  ./bin/fleet: "/usr/bin/fleet"
-config_files:
-  ./systemd/fleet.service: "/usr/lib/systemd/system/fleet.service"
-  ./fleet-server.yml: "/usr/share/fleet/fleet-server.yml"
+contents:
```
ruflin, line 13: Not sure how these changes are related?

systemd/fleet.service
```diff
@@ -4,7 +4,7 @@ After=network.target
 Requires=network.target
 
 [Service]
-ExecStart=/usr/bin/fleet -c /usr/share/fleet/fleet-server.yml
+ExecStart=/usr/bin/fleet-server -c /usr/share/fleet/fleet-server.yml
```
ruflin, line 7: I assume this would have to change if https://github.com/elastic/fleet-server/pull/57 is merged?

ruflin (approved):
LGTM. I'm not fully sure about the dev-tools changes on how they are related.

cmd/fleet/handleEnroll.go
aleksmaus: We don't, but yeah, could use a better wrapper if we have to do this in any other place.

dev-tools/rpm/nfpm.yaml
aleksmaus: This is not related, but the `make rpm` was broken, I just fixed it. Convenient when needed to build Linux binaries on Mac.

systemd/fleet.service
aleksmaus: yep, will adjust again, if we are changing the binary name once that PR landed

PR #62, opened 2021-01-15
[Fleet] Fix api key format for agent policy

nchaulet:
## Description 

The API key in the policy should be in the format `keyId:key` not base64 encoded this PR fix that.

### DIFF
--- a/internal/pkg/apikey/apikey.go
+++ b/internal/pkg/apikey/apikey.go
@@ -58,7 +58,7 @@ func (k ApiKey) Token() string {
 	return base64.StdEncoding.EncodeToString([]byte(s))
 }
 
-func (k ApiKey) Config() string {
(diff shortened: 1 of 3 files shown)


internal/pkg/apikey/apikey.go
```diff
@@ -58,6 +58,10 @@ func (k ApiKey) Token() string {
 	return base64.StdEncoding.EncodeToString([]byte(s))
 }
 
+func (k ApiKey) Config() string {
```
nchaulet, line 61: We can probably find a better name

blakerouse: Maybe `Agent()` being this is what gets sent to the agent.

blakerouse (approved):
Change looks good. Not hard change on name, if you want to leave it as `Config()` that is fine as well.

PR #64, opened 2021-01-19
Add release/snapshot building

blakerouse:
## What does this PR do?

Add `make release-manager-release` and `make release-manager-snapshot` targets that cross-build fleet-server. Output is placed into `build/distributions`.

## Why is it important?

So final build artifacts can be built.

## Checklist

## Logs

```
GOOS=darwin GOARCH=amd64 go build -ldflags="-w -s -X main.Version=8.0.0-SNAPSHOT" -o build/binaries/fleet-server-8.0.0-SNAPSHOT-darwin-x86_64/fleet-server .
GOOS=linux GOARCH=386 go build -ldflags="-w -s -X main.Version=8.0.0-SNAPSHOT" -o build/binaries/fleet-server-8.0.0-SNAPSHOT-linux-x86/fleet-server .
GOOS=linux GOARCH=amd64 go build -ldflags="-w -s -X main.Version=8.0.0-SNAPSHOT" -o build/binaries/fleet-server-8.0.0-SNAPSHOT-linux-x86_64/fleet-server .
GOOS=linux GOARCH=arm64 go build -ldflags="-w -s -X main.Version=8.0.0-SNAPSHOT" -o build/binaries/fleet-server-8.0.0-SNAPSHOT-linux-arm64/fleet-server .
GOOS=windows GOARCH=386 go build -ldflags="-w -s -X main.Version=8.0.0-SNAPSHOT" -o build/binaries/fleet-server-8.0.0-SNAPSHOT-windows-x86/fleet-server .
GOOS=windows GOARCH=amd64 go build -ldflags="-w -s -X main.Version=8.0.0-SNAPSHOT" -o build/binaries/fleet-server-8.0.0-SNAPSHOT-windows-x86_64/fleet-server .
```


.go-version
```diff
@@ -0,0 +1 @@
+1.14.12
```
ruflin, line 1: Nit: Should we go to 1.15.?

Makefile
```diff
@@ -14,33 +21,16 @@ help: ## - Show help message
...
-	@printf "${CMD_COLOR_ON} Build rpm\n${CMD_COLOR_OFF}"
-	@${DOCKER_BUILD} --ssh default  --target rpm -o ./  .
-
-.PHONY: image
-image:	## - Build the elastic fleet docker images
-	@printf "${CMD_COLOR_ON} Build the elastic fleet docker image\n${CMD_COLOR_OFF}"
-	${DOCKER_BUILD} --ssh default --target fleet -t fleet .
-
-.PHONY: run
-run: image ## - Run the smallest and secured golang docker image based on scratch
```
aleksmaus, line 28: nit: are we removing docker image build option?

ruflin (approved):
I did run both commands locally and tried to execute the darwin SNAPSHOT binary and seems to do what it should do.

aleksmaus (approved):

Makefile
blakerouse: Yes, we don't need to build in docker. Its simple cross architecture building, which golang does natively.

.go-version
blakerouse: This is matching beats. I think we should just map to beats, when beats moves we move.

ruflin: I think the nice part of being a separate binary is that we are not tied to what Beats has. I would rather tie it to the version the Elastic Agent uses if needed.

PR #68, opened 2021-01-20
Make Fleet Server components resilient if the indices do not exist

aleksmaus:
## What does this PR do?

The new package based approach installs the indices templates and ILM policies, but doesn't create the indexes. This change allows the fleet server to run if indexes do not exists without exiting with error.

The bootstrapping code is commented out now, expecting the bootstrap to happen from the integration package.
https://github.com/elastic/integrations/pull/544

### DIFF
--- a/internal/pkg/coordinator/monitor.go
+++ b/internal/pkg/coordinator/monitor.go
@@ -6,6 +6,7 @@ package coordinator
 
 import (
 	"context"
+	"errors"
 	"net"
 	"os"
 	"runtime"
@@ -184,7 +185,11 @@ func (m *monitorT) ensureLeadership(ctx context.Context) error {
 	leaders := map[string]model.PolicyLeader{}
 	policies, err := dl.QueryLatestPolicies(ctx, m.bulker, dl.WithIndexName(m.policiesIndex))
 	if err != nil {
-		return err
+		if errors.Is(err, es.ErrIndexNotFound) {
+			err = nil
+		} else {
+			return err
+		}
 	}
 	if len(policies) > 0 {
 		ids := make([]string, len(policies))
@@ -193,7 +198,11 @@ func (m *monitorT) ensureLeadership(ctx context.Context) error {
 		}
 		leaders, err = dl.SearchPolicyLeaders(ctx, m.bulker, ids, dl.WithIndexName(m.leadersIndex))
 		if err != nil {
-			return err
+			if errors.Is(err, es.ErrIndexNotFound) {
+				err = nil
+			} else {
+				return err
+			}
 		}
 	}
 
--- a/internal/pkg/coordinator/monitor.go
+++ b/internal/pkg/coordinator/monitor.go
@@ -186,10 +186,10 @@ func (m *monitorT) ensureLeadership(ctx context.Context) error {
 	policies, err := dl.QueryLatestPolicies(ctx, m.bulker, dl.WithIndexName(m.policiesIndex))
 	if err != nil {
 		if errors.Is(err, es.ErrIndexNotFound) {
-			err = nil
-		} else {
-			return err
+			m.log.Debug().Str("index", m.policiesIndex).Msg(es.ErrIndexNotFound.Error())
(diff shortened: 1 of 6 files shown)


internal/pkg/coordinator/monitor.go
```diff
@@ -184,7 +185,11 @@ func (m *monitorT) ensureLeadership(ctx context.Context) error {
 	leaders := map[string]model.PolicyLeader{}
 	policies, err := dl.QueryLatestPolicies(ctx, m.bulker, dl.WithIndexName(m.policiesIndex))
 	if err != nil {
-		return err
+		if errors.Is(err, es.ErrIndexNotFound) {
+			err = nil
```
ruflin, line 189: Should we have a debug log here to make it easier to debug issues?

internal/pkg/coordinator/monitor.go
```diff
@@ -193,7 +198,11 @@ func (m *monitorT) ensureLeadership(ctx context.Context) error {
 		}
 		leaders, err = dl.SearchPolicyLeaders(ctx, m.bulker, ids, dl.WithIndexName(m.leadersIndex))
 		if err != nil {
-			return err
+			if errors.Is(err, es.ErrIndexNotFound) {
+				err = nil
```
ruflin, line 202: Same as above, debug log?

internal/pkg/migrate/migrate.go
```diff
@@ -42,12 +45,19 @@ func MigrateEnrollmentAPIKeys(ctx context.Context, sv saved.CRUD, bulker bulk.Bu
 	}
 
 	var recs []model.EnrollmentApiKey
+	var resHits []es.HitT
 	res, err := bulker.Search(ctx, []string{dl.FleetEnrollmentAPIKeys}, raw, bulk.WithRefresh())
 	if err != nil {
-		return err
+		if errors.Is(err, es.ErrIndexNotFound) {
+			err = nil
```
ruflin, line 52: Debug log?

Could we directly return nil here or do we need to run the rest of the code below?

internal/pkg/monitor/monitor.go
```diff
@@ -291,7 +292,14 @@ func (m *simpleMonitorT) search(ctx context.Context, tmpl *dsl.Tmpl, params map[
 	}
 
 	if res.IsError() {
-		return nil, es.TranslateError(res.StatusCode, esres.Error)
+		err = es.TranslateError(res.StatusCode, esres.Error)
+	}
+
+	if err != nil {
+		if errors.Is(err, es.ErrIndexNotFound) {
+			return nil, nil
```
ruflin, line 300: Debug log?

internal/pkg/policy/monitor.go
```diff
@@ -125,6 +126,11 @@ LOOP:
 func (m *monitorT) process(ctx context.Context) error {
 	policies, err := m.policyF(ctx, m.bulker, dl.WithIndexName(m.policiesIndex))
 	if err != nil {
+		if errors.Is(err, es.ErrIndexNotFound) {
+			err = nil
```
ruflin, line 130: Debug log?

internal/pkg/policy/monitor.go
```diff
@@ -125,6 +126,11 @@ LOOP:
...
+		} else {
+			err = nil
```
ruflin, line 132: should this be `return err`?

ruflin (commented):
++ on being able to run if indices not exist and wait for until the indices are fully setup. But I think we should have at least some debug logs that indicate why things are not working yet.

I wonder for the methods if we could most of the time return / exist early and just do nothing or if we have to run all the code in the methods.

internal/pkg/policy/monitor.go
simitt: Also return here immediately? There won't be any policies yet.

aleksmaus: don't know what I was thinking yesterday. will update.

PR #73, opened 2021-01-27
Rename action input_id to input_type

aleksmaus:
## What does this PR do?

Rename action input_id to input_type. Instead of integration input id that is generated every time by kibana, rely on more stable integration input type.

Related Fleet Server package change: https://github.com/elastic/integrations/pull/573


nchaulet:
it is possbile that one `input_type` will target multiple inputs?

blakerouse:
@nchaulet Totally possible but that should be okay because 1 input type maps to only 1 application that Elastic Agent will run.

blakerouse (approved):
Looks good.

PR #80, opened 2021-02-02
Fix permissions on dependencies-report script

jmlrt:
This commit is adding execution permission to dependencies-report script
which are required during the unified release build.


blakerouse (approved):
Aw missed that, looks good.

PR #82, opened 2021-02-02
Reduce actions fetching interval if the full page of action documents was fetched

aleksmaus:
## What does this PR do?

Reduce actions fetching interval if the full page of action documents was fetched. Otherwise the monitor waits for 1 secs (by default) until the next poll.

Presently the UPGRADE action is creating individual action documents for each agent, which will be slow to roll out if we delay for 1 sec between each 10 actions. Ideally we would group as many agents under one document as possible.
Talked to @nchaulet and this is on his to do list for kibana.

### DIFF
--- a/internal/pkg/model/schema.go
+++ b/internal/pkg/model/schema.go
@@ -51,7 +51,7 @@ type Action struct {
 	// Date/time the action was created
 	Timestamp string `json:"@timestamp,omitempty"`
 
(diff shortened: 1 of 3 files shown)


scunningham (approved):
Better than a ❄️

blakerouse (approved):
Looks good.

PR #92, opened 2021-02-16
Change default bind port from 8000 to 8220

scunningham:
## What does this PR do?

Changes the default TCP/IP port that the fleet server binds to from 8000 to 8220.

## Why is it important?

8000 is commonly used.  8220 is not.

## Checklist

### DIFF
--- a/internal/pkg/config/input.go
+++ b/internal/pkg/config/input.go
@@ -10,6 +10,9 @@ import (
 	"time"
 )
 
+const kDefaultHost = "localhost"
+const kDefaultPort = 8220
+
 // Policy is the configuration policy to use.
 type Policy struct {
 	ID string `config:"id"`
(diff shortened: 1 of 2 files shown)


internal/pkg/config/input.go
```diff
@@ -10,6 +10,9 @@ import (
 	"time"
 )
 
+const kDefaultHost = "localhost"
+const kDefaultPort = 8220
```
aleksmaus, line 14: nit: not a big fan on 'k' prefixes in Go.

aleksmaus (approved):
Fonzi 👍

nchaulet (approved):
🚀  we probably want to have this as default value in the package too

aleksmaus:
> 🚀 we probably want to have this as default value in the package too

here: https://github.com/elastic/integrations/pull/699

PR #94, opened 2021-02-17
[Backport 7.x]  Change default bind port from 8000 to 8220 (#92)

scunningham:
[Backport 7.x]  Change default bind port from 8000 to 8220 (#92)


blakerouse (approved):
Backport looks good.

PR #105, opened 2021-02-25
Remove ES bootsrap code from Fleet Server

aleksmaus:
* Moved it to testing/esutil package because it is still used for
  integration testing indices bootstrapping


blakerouse (approved):
Nice, great to see this moved out of main.

scunningham (approved):
less is more

PR #112, opened 2021-03-01
Enable trace level logging

aleksmaus:
## What does this PR do?
Enable trace level logging.

Fleet Server has some useful trace level logs that are too noisy to surface to debug level.
The Elastic Agent would not be able to set the config lower than debug level, but this is still useful for development&diags when running Fleet Server standalone.


scunningham (approved):
grace.

blakerouse (approved):
Looks good.

Issue #115, opened 2021-03-03
Checkpoint Monitoring is broken after replacing .fleet-policies and .fleet-actions indices with aliases


aleksmaus:
Checkpoint Monitoring is broken after a change in .fleet indices bootstrapping by kibana that replaced .fleet-policies and .fleet-actions indices with aliases.

For example the API call
```
GET .fleet-actions/_stats?level=shards
```

returns 
```
  "indices" : {
    ".fleet-actions_1" : {
      "uuid" : "WdDo6Ji4Qva-ohKtQL3zPw",
 ```
and ".fleet-actions_1" doesn't match the index name as currently expected, so the global checkpoint is not found.

One possible issue down the road with aliasing the index is that there is nothing that enforces only one index under the given alias.
Adding the second index will break the seq_no based monitoring.

PR #116, opened 2021-03-03
Fix global checkpoint based monitoring

aleksmaus:
## What does this PR do?

Fixes the indices monitoring, that got broken due to replacing the original
.fleet-actions and .fleet-policies indices with aliases.

This change uses the first index global checkpoint value received from
stats and returns the error if there are more than one index for
alias for any reason.

## Checklist

## Related issues

- Closes https://github.com/elastic/fleet-server/issues/115

## Screenshots
before 
Screen Shot 2021-03-03 at 3 15 08 PM

after
Screen Shot 2021-03-03 at 3 15 27 PM

This unblocks the devs dependent on this feature ASAP. 
Will follow up with an issue that possibly reports degraded status to the agent instead and unit tests coverage.

### DIFF
--- a/internal/pkg/monitor/global_checkpoint.go
+++ b/internal/pkg/monitor/global_checkpoint.go
@@ -7,10 +7,15 @@ package monitor
 import (
 	"context"
 	"encoding/json"
+	"errors"
+	"fmt"
+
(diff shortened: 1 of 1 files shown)


blakerouse (approved):
Looks good.

Issue #121, opened 2021-03-04
Add an option to set a hard limit on the number of TCP connections the fleet server will accept.


scunningham:
**Describe the enhancement:**

To prevent a fleet server from allocating too many resources in a low resource environment, allow an optional hard limit on the number of concurrently accepted connections.  The option will be disabled by default.

PR #122, opened 2021-03-04
Add optional hard connection limit to the server.  No more than N con…

scunningham:
…nections will be active at any time.

## What does this PR do?

Adds optional hard limit to the number connections accepted at one time by server.

## Why is it important?

This will limit resource usage on the server.   Both CPU and memory usage is largely driven by the number of connected agents.

## Checklist

Add the 'hard_conn_limit' to the fleet_server
## Related issues

- 

https://github.com/elastic/fleet-server/issues/121


internal/pkg/config/input.go
```diff
@@ -57,6 +57,7 @@ type Server struct {
 	MaxHeaderByteSize int               `config:"max_header_byte_size"`
 	RateLimitBurst    int               `config:"rate_limit_burst"`
 	RateLimitInterval time.Duration     `config:"rate_limit_interval"`
+	HardConnLimit     int               `config:"hard_conn_limit"`
```
blakerouse, line 60: Is the `hard_` prefix needed.

What about `max_connections`?

internal/pkg/es/mapping.go
```diff
@@ -1,7 +1,3 @@
-// Copyright Elasticsearch B.V. and/or licensed to Elasticsearch B.V. under one
-// or more contributor license agreements. Licensed under the Elastic License;
-// you may not use this file except in compliance with the Elastic License.
```
aleksmaus, line 3: probably will not like removed license header

aleksmaus (approved):
👍

internal/pkg/config/input.go
scunningham: sure.

internal/pkg/es/mapping.go
scunningham: tnx

PR #130, opened 2021-03-11
Handle artifact request from endpoint  

scunningham:
## What does this PR do?

The security app generates multiple user defined artifacts for delivery to the endpoint.  There is an existing implementation in Kibana to server up these artifacts via HTTPS.  With the introduction of Fleet Server, endpoints will no longer have credentials nor necessarily a route to Kibana.  As such, these artifacts must be served from Fleet Server instead.

Artifacts were previously stored in saved objects.  There is ongoing work in Kibana to move the records from saved objects into a dedicated .fleet-artifacts indice.  Fleet-Server and Kibana have agreed on a [schema](https://github.com/elastic/kibana/issues/90513#issuecomment-783684463).  In addition, there is an agreed upon URL format that Kibana will employ when generating a schema:

```
api/fleet/artifacts/:id/:sha2
```

This structure is consistent with the already existing fleet [api](https://github.com/elastic/fleet-server/blob/master/cmd/fleet/router.go#L12)  structure.

Artifact records will be immutable.  This will allow forward caching in a Fleet Server without concern for stale records.

Implementation is straight forward.  A new policy will roll out to the agents, which in turn will trickle down to endpoint.  Endpoint will interpret the policy and detect that there are new artifacts available.  Relative URL's of the above form will be included in the schema.  Endpoint will open HTTPS connection to the Fleet Server and do a GET on the relative URL.  The Fleet Server will authenticate the API Key presented by the agent, do a lookup on the sha2 in its local cache, and return if has payload.   If there is a local cache miss, the Fleet Server will do a search in .fleet-artifacts indice and retrieve the artifact record if available.  Record will be cached on hit and returned to requestor.  Artifact records will expire out of the cache in 24 hours.

Note there is currently no additional authorization on the artifact.  This means that any endpoint with a valid API Key will be able to pull any artifact from the system, assuming the sha2 is known.  This may be addressed at later point; and is commented in the code.

To prevent unnecessary round trips to Elastic Search for a specific sha2, a throttle has been implemented that limits the number of requests for a specific sha2.  In addition, there is a limit to the overall number of parallel requests to Elastic Search.  This will not defeat a DDOS attack on random sha2, but will limit the damage to the Fleet Server and Elastic Search.

The initial implementation uses the in memory cache for all artifacts.  As the artifacts grow in size, this may become an issue, causing thrashing on the cache.  There is a TODO to spill large artifacts to disk storage in the cache code to limit the memory overhead.

## Why is it important?

Endpoint requires this functionality for GA.

## Checklist

~~- [ ] I have made corresponding change to the default configuration files~~
~~- [ ] I have added an entry in `CHANGELOG.next.asciidoc` or `CHANGELOG-developer.next.asciidoc`.~~

## Author's Checklist

- [ ]

## Related issues

- https://github.com/elastic/kibana/issues/90513

### DIFF
--- a/internal/pkg/dl/artifact.go
+++ b/internal/pkg/dl/artifact.go
@@ -7,40 +7,42 @@ package dl
 import (
 	"context"
 	"encoding/json"
-	"errors"
+
+	"github.com/rs/zerolog/log"
 
 	"github.com/elastic/fleet-server/v7/internal/pkg/bulk"
 	"github.com/elastic/fleet-server/v7/internal/pkg/dsl"
 	"github.com/elastic/fleet-server/v7/internal/pkg/model"
 )
 
-const (
-	artifactsIndexName = ".fleet-artifacts"
-)
-
 var (
-	QueryArtifactBySha2 = prepareQueryArtifactBySha2()
-	ErrConflict         = errors.New("Fail multiple artifacts for the same sha256")
+	QueryArtifactTmpl = prepareQueryArtifact()
 )
 
-func prepareQueryArtifactBySha2() *dsl.Tmpl {
+func prepareQueryArtifact() *dsl.Tmpl {
 	root := dsl.NewRoot()
 	tmpl := dsl.NewTmpl()
 
-	root.Query().Bool().Filter().Term(FieldEncodedSha256, tmpl.Bind(FieldEncodedSha256), nil)
+	must := root.Query().Bool().Must()
+	must.Term(FieldDecodedSha256, tmpl.Bind(FieldDecodedSha256), nil)
+	must.Term(FieldIdentifier, tmpl.Bind(FieldIdentifier), nil)
 	tmpl.MustResolve(root)
 	return tmpl
 }
 
-func FindArtifactBySha256(ctx context.Context, bulker bulk.Bulk, sha2 string) (*model.Artifact, error) {
+func FindArtifact(ctx context.Context, bulker bulk.Bulk, ident, sha2 string) (*model.Artifact, error) {
 
-	res, err := SearchWithOneParam(
+	params := map[string]interface{}{
+		FieldDecodedSha256: sha2,
+		FieldIdentifier:    ident,
+	}
+
+	res, err := Search(
 		ctx,
 		bulker,
-		QueryArtifactBySha2,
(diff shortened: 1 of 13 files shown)


aleksmaus (approved):
Looks good 👍

internal/pkg/dl/artifact.go
```diff
@@ -0,0 +1,65 @@
...
+
+	if err != nil {
+		return nil, err
+	}
+
+	if len(res.Hits) == 0 {
+		return nil, ErrNotFound
+	}
+
+	if len(res.Hits) > 1 {
```
ruflin, line 54: If this happens, I assume it means there is a bug in the Kibana code?

scunningham: Yes.  The sha2 is a primary key.  We really do not want more than one sha2 record in the index; it becomes unclear which is the correct record to pull.
But as I think about it, it could in theory happen due to multiple Kibana's trying to create the same artifact in the same way at the same time.  This is due to not using the sha2 to derive the documentId.  In addition, could be the case that N artifacts have the same body but are of different types, causing a collision.

Will work with @paul-tavares  on how to best address.

scunningham: Ok, after talking with Paul, will make a couple of changes:

1) Change the Fleet-Server code do a term query based on identity+sha2, instead of just sha2 to avoid the case where there are two records with the same payload, but different identities.
2) Tolerate record dupes in the Fleet Server.  We will log when we see more than one, pick the first and continue.
3) Validate the sha2 in the Fleet Server.  This is just good hygiene.
4) Consider using the sha2+identity to derive the document id in Kibana.  This will avoid race conditions across Kibanas as well as dupes.

ruflin: SGTM. I assume Kibana takes care of cleaning up "old" artifacts. We should make sure when artifacts are deleted that also all the dupes are cleaned up.

scunningham: Yep.  Spoke with @paul-tavares about this.  It is a bit complicated given that multiple policies will point to the same artifact records.  The plan is to implement a garbage collector that runs every N hours, interprets the current policies, and removes any artifacts that are no longer referenced.

PR #131, opened 2021-03-12
Add API status endpoint

blakerouse:
## What does this PR do?

Adds an `/api/status` endpoint to verify that Fleet Server is running properly. This is also used by the Elastic Agent that it can communicate with a Fleet Server.

## Why is it important?

Elastic Agent verifies that it can communicate with a Fleet Server when a policy included an updated connection URL to a Fleet Server. Without this endpoint present then Elastic Agent will not accept and apply the policy because it cannot verify that the Fleet Server it needs to communicate with is correct.

This mirrors the `/api/status` URL by Kibana, so nothing needs to change in Elastic Agent. Elastic Agent can switch from a Kibana to Fleet Server without any custom logic.

## Checklist

### DIFF
--- a/cmd/fleet/handleStatus.go
+++ b/cmd/fleet/handleStatus.go
@@ -17,9 +17,8 @@ import (
 func (rt Router) handleStatus(w http.ResponseWriter, _ *http.Request, _ httprouter.Params) {
 	status := rt.sm.Status()
 	resp := StatusResponse{
-		Name:    "fleet-server",
-		Version: rt.ver,
-		Status:  status.String(),
+		Name:   "fleet-server",
+		Status: status.String(),
 	}
 
 	data, err := json.Marshal(&resp)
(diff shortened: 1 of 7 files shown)


cmd/fleet/handleStatus.go
```diff
@@ -0,0 +1,43 @@
...
+	"encoding/json"
+	"net/http"
+
+	"github.com/elastic/elastic-agent-client/v7/pkg/proto"
+	"github.com/julienschmidt/httprouter"
+	"github.com/rs/zerolog/log"
+)
+
+func (rt Router) handleStatus(w http.ResponseWriter, _ *http.Request, _ httprouter.Params) {
+	status := rt.sm.Status()
```
scunningham, line 18: This appears to be unauthorized if I am reading this correctly.  We should avoid leaking the fleet server version in an unauthorized API. It gives an attacker knowledge about which software is executing, which is problematic if we have a vulnerability somewhere.

Is this version necessary here?

blakerouse: No its not, we can remove it. But I think Kibana expose version unauthorized as well.

scunningham: We should remove.  Kibana shouldn't be doing that either.  

FWIW, we had the same issue at Endgame, the web app was designed to expose too much pre-auth that provided a reliable signature for an attacker.

blakerouse:
@scunningham Removed the version from the endpoint.

PR #133, opened 2021-03-16
Cleanup and refactoring

ruflin:
I was going through the code base to get more familiar with it. Along the way I made some small changes / cleanups that I stumbled over. There are also parts where I added small tests which are more for me to understand the code on what it does. Here is the list:

* Remove savedObjectKey function as it is not needed anymore
 * add config options for monitoring to the default configuration for documentation purpose
* move the creation of the nowString to where it is actually used. Seems it is used only once.
* add test for convertActions to better understand the format on my end
* Add note to missing parts
* update t.Fatal should not be used in test go routine
* add basic token test
* add more entries to the config file about available config options
* remove unused env package
* remove not used `missing.txt` file.
* remove dsl readme and filed #136 instead


cmd/fleet/missing.txt
```diff
@@ -1,3 +1,5 @@
+# TODO: Can we convert this to Github issues, still needed?
```
ruflin, line 1: @ph Seems you created this file. I think we should convert these to gibhub issues. We have stats, what about the others?

ph: I think this come from the original directory when I've closed @scunningham's repository. 
I am not sure what "run as admin; doesn't drop creds" mean? is that dropping privileges?

scunningham: A bit cryptic on my part.  That was a note to myself that there is no reason for the fleet server to run with Elastic admin privileges.  We are addressing that with service accounts.

scunningham (approved):
Can remove the internal env package as well. No longer used.

ruflin:
Unused env package removed.

cmd/fleet/handleEnroll.go
```diff
@@ -347,6 +346,7 @@ func decodeEnrollRequest(data io.Reader) (*EnrollRequest, error) {
 
 	// Validate
 	switch req.Type {
+	// TODO: Should these be converted to constant? Need to be kept in sync with Kibana?
```
ruflin, line 349: @nchaulet @blakerouse Thoughts?

cmd/fleet/missing.txt
ruflin: I removed it.

ruflin:
Marked as ready for review. No real changes in this PR. More similar PR's might follow when I dig further into the code.

cmd/fleet/handleEnroll.go
blakerouse: @nchaulet Is this even used?

blakerouse (approved):
Clean ups look good.

cmd/fleet/handleEnroll.go
nchaulet: No it's not probably no reason to keep it :)

ruflin: Not sure what the paths forward here is. Can the validation part be removed?  I assume the decode we still need.

nchaulet: I think we could make the `type` field optional in Kibana (we already do not use it), remove it from the agent, and remove it from fleet server.

ruflin: I did not touch it for now and merged the PR. It sounds removing it might break things in Agent or Fleet.

Issue #101, opened 2021-02-23
[Discuss] Indexing permissions as part of the Elastic Agent policy


ruflin:
Currently all Elastic Agents enrolled into the fleet-server get the same permissions. To cover all the use cases these are the most permissive permissions. This proposal is to reduce the permissions given to an Elastic Agent based on the the policy. With this, each Elastic Agent would receive an API Key with the minimal permissions needed to get its job done.

#  Elastic Agent Policy contains permissions block

The fleet-server creates the API Keys for the Elastic Agents. Which permissions an Elastic Agent requires is based on the content of the policy. Because of this the policy should contain a section with the permissions it requires. This section could look similar to the following ([inspired by Elasticsearch API Key permissions](https://www.elastic.co/guide/en/elasticsearch/reference/current/security-api-create-api-key.html#security-api-create-api-key-example)):

```
policy: A
privileges: [
  {
    "names": ["logs-*", "metrics-*"],
    "privileges": ["create_doc"]
  }
]
```

The above would give `create_doc` permissions for the logs-* and metrics-* data streams. If an Elastic Agent is enrolled for the policy `A`, an Elasticsearch output API Key would be created with only the above permissions and added to the policy.

The above model also works for more complex cases. Lets assume we have a case where only 2 indices should be allowed to be written to and one index should have read permissions. This could look as following:

```
policy: B
permissions: [
  {
    "names": ["logs-nginx.access-default", "logs-nginx.error-default"],
    "privileges": ["create_doc"]
  },
  {
    "names": ["state-docs"],
    "privileges": ["create_doc", "read"]
  }
]
```

If an Elastic Agent is enrolled for policy `B`, the permissions to write to the nginx indices is given and read from the `state-docs` index.

In general, this model can be used to be as permissive or restrive as needed based on the Elasticearch permission model. The limitation on the maximum permissions that can be given to an Elastic Agent is the permissions the fleet-server user itself has.

# Change of policy

A change to a policy can mean the permission required on the Elastic Agent changes. For example at first only nginx access logs were collected but now also error logs. This means additional privileges for the error log data stream are required. The fleet-server must be able to hand out new API Keys with increased / reduced permissions in case of policy changes. In addition, old API Keys have to be invalidated.

# Same permissions for all processes per Elastic Agent

[...]

PR #169, opened 2021-03-25
Indexing permissions as part of the Elastic Agent policy

aleksmaus:
## What does this PR do?

Implements the Indexing permissions as part of the Elastic Agent policy
https://github.com/elastic/fleet-server/issues/101

1. Adds a new field ```policy_output_permissions_hash``` into the Agent document that stores the hash of ```output_permissions``` from the policy.
2. Retrieves the policy from .fleet-policies upon the agent enrollment and uses the ```output_permissions``` ```default``` namespace in order to create the role descriptors for the output API key generation.
3. Checks the ```output_permissions``` upon every checking. If the hash doesn't match with the stored in the agent document the new key is generated.

This PR:
1. doesn't invalidate the old API key see the comment for the next iteration https://github.com/elastic/fleet-server/issues/101#issuecomment-806286176
2. is not fully validated end-to-end with kibana for the ```output_permissions``` changes, pending on further kibana changes.

Should only be merged after the corresponding kibana changes:
https://github.com/elastic/kibana/pull/94591

## Why is it important?

Addresses: https://github.com/elastic/fleet-server/issues/101

## Checklist

## Related issues

- Addresses https://github.com/elastic/fleet-server/issues/101

## Screenshots

Screen Shot 2021-03-24 at 6 40 40 PM


internal/pkg/policy/output_permissions.go
```diff
@@ -0,0 +1,160 @@
...
+		}
+
+		idx := make([]interface{}, 0, len(permissions))
+
+		for _, permission := range permissions {
+			idx = append(idx, permission)
+		}
+
+		m := make(smap.Map)
+		m["index"] = idx
```
ruflin, line 94: I hope we can get rid of this bit as it is planned that Kibana will add the ` index` part to the output so we don't need to do the additional processing here.

aleksmaus: Yeah, Alejandro reached out about the upcoming change. Might add more things here then.

internal/pkg/dl/policies.go
```diff
@@ -79,3 +90,18 @@ func CreatePolicy(ctx context.Context, bulker bulk.Bulk, policy model.Policy, op
 	}
 	return bulker.Create(ctx, o.indexName, "", data)
 }
+
+// FindPolicyByID find policy by ID
+func FindPolicyByID(ctx context.Context, bulker bulk.Bulk, policyID string) (policy model.Policy, err error) {
```
blakerouse, line 95: The `policy.Monitor` keeps the latest revision of each policy in memory. It would be better to fetch the policy from there instead of going to elasticsearch, as fleet-server already has it cached in memory.

blakerouse (approved):
Overall this change looks good. See my inline comment about fetching the policy from elasticsearch.

internal/pkg/dl/policies.go
aleksmaus: Sounds good, was not aware, will update.

aleksmaus:
Removed the output key generation upon enrollment. Not this is driven by the policy update pushes only, one code path.
Updated to the latest kibana changes, removed output permissions to the role descriptors transformation.

ruflin:
@aleksmaus I see this got merged but https://github.com/elastic/kibana/pull/94591 is still open. How is the current behaviour?

PR #171, opened 2021-03-25
Increase the monitor throughput: increase the default fetch size, make it configurable.

aleksmaus:
## What does this PR do?

Increases the monitor throughput:

* Increase the monitor fetch size to 1000 by default, the max number of documents fetched at once.
* Makes the monitor fetch size configurable.
* Decreased the pause between fetches from 50ms to 10ms.

## Why is it important?

Tuning of the throughput.

## Checklist

### DIFF
--- a/internal/pkg/config/config.go
+++ b/internal/pkg/config/config.go
@@ -27,8 +27,6 @@ type Config struct {
 	Inputs  []Input `config:"inputs"`
 	Logging Logging `config:"logging"`
 	HTTP    HTTP    `config:"http"`
-	Cache   Cache   `config:"cache"`
-	Monitor Monitor `config:"monitor"`
 }
 
 // InitDefaults initializes the defaults for the configuration.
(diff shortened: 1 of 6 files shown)


internal/pkg/config/config.go
```diff
@@ -27,13 +27,17 @@ type Config struct {
 	Inputs  []Input `config:"inputs"`
 	Logging Logging `config:"logging"`
 	HTTP    HTTP    `config:"http"`
+	Cache   Cache   `config:"cache"`
+	Monitor Monitor `config:"monitor"`
```
blakerouse, line 31: This also needs to be added under the input and not at this top-level.

blakerouse (changes requested):
The location of the configuration options needs to be adjusted.

aleksmaus:
Updated, moved the monitor config into input config.

blakerouse (approved):
Thanks for the fixes.

PR #180, opened 2021-03-29
Fix issue with rolling out policies from the coordinator.

### LINKED ISSUE #176: [Fleet] creating a policy without integration break fleet server
## Description 

Blocking https://github.com/elastic/beats/issues/24725

When creating a policy without integrations, looks like Fleet server is in a stuck phase where it's not possible to create new policy change, new policy with integrations.

## How to reproduce

1. create a policy without any integrations
2. Try to assign an agent to that policy or to trigger a policy change of a previous policy nothings happen.

blakerouse:
## What does this PR do?

There was an issue where it was possible for the policy coordinator to get stuck because the leader election happened at the same time the coordinator was starting. It was possible that pushing onto the coordinator channel in `Run` would block forever because `Update` was called before the `Run` could start the `for` loop. This removes that logic and removes the requirement of a buffered channel.

There was also an issue where the `QueryLatestPolicies` would only return the first 10 policies. If you had more than 10 policies the 10+ where never picked up by Fleet Server. This increases the size to the maximum amount of 10000.

## Why is it important?

So policies will pass through the coordinator and be rolled out to Elastic Agents.

## Checklist

## Related issues

- Closes #176

### DIFF
--- a/internal/pkg/coordinator/v0.go
+++ b/internal/pkg/coordinator/v0.go
@@ -76,12 +76,7 @@ func (c *coordinatorZeroT) updatePolicy(p model.Policy) error {
 	if err != nil {
 		return err
 	}
-	if p.CoordinatorIdx == 0 {
-		p.CoordinatorIdx = 1
-		p.Data = newData
-		c.policy = p
-		c.out <- p
-	} else if string(newData) != string(p.Data) {
+	if p.CoordinatorIdx == 0 || string(newData) != string(p.Data) {
 		p.CoordinatorIdx += 1
 		p.Data = newData
 		c.policy = p
--- a/internal/pkg/coordinator/v0.go
+++ b/internal/pkg/coordinator/v0.go
@@ -28,7 +28,7 @@ func NewCoordinatorZero(policy model.Policy) (Coordinator, error) {
 	return &coordinatorZeroT{
 		log:    log.With().Str("ctx", "coordinator v0").Str("policyId", policy.PolicyId).Logger(),
 		policy: policy,
-		in:     make(chan model.Policy, 1),
+		in:     make(chan model.Policy),
 		out:    make(chan model.Policy),
 	}, nil
 }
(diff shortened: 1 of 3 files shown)


internal/pkg/coordinator/v0.go
```diff
@@ -77,7 +70,27 @@ func (c *coordinatorZeroT) Output() <-chan model.Policy {
...
+		p.CoordinatorIdx = 1
+		p.Data = newData
+		c.policy = p
+		c.out <- p
+	} else if string(newData) != string(p.Data) {
+		p.CoordinatorIdx += 1
+		p.Data = newData
+		c.policy = p
+		c.out <- p
+	}
```
aleksmaus, line 89: is this the same as:
```
	if p.CoordinatorIdx == 0 || string(newData) != string(p.Data) {
		p.CoordinatorIdx += 1
		p.Data = newData
		c.policy = p
		c.out <- p
	}
```
?

aleksmaus (approved):
couple of questions, otherwise looks good

internal/pkg/coordinator/v0.go
blakerouse: yes updated

internal/pkg/dl/policies.go
```diff
@@ -27,7 +27,7 @@ func prepareQueryLatestPolicies() []byte {
 	root := dsl.NewRoot()
 	root.Size(0)
 	policyId := root.Aggs().Agg(FieldPolicyId)
-	policyId.Terms("field", FieldPolicyId, nil)
+	policyId.Terms("field", FieldPolicyId, nil).Size(10000)
```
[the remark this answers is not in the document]
blakerouse: We could, even a 100 policies is a lot so I just set it at the maximum at the moment.

ruflin:
Would be good to have some tests for these cases.

PR #205, opened 2021-04-07
Update graceful handling for missing indices to accommodate the new fleet system indices plugin

aleksmaus:
## What does this PR do?

Graceful handling for the case where the indices are not found.

Tested full stack with ES system indices plugin PR https://github.com/elastic/elasticsearch/pull/70689
and kibana update for this https://github.com/elastic/kibana/pull/96338 working end to end with osquerybeat.
Found issues where fleet server doesn't handle the case where the index doesn't exists.

The ES system indices plugin no longer creates the indices. The indices are now created when the first document is pushed, and the fleet server doesn't handle the missing index case gracefully (as some point before it did).

This change can be merged before the related tickets.

## Why is it important?

This is important in order to adopt the fleet system indices plugin.

## Checklist

### DIFF
--- a/internal/pkg/dl/actions.go
+++ b/internal/pkg/dl/actions.go
@@ -6,9 +6,11 @@ package dl
 
 import (
 	"context"
+	"errors"
 
 	"github.com/elastic/fleet-server/v7/internal/pkg/bulk"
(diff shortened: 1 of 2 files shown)


scunningham (approved):
Should we log something?

PR #207, opened 2021-04-07
Propagate enrollment_token to agent during bootstrap.

blakerouse:
## What does this PR do?

This propagates the enrollment token that the Elastic Agent should use to enroll itself into the bootstrapped Fleet Server.

## Why is it important?

So the bootstrapping Elastic Agent does not need to know the enrollment token. Being that a service token or user/pass is provided there is no reason the Elastic Agent even needs the enrollment token. This removes that requirement from Elastic Agent as it can get this information from Fleet Server.

## Checklist


cmd/fleet/handleCheckin.go
```diff
@@ -255,12 +256,19 @@ func (ct *CheckinT) resolveSeqNo(ctx context.Context, req CheckinRequest, agent
...
 
-	return dl.FindActions(ctx, ct.bulker, dl.QueryAgentActions, map[string]interface{}{
+	actions, err := dl.FindActions(ctx, ct.bulker, dl.QueryAgentActions, map[string]interface{}{
 		dl.FieldSeqNo:      seqno.Get(0),
 		dl.FieldMaxSeqNo:   ct.gcp.GetCheckpoint(),
 		dl.FieldExpiration: now,
 		dl.FieldAgents:     []string{agentId},
 	})
+	if err != nil {
+		if errors.Is(err, es.ErrIndexNotFound) {
```
aleksmaus, line 309: why? this check is already in findActions
https://github.com/elastic/fleet-server/blob/master/internal/pkg/dl/actions.go#L89

internal/pkg/coordinator/monitor.go
```diff
@@ -202,11 +202,9 @@ func (m *monitorT) ensureLeadership(ctx context.Context) error {
 		}
 		leaders, err = dl.SearchPolicyLeaders(ctx, m.bulker, ids, dl.WithIndexName(m.leadersIndex))
 		if err != nil {
-			if errors.Is(err, es.ErrIndexNotFound) {
-				m.log.Debug().Str("index", m.leadersIndex).Msg(es.ErrIndexNotFound.Error())
-				return nil
+			if !errors.Is(err, es.ErrIndexNotFound) {
+				return err
```
aleksmaus, line 206: don't want to log anymore if the index is missing?

aleksmaus (approved):
one comment in fetchAgentPendingActions, otherwise lgtm

internal/pkg/coordinator/monitor.go
blakerouse: No because this monitor creates a document, that will create the index. So no need to log it.

cmd/fleet/handleCheckin.go
blakerouse: hmm maybe I had an outdated master when I was testing, will remove and test before merging

PR #215, opened 2021-04-08
Preprocess policy on new revision to pre-calculate hashes and minimize memory allocations.

scunningham:
## What does this PR do?

Rolling out a new policy caused a lot of memory and cpu thrashing. This PR moves much of the expensive calculations to happen only once on a policy change.

## Why is it important?

Fleet Server is memory bursty on policy change.  This is an issue in low memory environments.

## Checklist

- [ x] I have added an entry in `CHANGELOG.next.asciidoc` or `CHANGELOG-developer.next.asciidoc`.


cmd/fleet/handleCheckin.go
```diff
@@ -429,7 +566,7 @@ func parsePolicy(ctx context.Context, bulker bulk.Bulk, agentId string, p model.
 
 	return &resp, nil
 }
-
+*/
```
blakerouse, line 569: Should this just be removed? Why leave it in and commented out?

internal/pkg/policy/parsed_policy.go
```diff
@@ -0,0 +1,82 @@
...
+
+type RoleMapT map[string]RoleT
+
+type ParsedPolicy struct {
+	Policy model.Policy
+	Fields map[string]json.RawMessage
+	Roles  RoleMapT
+}
+
+func NewParsedPolicy(p model.Policy) (*ParsedPolicy, error) {
```
blakerouse, line 27: Should add unit tests for this function so we can ensure that a `model.Policy` converts to a `ParsedPolicy` correctly.

blakerouse (commented):
Overall looks good, like the idea of pre-parsing the policy so its not done per Agent.

Should add some unit tests around the parsed policy.

aleksmaus (approved):
👍

cmd/fleet/handleCheckin.go
scunningham: Oversight.  Removed it; thanks.

internal/pkg/policy/parsed_policy.go
scunningham: Will do.

PR #226, opened 2021-04-13
Add ability to communicate with elasticsearch using a service token

blakerouse:
## What does this PR do?

Adds the ability to communicate with elasticsearch using a service token.

Set the `service_token` key in the `output` defined in `fleet-server.yml`.

## Why is it important?

So Fleet Server can communicate using a service token versus using a username and password.

## Checklist


go.mod
```diff
@@ -29,5 +29,6 @@ require (
 replace (
 	github.com/Shopify/sarama => github.com/elastic/sarama v1.19.1-0.20200629123429-0e7b69039eec
 	github.com/dop251/goja => github.com/andrewkroh/goja v0.0.0-20190128172624-dd2ac4456e20
+	github.com/elastic/go-elasticsearch/v8 => github.com/blakerouse/go-elasticsearch/v8 v8.0.0-20210413134915-cee76d43c5cf
```
ruflin, line 32: @blakerouse Should we get it in directly with this "fork"? I think it is unlikely that we can use a stable release soon enough.

blakerouse: @ruflin really? I just pushed this so others can test it. I didn't think getting it merged into go-elasticsearch was going to be a big deal. Its a very small change.

ruflin: Looks like I was wrong and it is already in!

blakerouse: Yep, updated to use new upstream version.

aleksmaus (approved):

PR #252, opened 2021-04-20
Bump elastic stack version automation

v1v:
## What does this PR do?

Script to bump the versions:

```
$ /bump-stack-version.sh 8.0.0-3beeb19e
Update stack with version 8.0.0-3beeb19e
```

And enable automerge with `mergify`

This PR is one part of the required changes to enable the automation. 

## Why is it important?

Handle the stack versions on a PR basis to ensure breaking changes are handled correctly. Unfortunately we cannot use `dependabot` since those versions are handled differently.

## Issues

Similar to https://github.com/elastic/beats/pull/25148

## Test

I tested this PR in my local CI environment and produced -> https://github.com/v1v/fleet-server/pull/8

### DIFF
--- a/.ci/bump-stack-version.sh
+++ b/.ci/bump-stack-version.sh
@@ -1,0 +1,31 @@
+#!/usr/bin/env bash
+#
+# Given the stack version this script will bump the version.
+#
+# This script is executed by the automation we are putting in place
+# and it requires the git add/commit commands.
+#
+set -euo pipefail
+MSG="parameter missing."
+VERSION=${1:?$MSG}
+
+OS=$(uname -s| tr '[:upper:]' '[:lower:]')
+
+if [ "${OS}" == "darwin" ] ; then
+	SED="sed -i .bck"
+else
+	SED="sed -i"
+fi
+
+echo "Update stack with version ${VERSION}"
+${SED} -E -e "s#(ELASTICSEARCH_VERSION)=[0-9]+\.[0-9]+\.[0-9]+(-[a-f0-9]{8})?#\1=${VERSION}#g" dev-tools/integration/.env
+
+exit 0
+echo "Commit changes"
+git config user.email
+git checkout -b "update-stack-version-$(date "+%Y%m%d%H%M%S")"
+git add dev-tools/integration/.env
+git commit -m "bump stack version ${VERSION}"
+git --no-pager log -1
+
+echo "You can now push and create a Pull Request"
--- a/.ci/bump-stack-version.sh
+++ b/.ci/bump-stack-version.sh
@@ -20,7 +20,6 @@ fi
 echo "Update stack with version ${VERSION}"
 ${SED} -E -e "s#(ELASTICSEARCH_VERSION)=[0-9]+\.[0-9]+\.[0-9]+(-[a-f0-9]{8})?#\1=${VERSION}#g" dev-tools/integration/.env
 
-exit 0
 echo "Commit changes"
 git config user.email
 git checkout -b "update-stack-version-$(date "+%Y%m%d%H%M%S")"
--- a/.ci/bump-stack-version.sh
+++ b/.ci/bump-stack-version.sh
@@ -28,7 +28,6 @@ echo "Update stack with version ${VERSION}"
 ${SED} -E -e "s#(ELASTICSEARCH_VERSION)=[0-9]+\.[0-9]+\.[0-9]+(-[a-f0-9]{8})?#\1=${VERSION}#g" dev-tools/integration/.env
 
 echo "Commit changes"
-git config user.email
 if [ "$CREATE_BRANCH" = "true" ]; then
 	git checkout -b "update-stack-version-$(date "+%Y%m%d%H%M%S")"
 else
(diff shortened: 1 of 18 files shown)


.ci/bump-stack-version.sh
```diff
@@ -0,0 +1,31 @@
...
+if [ "${OS}" == "darwin" ] ; then
+	SED="sed -i .bck"
+else
+	SED="sed -i"
+fi
+
+echo "Update stack with version ${VERSION}"
+${SED} -E -e "s#(ELASTICSEARCH_VERSION)=[0-9]+\.[0-9]+\.[0-9]+(-[a-f0-9]{8})?#\1=${VERSION}#g" dev-tools/integration/.env
+
+exit 0
```
v1v, line 23:
```suggestion
```

ruflin:
Are these builds coming form the "official" snapshot builds? How long are these around? I remember in the past some of these got cleaned up after a few days / weeks.

I'm also challenging if we should test master against a specific commit instead of the most recent snapshot. It is better for the builds itself to test against a specific version. But as we are unlikely to update it often, it also means we don't uncover a breaking change in ES immediately and maybe only a few weeks later. Even though a red build is not great for the fleet-server team itself, I think overall it is much better to uncover these issues early. What would be nice is if we would have an option to select a specific hash if needed in case things break.

In contrast for the stable branches, ++ on having a fixed version.

v1v:
Thanks @ruflin , I just put you in the loop in one of the internal github issues. But for clarity I'll answer here some of those questions.

> Are these builds coming form the "official" snapshot builds? How long are these around? I remember in the past some of these got cleaned up after a few days / weeks.

Yes, the official snapshot builds.
Those builds are kept forever for the time being, I already asked the infra team to clarify this, but so far, I was able to pull a docker image that was built 6 months ago.


> I'm also challenging if we should test master against a specific commit instead of the most recent snapshot. It is better for the builds itself to test against a specific version. But as we are unlikely to update it often, it also means we don't uncover a breaking change in ES immediately and maybe only a few weeks later. Even though a red build is not great for the fleet-server team itself, I think overall it is much better to uncover these issues early. What would be nice is if we would have an option to select a specific hash if needed in case things break.

The automation can also do a couple of things once the PR with the bump has been created.
* merge automatically if the CI passes.
* reuse the same PR to bump the versions in case the build did not successfully run in the CI.
* use one PR per bump to keep track.

The current approach will run on a daily basis.

.ci/bump-stack-version.sh
```diff
@@ -0,0 +1,30 @@
...
+else
+	SED="sed -i"
+fi
+
+echo "Update stack with version ${VERSION}"
+${SED} -E -e "s#(ELASTICSEARCH_VERSION)=[0-9]+\.[0-9]+\.[0-9]+(-[a-f0-9]{8})?#\1=${VERSION}#g" dev-tools/integration/.env
+
+echo "Commit changes"
+git config user.email
+git checkout -b "update-stack-version-$(date "+%Y%m%d%H%M%S")"
```
v1v, line 25: In order to reuse an existing PR then the branch creating should only happen if set.

In other words,

There is a requirement to support:
1) Reusing an existing PR if there is already an open PR with a previous bump.
2) Creating a new PR if the above is not true or it's configured to create always independently.

urso:
We should also add depencencies in the `go.mod` files. E.g. go-elasticsearch and beats dependencies should be updated, so to track 'master'.

ruflin:
@urso I assume we can do that as a follow up to tackle this in small steps? Right now these are also not tied together. Long term, ++ on getting automated updates for these too, maybe even in separate PRs to not tied problems together.

.ci/bump-stack-version.sh
```diff
@@ -0,0 +1,41 @@
...
+	SED="sed -i .bck"
+else
+	SED="sed -i"
+fi
+
+echo "Update stack with version ${VERSION}"
+${SED} -E -e "s#(ELASTICSEARCH_VERSION)=[0-9]+\.[0-9]+\.[0-9]+(-[a-f0-9]{8})?#\1=${VERSION}#g" dev-tools/integration/.env
+
+echo "Commit changes"
+git config user.email
```
v1v, line 31: I'll remove this since in my local environment it fails for some reason

```
[2021-04-23T09:04:17.319Z] + echo 'Commit changes'

[2021-04-23T09:04:17.319Z] Commit changes

[2021-04-23T09:04:17.319Z] + git config user.email

script returned exit code 1
```

v1v, line 31:
```suggestion
```

.ci/bump-stack-version.sh
```diff
@@ -0,0 +1,40 @@
...
+${SED} -E -e "s#(ELASTICSEARCH_VERSION)=[0-9]+\.[0-9]+\.[0-9]+(-[a-f0-9]{8})?#\1=${VERSION}#g" dev-tools/integration/.env
+
+echo "Commit changes"
+if [ "$CREATE_BRANCH" = "true" ]; then
+	git checkout -b "update-stack-version-$(date "+%Y%m%d%H%M%S")"
+else
+	echo "Branch creation disabled."
+fi
+git add dev-tools/integration/.env
+git commit -m "bump stack version ${VERSION}"
```
v1v, line 37:
```suggestion
git diff --staged --quiet || git commit -m "bump stack version ${VERSION}"
```

ruflin (approved):
LGTM. Lets try it out.

PR #254, opened 2021-04-20
Wait on index when polling global checkpoints

aleksmaus:
## What does this PR do?

Adopts the Elasticsearch polling API change that now handles the waiting on index if it doesn't exists.
https://github.com/elastic/elasticsearch/pull/71890

ATTN!!! This PR integration tests will fail until the Elasticsearch docker images are updated with the corresponding ES change. Merging it before the change will break fleet server for a lot of good people out there.

## Why is it important?

The Fleet Server would poll every 3 secs for each monitored index until it is created otherwise.
This change reduces the number of http requests the Fleet Server would have to do to detect the very document in the index and let Eleasticsearch handle the wait.

## Checklist

### DIFF
--- a/internal/pkg/es/fleet_global_checkpoints.go
+++ b/internal/pkg/es/fleet_global_checkpoints.go
@@ -47,6 +47,7 @@ type GlobalCheckpointsRequest struct {
 
 	Index          string
 	WaitForAdvance *bool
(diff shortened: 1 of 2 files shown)


scunningham (approved):
looks good.

PR #255, opened 2021-04-20
Add script to avoid race condition on updating policy metadata in age…

scunningham:
…nt record

## What does this PR do?

Works around a race condition where the policy that an agent is assigned to could change while updating the metadata related to the previous policy.

**NOTE** Not targeted for 7.13.

## Why is it important?

Without this change, an agent could end up with an inconsistent revision_id and/or coordinator_id on a newly assigned policy.

## Checklist

- [ x] I have added an entry in `CHANGELOG.next.asciidoc` or `CHANGELOG-developer.next.asciidoc`.

Assign a new policy to an agent.  Validate that the revision and the coordinator idx are set appropriately.

### DIFF
--- a/cmd/fleet/handleAck.go
+++ b/cmd/fleet/handleAck.go
@@ -5,11 +5,13 @@
 package fleet
 
 import (
+	"bytes"
 	"context"
 	"encoding/json"
 	"errors"
 	"io/ioutil"
 	"net/http"
+	"strconv"
 	"strings"
 	"time"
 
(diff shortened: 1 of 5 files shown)


cmd/fleet/handleAck.go
```diff
@@ -295,3 +272,41 @@ func _getAPIKeyIDs(agent *model.Agent) []string {
...
+	dl.FieldPolicyRevisionIdx +
+	` = params.rev;ctx._source.` +
+	dl.FieldPolicyCoordinatorIdx +
+	`= params.coord;ctx._source.` +
+	dl.FieldUpdatedAt +
+	` = params.ts;} else {ctx.op = \"noop\";}","params": {"id":"`
+
+func makeUpdatePolicyBody(policyId string, newRev, coordIdx int64) []byte {
+
+	var buf bytes.Buffer
```
aleksmaus, line 297: would the strings.Builder be faster?

aleksmaus (approved):

blakerouse (approved):
Looks good to me. Only question is why not for 7.13? This seems like a bug that really needs to be fixed in 7.13?

cmd/fleet/handleAck.go
scunningham: It is faster, but you get an extra copy because there's no Bytes() accessor.

PR #290, opened 2021-04-27
Report failure running under Elastic Agent

### LINKED ISSUE #235: [Windows and Linux ARM64]: Unable to install agent with Fleet Server URL.
**Kibana version:** 8.0 Snapshot Kibana cloud environment

**Host OS and Browser version:** Windows and Ubuntu ARM64, All

**Build Details:**
```
  [Windows]Artifact link used: https://snapshots.elastic.co/8.0.0-c711755c/downloads/beats/elastic-agent/elastic-agent-8.0.0-SNAPSHOT-windows-x86_64.zip
  [Linux]Artifact link used: https://snapshots.elastic.co/8.0.0-c711755c/downloads/beats/elastic-agent/elastic-agent-8.0.0-SNAPSHOT-linux-arm64.tar.gz
  Build: 41953
  Commit: cb3c4e3a212255a9e9b8c89e784e0e452b661233
```

**Preconditions:**
1. 8.0 Snapshot Kibana cloud environment should be available.
2. Fleet Server should be setup for Windows and Linux endpoints.

**Steps to reproduce:**
1. Login to Kibana environment.
2. Run install command for Agent installation:
```
.\elastic-agent.exe install -f --url=http://10.0.5.233:8220 --enrollment-token=VUJFWnozZ0I3Y3RGakhkRVRNcUI6ZDc1SXljaW9TaENndkYxNWc5ekxaQQ== --insecure
```
3. Observe repetitive "Remote server is not ready to accept connections, will retry in a moment" errors.

**Expected Result:**
User should be able to install agent with Fleet Server URL.

**Screenshots:**
Fleet Server Agent

blakerouse:
## What does this PR do?

When running under Elastic Agent the Fleet Server will report the failure to Elastic Agent instead of crashing.

## Why is it important?

So that the process is not in a constant restart loop when an error occurs.

## Checklist

## Related issues

- Closes #235

### DIFF
--- a/cmd/fleet/main.go
+++ b/cmd/fleet/main.go
@@ -257,7 +257,7 @@ func (a *AgentMode) Run(ctx context.Context) error {
 			err := a.srv.Run(srvCtx)
 			if err != nil && err != context.Canceled {
 				// report the status over the reporter (logs and reports)
-				reporter.Status(proto.StateObserved_FAILED, err.Error(), nil)
+				reporter.Status(proto.StateObserved_FAILED, fmt.Sprintf("Error: %s", err), nil)
 				if sleep.WithContext(srvCtx, kAgentModeRestartLoopDelay) == context.Canceled {
 					// context cancelled while sleeping
 					res <- err
--- a/cmd/fleet/main.go
+++ b/cmd/fleet/main.go
@@ -43,7 +43,7 @@ import (
 )
 
 const (
-	kAgentMode = "agent-mode"
+	kAgentMode                 = "agent-mode"
 	kAgentModeRestartLoopDelay = 2 * time.Second
 )
 
--- a/cmd/fleet/main.go
+++ b/cmd/fleet/main.go
@@ -11,6 +11,7 @@ import (
 	"os"
 	"runtime/debug"
 	"sync"
+	"time"
 
 	"github.com/elastic/go-ucfg"
 	"github.com/elastic/go-ucfg/yaml"
@@ -28,6 +29,7 @@ import (
 	"github.com/elastic/fleet-server/v7/internal/pkg/profile"
 	"github.com/elastic/fleet-server/v7/internal/pkg/reload"
 	"github.com/elastic/fleet-server/v7/internal/pkg/signal"
+	"github.com/elastic/fleet-server/v7/internal/pkg/sleep"
 	"github.com/elastic/fleet-server/v7/internal/pkg/status"
 	"github.com/elastic/fleet-server/v7/internal/pkg/ver"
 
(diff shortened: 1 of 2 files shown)


cmd/fleet/metrics.go
```diff
@@ -36,9 +36,12 @@ var (
...
 	registry := monitoring.GetNamespace("info").GetRegistry()
-	monitoring.NewString(registry, "version").Set(f.ver)
-	monitoring.NewString(registry, "name").Set("fleet-server")
-	metrics.SetupMetrics("fleet-server")
+	if registry.Get("version") == nil {
+		monitoring.NewString(registry, "version").Set(f.ver)
+	}
+	if registry.Get("name") == nil {
+		monitoring.NewString(registry, "name").Set("fleet-server")
+	}
```
blakerouse, line 44: This was required to be changed, because it cannot be called more than once per process. But the version is not know until it reaches this far.

If it is called more than once it will panic.

cmd/fleet/metrics.go
```diff
@@ -83,6 +86,7 @@ func (rt *routeStats) Register(registry *monitoring.Registry) {
 }
 
 func init() {
+	metrics.SetupMetrics("fleet-server")
```
blakerouse, line 89: This was required to be changed, because it cannot be called more than once per process.

If it is called more than once it will panic.

cmd/fleet/main.go
```diff
@@ -245,7 +248,21 @@ func (a *AgentMode) Run(ctx context.Context) error {
...
+	res := make(chan error)
+	go func() {
+		for {
+			err := a.srv.Run(srvCtx)
+			if err == nil || err == context.Canceled {
+				res <- err
+				return
+			}
+			// sleep some before calling Run again
+			sleep.WithContext(srvCtx, kAgentModeRestartLoopDelay)
```
aleksmaus, line 262: if the context is cancelled on `sleep`, should we bail out and not attempt another `srv.Run`?
maybe 
```
			err = sleep.WithContext(srvCtx, kAgentModeRestartLoopDelay)
			if err != nil {
                                 res <- err
				return
			}
```

aleksmaus: thinking out loud: should we rethink how we start http server for the next version maybe? 
currently it's tied the fleet server, if something failed to start the whole server is down for 2 seconds.
it would be more graceful if the `/api/status` route for example would still be available reflecting the status of the fleet server if it has issues starting

blakerouse: I just thought it made the code path simpler, calling `srv.Run` with a cancelled context should just make it return the error, so that is why I keep it this way. I can add the check if you want.

Yeah that probably would be nice to change that in 7.14.

aleksmaus (approved):

PR #324, opened 2021-05-04
Add error response to be JSON

### LINKED ISSUE #323: Fleet Server doesn't response with formatted error body
Elastic Agent expects the Fleet Server to respond with a JSON response when an error occurs. This allows the Elastic Agent to output an understandable error to a user when enrolling or in other cases.

The Fleet Server should respond with the following matching struct:

```
type errResp struct {
	StatusCode int    `json:"statusCode"`
	Error      string `json:"error"`
	Message    string `json:"message"`
}
```

blakerouse:
## What does this PR do?

Any time the API returns an error to an HTTP request it should be JSON so it can be decoded by the Elastic Agent.

## Why is it important?

So proper and easy to understand error messages are reported by the Elastic Agent.

## Checklist

## Related issues

- Closes #323

### DIFF
--- a/cmd/fleet/metrics.go
+++ b/cmd/fleet/metrics.go
@@ -101,29 +101,35 @@ func init() {
 }
 
 // Increment error metric, log and return code
-func (rt *routeStats) IncError(err error) (int, zerolog.Level) {
+func (rt *routeStats) IncError(err error) (int, string, zerolog.Level) {
 	lvl := zerolog.DebugLevel
 
 	incFail := true
 
 	var code int
+	var errStr string
 	switch err {
 	case ErrAgentNotFound:
+		errStr = "AgentNotFound"
 		code = http.StatusNotFound
 		lvl = zerolog.WarnLevel
 	case limit.ErrRateLimit:
+		errStr = "RateLimit"
 		code = http.StatusTooManyRequests
 		rt.rateLimit.Inc()
 		incFail = false
 	case limit.ErrMaxLimit:
+		errStr = "MaxLimit"
 		code = http.StatusTooManyRequests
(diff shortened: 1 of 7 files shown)


scunningham (changes requested):
You are leaking raw strings to an untrusted client.  You should not provide that much information to the client.  What is the reasoning here?

blakerouse:
@scunningham You talking about the `err.Error()`. It would be nice to give a reason, at the moment there is no information in the client.

Let's say it's a version mismatch, its unknown currently what the issue. We need to provide something.

scunningham:
It is generally bad practice to leak out the underlying raw errors on the interface, particularly to clients that should be considered untrusted.  It would be fine to translate well known errors into documented error strings, but since  the error could come out of any subsystem, this implementation could expose specifics.

A cleaner strategy would be to log verbose errors in the fleet server, translate known useful errors to pre-baked user facing error strings, and those that are not useful should be generic errors.  Customer, if have access, can refer to the fleet server log, which should contain metadata about the transaction (src port, ip, agent_id if available, etc).

blakerouse:
@scunningham How about now? Updated it to force the message text.

cmd/fleet/metrics.go
```diff
@@ -101,33 +101,48 @@ func init() {
...
 
 	incFail := true
 
 	var code int
 	var errStr string
+	var msgStr string
 	switch err {
 	case ErrAgentNotFound:
 		errStr = "AgentNotFound"
+		msgStr = "agent could not be cound"
```
ruflin, line 115:
```suggestion
		msgStr = "agent could not be found"
```

scunningham (approved):
thanks!

PR #375, opened 2021-05-19
Log version and commit at boot

scunningham:
## What does this PR do?

Adds the short form of the git commit id and the version to the boot message.

## Why is it important?

It remains difficult to know what specific version of code is installed in the cloud during beta cycles.  
This will allow the developer to examine the logs to figure out the specific version of the code used in the build.

## Checklist

## Logs

{"log.level":"info","version":"7.13.0","commit":"34e4a3c","pid":68004,"ppid":19730,"exe":"./bin/fleet-server","args":[],"@timestamp":"2021-05-19T09:52:25.125Z","message":"boot"}

### DIFF
--- a/cmd/fleet/main.go
+++ b/cmd/fleet/main.go
@@ -68,7 +68,26 @@ func makeCache(cfg *config.Config) (cache.Cache, error) {
 	return cache.New(cacheCfg)
 }
 
-func getRunCommand(version string) func(cmd *cobra.Command, args []string) error {
+func initLogger(cfg *config.Config, version, commit string) (*logger.Logger, error) {
+	l, err := logger.Init(cfg)
+	if err != nil {
+		return nil, err
+	}
+
+	log.Info().
+		Str("version", version).
+		Str("commit", commit).
+		Int("pid", os.Getpid()).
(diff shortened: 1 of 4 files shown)


cmd/fleet/main.go
```diff
@@ -68,7 +68,26 @@ func makeCache(cfg *config.Config) (cache.Cache, error) {
...
+		return nil, err
+	}
+
+	log.Info().
+		Str("version", version).
+		Str("commit", commit).
+		Int("pid", os.Getpid()).
+		Int("ppid", os.Getppid()).
+		Str("exe", os.Args[0]).
+		Strs("args", os.Args[1:]).
```
ruflin, line 83: Could these args contain confidential info?

scunningham: Question for @blakerouse .  The args are currently being logged with or without this change. We can remove them, but I am hesitant as they are very useful for debugging.

blakerouse: The args do not container any credentials in them. All credentials are passed over GRPC and the certificates for the mTLS for local GRPC is done through Stdin. I don't see any danger keeping the arguments, and they are not to have for debugging.

PR #420, opened 2021-06-02
Propagate checkin status to agent record

scunningham:
## What does this PR do?

Writes out the checkin status field to the last_checkin_status field in the agent record.  This is necessary to support UX changes.

## Why is it important?

UX needs it.

## Checklist

Setup local fleet server and point an agent at it.  You can validate field change using the dev tools in kibana.

## Screenshots

Screen Shot 2021-06-02 at 3 40 57 PM

### DIFF
--- a/cmd/fleet/handleCheckin.go
+++ b/cmd/fleet/handleCheckin.go
@@ -176,6 +176,15 @@ func (ct *CheckinT) _handleCheckin(w http.ResponseWriter, r *http.Request, id st
 		return err
 	}
 
+	log.Debug().
+		Str("agentId", id).
+		Str("reqId", reqId).
+		Str("status", req.Status).
+		Str("seqNo", seqno.String()).
+		RawJSON("meta", rawMeta).
+		Uint64("bodyCount", readCounter.Count()).
+		Msg("checkin start long poll")
+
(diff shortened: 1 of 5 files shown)


jen-huang:
@ph / @urso This seems like a good candidate to backport to 7.13, what do you think?

scunningham:
I wouldn't backport unless we were backporting the new UX as well.  Not sure we are buying ourselves much otherwise.

jen-huang:
@scunningham We already have code in Fleet that looks at `last_checkin_status` from pre-Fleet Server days, not just for new UX, so this seems like a regression with agent status reporting that we didn't catch in testing.

jfsiii (approved):
🚀 Thanks for getting this in so quickly. I ran this branch locally and, without any Kibana changes, it now shows an `Unhealthy` badge for an agent with `last_checkin_status` of `degraded`.

Screen Shot 2021-06-03 at 9 33 16 AM
Screen Shot 2021-06-03 at 9 33 28 AM

Without these changes, the UI marks the agents as `Healthy`

ph:
@scunningham based on the discussion from @jen-huang and the comment from @jfsiii I would backport this to 7.13.x too. If it's fine with you I will just add the label.

PR #436, opened 2021-06-07
Update go-version

michel-laterman:
## What does this PR do?

Update go version to 1.16.5

- Relates https://github.com/elastic/beats/issues/26182


urso (approved):

urso:
Looks like the go.mod is incomplete?

michel-laterman:
I'm not really sure what's going on, I can recreate the issue locally but the steps that go says to take to resolve the issue (`go mod download github.com/elastic/go-licenser go.elastic.co/go-licence-detector`) do not fix it

urso:
The recent problem was introduced with introducing `go get ...`. When doing go get you update the dependency you get. In that case the `go-licenser` was updated. `go get` is run in the current working directory, which is contains the fleet-server source. This motivates the go tool to update your dependencies unfortunately. This might update `go.sum` and `go.mod` as well. The newer `go-licenser` version also changes the format it seems, resulting in the huge diff on the notice file.

PR #443, opened 2021-06-08
Configure cache

scunningham:
## What does this PR do?

Makes the TTL values on the cache objects configurable.  Allows cache to be reconfigured without restarting the process.  Cache is dropped on reconfiguration.

Increate the TTL on the api key to 15m with optional jitter.   This is to work around latency issues at scale with validating API .  In addition, validate that 'enabled' flag on the agent during auth.   Between the 'enabled' flag, and the ability to drop the cache by reconfiguration in a break glass scenario, it seems minimize auth round trips to elastic.  This takes load off the fleet-server as well as elastic.

Also, cache the case when an api key is not valid.  This avoids roundtrip for an already disabled api key.

## Why is it important?

The server runs into issues at large scale while trying to validate many ApiKeys at once.  Particularly when a policy changes and there are many GET requests for artifacts. 

## Checklist

### DIFF
--- a/internal/pkg/cache/cache.go
+++ b/internal/pkg/cache/cache.go
@@ -6,25 +6,61 @@ package cache
 
 import (
 	"fmt"
+	"math/rand"
+	"sync"
 	"time"
 
 	"github.com/dgraph-io/ristretto"
+	"github.com/rs/zerolog"
 	"github.com/rs/zerolog/log"
 
 	"github.com/elastic/fleet-server/v7/internal/pkg/apikey"
 	"github.com/elastic/fleet-server/v7/internal/pkg/model"
 )
 
+type Cache interface {
+	Reconfigure(Config) error
+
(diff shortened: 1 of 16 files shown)


internal/pkg/cache/cache.go
```diff
@@ -87,26 +162,56 @@ func (c Cache) GetAction(id string) (model.Action, bool) {
...
-	ok := c.cache.SetWithTTL(scopedKey, key.Key, int64(cost), ttl)
+
+	// Use the valid key as the payload of the record;
+	// If caller has marked key as not enabled, use empty string.
+	val := key.Key
+	if !enabled {
+		val = ""
+	}
+
+	// If enabled, jitter allows us to randomize the expirtion of the artifact
```
aleksmaus, line 178: typo "expirtion"

aleksmaus (approved):
LGTM

PR #445, opened 2021-06-09
Add optional jitter to long poll

scunningham:
## What does this PR do?

Adds an optional jitter to the long poll, defined as 30s by default.

## Why is it important?

Adding a bit of jitter helps smooth out load over time at scale. The server performs better if the load is spread out avoiding spikes, particularly as we add policy rollout, etc.

## Checklist

## Screenshots

Test was run with 100k endpoints horde emulation, deploying at a rate of 20k per second.  This is a two hour snapshot of the rollout without the jitter:

jitter_off

Adding the 30s second jitter significantly smoothed out the load over time.

jitter_30s

### DIFF
--- a/NOTICE.txt
+++ b/NOTICE.txt
@@ -72,11 +72,11 @@ SOFTWARE.
 
 --------------------------------------------------------------------------------
 Dependency : github.com/dgraph-io/ristretto
-Version: v0.0.3
+Version: v0.1.0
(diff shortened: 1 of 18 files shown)


james-elastic (approved):
I think the default could be 42 :)

urso (approved):

PR #457, opened 2021-06-14
Remove indirect build dependency on go-license-detector.

scunningham:
## What does this PR do?

Tweaks the dependencies-report script to explicitly download required version of the go-license-detector tool and invoke it directly.   Previous version expected the dependencies to be included in the go module files that it was analyzing.  This broke when newer version of 'go tidy' more aggressively strips modules that are not part of the targeted program's dependency tree.

## Why is it important?

The release build process invokes the 'dependencies-report' script, which is currently broken.

## Checklist


michel-laterman:
If we want to take this approach we'll also need the go-licenser installed in a similar manner.
Maybe we should have a make command that will install both tools?

scunningham:
Open to suggestions.  I am having similar issues with the licenser.

Makefile
```diff
@@ -42,9 +45,9 @@ clean: ## - Clean up build artifacts
 .PHONY: generate
 generate: ## - Generate schema models
 	@printf "${CMD_COLOR_ON} Installing module for go generate\n${CMD_COLOR_OFF}"
-	go install github.com/aleksmaus/generate/...
+	env GOBIN=${GOBIN} go install github.com/aleksmaus/generate/cmd/schema-generate@latest
```
EricDavisX, line 48: is it in the plan prior to 7.14 to remove alexmaus personal git directory as part of source?

scunningham: That would be wise.  @aleksmaus can handle that when he returns.  Also noticed a [tool](https://github.com/elastic/fleet-server/blob/master/dev-tools/common.bash#L43) in Andrew Kroh's personal space.  Not sure where that came from.

EricDavisX:
> Open to suggestions. I am having similar issues with the licenser.

If it works as is, I prefer to merge and iterate since it is blocking the build for all erryone

scunningham:
> If it works as is, I prefer to merge and iterate since it is blocking the build for all erryone

Just waiting for a review.  Anybody?  Beuller?

Makefile
aleksmaus: This was a fork and heavily customized schema code generator for our specific needs.
FWIW, we pull bunch of third party dependencies from other people personal spaces ;-)

Can move it at some point into elastic org sure.

aleksmaus (approved):
LGTM

PR #472, opened 2021-06-21
Add unenroller based on unenroll_timeout from policy

### LINKED ISSUE #446: Unenroll ephemeral Agents
The Policy for Agents will include a optional TTL (name of the setting to be defined) value for ephemeral Agents. Each Agent enrolled to a policy that includes a TTL > 0, will be an ephemeral Agent.

Fleet Server is required to automatically unenroll an Agent that didn't check in within the TTL limit (the TTL must be >= the required checking duration).

So users can still inspect the current Agent, we will not directly unenroll the Agent. The Agent will be marked as "Inactive", and its API keys will be revoked. The Agent is soft-deleted.

blakerouse:
## What does this PR do?

Adds an unenroller that automatically unenrolls agents when the policy has `unenroll_timeout` set and agents `last_checkin` timestamp is older than the defined timeout.

## Why is it important?

To support ephemeral Elastic Agent's, having the Elastic Agent's list automatically cleaned up.

## Checklist

## Related issues

- Closes #446

### DIFF
--- a/internal/pkg/coordinator/monitor.go
+++ b/internal/pkg/coordinator/monitor.go
@@ -7,6 +7,7 @@ package coordinator
 import (
 	"context"
 	"errors"
+	"github.com/elastic/fleet-server/v7/internal/pkg/apikey"
 	"net"
 	"os"
 	"runtime"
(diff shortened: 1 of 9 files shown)


blakerouse:
Forgot this needs to invalidate the API keys for the unenrolled Elastic Agents, I will fix that and push a new revision.

internal/pkg/coordinator/monitor_integration_test.go
```diff
@@ -108,6 +110,86 @@ func TestMonitorLeadership(t *testing.T) {
...
+	go func() {
+		err := pm.Run(ctx)
+		wg.Done()
+		if err != nil && err != context.Canceled {
+			t.Fatal(err)
+		}
+	}()
+
+	// wait 2s to ensure unenroller runs
+	<-time.After(2 * time.Second)
```
aleksmaus, line 180: every 2 secs add up to the increase of the test running time. not a fan of these kind of hardcoded waits

internal/pkg/coordinator/monitor.go
```diff
@@ -394,3 +425,56 @@ func runCoordinatorOutput(ctx context.Context, cord Coordinator, bulker bulk.Bul
...
 	}
 }
+
+func runUnenroller(ctx context.Context, bulker bulk.Bulk, policyId string, unenrollTimeout time.Duration, l zerolog.Logger, checkInterval time.Duration, agentsIndex string) {
+	t := time.NewTimer(checkInterval)
+	defer t.Stop()
+	for {
+		select {
+		case <-t.C:
+			t.Reset(checkInterval)
```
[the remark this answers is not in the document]
blakerouse: Fixed

internal/pkg/coordinator/monitor.go
```diff
@@ -394,3 +425,56 @@ func runCoordinatorOutput(ctx context.Context, cord Coordinator, bulker bulk.Bul
...
+						Id:    agent.Id,
+						Index: agentsIndex,
+						Body:  body,
+					}
+				}
+				if err != nil {
+					// hit an error; retry loop
+					continue
+				}
+				if len(agentIds) > 0 {
```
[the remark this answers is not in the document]
blakerouse: Changed the flow to combine API key invalidate with updating of Agent so they align with each other.

internal/pkg/coordinator/monitor_integration_test.go
```diff
@@ -108,6 +110,86 @@ func TestMonitorLeadership(t *testing.T) {
...
+	require.NoError(t, err)
+
+	// start the monitors
+	var wg sync.WaitGroup
+	wg.Add(2)
+	go func() {
+		err := pim.Run(ctx)
+		wg.Done()
+		if err != nil && err != context.Canceled {
+			t.Fatal(err)
```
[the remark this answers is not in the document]
blakerouse: Fixes

internal/pkg/coordinator/monitor_integration_test.go
blakerouse: Fixed

blakerouse:
@aleksmaus Thanks for the review, its ready for another look.

internal/pkg/coordinator/monitor.go
```diff
@@ -394,3 +426,72 @@ func runCoordinatorOutput(ctx context.Context, cord Coordinator, bulker bulk.Bul
...
+		dl.FieldUpdatedAt:        now,
+	}
+	body, err := fields.Marshal()
+	if err != nil {
+		return err
+	}
+	apiKeys := getAPIKeyIDs(agent)
+	if len(apiKeys) > 0 {
+		err = apikey.Invalidate(ctx, bulker.Client(), apiKeys...)
+		if err != nil {
```
aleksmaus, line 480: does it need to stop invalidation and unenrollment if it fails to invalidate one of the api keys?

aleksmaus (approved):
One more question, please respond.
Otherwise LGTM

internal/pkg/coordinator/monitor.go
blakerouse: I think this is the correct behavior, because if it fails to invalidate then it will stop. The loop will run again after the check interval and then hopefully that time it will succeed. If it keeps failing then it will not unenroll Elastic Agent's which will then be obvious on the listing and that would prompt them to check on what is going wrong.

urso:
Looks like we still should add tests for the reconfiguration itself. Is it possible for the old unenroller to block during shutdown?

PR #477, opened 2021-06-22
Move Fleet Server specific config from elastic search section to server section.

scunningham:
The elastic search section is an "output" section and has the side effect of being
applied to all agents when changed in Kibana.  The bulk section only applies to Fleet
Server so should be located in the input section configuration the server.

## What does this PR do?

Moves the bulk config section around so more easily configurable in Kibana.

## Why is it important?

Previous location was not strictly correct and had operational side effect of being applied to every policy when changed in Kibana.

## Checklist


aleksmaus (approved):
👍

PR #511, opened 2021-07-01
Enable http2 in Fleet Server.

scunningham:
## What does this PR do?

Enables http2 on Fleet Server's HTTP interface.  The protocol is not supported implicitly when creating a server with the more sophisticated API calls.

## Why is it important?

Agent expects HTTP2 now.  System is broken.

## Checklist

### DIFF
--- a/cmd/fleet/server.go
+++ b/cmd/fleet/server.go
@@ -88,20 +88,33 @@ func runServer(ctx context.Context, router *httprouter.Router, cfg *config.Serve
 		return err
 	}
 
-	defer ln.Close()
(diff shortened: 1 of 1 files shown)


blakerouse (approved):
Looks good, thanks for the comments.

blakerouse:
After more trial and error I was able to determine that Elastic Agent 7.14 (what is in the repo does work, with or without this change). I was also able to confirm using curl that this does allow the upgrade to http2, but the current code in Elastic Agent 7.14 does not actually upgrade the connection (that is a client-side issue only, so needs to be fixed in Elastic Agent). Going to merge this for Fleet Server and get it into 7.14, as this change does work with current 7.14.

PR #520, opened 2021-07-05
Bump version to 7.13.4

andresrc:
Part of 7.13.3 release process.

To be merged on release day


urso:
7.13 should not have been moved to go1.16 yet. Let's ignore the error and undo the update to go.sum. We have had issues with checksums being added and removed by different tools when moving to go1.16.

urso (approved):

PR #523, opened 2021-07-07
Policy changes can be dropped during policy rollouts to large number …

### LINKED ISSUE #522: A policy that is updated frequently in a large fleet may not be updated.
fleet-server 7.14  8fc011454055c1ed9f8f2c5a21a9dd12b07ddf68

Steps to reproduce:

0) Configure Fleet-Server to support 5k endpoints or higher.
1) Use Horde to enroll 5k endpoints.
2) Change the policy every 30 seconds.

Result:  The first policy change will be rolled out, but the following changes will be ignored.   The policy will appear stuck.

scunningham:
…of agents.  Aggregates changes to avoid missing updates.

## What does this PR do?

Works around race condition where internal notifications on policy update were dropped.

## Why is it important?

Policies can appear stuck at scale.

## Checklist

Follow instructions in related issue.  Check the logs.  

## Related issues

- Closes #522 

## Logs

```
10:53:08.821 INF Received hits on local monitor ctx="policy agent monitor" nHits=1 service.name=fleet-server
10:53:09.052 INF Received hits on local monitor ctx="policy agent monitor" nHits=1 service.name=fleet-server
10:54:44.821 INF policy rollout end coordinatorIdx=1 ctx="policy agent monitor" policyId=d43cd250-df22-11eb-9c07-7d2f0052b4b7 revisionIdx=18 service.name=fleet-server tdiff=99004.626266
10:54:44.822 INF Hits dispatched to main loop ctx="policy agent monitor" nHits=2 service.name=fleet-server
10:54:44.823 INF policy revised ctx="policy agent monitor" ncoord=1 nrev=19 ocoord=1 orev=18 policyId=d43cd250-df22-11eb-9c07-7d2f0052b4b7 service.name=fleet-server
10:54:44.824 INF policy rollout begin coordinatorIdx=1 ctx="policy agent monitor" nSubs=4940 policyId=d43cd250-df22-11eb-9c07-7d2f0052b4b7 revisionIdx=19 service.name=fleet-server throttle=20
10:56:23.628 INF policy rollout end coordinatorIdx=1 ctx="policy agent monitor" policyId=d43cd250-df22-11eb-9c07-7d2f0052b4b7 revisionIdx=19 service.name=fleet-server tdiff=98801.968635
```

### DIFF
--- a/internal/pkg/policy/monitor.go
+++ b/internal/pkg/policy/monitor.go
@@ -101,6 +101,13 @@ func (m *monitorT) Run(ctx context.Context) error {
 	s := m.monitor.Subscribe()
 	defer m.monitor.Unsubscribe(s)
 
+	// The output from m.monitor times out if we don't pull data off quickly enough.
+	// Rollout can take a while; append upates here until we can attend to it.
+	// TODO: This is a workaround for 7.14.  Rollout strategy will be reconsidered for 7.15.
+	subCtx, cfunc := context.WithCancel(ctx)
+	defer cfunc()
+	outputCh := m.monitorOutputChannel(subCtx, s.Output())
+
 	close(m.startCh)
 LOOP:
 	for {
@@ -111,7 +118,7 @@ LOOP:
 			if err := m.process(ctx); err != nil {
 				return err
 			}
-		case hits := <-s.Output():
+		case hits := <-outputCh:
 			policies := make([]model.Policy, len(hits))
 			for i, hit := range hits {
 				err := hit.Unmarshal(&policies[i])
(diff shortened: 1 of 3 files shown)


internal/pkg/policy/monitor.go
```diff
@@ -128,6 +135,35 @@ LOOP:
...
+
+		for {
+			select {
+			case <-ctx.Done():
+				m.log.Info().Msg("Exit policy monitor local")
+				return
+			case nHits := <-outputCh:
+				m.log.Info().Int("nHits", len(nHits)).Msg("Received hits on local monitor")
+				hits = append(hits, nHits...)
+				outCh = localOutputCh
```
blakerouse, line 155: Why is `outCh` set in this case? Couldn't it be set at the top of the go func or just use `localOutputCh` directly. Or am I missing something about this flow.

blakerouse (approved):
Change looks good, just one comment on the flow of the channel in the go routine.

internal/pkg/policy/monitor.go
scunningham: The localOutputCh  has no buffer by design.  If we use it directly it will block and we won't pickup any additional changes on outputCh.

We could make localOutpuCh a buffered channel, but how big is the buffer?  You may still hit the max size and block.  

This method allows us to continue to add data to the 'hits' array until the consuming loop is able to retrieve the data.  Switching 'outCh' to null once consumed allows the loop to spin properly until additional data is available.

internal/pkg/policy/monitor.go
```diff
@@ -128,6 +135,35 @@ LOOP:
 	return nil
 }
 
+// Aggegates changes from the output channel until the main loop can process.
+func (m *monitorT) monitorOutputChannel(ctx context.Context, outputCh <-chan []es.HitT) chan []es.HitT {
```
aleksmaus, line 139: nit:
maybe?
```
func (m *monitorT) monitorOutputChannel(ctx context.Context, outputCh <-chan []es.HitT) <-chan []es.HitT {
```

aleksmaus (approved):
LGTM

internal/pkg/policy/monitor.go
blakerouse: Thanks for the explanation. I missed the `outCh = nil` in the last case, that make sense now.

blakerouse (approved):
Looks good.

urso:
Have we tried to reproduce it in a test putting the subsystem under pressure? Any test we should add here?

PR #555, opened 2021-07-14
refactor rollout to be more deterministic

scunningham:
## What does this PR do?

Reworks the policy rollout to be fairer and more controlled.  Avoids bursts when a bunch of agents come online at once.

## Why is it important?

The server becomes overwhelmed if too many agents checkin at the same time after a policy change.

## Checklist

### DIFF
--- a/internal/pkg/policy/monitor.go
+++ b/internal/pkg/policy/monitor.go
@@ -7,9 +7,7 @@ package policy
 import (
 	"context"
 	"errors"
-	"fmt"
 	"sync"
-	"sync/atomic"
 	"time"
 
 	"github.com/rs/zerolog"
@@ -22,7 +20,30 @@ import (
 	"github.com/elastic/fleet-server/v7/internal/pkg/monitor"
 )
 
-var gCounter uint64
+const cloudPolicyId = "policy-elastic-agent-on-cloud"
+
+/*
+Design should have the following properites
+
+Policy rollout scheduling should...
+1) be fair; delivered in first come first server order.
+2) be throttled to avoid uncontrolled impact on resources, particularly CPU.
+3) adapt to subscribers that drop offline.
+4) attempt to deliver the latest policy to each subscriber at the time of delivery.
+5) prioritize delivery to agents that supervise fleet-servers.
+
+This implementation addresses the above issues by queuing subscription requests per
+policy, and moving requests to the pending queue when the requirement is met; ie.
+the policy is updateable.
+
+If the subscription is unsubscribed (ie. the agent drops offline), this implementation
+will remove the subscription request from its current location in either the waiting
+queue on the policy or the pending queue.
+
+Ordering is achieved with a simple double linked list implementation that allows object
+migration across queues, and O(1) unlink without knowledge about which queue the subscription
+is in.
+*/
 
 type Subscription interface {
 	// Output returns a new policy that needs to be sent based on the current subscription.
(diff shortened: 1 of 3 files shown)


internal/pkg/policy/monitor.go
```diff
@@ -22,7 +20,28 @@ import (
...
-var gCounter uint64
+/*
+Design should have the following properites
+
+Policy rollout scheduling should...
+1) be fair; delivered in first come first server order.
+2) be throttled to avoid uncontrolled impact on resources, particularly CPU.
+3) adapt to subscribers that drop offline.
+4) attempt to deliver the latest policy to each subscriber at the time of delivery.
+5) prioritize delivery to agents that supervisor fleet-servers.
```
aleksmaus, line 31: ```to agents that supervise fleet-servers```

internal/pkg/policy/monitor.go
```diff
@@ -96,72 +105,88 @@ func NewMonitor(bulker bulk.Bulk, monitor monitor.Monitor, throttle time.Duratio
...
+			if err := m.processHits(ctx, hits); err != nil {
 				return err
 			}
+			startDeploy()
+		case <-ticker.C:
+			if done := m.dispatchPending(); done {
+				stopDeploy()
+			}
+		case <-ctx.Done():
+			break LOOP
```
aleksmaus, line 162: this will leak the ticker if it was started by any of the previous ```case```s

internal/pkg/policy/monitor.go
```diff
@@ -224,131 +291,93 @@ func (m *monitorT) groupByLatest(policies []model.Policy) map[string]model.Polic
...
+	for sub := iter.Next(); sub != nil; sub = iter.Next() {
+		if sub.isUpdate(&newPolicy) {
+
+			// Unlink the target node from the list
+			iter.Unlink()
+
+			// Push the node onto the pendingQ
+			// HACK: if update is for cloud agent, put on front of queue
+			// not at the end for immediate delivery.
+			if newPolicy.PolicyId == "policy-elastic-agent-on-cloud" {
```
aleksmaus, line 343: const?

scunningham: yep, fixed. tnx

internal/pkg/policy/monitor.go
scunningham: whoops fixed.

internal/pkg/policy/monitor.go
scunningham: tnx

internal/pkg/policy/monitor.go
```diff
@@ -224,131 +295,93 @@ func (m *monitorT) groupByLatest(policies []model.Policy) map[string]model.Polic
...
+	for sub := iter.Next(); sub != nil; sub = iter.Next() {
+		if sub.isUpdate(&newPolicy) {
+
+			// Unlink the target node from the list
+			iter.Unlink()
+
+			// Push the node onto the pendingQ
+			// HACK: if update is for cloud agent, put on front of queue
+			// not at the end for immediate delivery.
+			if newPolicy.PolicyId == cloudPolicyId {
```
blakerouse, line 347: Yeah this is an ugly hack. I think you should also add a check to move to the front of the queue `newPolicy.is_default_fleet_server`.

But I think long term we need a way for a priority to be set on the policy in the index. So the cloud policy and the default fleet server policy can set a higher priority than the other policies.

blakerouse (approved):
Overall I like this change, other than the hack of course. But I think the hack is okay for now, until we can come up with a better solution.

PR #591, opened 2021-07-27
Improve PR description template

ruflin:
One area I think we can improve as a team is on descriptions of pull requests. A pull request description should not only reference an issue that it solves but also be able to stand on its own. It is important to describe how a certain problem is solved. What approach was taken and why. This allows anyone reviewing a PR to build a mental model on what change is in the pull request and if it aligns with the described solution.

There are cases where the PR description takes more time then the actual fix in code. Anyone reading the PR description should understand why the changes was made and what the change contains.

I opened to discuss a change to the template with these two question on the top. The template itself from my perspective is mostly meant as a guideline/helper but it should be up to each engineer to decide what is really needed for this PR.

This PR is made on the fleet-server repository as a first trial but the idea is to use the same template for Beats and other repositories.

I would like to hear the opinion from others on how they think about these templates. TBH in most cases I personally just wipe the template an use my own description.


ruflin:
@urso @exekias @blakerouse @masci @andresrc What is your take on this? Is the template even useful or should it just be bullet list as a "helper"?

andresrc (approved):

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -8,6 +8,15 @@ Please label this PR with one of the following labels, depending on the scope of
...
+## What is the problem this PR solves?
+
+// Please do not just reference an issue
+
+## How does this PR solve the problem?
+
+// Explain how you solved the problem in your code.
+
+
 ## What does this PR do?
```
blakerouse, line 20: I think the 2 new sections, which I like would satisfy this section. What would we add to this section that would not probably be already explained in the previous 2 sections?

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -8,6 +8,15 @@ Please label this PR with one of the following labels, depending on the scope of
 - Docs
 -->
 
+## What is the problem this PR solves?
```
blakerouse, line 11: I always took this as the `Why is it important?` section. Should that section be removed in place of this section?

ruflin: I was looking at these 2 a bit different. This describes the problem itself without rating it on importance. But I agree, often this part will contain content around why it is important because it blocks xyz or fixes memory leak etc. 

I'll remove the "Why is it important?" section.

.github/PULL_REQUEST_TEMPLATE.md
ruflin: Agree, will remove it too.

ruflin:
Updated the PR with a few details and made it ready for review. Let me know what you think.

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -8,45 +8,36 @@ Please label this PR with one of the following labels, depending on the scope of
...
+This checklist is to help creators of PRs to find parts which might not be directly related to code change but still need to be addressed. Anything that does not apply to the PR should be removed from the checklist.
 -->
 
-- [ ] My code follows the style guidelines of this project
 - [ ] I have commented my code, particularly in hard-to-understand areas
 - [ ] I have made corresponding changes to the documentation
 - [ ] I have made corresponding change to the default configuration files
 - [ ] I have added tests that prove my fix is effective or that my feature works
 - [ ] I have added an entry in `CHANGELOG.next.asciidoc` or `CHANGELOG-developer.next.asciidoc`.
+- [ ] Documentations added / updated for the change.
```
blakerouse, line 37: How is this different from `I have made corresponding changes to the documentation`?

ruflin: It's not 🤦‍♂️  Fixing.

blakerouse (approved):
Looks good!

Issue #602, opened 2021-07-29
Add Changelog for fleet-server


ruflin:
Fleet-server requires its own changelog. It needs to be decided what the right format is and how it fits into Elastic Agent release notes.

ruflin:
I opened https://github.com/elastic/fleet-server/pull/672 with an initial version. This is not the end goal but at least should give us a place to track changes.

PR #665, opened 2021-08-26
Add index refresh call after fetch if the list of hits has holes

aleksmaus:
## What does this PR do?

Adds index refresh call after fetch if the list of hits/results has holes.

After discussion with Elasticsearch team it looks like the current global checkpoints
API doesn't not gurantee the list of document without holes on
subsequent index search and there is still a slim chance of this with
our current implementation. Elastisearch team is working on getting the
refresh implemented on thier side, and for now recommended to do the
refresh before fetch if there are holes detected in the result.

Added the check for the holes and refetch, here are the steps:
1. Call Global checkpoints = 5
2. Search = 1, 2, 3, 5.
3. Manual refresh
4. Search and get 4,5
5. Return to step 1

## Why is it important?

Covers the edge case where the fetched documents lists after checkpoint API could still potentially have holes.
In practice we haven't seen this happening, but this change was recommended by Elasticsearch team.

## Checklist

### DIFF
--- a/internal/pkg/monitor/monitor.go
+++ b/internal/pkg/monitor/monitor.go
@@ -259,6 +259,14 @@ func (m *simpleMonitorT) Run(ctx context.Context) (err error) {
 			}
 		}
 
+		// This is an example of steps for fetching the documents without "holes" (not-yet-indexed documents in between)
+		// as recommended by Elasticsearch team on August 25th, 2021
+		// 1. Call Global checkpoints = 5
+		// 2. Search = 1, 2, 3, 5.
+		// 3. Manual refresh
+		// 4. Search and get 4,5
+		// 5. Return to step 1
+
 		// Fetch up to known checkpoint
 		count := m.fetchSize
 		for count == m.fetchSize {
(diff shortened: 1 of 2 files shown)


internal/pkg/monitor/monitor.go
```diff
@@ -308,6 +368,37 @@ func (m *simpleMonitorT) fetch(ctx context.Context, maxCheckpoint sqn.SeqNo) ([]
 	return hits, nil
 }
 
+// Refreshes index. This is temporary code
+// TODO: Remove this when the refresh is properly imlemented on Eleasticsearch side
```
scunningham, line 372: spelling imlemented

scunningham (approved):
Can you change comment to indicate when we expect ES to fix and this will become deprecated.

ruflin:
* Is my understanding correct that if a user hits this, some changes might be missed so this would be a bug? 
* Does the Elasticsearch team have an issue to track this? Would be nice to link to it in the code based for any future engineer stumbling over it + having an issue on our end to track this change.
* Wanted to ask for an entry in the changelog but seems like we still don't have one :-( https://github.com/elastic/fleet-server/issues/602

EricDavisX:
It looks ok to me for 7.15 release if you feel good.  Thanks for adding tests!  And ++ to editing comment to cite the ES side bug this works around so we can (at least periodically) check to see when /if this can be removed

ruflin:
+1 on 7.15 as this might cause some issues if someone hits it so I consider it a bug.

blakerouse (approved):
Looks good, tested well. +1 for 7.15 as well.

aleksmaus:
> Can you change comment to indicate when we expect ES to fix and this will become deprecated.

Reached out to elasticsearch team. The issue for "refresh" falls under phase 2 of https://github.com/elastic/elasticsearch/issues/71449

Once the phase is complete we can remove the refreshes from fleet-server.

tbrooks8:
We're looking at phase 2 currently. We had a prototype (https://github.com/elastic/elasticsearch/pull/73134). However, we have slightly reconsidered the design based on work we plan on doing for multi-shard support. We would hope that the refresh work could go into 7.16, but we need to sync with the es-search team to ensure that our designs make sense.

PR #672, opened 2021-08-30
Add changelog file to fleet-server

ruflin:
Currently fleet-server does not contain a changelog. With this I am adding the simplest changelog file possible. We need to find a better way on how to handle the changelog but for now I think it is more important to have a changelog then not to have one at all.

Closes https://github.com/elastic/fleet-server/issues/602

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,0 +1,16 @@
+# Changelog
+
+All notable changes to this project will be documented in this file.
+
+
+## [Unreleased]
+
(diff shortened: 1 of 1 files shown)


ruflin:
@jlind23 Curious to hear your take on this. How do we track best changes in fleet-server for now?

ruflin:
Closing this for now so we can continue the discussion but don't need an open PR for this. Can always reopen.

PR #681, opened 2021-08-31
Add "hole" detection and refresh for the agent actions fetch upon initial check in

aleksmaus:

This covers one more edge case for actions in multiple replica shards environment per discussion with Elasticsearch team

The edge use case:
1. the monitor fetches from one replica shard
2. result has no holes
3. the index refresh is skipped if there is no holes
4. the seqno is advanced forward as expected
5. the agent upon check-in, by the random draw of luck fetches actions from different replica shard that doesn't have all the documents indexed in that range
6. the result can have holes

In order to cover this use case need to add the holes detection and refresh upon the agent initial actions fetch during checkin.

* Extracted HasHoles function into es package so it can be reused in two places: the monitor and the checking handler
* Extracted es.Refresh function so it can be reused in two places: the monitor and the checking handler
* Added FindAgentActions wrapper function with the holes check and refresh that is used to fetch the initial agent's actions upon check in.

## Checklist

### DIFF
--- a/cmd/fleet/handleCheckin.go
+++ b/cmd/fleet/handleCheckin.go
@@ -375,14 +375,8 @@ func (ct *CheckinT) resolveSeqNo(ctx context.Context, req CheckinRequest, agent 
 }
 
 func (ct *CheckinT) fetchAgentPendingActions(ctx context.Context, seqno sqn.SeqNo, agentId string) ([]model.Action, error) {
-	now := time.Now().UTC().Format(time.RFC3339)
 
-	actions, err := dl.FindActions(ctx, ct.bulker, dl.QueryAgentActions, map[string]interface{}{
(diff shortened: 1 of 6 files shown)


blakerouse (commented):
Question is does this also affect policies or is it only an issue for actions? If not for policies can you explain why not, that would really help me understand why this is needed.

aleksmaus:
> Question is does this also affect policies or is it only an issue for actions? If not for policies can you explain why not, that would really help me understand why this is needed.

don't remember how the policies are done, you might remember better, but I think you are always interested in the latest revision of the policy to roll out.

scunningham (approved):

blakerouse (approved):
Looks good.

Thanks for adding the comments about when this can be removed.

PR #683, opened 2021-08-31
Fix panic caused by Agent check-in when Fleet Server is upgraded from 7.14.0 previous

blakerouse:
## What is the problem this PR solves?

// Please do not just reference an issue. Explain WHAT the problem this PR solves here.

When running on previous version of Fleet Server the version of the Elastic Agent was not stored. With 7.15 the version was added, but that means that on check-in it needs to handle the case that the metadata is not present.

## How does this PR solve the problem?

// Explain HOW you solved the problem in your code. It is possible that during PR reviews this changes and then this section should be updated.

This adds a check to ensure that the metadata is present before checking the version field. If not present then the version will be updated correctly in 7.15 as it should.

## Checklist

- [x] I have commented my code, particularly in hard-to-understand areas
- ~~[ ] I have made corresponding changes to the documentation~~
- ~~[ ] I have made corresponding change to the default configuration files~~
- ~~[ ] I have added tests that prove my fix is effective or that my feature works~~
- ~~[ ] I have added an entry in `CHANGELOG.next.asciidoc` or `CHANGELOG-developer.next.asciidoc`.~~

## Related issues

- Closes #682

PR #688, opened 2021-09-01
Encapsulate the agent new version safe check, full unit test coverage

aleksmaus:

Addresses comment on previous PR:
https://github.com/elastic/fleet-server/pull/683#issuecomment-909984620

* Extract the get agent new version into a function
* Add full unit test coverage for that part

## Checklist

## Related issues

- Relates https://github.com/elastic/fleet-server/pull/683#issuecomment-909984620

### DIFF
--- a/internal/pkg/model/ext.go
+++ b/internal/pkg/model/ext.go
@@ -32,10 +32,9 @@ func (m *Agent) GetNewVersion(ver string) string {
 		return ""
 	}
 
-	var newVer string
 	if m.Agent == nil || ver != m.Agent.Version {
-		newVer = ver
+		return ver
 	}
 
-	return newVer
+	return ""
 }
--- a/internal/pkg/model/ext.go
+++ b/internal/pkg/model/ext.go
@@ -26,8 +26,8 @@ func (m *Server) SetTime(t time.Time) {
 	m.Timestamp = t.Format(time.RFC3339Nano)
 }
 
-// GetNewVersion returns Agent version if it is different from ver, otherwise return empty string
-func (m *Agent) GetNewVersion(ver string) string {
+// CheckDifferentVersion returns Agent version if it is different from ver, otherwise return empty string
+func (m *Agent) CheckDifferentVersion(ver string) string {
 	if m == nil {
 		return ""
 	}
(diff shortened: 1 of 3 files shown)


internal/pkg/model/ext.go
```diff
@@ -25,3 +25,16 @@ func (m *Server) Time() (time.Time, error) {
 func (m *Server) SetTime(t time.Time) {
 	m.Timestamp = t.Format(time.RFC3339Nano)
 }
+
+// GetNewVersion returns Agent version if it is different from ver, otherwise return empty string
+func (m *Agent) GetNewVersion(ver string) string {
+	if m == nil {
```
ruflin, line 31: Is this an expected scenario or should we log an error in this case?

internal/pkg/model/ext.go
```diff
@@ -25,3 +25,16 @@ func (m *Server) Time() (time.Time, error) {
 func (m *Server) SetTime(t time.Time) {
 	m.Timestamp = t.Format(time.RFC3339Nano)
 }
+
+// GetNewVersion returns Agent version if it is different from ver, otherwise return empty string
+func (m *Agent) GetNewVersion(ver string) string {
```
ruflin, line 30: Nit: It works also for older versions and only new :-)

ruflin (approved):
LGTM

scunningham (approved):

internal/pkg/model/ext.go
aleksmaus: > Nit: It works also for older versions and only new :-)
yeah might be the best name for the function, open to suggestions.

internal/pkg/model/ext.go
aleksmaus: it's not expected, the function is currently called only after successfully authenticated agent (agent is normally not nil). just additional nil safety check.

ruflin: `CheckDifferentVersion` or `GetDifferentVersion` -> took what you have in the description :-)

blakerouse (approved):
Looks good. Thanks for cleaning it up more and adding tests.

internal/pkg/model/ext.go
aleksmaus: ```CheckDifferentVersion``` sounds good, will update

PR #692, opened 2021-09-07
Fix issues with config reloading

### LINKED ISSUE #691: Agent Mode restart on failure uses the wrong configuration
When Elastic Agent sends a new configuration to Fleet Server and that configuration fails to load, then Fleet Server will rollback to the very first configuration that Elastic Agent ever sent. So if Elastic Agent sent 5 configurations before the last bad one it would result it going back all the way to the 1st configuration not the 5th configuration.

Fleet Server should take the configuration from Elastic Agent as final, being that it should not rollback to a good configuration on its own. Only the Elastic Agent should have control of that behavior. Instead it should keep trying to use that last configuration.

There is also an issue where changing the `fleet.agent.id` and/or any settings in `output` will not result in Fleet Server actually reloading.

blakerouse:

There are 3 problems that this PR solves:

1. Configurations changes to the `fleet.*` or the `output.*` sections of the Fleet configurations would not cause the Fleet Server to restart the required sections of the server for those changes to actually take affect.
2. In the case that the configuration change fails (maybe a transient elasticsearch connection problem) then the Fleet Server would roll all the way back to its initial configuration sent from Elastic Agent.
3. No integration tests existed for Fleet Server being controlled by Elastic Agent.

This solves each problem as follows:

1. It adjusts the configuration change checking code to check for the `fleet.*` and `output.*` sections.
2. It always stores the last given configuration in the FleetServet struct so on restart it always goes back to the last configuration that Elastic Agent provided it.
3. Integration test was added.

This PR also has a side-effect change that I needed to make to get the `libbeat/v7/x-pack/elastic-agent/core/server` working (which controls the AgentMode, just like the real Elastic Agent does). It seems that the `ristretto` module that Fleet Server uses for caching registers global flags because it includes `golang/glog` and the `libbeat/v7/x-pack/elastic-agent/core/server` includes the `logp` from libbeat registering the same flag. So to get the integration tests working I had to remove the `ristretto` module only when running integration tests.

`make test-int`

## Checklist

## Related issues

- Closes #691

### DIFF
--- a/cmd/fleet/main.go
+++ b/cmd/fleet/main.go
@@ -511,6 +511,7 @@ LOOP:
 		}
 
 		curCfg = newCfg
+		f.cfg = curCfg
 
 		select {
 		case newCfg = <-f.cfgCh:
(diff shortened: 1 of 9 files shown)


cmd/fleet/main.go
```diff
@@ -511,6 +511,7 @@ LOOP:
 		}
 
 		curCfg = newCfg
+		f.cfg = curCfg
```
aleksmaus, line 514: doesn't look like curCfg is used after this point

why not just do
```
f.cfg = newCfg
```

and remove 
```
curCfg = newCfg
```

internal/pkg/cache/impl.go
```diff
@@ -0,0 +1,16 @@
...
+// or more contributor license agreements. Licensed under the Elastic License;
+// you may not use this file except in compliance with the Elastic License.
+
+package cache
+
+import (
+	"time"
+)
+
+type CacheImpl interface {
```
aleksmaus, line 11: nit: maybe different name for the interface, having "Impl" in it a bit odd. 
```
type Cacher interface
```

aleksmaus (approved):

cmd/fleet/main.go
blakerouse: it is used in the for loop for checking to see if the configuration change required a reload

aleksmaus: ughh, didn't notice the bigger loop it's in ... long function

PR #713, opened 2021-09-15
Do not require that the default output use the default key in the outputs of policy

blakerouse:

Kibana will adjust its policies to not always provide a `outputs.default`. Instead the `outputs` section can use any key, but it can only have one that is the default output for the policy. It determines the default by ensuring that its an `elasticsearch` type and that it does not define a `fleet_server.service_token` on the output.

This is Step 1 in supporting multiple outputs. This will allow support a different output in policies, but does not yet add support for a remote cluster or multiple outputs in the policy.

See: https://github.com/elastic/kibana/pull/111002

## Checklist


aleksmaus (approved):
lgtm, thanks!

PR #762, opened 2021-10-06
Server and cache defaults based on env

michalpristas:

This PR takes available RAM as a reference point for setting defaults.
Defaults are based on our recommended values listed here: https://www.elastic.co/guide/en/fleet/current/fleet-server.html#resource-requirements-by-number-agents

This PR introduces couple of configuration files. each of them sets defaults for env based on ram size range. 
Based on env correct file is picked upon Unpacking configuration. 
These files are built in using packer to avoid runtime dependencies on these config files.
Configuration can be overridden by config values coming from fleet. 

In case config file changes `make defaults` can be used to update `env_defaults.go` file. `defaults` is also part of `check` target.

Build and run on your env. 
I added fmt.Print additionally to check values.

## Checklist

### DIFF
--- a/dev-tools/buildlimits/buildlimits.go
+++ b/dev-tools/buildlimits/buildlimits.go
@@ -72,17 +72,23 @@ const (
 	defaultAckMaxBody  = 1024 * 1024 * 2
 )
 
+type valueRange struct {
+	Min int ` + "`config:\"min\"`" + `
+	Max int ` + "`config:\"max\"`" + `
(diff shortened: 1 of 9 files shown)


blakerouse:
I think this looks really good. I would have preferred to see all these settings in the Fleet Server package, but being we need to have it change automatically based on host size, this works well.

What about the settings in the Fleet Server integration package, like max connections? Is that going away?

PR #798, opened 2021-10-21
Change agentId field to a value that elastic will map.

scunningham:

While errors are being logged when a checkin fails, the "id" field is not mapped making it difficult to track individual agents in the logs.  Use a field name that will be mapped.

Renames the field.

## Checklist

### DIFF
--- a/cmd/fleet/handleAck.go
+++ b/cmd/fleet/handleAck.go
@@ -64,6 +64,7 @@ func (rt Router) handleAcks(w http.ResponseWriter, r *http.Request, ps httproute
 
 		log.WithLevel(resp.Level).
 			Err(err).
+			Str("agentId", id).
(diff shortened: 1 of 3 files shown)


cmd/fleet/handleAck.go
```diff
@@ -64,6 +64,7 @@ func (rt Router) handleAcks(w http.ResponseWriter, r *http.Request, ps httproute
 
 		log.WithLevel(resp.Level).
 			Err(err).
+			Str("agentId", id).
```
aleksmaus, line 67: nit: maybe agent_id?

scunningham: Kept to be in alignment with the rest of the app.   Later PR to standardize it.

We have an additional issue that ECS is formatted differently; lower case dot formatted.  But we don't want conflict either, so no good answers.

PR #805, opened 2021-10-25
Normalize logging

scunningham:

This PR fixes several issues:
- Another pass at normalizing logging using field names that ElasticSearch will index.
- Uses the proper TLS structure for reading the configuration; thus enabling optional client validation.
- Handles case where enrollment response fails to write, removing orphaned records.
- Enables HTTP logging at the INFO level, adding addition logging if DEBUG enabled.

Refines the logging with constants.  Works the edge cases.

## Checklist

### DIFF
--- a/cmd/fleet/handleEnroll.go
+++ b/cmd/fleet/handleEnroll.go
@@ -27,6 +27,7 @@ import (
 	"github.com/julienschmidt/httprouter"
 	"github.com/miolini/datacounter"
 	"github.com/pkg/errors"
+	"github.com/rs/zerolog"
 	"github.com/rs/zerolog/log"
 )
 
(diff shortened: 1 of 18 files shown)


cmd/fleet/handleEnroll.go
```diff
@@ -76,53 +77,44 @@ func (rt Router) handleEnroll(w http.ResponseWriter, r *http.Request, ps httprou
...
-		Str("mod", kEnrollMod).
-		Str("agentId", enrollResponse.Item.ID).
-		Str("policyId", enrollResponse.Item.PolicyId).
-		Str("apiKeyId", enrollResponse.Item.AccessApiKeyId).
-		Str(EcsHttpRequestId, reqId).
-		Int(EcsHttpResponseBodyBytes, numWritten).
-		Int64(EcsEventDuration, time.Since(start).Nanoseconds()).
-		Msg("success enroll")
+		// Remove ghost artifacts; agent will never receive the paylod
+		rt.et.wipeGhosts(r.Context(), zlog, resp)
```
aleksmaus, line 113: the cleanup will stop if the request is cancelled, this will need to be revisited. agreed to address after this PR is merged.

cmd/fleet/handleEnroll.go
```diff
@@ -238,11 +238,96 @@ func _enroll(ctx context.Context, bulker bulk.Bulk, c cache.Cache, req EnrollReq
...
+	if err := et.bulker.Delete(ctx, dl.FleetAgents, resp.Item.ID); err != nil {
+		zlog.Error().Err(err).Msg("ghost agent record failed to delete")
+	} else {
+		zlog.Info().Msg("ghost agent record deleted")
+	}
+
+	invalidateApiKey(ctx, zlog, et.bulker, resp.Item.AccessApiKeyId)
+}
+
+func invalidateApiKey(ctx context.Context, zlog zerolog.Logger, bulker bulk.Bulk, apikeyId string) error {
```
aleksmaus, line 259: this has a potential to hold the request go routine alive, that might cause the go routines exhaustion at scale.
agreed this will be addressed in the follow up PRs probably as standalone GC routine/pool

aleksmaus (approved):

cmd/fleet/handleEnroll.go
scunningham: Agreed, will address in subsequent PR.

jlind23:
@scunningham could we backport it to 7.15.2?

ruflin:
This was not backported to 8.0. Added the label now. We need to check if other backports are missing too.

ruflin:
This also has not made it into 7.16 or 7.15 as the wrong backport labels were used. Talked to Victor to clean this up.

cmd/fleet/handleEnroll.go
ruflin: Is there already a follow up PR for this?

aleksmaus: not yet, was planning to address that

scunningham:
> This also has not made it into 7.16 or 7.15 as the wrong backport labels were used. Talked to Victor to clean this up.
> 
> I don't think we should backport to 7.15, only 7.16 now.

Purposely not back-ported.  Security team gets beat up for landing features post feature freeze. This is not a bugfix or a security issue, so not back-ported.  We can chose back-port anyhow, but I would not do it without good reason.

Issue #819, opened 2021-10-29
`agent.id` field isn't populated in `.fleet-agents` when upgrading from 7.14


joeypoon:
works if enrolled in 7.15 or above and then upgrading but if agent enrolled from <=7.14 and then upgraded, `agent.id` isn't populated.

PR #823, opened 2021-11-01
Support agent record migration from 7.14 agents.

scunningham:

Agent records created during 7.14 do not have the new "agent" sub-object used for transforms.  This code performs the migration.

If the agent record does not exist, create it.

## Checklist

## Related issues

- Relates https://github.com/elastic/fleet-server/issues/819


aleksmaus (approved):
👍

PR #827, opened 2021-11-02
Migrate agent.id field from 7.14 to 7.15+

scunningham:

Fix missing agent record migration added in 7.15.

At boot, query for agent records that do not contain agent.id field and fix them up.  Should be a fast noop when no changes necessary.

In addition, fixes issue where agent.version was being written at root instead of as a child of the agent object.

Upgrade from 7.14, or enroll with 7.15 or later, and delete the "agent" object from the agent record in .fleet-agents.  This has been tested at scale, supports retry version conflict.  Once migration has occurred, typical response time from Elastic is less then 10ms (not including TLS roundtrip).

## Checklist

## Related issues

- Closes https://github.com/elastic/fleet-server/issues/819

### DIFF
--- a/cmd/fleet/auth.go
+++ b/cmd/fleet/auth.go
@@ -137,10 +137,10 @@ func authAgent(r *http.Request, id *string, bulker bulk.Bulk, c cache.Cache) (*m
 	}
 
 	// validate that the id in the header is equal to the agent id record
-	if id != nil && *id != agent.Agent.Id {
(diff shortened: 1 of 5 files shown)


aleksmaus (approved):

ruflin:
@scunningham This is not backported so it is not in the 8.0 or any of the 7.x branches. Which branches should this go into? 7.16 and 8.0?

PR #844, opened 2021-11-05
Improve some of the log message

ruflin:
This was coming out of the debugging session around fleet-server where some of the log messages were not too clear to me on what these mean.

### DIFF
--- a/cmd/fleet/main.go
+++ b/cmd/fleet/main.go
@@ -92,7 +92,7 @@ func initLogger(cfg *config.Config, version, commit string) (*logger.Logger, err
 		Int("ppid", os.Getppid()).
 		Str("exe", os.Args[0]).
 		Strs("args", os.Args[1:]).
-		Msg("Fleet server logger initialised")
+		Msg("Boot fleet-server")
 	log.Debug().Strs("env", os.Environ()).Msg("environment")
 
 	return l, err
--- a/cmd/fleet/main.go
+++ b/cmd/fleet/main.go
@@ -61,7 +61,7 @@ func installSignalHandler() context.Context {
 
 func makeCache(cfg *config.Config) (cache.Cache, error) {
 	cacheCfg := makeCacheConfig(cfg)
-	log.Info().Interface("cfg", cacheCfg).Msg("makeCache")
+	log.Info().Interface("cfg", cacheCfg).Msg("Setting cache config options")
 	return cache.New(cacheCfg)
 }
 
@@ -92,7 +92,7 @@ func initLogger(cfg *config.Config, version, commit string) (*logger.Logger, err
 		Int("ppid", os.Getppid()).
 		Str("exe", os.Args[0]).
 		Strs("args", os.Args[1:]).
-		Msg("boot")
+		Msg("Fleet server logger initialised")
 	log.Debug().Strs("env", os.Environ()).Msg("environment")
 
 	return l, err
--- a/cmd/fleet/router.go
+++ b/cmd/fleet/router.go
@@ -79,7 +79,7 @@ func NewRouter(bulker bulk.Bulk, ct *CheckinT, et *EnrollerT, at *ArtifactT, ack
 
 	// Install routes
 	for _, rte := range routes {
-		log.Info().
+		log.Debug().
 			Str("method", rte.method).
 			Str("path", rte.path).
 			Msg("Server install route")
@@ -90,6 +90,8 @@ func NewRouter(bulker bulk.Bulk, ct *CheckinT, et *EnrollerT, at *ArtifactT, ack
 			logger.HttpHandler(rte.handler),
 		)
 	}
+
+	log.Info().Msg("fleet-server routes set up")
 
 	return router
 }
--- a/cmd/fleet/router.go
+++ b/cmd/fleet/router.go
@@ -79,10 +79,10 @@ func NewRouter(bulker bulk.Bulk, ct *CheckinT, et *EnrollerT, at *ArtifactT, ack
 
 	// Install routes
 	for _, rte := range routes {
-		log.Debug().
+		log.Info().
 			Str("method", rte.method).
 			Str("path", rte.path).
-			Msg("Server install route")
+			Msg("fleet-server route added")
 
 		router.Handle(
 			rte.method,
--- a/internal/pkg/bulk/helpers.go
+++ b/internal/pkg/bulk/helpers.go
@@ -6,6 +6,7 @@ package bulk
 
 import (
 	"encoding/json"
+	"io/ioutil"
 
 	"github.com/elastic/fleet-server/v7/internal/pkg/es"
 	"github.com/elastic/go-elasticsearch/v7/esapi"
@@ -36,6 +37,13 @@ func parseError(res *esapi.Response) error {
 
 	if err := decoder.Decode(&e); err != nil {
 		log.Error().Err(err).Msg("Cannot decode Elasticsearch error body")
+		bodyBytes, readErr := ioutil.ReadAll(res.Body)
+		if readErr != nil {
+			log.Debug().Err(readErr).Msg("Error reading error response body from Elasticsearch")
+		} else {
+			log.Debug().Err(err).Bytes("body", bodyBytes).Msg("Error content")
+		}
+
 		return err
 	}
 
--- a/internal/pkg/bulk/helpers.go
+++ b/internal/pkg/bulk/helpers.go
@@ -35,7 +35,7 @@ func parseError(res *esapi.Response) error {
 	decoder := json.NewDecoder(res.Body)
 
 	if err := decoder.Decode(&e); err != nil {
-		log.Error().Err(err).Msg("Cannot decode error body")
+		log.Error().Err(err).Msg("Cannot decode Elasticsearch error body")
 		return err
 	}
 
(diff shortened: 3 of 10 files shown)


internal/pkg/bulk/helpers.go
```diff
@@ -35,7 +35,7 @@ func parseError(res *esapi.Response) error {
 	decoder := json.NewDecoder(res.Body)
 
 	if err := decoder.Decode(&e); err != nil {
-		log.Error().Err(err).Msg("Cannot decode error body")
+		log.Error().Err(err).Msg("Cannot decode Elasticsearch error body")
```
ruflin, line 38: @scunningham The error I have seen in the logs related to this is as following:

```
invalid character '<' looking for beginning of value
```

WDYT about logging the full response body here so in case of error we can better figure out what happened?

cmd/fleet/router.go
```diff
@@ -79,7 +79,7 @@ func NewRouter(bulker bulk.Bulk, ct *CheckinT, et *EnrollerT, at *ArtifactT, ack
 
 	// Install routes
 	for _, rte := range routes {
-		log.Info().
+		log.Debug().
```
scunningham, line 82: why make these debug?  useful in production to know specifically what was installed.  we have the unfortunate problem that is near to impossible to turn the server to debug at boot time in cloud so we will never see these.

internal/pkg/bulk/helpers.go
scunningham: I'm ok with it.  It's not parsing so no idea what is in the body so reluctant to log out.  Maybe in debug mode.

internal/pkg/coordinator/monitor.go
```diff
@@ -434,7 +434,7 @@ func runUnenroller(ctx context.Context, bulker bulk.Bulk, policyId string, unenr
 		Dur("checkInterval", checkInterval).
 		Dur("unenrollTimeout", unenrollTimeout).
 		Msg("unenroll monitor start")
-	defer l.Info().Msg("unenroll monitor exit")
+	defer l.Info().Msg("Unenroll monitor exit")
```
scunningham, line 437: Are we capitalizing log messages now?  I wish folks would make up their mind.

internal/pkg/monitor/monitor.go
```diff
@@ -186,12 +186,12 @@ func (m *simpleMonitorT) loadCheckpoint() sqn.SeqNo {
 
 // Run runs monitor.
 func (m *simpleMonitorT) Run(ctx context.Context) (err error) {
-	m.log.Info().Msg("start")
+	m.log.Info().Msg("Starting index monitor")
```
scunningham, line 189: Typically add a key to m.log showing context.  Unfortunately elastic logging UI does a not great job of showing other keys besides messages, and won't necessarily index the fields.  This makes it difficult to use structured logging properly.

scunningham (approved):
Looks good.  i only quite on "boot" because I've done that for 10 years.  But fine by me.

internal/pkg/bulk/helpers.go
ruflin: I have the suspicion that the response does not come from Elasticsearch but a proxy or something in between and is why it is not valid.

internal/pkg/coordinator/monitor.go
ruflin: I see, we are not consistent at all. The way I think of it is when I see an log message is that it should be a sentence I read. Sentences start with an uppercase letter. I think that is the way we did it mostly in Beats. But this is total personal preference. Looking at the fleet-server code, we don't do it.

What is your preference and why?

internal/pkg/monitor/monitor.go
ruflin: That is what I struggled during debugging today. There are so many other meta fields, the relevant ones are hard to find. Also I think there is some value to have the most important info also available directly in the message itself (again, personal preference).

internal/pkg/coordinator/monitor.go
scunningham: Honestly don't care.  I prefer starting with a capital letter.  Blake didn't do it that way and  wanted lower case.  I just ask that we are consistent.

internal/pkg/bulk/helpers.go
scunningham: Fine by me.  Have to treat it as a Bytes() payload.

cmd/fleet/router.go
ruflin: As the routes are hardcoded I expect that always all of them are always running and there is no dynamic part to it. Happy to move it back to Info.

cmd/fleet/main.go
```diff
@@ -92,7 +92,7 @@ func initLogger(cfg *config.Config, version, commit string) (*logger.Logger, err
 		Int("ppid", os.Getppid()).
 		Str("exe", os.Args[0]).
 		Strs("args", os.Args[1:]).
-		Msg("boot")
+		Msg("Fleet server logger initialised")
```
ruflin, line 95: @scunningham "Boot fleet server"? There is one problem in Cloud that it is not directly visible if it is an elastic-agent or fleet-server log. Maybe we should add a component or something to each log message?

internal/pkg/monitor/monitor.go
scunningham: I would prefer if elastic logging supported a better way to aggregate actual messages from the application.  All the information is there, it is annoying and wasteful to have to duplicate it because the UI doesn't handle structured logging well.

Zerolog does a nice job with its pretty print on fields specifically from the app.  Would like to see elastic do something like that.

That said, is what it is.

cmd/fleet/main.go
scunningham: You can see it is from the fleet server in one of the other fields. but yeah it is tricky.
We could formalize a component type field, we use it haphazardly in a couple places
context: fleetserver.enroll
Or some such.

cmd/fleet/router.go
scunningham: Sure.  I know what the routes are cause I wrote them.   I think more about support folks in this case.

internal/pkg/monitor/monitor.go
ruflin: @weltenwort Any ideas on how to "fix" the above in the Logs UI?

weltenwort: Yes, actually. But it's a bigger effort, obviously :grimacing: My vision would be that an integration could supply some form of configuration that the Logs UI can use to optimize the UX for these structured logs.

It's beyond of the scope of a comment in a PR review, but with the reconstitution of the infra ui team we might be able to get around to following up on that :crossed_fingers:

cc @mukeshelastic

internal/pkg/policy/monitor.go
```diff
@@ -306,7 +306,7 @@ func (m *monitorT) updatePolicy(pp *ParsedPolicy) bool {
 		Logger()
 
 	if newPolicy.CoordinatorIdx <= 0 {
-		zlog.Info().Msg("ignore policy that has not pass through coordinator")
+		zlog.Info().Str(logger.PolicyId, newPolicy.PolicyId).Msg("ignore policy that has not passed through coordinator")
```
michel-laterman, line 309:
```suggestion
		zlog.Info().Str(logger.PolicyId, newPolicy.PolicyId).Msg("Ignore policy that has not passed through coordinator")
 ```

**Review by michel-laterman — APPROVED** [review]
mostly nitpicks to make changes consistent

File: internal/pkg/monitor/monitor.go (hunk 4 above)
Comment (ruflin) [reply internal/pkg/monitor/monitor.go:189]: @weltenwort Great to hear, please pull me in if this conversation continues somewhere, happy to chime in.

File: internal/pkg/policy/monitor.go (hunk 6 above)
Comment (ruflin) [reply internal/pkg/policy/monitor.go]: Thanks for the fix, applied it directly locally to my diff.

File: internal/pkg/ver/check.go [hunk 7]
```diff
@@ -57,7 +57,7 @@ func checkCompatibility(fleetVersion, esVersion string) error {
 			Msg("failed elasticsearch version check")
 		return ErrUnsupportedVersion
 	}
-	log.Info().Str("fleet_version", fleetVersion).Str("elasticsearch_version", esVersion).Msg("versions are compatible")
+	log.Info().Str("fleet_version", fleetVersion).Str("elasticsearch_version", esVersion).Msg("Elasticsearch compatibility check succseful")
```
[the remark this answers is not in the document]
ruflin: Thanks for the fix, directly applied to my local diff.

internal/pkg/bulk/opRead.go
```diff
@@ -86,7 +86,7 @@ func (b *Bulker) flushRead(ctx context.Context, queue queueT) error {
 	res, err := req.Do(ctx, b.es)
 
 	if err != nil {
-		log.Error().Err(err).Str("mod", kModBulk).Msg("Fail MgetRequest req.Do")
+		log.Error().Err(err).Str("mod", kModBulk).Msg("Error sending Mget request to Elasticsearch")
```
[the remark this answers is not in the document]
ruflin: applied to my diff.

ruflin:
PR reviews added. @michel-laterman Would be great to get another quick look from you on it.

michel-laterman (approved):
lgtm

ruflin:
@Mergifyio backport 8.0

PR #854, opened 2021-11-10
[mergify] use backport- labels

v1v:
### What

Use the `backport-` labels approach similar to Beats, e2e-testing

### Why

Same user experience

### Question

Or we can apply the existing `Fleet-Server` backport labels in Beats and e2e-testing

### DIFF
--- a/.mergify.yml
+++ b/.mergify.yml
@@ -17,7 +17,7 @@ pull_request_rules:
     conditions:
       - merged
       - base=master
-      - label=v8.0.0
+      - label=backport-v8.0.0
     actions:
       backport:
         assignees:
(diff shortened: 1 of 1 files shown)


cachedout (approved):

ruflin (approved):
LGTM. Thanks for the cleanup. This also fixes confusions like https://github.com/elastic/fleet-server/pull/805 The problem was having both labels around.

Also with this change we have a "single purpose" label for backports and potentially conflicting with other use cases of the v labels.

Lets communicate this to the Elastic Agent team as soon as the changes are in.

aleksmaus (approved):

PR #855, opened 2021-11-11
Periodic expired actions cleanup

aleksmaus:

Due to current limitations of the index changes monitoring with utilizing Elasticsearch document sequence numbers for fleet actions, the .fleet-actions index can't be rolled over or have multiple shards.
Until this problem is solved on Elasticsearch side we have to continue using the same single index/shard, that allows the implementation to rely on the guaranteed incremental document sequence number. In order to reduce the number of documents stored in the index over time we need to introduce some kind of cleanup procedure for removal of the old documents.

This PR introduces scheduled cleanup for the expired fleet actions.
The current default is to remove the expired document where the expiration date/time is older than 30 day from the moment the cleanup job runs. The cleanup runs first within about 10 seconds from the start of the fleet server and then about every hour after that, the interval is splayed by 10% by default.

I believe the .fleet-policies would have to do the same, will add to this PR later.

This PR introduces scheduled cleanup for the expired fleet actions.
The current default is to remove the expired document where the expiration date/time is older than 30 day from the moment the cleanup job runs. The cleanup runs first within about 10 seconds from the start of the fleet server and then about every hour after that, the interval is splayed by 10% by default.

## Checklist

### DIFF
--- a/internal/pkg/gc/actions.go
+++ b/internal/pkg/gc/actions.go
@@ -106,14 +106,22 @@ func cleanupActions(ctx context.Context, index string, bulker bulk.Bulk, opts ..
 			if err != nil {
 				// The error is logged
 				log.Debug().Err(err).Msg("failed to delete actions")
+				if eserr, ok := err.(*es.ErrElastic); ok {
+					if eserr.Status == http.StatusNotFound {
+						err = nil
+					}
+				}
+				if err != nil {
+					return err
+				}
 			}
 			for i, r := range res {
 				if r.Error != nil {
 					err = es.TranslateError(r.Status, r.Error)
 					if err != nil {
 						log.Debug().Err(err).Str("action_id", hits[i].Id).Msg("failed to delete action")
-						if r.Status == http.StatusNotFound {
-							err = nil
+						if r.Status != http.StatusNotFound {
+							return err
 						}
 					}
 				}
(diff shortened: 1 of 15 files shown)


internal/pkg/gc/actions.go
```diff
@@ -0,0 +1,134 @@
...
+		}
+
+		if len(hits) == 0 {
+			log.Debug().Msg("no more expired actions found, done cleaning")
+			return nil
+		}
+
+		log.Debug().Int("count", len(hits)).Msg("delete expired actions")
+		for _, hit := range hits {
+			err = bulker.Delete(ctx, index, hit.Id)
```
scunningham, line 100: multi delete? https://github.com/elastic/fleet-server/blob/master/internal/pkg/bulk/opMulti.go#L25

internal/pkg/wait/wait.go
```diff
@@ -0,0 +1,21 @@
...
+// you may not use this file except in compliance with the Elastic License.
+
+package wait
+
+import (
+	"context"
+	"time"
+)
+
+func WithContext(ctx context.Context, d time.Duration) error {
```
scunningham, line 12: Handy.

scunningham (changes requested):

internal/pkg/gc/actions.go
aleksmaus: Updated to use MDelete

scunningham (approved):
👍

james-elastic (approved):

PR #883, opened 2021-11-16
Retry reads when ES unavailable

lykkin:

Fleet server will now gracefully handle situations where it can't issue reads to Elasticsearch. Issue detailed in https://github.com/elastic/obs-dc-team/issues/627.

Wraps the calls responsible for reading from ES in logic that will detect expected availability errors (specifically parsing the error message with a regex) and retry up to a maximum number of times with a linearly growing backoff.

Currently the backoff grows in 30 second increments, up to 5 minutes, and it will attempt 20 times before passing the error back to the caller.

There are unit tests for the new code.
To run a manual integration test: run an ES instance locally, and wire the fleet server up to it. After things have settled, kill the ES instance; log lines explaining the retry actions should appear in fleet server.

## Checklist

## Related issues
Closes https://github.com/elastic/obs-dc-team/issues/627


scunningham:
@lykkin and I discussed this in Slack.  Going to implement the application level retires here and defer the Bulk engine retry logic until we have a better use case.

internal/pkg/bulk/retry.go
```diff
@@ -0,0 +1,61 @@
...
+	if err != nil {
+		log.Warn().Err(err).Msg("Error occured while parsing error message")
+		return false
+	}
+	return match
+}
+
+type retryActionT func() respT
+
+func RetryWithBackoff(action retryActionT, requestRetryLimit int, shouldRetry RequestRetryPredicateT, requestRetryDelayIncrement time.Duration) respT {
```
ruflin, line 27: Do we need a requestRetryLimit? What happens after the limit is reached?

scunningham: As mentioned below, we are going to solve this problem higher up for now.  In the above code, when retry limit it hit the function will error out, and you would still have to handle it upstream.

internal/pkg/coordinator/monitor.go
```diff
@@ -188,10 +189,29 @@ func (m *monitorT) handlePolicies(ctx context.Context, hits []es.HitT) error {
...
+}
+
 // ensureLeadership ensures leadership is held or needs to be taken over.
 func (m *monitorT) ensureLeadership(ctx context.Context) error {
 	m.log.Debug().Msg("ensuring leadership of policies")
 	err := dl.EnsureServer(ctx, m.bulker, m.version, m.agentMetadata, m.hostMetadata, dl.WithIndexName(m.serversIndex))
+
+	// If we are doing a read on an unavailable ES instance ignore it and we will check back later.
+	if isAvailabilityError(err) {
+		log.Info().Err(err).Msgf("Encountered availability error while attempting ES read; continuing to retry.")
```
ruflin, line 211: Could we also log the hostname we tried to connect to?

Lets used Elasticsearch and not ES as an abbreviation as this is user facing.

internal/pkg/coordinator/monitor.go
```diff
@@ -188,10 +189,29 @@ func (m *monitorT) handlePolicies(ctx context.Context, hits []es.HitT) error {
 	return nil
 }
 
+func isAvailabilityError(e error) bool {
+	if e == nil {
+		return false
+	}
+	match, err := regexp.MatchString("connect: connection refused$|net/http: timeout awaiting response headers$", e.Error())
```
ruflin, line 196: This looks potentially error prone. Is there any other way to check for this like status code or similar?

lykkin: yeah, i agree, i suspect it'll be platform dependent, since i believe that these messages come from the OS level.

unfortunately these cases don't make a full http roundtrip, so there aren't proper status codes to check. we could do something internally. for example, we could check for a 0 status code, which is the default the status code will be set to coming out of libbeat.

ruflin:
I want to add here some more general thoughts around error handling in fleet-server. This PR might not solve all of it but we should make sure it is heading in this direction.

fleet-server is a service. In case of errors it should be able to handle it and recover from it eventually. There are 2 types of errors: The known ones and the unknown ones. For the known ones we can have special handling in place to improve the situation but this will not always be possible. For the unknown ones there are 2 options, keep waiting and retrying or stop the service. Stopping the service is radical and it assumes elastic-agent will start a fresh instance of fleet-server. Likely the better option is to just keep waiting and logging what the error is so it can be fixed.

The state we should never end up in is that fleet-server is in a deadlock and the only way out is the user restarting it.

internal/pkg/coordinator/monitor.go
```diff
@@ -157,6 +158,7 @@ func (m *monitorT) handlePolicies(ctx context.Context, hits []es.HitT) error {
 		var policy model.Policy
 		err := hit.Unmarshal(&policy)
 		if err != nil {
+			m.log.Debug().Err(err).Msg("Failed to deserialize policy json")
```
ruflin, line 161: What is the reason you put this on Debug level? Do you expect this to happen often?

internal/pkg/coordinator/monitor.go
```diff
@@ -192,7 +195,9 @@ func (m *monitorT) handlePolicies(ctx context.Context, hits []es.HitT) error {
 func (m *monitorT) ensureLeadership(ctx context.Context) error {
 	m.log.Debug().Msg("ensuring leadership of policies")
 	err := dl.EnsureServer(ctx, m.bulker, m.version, m.agentMetadata, m.hostMetadata, dl.WithIndexName(m.serversIndex))
+
 	if err != nil {
+		m.log.Debug().Err(err).Str("eshost", m.hostMetadata.Name).Msg("Failed to ")
 		return err
```
ruflin, line 201: If this error happens, an error on the info level will be logged anyways. Why not directly return the error with the info from the debug log message inside so that also on the info level we get the elasticsearch host information. With this you can remove the Debug log level.

internal/pkg/coordinator/monitor.go
```diff
@@ -204,6 +209,7 @@ func (m *monitorT) ensureLeadership(ctx context.Context) error {
 			m.log.Debug().Str("index", m.policiesIndex).Msg(es.ErrIndexNotFound.Error())
 			return nil
 		}
+		m.log.Debug().Err(err).Msg("Encountered error while querying policies")
```
ruflin, line 212: Same here as above, it will be logged anyways so you can skip the debug log message and enahnce the error itself.

internal/pkg/coordinator/monitor.go
```diff
@@ -214,6 +220,7 @@ func (m *monitorT) ensureLeadership(ctx context.Context) error {
 		leaders, err = dl.SearchPolicyLeaders(ctx, m.bulker, ids, dl.WithIndexName(m.leadersIndex))
 		if err != nil {
 			if !errors.Is(err, es.ErrIndexNotFound) {
+				m.log.Debug().Err(err).Msg("Encountered error while fetching policy leaders")
```
ruflin, line 223: See above.

ruflin (commented):
Overall change LGTM but left some comments related to logging. In the two error cases, we know continue and keep retrying which I think is good.

As a follow up (separate PR), there is one case I stumbled over on line 114:

```
	// Ensure leadership on startup
	err = m.ensureLeadership(ctx)
	if err != nil {
		return err
	}
```

It seems here we could also end up in a limbo state? Should we fully bail out here? Lets first get the changes in you made and then discuss the above as a follow up.

internal/pkg/coordinator/monitor.go
lykkin: i figured additional information could be placed at specific error points with more information, returning a formatted error would achieve the same thing though. i'll update the pr.

lykkin:
> Overall change LGTM but left some comments related to logging. In the two error cases, we know continue and keep retrying which I think is good.
> 
> As a follow up (separate PR), there is one case I stumbled over on line 114:
> 
> ```
> 	// Ensure leadership on startup
> 	err = m.ensureLeadership(ctx)
> 	if err != nil {
> 		return err
> 	}
> ```
> 
> It seems here we could also end up in a limbo state? Should we fully bail out here? Lets first get the changes in you made and then discuss the above as a follow up.

the block at 114 is done outside the polling loop (i.e. on start up). it seems like bailing out on start up would be reasonable, since it is more likely to be a misconfiguration.

internal/pkg/coordinator/monitor.go
```diff
@@ -132,15 +133,15 @@ func (m *monitorT) Run(ctx context.Context) (err error) {
 		case hits := <-s.Output():
 			err = m.handlePolicies(ctx, hits)
 			if err != nil {
-				return err
+				m.log.Info().Err(err).Msgf("Encountered an error while policy leadership changes; continuing to retry.")
```
michalpristas, line 136: i think these should be at least warn?

internal/pkg/coordinator/monitor.go
```diff
@@ -192,8 +195,9 @@ func (m *monitorT) handlePolicies(ctx context.Context, hits []es.HitT) error {
 func (m *monitorT) ensureLeadership(ctx context.Context) error {
 	m.log.Debug().Msg("ensuring leadership of policies")
 	err := dl.EnsureServer(ctx, m.bulker, m.version, m.agentMetadata, m.hostMetadata, dl.WithIndexName(m.serversIndex))
+
 	if err != nil {
-		return err
+		return errors.New(fmt.Sprintf("Failed to check server status on Elasticsearch (%s): %s", m.hostMetadata.Name, err.Error()))
```
michalpristas, line 200: can we wrap existing error with error message rather than using `err.Error()` this way we may be losing some context

ruflin:
@lykkin Lets take the discussion around `m.ensureLeadership(ctx)` to a separate thread.

internal/pkg/coordinator/monitor.go
lykkin: i was thinking the same, but i followed the [levels in the index monitor](https://github.com/elastic/fleet-server/blob/master/internal/pkg/monitor/monitor.go#L224-L236), though at this point i think these log lines are providing a different function. two independent notes seems like enough to up the log level.

ruflin (approved):
Change LGTM. Lets get this into master and 8.0, do some manual testing with the full build and then backport to 7.16.

@EricDavisX As soon as this gets in, it would be great if the team could do some test with it. The way to confirm this works as expected is setting up an Elastic Agent with fleet-server and temporarly cut the connection to Elasticsearch and let but it up again. Everything should go back to normal afterwards again.

EricDavisX:
Agree on testing and merging backports *after we've seen it working in 8.0 (ideally).  

@dikshachauhan-qasource @amolnater-qasource can you track this and hit it with the next snapshot after merge?

to expand on Nicolas' test info... I'd say we should do it on cloud as well as on-prem.  And in both cases, validate the less-than 5 minute long outage and reconnection and the greater-than 5 minutes off-line reconnect.

amolnater-qasource:
Hi @EricDavisX 
Sure we will run this test whenever new 8.0 Snapshot build will be available.

Thanks

EricDavisX:
It is testable now on 8.0 and 7.16 - Please give it a shot!

amolnater-qasource:
Hi @ruflin @EricDavisX 
We have revalidated this issue on latest self-managed 7.16.0 Snapshot.
**Steps performed and Observations:**
1. Setup 7.16.0 Snapshot self-managed environment.
2. Installed Fleet Server and observed it running Healthy.
3. Terminated elasticsearch.bat and kept it off for 5-7 minutes for first test _(Second test for 25-30 minutes)_ .
4. Restarted elasticsearch.bat and observed below error log:
```
05:40:17.965
elastic_agent
[elastic_agent][error] Could not communicate with fleet-server Checking API will retry, error: status code: 400, fleet-server returned an error: BadRequest
```

For detailed logs, please find below attached debug level logs for self-managed test:
[logs.zip](https://github.com/elastic/fleet-server/files/7641457/logs.zip)

We have done this test for cloud kibana too:
- We restarted the deployment and observed error logs under Logs tab.
- No error logs were observed on restarting kibana deployment.

Further we restarted agent and restarted machine, however we assume that we didn't get expected error logs.

**Query:** Is there any way than restarting deployment to stop elasticsearch on a cloud build?

Logs for cloud build elastic-agent are shared below:
[logs.zip](https://github.com/elastic/fleet-server/files/7641687/logs.zip)

**Build details 7.16.0 self-managed and cloud Kibana:**
Build: 46184
Commit: dfeddea20881915a2654795def35043f6fdc3fc7
Artifact Link: https://snapshots.elastic.co/7.16.0-80ab1dd0/summary-7.16.0-SNAPSHOT.html

Please let us know if we are missing anything.
Thanks

ruflin:
@amolnater-qasource I tested the above on prem today myself. I'm wondering how long you waited for it to become available again. I had the interesting case that the Elastic Agents without fleet-server took about 5 min to show up again and the one with fleet-server took around 10min.

jlind23:
@lykkin could you chime in please?
It does seem to be linked to the "exponential" back off delay so nothing weird for me but I would rather have your acknowledgement.

amolnater-qasource:
Hi @ruflin 
We have tested this upon a Fleet Server, in which we have added integrations like System and Endpoint Security(including Fleet Server).
We are blocked to test multiple agents installed on a fleet-server on self-managed kibana because of reported issue: https://github.com/elastic/beats/issues/28152

Further details for fleet-server(above described) shared below:

<BODY TEXT="#000000">
<TABLE FRAME=VOID CELLSPACING=0 COLS=3 RULES=NONE BORDER=0>
	<COLGROUP><COL WIDTH=86><COL WIDTH=86><COL WIDTH=86></COLGROUP>
	<TBODY>
		<TR>
			<TD STYLE="border-top: 1px solid #000000; border-bottom: 1px solid #000000; border-left: 1px solid #000000; border-right: 1px solid #000000" WIDTH=86 HEIGHT=50 ALIGN=CENTER VALIGN=MIDDLE><B>S.No.</B></TD>
			<TD STYLE="border-top: 1px solid #000000; border-bottom: 1px solid #000000; border-left: 1px solid #000000; border-right: 1px solid #000000" WIDTH=86 ALIGN=CENTER VALIGN=MIDDLE><B>ES off duration</B></TD>
			<TD STYLE="border-top: 1px solid #000000; border-bottom: 1px solid #000000; border-left: 1px solid #000000; border-right: 1px solid #000000" WIDTH=86 ALIGN=CENTER VALIGN=MIDDLE><B>Fleet-Server back Healthy</B></TD>
		</TR>
		<TR>
			<TD STYLE="border-top: 1px solid #000000; border-bottom: 1px solid #000000; border-left: 1px solid #000000; border-right: 1px solid #000000" HEIGHT=18 ALIGN=CENTER VALIGN=MIDDLE SDVAL="1" SDNUM="1033;">1</TD>
			<TD STYLE="border-top: 1px solid #000000; border-bottom: 1px solid #000000; border-left: 1px solid #000000; border-right: 1px solid #000000" ALIGN=CENTER VALIGN=MIDDLE>5-7 minutes</TD>
			<TD STYLE="border-top: 1px solid #000000; border-bottom: 1px solid #000000; border-left: 1px solid #000000; border-right: 1px solid #000000" ALIGN=CENTER VALIGN=MIDDLE>4-5 minutes</TD>
		</TR>
		<TR>
			<TD STYLE="border-top: 1px solid #000000; border-bottom: 1px solid #000000; border-left: 1px solid #000000; border-right: 1px solid #000000" HEIGHT=34 ALIGN=CENTER VALIGN=MIDDLE SDVAL="2" SDNUM="1033;">2</TD>
			<TD STYLE="border-top: 1px solid #000000; border-bottom: 1px solid #000000; border-left: 1px solid #000000; border-right: 1px solid #000000" ALIGN=CENTER VALIGN=MIDDLE>25-30 minutes</TD>
			<TD STYLE="border-top: 1px solid #000000; border-bottom: 1px solid #000000; border-left: 1px solid #000000; border-right: 1px solid #000000" ALIGN=CENTER VALIGN=MIDDLE>5-7 minutes</TD>
		</TR>
	</TBODY>
</TABLE>
</BODY>

</HTML>

Please let us know if anything else is required from our end.
Thanks

ruflin:
@amolnater-qasource Not sure if there is a misunderstanding. When you state above "revalidate issue" do you mean it still exists or it is resolved? My initial understanding was that it still exists even though the fix made it in. But now in the table you show after which time fleet-server is back healthy, so it means it recovers as expected?

amolnater-qasource:
Hi @ruflin 
>When you state above "revalidate issue" do you mean it still exists or it is resolved.

We meant to re-test this ticket, for which we found that Fleet server come back healthy after stopping and starting elasticsearch for the shared duration.

>My initial understanding was that it still exists even though the fix made it in.

We were not sure about what kind of retry error logs from fleet-server we expect which should be there when elasticsearch is stopped.
For this reason we shared the logs at https://github.com/elastic/fleet-server/pull/883#issuecomment-984580294

> the table you show after which time fleet-server is back healthy, so it means it recovers as expected?

Yes fleet-server came back Healthy and working as expected.

Thanks

jlind23:
@lykkin could you also enhance this PR by adding a more appropriate error message as stated by @ruflin?

PR #884, opened 2021-11-16
API keys cleanup

aleksmaus:

Adds the API keys cleanup/invalidation.

1. If the API key needs to be updated, the old API keys are appended to the .fleet-agents ```default_api_key_history``` property.
2. When the policy ack is received the keys from the history are invalidated and the agent document ```default_api_key_history``` is reset

1. Enroll the agent, check the ```default_api_key``` and ```default_api_key_id```
2. Update the policy, by adding new integration. The permissions change will cause the API key to be regenerated and the old keys to be appended to the history. Should be able to observer the ```default_api_key_history``` updated with the previous key before the policy change ack is recieved:

Screen Shot 2021-11-16 at 4 05 14 PM

3. When the policy ack is received, should be able to see in the fleet server log, the ```"Invalidate old API keys"``` message.

Screen Shot 2021-11-16 at 4 06 13 PM

4. Check the agent record again after that, confirm that the history is cleared.

Screen Shot 2021-11-16 at 4 08 37 PM

5. Check the API keys
```
GET /_security/api_key
```
Confirm that the old key id invalidated:
Screen Shot 2021-11-16 at 4 07 38 PM

@scunningham there are still probably some edge cases, but this should improve the situation.
for example:
* what happens if the agent never acks?
* how should we handle invalidation errors? 
* what if only some keys fail to invalidate? how many time should we retry? or how long should we keep the key in the history?

## Checklist

### DIFF
--- a/cmd/fleet/handleAck.go
+++ b/cmd/fleet/handleAck.go
@@ -259,6 +259,18 @@ func (ack *AckT) handlePolicyChange(ctx context.Context, zlog zerolog.Logger, ag
 		return nil
 	}
 
+	sz := len(agent.DefaultApiKeyHistory)
+	if sz > 0 {
+		ids := make([]string, sz)
+		for i := 0; i < sz; i++ {
+			ids[i] = agent.DefaultApiKeyHistory[i].Id
+		}
+		log.Info().Strs("ids", ids).Msg("Invalidate old API keys")
+		if err := ack.bulk.ApiKeyInvalidate(ctx, ids...); err != nil {
+			log.Info().Err(err).Strs("ids", ids).Msg("Failed to invalidate API keys")
+		}
+	}
+
 	body := makeUpdatePolicyBody(
 		agent.PolicyId,
 		currRev,
@@ -360,7 +372,7 @@ func _getAPIKeyIDs(agent *model.Agent) []string {
 //
 // WARNING: This assumes the input data is sanitized.
 
-const kUpdatePolicyPrefix = `{"script":{"lang":"painless","source":"if (ctx._source.policy_id == params.id) {ctx._source.` +
+const kUpdatePolicyPrefix = `{"script":{"lang":"painless","source":"if (ctx._source.policy_id == params.id) {ctx._source.remove('default_api_key_history');ctx._source.` +
 	dl.FieldPolicyRevisionIdx +
 	` = params.rev;ctx._source.` +
 	dl.FieldPolicyCoordinatorIdx +
@@ -371,7 +383,7 @@ const kUpdatePolicyPrefix = `{"script":{"lang":"painless","source":"if (ctx._sou
 func makeUpdatePolicyBody(policyId string, newRev, coordIdx int64) []byte {
 
 	var buf bytes.Buffer
-	buf.Grow(384)
+	buf.Grow(410)
 
 	//  Not pretty, but fast.
 	buf.WriteString(kUpdatePolicyPrefix)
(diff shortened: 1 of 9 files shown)


aleksmaus:
Once we agree on the schema change I'll open PR to elasticsearch to update the agent mapping.

ruflin:
@aleksmaus Are we talking here about the API which communicates to fleet-server or the ES shipping API key or both?

aleksmaus:
> @aleksmaus Are we talking here about the API which communicates to fleet-server or the ES shipping API key or both?

this is for elasticsearch API keys, that we regenerate on each policy permissions change

ruflin:
I like the direction of this approach. Having the information persisted in Elasticsearch around the status of the API keys and some history also allows potential future cleanup for error/edge cases and also gives Fleet visibility into it. This means assuming a user in Fleet would have enough permissions, we could also show mismatches on the UI (@joshdover ).

cmd/fleet/handleAck.go
```diff
@@ -360,7 +372,7 @@ func _getAPIKeyIDs(agent *model.Agent) []string {
 //
 // WARNING: This assumes the input data is sanitized.
 
-const kUpdatePolicyPrefix = `{"script":{"lang":"painless","source":"if (ctx._source.policy_id == params.id) {ctx._source.` +
+const kUpdatePolicyPrefix = `{"script":{"lang":"painless","source":"if (ctx._source.policy_id == params.id) {ctx._source.remove('default_api_key_history');ctx._source.` +
```
scunningham, line 375: I'm sorry you had to see this.  Look away.

cmd/fleet/handleCheckin.go
```diff
@@ -476,10 +478,17 @@ func processPolicy(ctx context.Context, zlog zerolog.Logger, bulker bulk.Bulk, a
 			dl.FieldDefaultApiKeyId:             defaultOutputApiKey.Id,
 			dl.FieldPolicyOutputPermissionsHash: pp.Default.Role.Sha2,
 		}
+		if agent.DefaultApiKeyId != "" {
+			fields[dl.FieldDefaultApiKeyHistory] = model.DefaultApiKeyHistoryItems{
+				ApiKey:    agent.DefaultApiKey,
```
scunningham, line 483: Do we need to keep the old key material in the history?

scunningham (approved):

cmd/fleet/handleCheckin.go
aleksmaus: > Do we need to keep the old key material in the history?

Not really at the moment. Just added all the default_api_key related properties there. Can remove, let me know.

scunningham: Fewer secrets lying around the better I think.

aleksmaus: Done.
Screen Shot 2021-11-17 at 10 20 16 AM

scunningham:
👍

cmd/fleet/handleCheckin.go
aleksmaus: image

model/schema.json
```diff
@@ -442,6 +442,24 @@
           "description": "API key the Elastic Agent uses to authenticate with elasticsearch",
           "type": "string"
         },
+        "default_api_key_history": { 
+          "description": "Default API Key History",
+          "type": "array",
```
ruflin, line 447: Could we maybe set this to index: false instead? https://www.elastic.co/guide/en/elasticsearch/reference/current/mapping-index.html The potential problem I see with this structure if we index is that each entry creates a separate document in lucene which creates unnecessary overhead if we never have to query the data.

aleksmaus: yep. will have that set to false for elasticsearch PR for .fleet-agents mapping.

PR #898, opened 2021-11-19
instrumentation: Add initial APM tracing

### LINKED ISSUE #850: Add APM instrumentation to Fleet Server
**Describe the enhancement:**
Add apm instrumentation to the Fleet Server. At least following configuration options should be supported:
* environment
* hosts
* secret-token | api-key

Turning on APM tracing will help gain more insights into the Fleet Server performance and errors, especially in combination with https://github.com/elastic/beats/issues/28895.

cc @elastic/apm-server @jlind23

marclop:

Adds support to enable instrumentation via the APM Go agent. New config
options have been added to the `Server` input which could be set up in
the `fleet-server` integration configuration.

The added instrumentation covers the `fleet-server` http server and the
`go-elasticsearch` client.

A sample of the configuration that's been added (`instrumentation`):

```yaml
inputs:
  - type: fleet-server
    server:
      host: localhost
      port: 8220
      instrumentation:
        enabled: true
        hosts: ["localhost:8200"]
        environment: production
        secret_token: token
        api_key: apikey
```

By using the `go.elastic.co/apm` client.

1. Build `elastic-agent` and patch it with a locally built version of `fleet-server`.
2. Run fleet server in agent mode.
3. Go to Kibana and install the APM Integration
4. Modify fleet server and add these custom settings:
```yml
server.instrumentation:
  enabled: true
  environment: test
```

image

5. Navigate to the APM App and verify that traces are shown and have the following metadata set:

Screen Shot 2021-11-19 at 7 22 38 PM
image

6. Verify `elasticsearch` shows up as a dependency in the APM Service map

image

## Checklist

~- [ ] I have made corresponding changes to the documentation~
~- [ ] I have added an entry in `CHANGELOG.next.asciidoc` or `CHANGELOG-developer.next.asciidoc`.~

## Related issues

Closes #850


cmd/fleet/main.go
```diff
@@ -860,3 +884,57 @@ func (f *FleetServer) Reload(ctx context.Context, cfg *config.Config) error {
...
+func (f *FleetServer) initTracer(cfg config.Instrumentation) error {
+	// TODO(marclop) is there an APM Go agent environment variable that could
+	// be set? If so, we could use that when cfg.Enabled == false and has not
+	// been specified in the config.
+	if !cfg.Enabled {
+		return nil
+	}
+
+	log.Info().Msg("fleet-server instrumentation is enabled")
+	apm.DefaultTracer.Close()
```
axw, line 897: maybe just do this once in a package-level init func?

cmd/fleet/main.go
```diff
@@ -846,6 +866,10 @@ func (f *FleetServer) runSubsystems(ctx context.Context, cfg *config.Config, g *
 	router := NewRouter(ctx, bulker, ct, et, at, ack, sm)
 
 	g.Go(loggedRunFunc(ctx, "Http server", func(ctx context.Context) error {
+		var router http.Handler = router
+		if f.tracer != nil {
+			router = apmhttp.Wrap(router, apmhttp.WithTracer(f.tracer))
```
axw, line 871: You should use https://github.com/elastic/apm-agent-go/tree/master/module/apmhttprouter so the transaction names are based on the registered route and not the request URL path.

cmd/fleet/main.go
```diff
@@ -860,3 +883,57 @@ func (f *FleetServer) Reload(ctx context.Context, cfg *config.Config) error {
...
+
+	log.Info().Msg("fleet-server instrumentation is enabled")
+
+	transport, err := apmtransport.NewHTTPTransport()
+	if err != nil {
+		return err
+	}
+
+	if cfg.Hosts != nil {
+		if len(cfg.Hosts) > 0 {
```
aleksmaus, line 903:
```
	if cfg.Hosts != nil {
		if len(cfg.Hosts) > 0 {
```
can be replaced with just
```
if len(cfg.Hosts) > 0 {
```

cmd/fleet/main.go
```diff
@@ -783,7 +802,11 @@ func (f *FleetServer) runSubsystems(ctx context.Context, cfg *config.Config, g *
 	g.Go(loggedRunFunc(ctx, "Elasticsearch GC", sched.Run))
 
 	// Monitoring es client, longer timeout, no retries
-	monCli, err := es.NewClient(ctx, cfg, true, es.WithUserAgent(kUAFleetServer, f.bi))
+	options := []es.ConfigOption{es.WithUserAgent(kUAFleetServer, f.bi)}
+	if cfg.Inputs[0].Server.Instrumentation.Enabled {
+		options = append(options, es.InstrumentRoundTripper())
+	}
```
aleksmaus, line 808: nit: adding instrumentation to es client seems repeated in two places, maybe extract into a function?

scunningham:
This seems like the right level of integration.

aleksmaus (approved):
LGTM

cmd/fleet/main.go
```diff
@@ -860,3 +879,63 @@ func (f *FleetServer) Reload(ctx context.Context, cfg *config.Config) error {
 	}
 	return nil
 }
+
+func (f *FleetServer) initTracer(cfg config.Instrumentation) error {
+	// TODO(marclop) is there an APM Go agent environment variable that could
```
axw, line 884: There's two environment variables which may be relevant depending on what you want to do:
 - [`ELASTIC_APM_ACTIVE`](https://www.elastic.co/guide/en/apm/agent/go/current/configuration.html#config-active)
 - [`ELASTIC_APM_RECORDING`](https://www.elastic.co/guide/en/apm/agent/go/current/configuration.html#config-recording)

I guess we want `ELASTIC_APM_ACTIVE`? Not creating a tracer is also a valid choice though.

cmd/fleet/main.go
```diff
@@ -396,6 +404,7 @@ type FleetServer struct {
 	cfgCh    chan *config.Config
 	cache    cache.Cache
 	reporter status.Reporter
+	tracer   *apm.Tracer
```
axw, line 407: Seems to me it would be more fitting with the existing code to pass the tracer into runSubsystems, and have that flush and close it when the context is cancelled.

axw (approved):
Looks good!

cmd/fleet/main.go
marclop: I decided to remove that comment and to not handle any of these environment variables since the tracer constructor will already behave the way we'd like. 
Let me know if you'd rather not create a tracer object when `ELASTIC_APM_ACTIVE` is set to `false`, but the overhead here should be minimal.

axw: SGTM

aleksmaus (approved):
LGTM

PR #960, opened 2021-12-06
instrumentation: Add "tls" configuration options

marclop:

Adds TLS configuration options for the APM instrumentation, using env
vars to configure the APM HTTP Tracer since it currently doesn't support
setting those values in Golang. We'll follow up on this once the apm
tracer has a function to create a new tracer with configurable settings
via config struct.

Adds a TLS config struct and sets the relevant values as env vars for the APM tracer to configure itself.

## Checklist

~- [ ] I have made corresponding changes to the documentation~
~- [ ] I have made corresponding change to the default configuration files~
~- [ ] I have added tests that prove my fix is effective or that my feature works~
~- [ ] I have added an entry in `CHANGELOG.next.asciidoc` or `CHANGELOG-developer.next.asciidoc`.~

## Related issues

Follow up from #898

### DIFF
--- a/cmd/fleet/main.go
+++ b/cmd/fleet/main.go
@@ -895,6 +895,26 @@ func (f *FleetServer) initTracer(cfg config.Instrumentation) (*apm.Tracer, error
 
 	log.Info().Msg("fleet-server instrumentation is enabled")
 
+	// TODO(marclop): Ideally, we'd use apmtransport.NewHTTPTransportOptions()
+	// but it doesn't exist today. Update this code once we have something
+	// available via the APM Go agent.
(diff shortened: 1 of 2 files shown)


stuartnelson3 (approved):

axw (approved):

scunningham (commented):
Looks ok.  Only side effect is that it will step on any existing environment variables with overlapping names.

marclop:
@scunningham I thought about it too, but it seems like it's a non-issue since in general, explicitly configuring something in the file will take precedence over environment variables anyway.

Issue #466, opened 2021-06-16
Introduce "max agents" config options


ruflin:
fleet-server provides quite a list of config options to be tweaked: https://github.com/elastic/fleet-server/blob/master/fleet-server.yml#L24 For most standard users I would assume this is overwhelming. One of the factors driving all these config options is max number of agents that should connect to a fleet-server. This is likely a number that users might know. 

Can we provide good defaults for certain numbers / ranges of "max agents"? If a user sets "100" or "1000" internally the other values are automatically adjusted and optimised.

Of course there is much more variation and optimisation that needs to be done for expert users and these could use the existing config options. The above is to simplify it for most users.

jlind23:
Rescoping it to 8.2 as discussed with @lykkin

Issue #571, opened 2021-07-19
[Fleet] Fleet Server max connections setting is misleading


mostlyjason:
A user filed a discuss issue that highlighted how confusing our settings page is for fleet server scaling https://discuss.elastic.co/t/elastic-agents-error-after-installation-fleet-server-returned-an-error-maxlimit/278716. Changing the max connection setting will not allow more agents to enroll. Since its such a low limit, its very likely that users will run into this problem. The longer term solution is to more gracefully limit connections https://github.com/elastic/fleet-server/issues/153

Can we consider a quick update to the package to make the limits more clear? Users also need to set `server.limits.checkin_limit.max`. We could add an input box for this setting to make it more visible and intuitive. Ideally, the "Max connections" setting should not be set directly by users. It should be calculated as the sum of the max limits for `server.limits.*.max`. If we cannot easily do that in the short term, we could provide input boxes for both of these settings and document how to set them. That will at least make them more visible and help users learn about the difference.

![image](https://user-images.githubusercontent.com/6370704/126210202-a746b192-e0ce-408f-a109-5fecbada2208.png)
![image](https://user-images.githubusercontent.com/6370704/126210229-0f03b1e4-a9c1-4542-bd8b-1ad68cd2c957.png)

CC @nimarezainia @ruflin @scunningham

scunningham:
Thanks @mostlyjason .  This is accurate.

EricDavisX:
linking this to the original 'max settings' ticket which is still open:
https://github.com/elastic/fleet-server/issues/466

nimarezainia:
@mostlyjason I agree with your statement but I wonder if this absolutely needs to be configurable or just clearly stated for the user to know. 

The main variable here the user would need to know about is the number of agents that can enroll into the fleet-server - given resources (cpu/mem) they have made available to it. Number of connections and other artifacts should be calculated or looked up in a table that we have statically created based on our testing numbers.

**Options:**
1) As the user adjusts the slider (allocating more/less resources) this "_number of agents_" also automatically adjusts accordingly.
        - the user workflow for this is: "I have this much resources, how many agents can I get?"
        - it may facilitate more of a trial and error, the user keeps adjusting the slider to see what they get
        - of course the user can look at our documentation, pick the compute resourcing they need for a certain fleet size and then move the slider to that resourcing combination.
        - with metrics based auto-scaling: once we scale horizontally (or vertically even) we are essentially moving the slider which should result in "_number of agents_" changing
 
[...]

PR #1014, opened 2021-12-20
added agent limit configuration setting

lykkin:

[Limit specs](https://docs.google.com/document/d/1ZU5ov0mtu8IMJY0-fGyi7X2hVcB60ntMYGRYWwXU1_o/edit#heading=h.cbor9y67kmyz)

In an attempt to make fleet server sizing more simple, we are introducing more intuitive configuration options. Previously we attempted to size the server based on the amount of RAM available to the machine hosting the server, though this became difficult to configure in the case where we overshot/undershot the estimation.

Adds a configuration setting, `server.limits.max_agents` under the fleet-server input. This will be loaded from the user's config file then used to calculate the proper limits associated with the minimal limit that supports the configured value.

This may be moved over to a continuous implementation that uses formulae per agent to derive the limits.

Modified automated tests to pick up the changes, can be run with `make test`

## Checklist

## Related issues
Closes https://github.com/elastic/fleet-server/issues/466


internal/pkg/config/env_defaults.go
```diff
@@ -156,19 +159,27 @@ func init() {
...
+func loadLimitsForAgents(agentLimit int) *envLimits {
 	for _, l := range defaults {
-		// get max possible config for current env
-		if l.RAM.Min < currentRAM && currentRAM <= l.RAM.Max {
+		// get nearest limits for configured agent numbers
+		if l.Agents.Min < agentLimit && agentLimit <= l.Agents.Max {
+			log.Info().Msgf("Using system limits for %d to %d agents for a configured value of %d agents", l.Agents.Min, l.Agents.Max, agentLimit)
+			ramSize := int(memory.TotalMemory() / 1024 / 1024)
+			if ramSize < l.RecommendedRAM {
+				log.Warn().Msgf("Detected %d MB of system RAM, which is lower than the recommended amount (%d MB) for the configured agent limit", ramSize, l.RecommendedRAM)
```
lykkin, line 177: i attempted to pass the logger down from the call point, but the logger requires the config to be parsed and loaded before it knows how to configure itself. if we are concerned about consistency of logs (i.e. these won't be using the ecs schema), then it will be worth loading the logging config first before the rest of the configuration file. thoughts?

similarly, since this code first runs before the config is first built, these log lines will always show up regardless of configuration settings. will this be potentially confusing?

jlind23:
@lykkin Could you also take care of: https://github.com/elastic/fleet-server/issues/571?

jlind23:
@ph @ruflin can i have your eyes there to make this move?

cmd/fleet/main.go
```diff
@@ -430,8 +434,21 @@ func NewFleetServer(cfg *config.Config, bi build.Info, reporter status.Reporter)
 
 type runFunc func(context.Context) error
 
-// Run runs the fleet server.
+// Run runs the fleet server
 func (f *FleetServer) Run(ctx context.Context) error {
+	// Set up a noop consumer for the ready signal
```
ruflin, line 439: Trying to understand how this change is related?

go.mod
```diff
@@ -30,7 +29,10 @@ require (
 	golang.org/x/time v0.0.0-20210611083556-38a9dc6acbc6
 )
 
-require go.elastic.co/ecszerolog v0.1.0
+require (
```
ruflin, line 32: Why do we multiple require blocks in the go.mod file and why did it change as part of this PR?

internal/pkg/config/defaults/gt10000_limits.yml
```diff
@@ -1,6 +1,7 @@
-ram:
-  min: 8192
-  max: 16384
+num_agents:
```
ruflin, line 1: Can you link to where these numbers are coming from, best in the PR description?

internal/pkg/config/defaults/lte50_limits.yml
```diff
@@ -1,5 +1,6 @@
-ram:
-  max: 1024
+num_agents:
+  max: 50
+recommended_min_ram: 0
```
ruflin, line 3: No ram for 50 agents seems to be a bit too little?

internal/pkg/config/env_defaults.go
```diff
@@ -133,13 +135,14 @@ var defaults []*envLimits
 
 func init() {
 	// Packed Files
-	// internal/pkg/config/defaults/1024_limits.yml
-	// internal/pkg/config/defaults/2048_limits.yml
-	// internal/pkg/config/defaults/4096_limits.yml
-	// internal/pkg/config/defaults/8192_limits.yml
-	// internal/pkg/config/defaults/base_limits.yml
+	// internal/pkg/config/defaults/gt10000_limits.yml
```
ruflin, line 138: I'm ok with hardcoding it for now but it would be nice if we could come up with a formula instead mid term.

cmd/fleet/main.go
```diff
@@ -430,8 +434,21 @@ func NewFleetServer(cfg *config.Config, bi build.Info, reporter status.Reporter)
...
+	c := make(chan struct{})
+	go func() {
+		for range c {
+		}
+	}()
+	return f.RunWithReadyChannel(ctx, c)
+}
+
+// RunWithReadyChannel runs the fleet server with a ready signal channel
+func (f *FleetServer) RunWithReadyChannel(ctx context.Context, readyCh chan struct{}) error {
```
ph, line 449: I don't understand this change, we are creating a channel to delay in some way the startup is that correct?

internal/pkg/config/config.go
```diff
@@ -10,6 +10,7 @@ import (
 	"github.com/elastic/go-ucfg"
 	"github.com/elastic/go-ucfg/flag"
 	"github.com/elastic/go-ucfg/yaml"
+	"github.com/rs/zerolog/log"
```
ph, line 13: I will need to dig into how the logger works and is configured in the fleet-server, something looks weird to me with the `zerolog/ecszerolog` and the `logger package`.

internal/pkg/config/env_defaults.go
ph: This might be indeed confusing for the users, Looking at the logging and how its used quickly I think some work might be needed to clean up the logging initialization.  I would prefer to have log and create an issue for clean up the log in a more general way, WDYT @ruflin?

ph (commented):
I've looked at the code, I don't know all the details and how the limits are selected, and why it's code generated. It seems to me we could also do a better job at validating the values, ie: zero values and negative values. This could lead to a fragile defining of configuration.

I will need to dig deeper in the code to understand it, thanks @jlind23 I have collected a few question I that I will need to followup.

internal/pkg/config/env_defaults.go
ruflin: There definitively seems to be a smell here.

@ph As you know, elastic-agent and fleet-server have a bit a different history. If we can align logging would be great.

cmd/fleet/main.go
lykkin: when adding the limit calculation in response to a config being passed down there was a race condition introduced with one of the integration tests that read off the fleet input. i added the ready channel to the `Run` command so the test could use it, and added this method to interface cleanly with the existing code.

go.mod
lykkin: accidental, i'll revert this change.

internal/pkg/config/defaults/lte50_limits.yml
lykkin: this effectively says there is no minimum, which matches up with the original limit calculation based on resources available. we could have a minimum for 0-50 agents, but there's nothing in the specs about an absolute minimum from what i saw.

internal/pkg/config/env_defaults.go
lykkin: yeah, agreed. i'll have to do some more analysis on how much memory/cpu is used by different configurations of agents.

i'm not certain if it exists in fleet-server, but some kind of circuit breaker code could be helpful if we commonly run into situations where the the server is having some issues/gets swamped. i'm interested in what other folks have to say on the topic, as i'm sure there are some implications around leader election at the least.

cmd/fleet/main.go
lykkin: when adding the limit calculation in response to a config being passed down there was a race condition introduced with one of the integration tests that read off the fleet input. the existing code doesn't really change in behavior (hopefully), as i converted the `Run` method to use this one and just eat all the signals from the ready channel.

i'm not 100% sure if this is the cleanest way to implement this functionality, it was just the first thing that came to mind.

internal/pkg/config/config.go
lykkin: the `zerolog/ecszerolog` package sets up the `zerolog` globals necessary to adhere to the ecs format, so as long as it's added before/alongside the logger then it should be "functional". ideally, i would like to move more toward a dependency injection pattern, where we pass things down through either the context/parameters/data members.

internal/pkg/config/defaults/gt10000_limits.yml
lykkin: 👍 added a link in the description

cmd/fleet/main.go
ph: Can we investigate the root cause, the goroutine/channel will eventually be flaky.

lykkin: I updated the PR to protect the test from the race condition by introducing a mutex to the config. currently it only protects against writes in the methods of the config, and doesn't protect against things reaching directly into the config to set things.

This feels like I'm violating some assumption about how the config object is being built and used, since it wasn't threadsafe to begin with. I'm going to dig around to see if I can replace any direct access of the config with some threadsafe getters/setters, but wanted to see what you thought about the approach.

jlind23:
@ph @lykkin what should we do with this PR? It has been stale for a while now.

lykkin:
> @ph @lykkin what should we do with this PR? It has been stale for a while now.

i'm currently prioritizing the logstash PR over this. i'll revisit to address comments after.

internal/pkg/config/config.go
lykkin: i think punting on the logging updates makes sense for now, since adjusting the bootstrapping process seems out of scope for this change set. WDYT?

ruflin:
Why is this planned to be backported to 8.1? This does not seem to be a bugfix.

jlind23:
Shouldn't be backported to 8.1 indeed.

cmd/fleet/main.go
```diff
@@ -476,12 +480,21 @@ func (f *FleetServer) Run(ctx context.Context) error {
...
+			err = f.reporter.Status(proto.StateObserved_CONFIGURING, "Re-configuring", nil)
 		} else {
 			started = true
-			f.reporter.Status(proto.StateObserved_STARTING, "Starting", nil)
+			err = f.reporter.Status(proto.StateObserved_STARTING, "Starting", nil)
+		}
+
+		if err != nil {
+			return fmt.Errorf("error computing status: %w", err)
+		}
```
ph, line 493: thanks++

lykkin:
For reference, the linting errors aren't related to the changes being made, so I am going to punt on fixing those for now.

ruflin:
@ph @lykkin Why is this backported? Is this a bugfix?

narph (commented):
besides the pending lint errors and the already raised concerns about `go.elastic.co/ecszerolog` (might worth creating an issue in the future for that) it looks good to me

lykkin:
Created https://github.com/elastic/fleet-server/issues/1278 to track the logging issues, figure this will be a good issue to bash out next week during the clean up work

jlind23:
@lykkin Is there anything else pending on this? Changing the targeted release to 8.3
cc @ph

internal/pkg/config/config.go
```diff
@@ -48,8 +64,32 @@ func (c *Config) Validate() error {
...
 }
 
+// LoadServerLimits should be called after initialization, so we may access the user defined
+// agent limit setting.
+func (c *Config) LoadServerLimits() error {
+	c.m.Lock()
+	defer c.m.Unlock()
+	err := c.Validate()
+	if err != nil {
+		log.Warn().Msgf("failed to validate while calculating limits, %s", err.Error())
```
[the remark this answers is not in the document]
lykkin: I agree, since it returns at that point it should be at an error level. Will update.

Issue #1020, opened 2021-12-22
Revamp integration tests to support the username/password removal


aleksmaus:
**Describe the enhancement:**

Revamp integration tests to support the username/password removal
https://github.com/elastic/fleet-server/pull/1006

PR #1039, opened 2021-12-29
Adjust integration tests to make them work with fleet service token.

aleksmaus:

Revamp integration tests to support the username/password removal.
Addresses: https://github.com/elastic/fleet-server/issues/1020

Use username/password auth only for bulker integration tests and for the test setup where we need elevated permissions to be able to do perform indices setup with ```refresh``` option, in order to ensure the correct state of the index/datastream before the integration test

Here is the list of all the changes:
* Update the integration token generating script to cache the token on disk, in order to allow to run individual integration tests repetitively without starting the new instance of elastic search every time. This is huge time saver during development.
* Switch the tests to using existing .fleet-* indices.
* Execute all integration tests serially, added ```-p 1``` test option. This is needed now since the tests in different packages can use shared .fleet-* indices
* Use username/password authenticated elastic search client for cleaning up the .fleet-* indices before running the test. This client is used for the bunker tests as well, since the tests is generic and relies on it’s own test mapping.
* Move out the global checkpoint operations from the monitor package into it’s own checkpt package, and expose the API since this was needed for the actions integration tests which now doesn’t necessarily starts with the new index with -1 checkpoint value.
* Remove ```refresh``` options from migration query. The fleet service token doesn't have permissions for ```refresh``` operations.

## Checklist

## Related issues

- Closes https://github.com/elastic/fleet-server/issues/1020


internal/pkg/dl/migration.go
```diff
@@ -74,7 +74,6 @@ func updateAgentMetadata(ctx context.Context, bulker bulk.Bulk, body []byte) (in
 	opts := []func(*esapi.UpdateByQueryRequest){
 		client.UpdateByQuery.WithBody(reader),
 		client.UpdateByQuery.WithContext(ctx),
-		client.UpdateByQuery.WithRefresh(true),
 		client.UpdateByQuery.WithConflicts("proceed"),
```
aleksmaus, line 77: had to remove 
```
client.UpdateByQuery.WithRefresh(true)
```
The migration is broken without this change. The fleet service token doesn't have permissions for refresh.

aleksmaus:
Will do another pass tomorrow and change this from draft to PR.

internal/pkg/dl/migration.go
scunningham: Probably should comment it out with a warning rather than just delete.

scunningham (changes requested):

aleksmaus:
Created PR for Elasticsearch https://github.com/elastic/elasticsearch/pull/82125. 
Can optimize the integration tests setup code once the ES change lands.

It looks like the ```_bulk``` requests still work without this permission but any other requests fail. That's why most of the code that relies on the ```_bulk``` operations still works.
Example of bulk request/response with service token:
```
POST /_bulk?refresh=true HTTP/1.1
Host: localhost:9200
User-Agent: go-elasticsearch/7.15.1 (darwin amd64; Go 1.17.5)
Content-Length: 323
Authorization: Bearer AAEAAWVsYXN0aWMvZmxlZXQtc2VydmVyL3Rva2VuMTpDV3RsOUtYMlFWV0x0TmtMYkpvMV9R
Content-Type: application/json
X-Elastic-Client-Meta: es=7.15.1,go=1.17.5,t=7.15.1,hc=1.17.5
X-Elastic-Product-Origin: fleet
Accept-Encoding: gzip

{"create":{"_id":"c7679bvblarusq2fasng","_index":".fleet-actions"}}
{"action_id":"ee21d4e3-9cbc-4949-b73c-7f738ff95b41","data":{"ba4cf666-696c-4660-92ba-254bbbf31c90":"582b34f1-2b31-44b7-bfb6-04501d09b478"},"expiration":"2021-12-29T14:51:05Z","input_type":"osquery","@timestamp":"2021-12-29T14:46:05Z","type":"APP_ACTION"}
HTTP/1.1 200 OK
X-elastic-product: Elasticsearch
content-type: application/json
content-encoding: gzip
content-length: 204

{"took":10,"errors":false,"items":[{"create":{"_index":".fleet-actions-7","_id":"c7679bvblarusq2fasng","_version":1,"result":"created","forced_refresh":true,"_shards":{"total":1,"successful":1,"failed":0},"_seq_no":42,"_primary_term":1,"status":201}}]}
```

This seems to be a bug on Elasticsearch side. If it gets fixed it can break fleet server unless we add the permission that allows fleet server ```refresh``` ops.

aleksmaus:
* Updated integration tests setup to use the token for setup.
* Rolled back the change in the migration call, did put the refresh back (@scunningham).
* Tested locally against Elasticsearch with the fleet server token permissions updated.

This PR will fail in CI for now until the Elasticsearch image with the permissions change  is published and picked up for our integration tests.

scunningham (approved):
Nice reWamp.

aleksmaus:
> Nice reWamp.

image

PR #1059, opened 2022-01-04
 Fix 8.0 branch NOTICE.txt

aleksmaus:

Looks like the 8.0 branch is in failed state due to notice update. This fixes it.

Ran ```make check```, updated NOTICE.txt


aleksmaus:
This causes the automation elasticsearch tags update PRs to fail. 
@michel-laterman

scunningham (approved):

PR #1067, opened 2022-01-06
chore: update year 2022

v1v:

Update Copyright year, otherwise the builds will fail since the linting verifies this


jlind23 (approved):

aleksmaus:
Thanks!

PR #1068, opened 2022-01-07
Seperate unenrollment canceller from policy

### LINKED ISSUE #735: [Fleet]: Even after setting Unenrollment timeout field back to blank, installed agents get unenrolled after getting offline.
**Kibana version:**  7.15.0 BC-5 Kibana Cloud environment

**Host OS and Browser version:** All, All

**Build details:**
```
Build: 43957
Commit: 0239ff6864dd9930cfe9bcd9a679272f2b7465c2
Artifact Link: https://staging.elastic.co/7.15.0-edd09abf/downloads/beats/elastic-agent/elastic-agent-7.15.0-windows-x86_64.zip
```

**Preconditions:**
1. 7.15.0 BC-5 Kibana Cloud environment should be available.
2. Agent must be installed with Default policy having System integration.

**Steps to reproduce:**
1. Login to Kibana environment.
2. Set unenrollment timeout from Default Policy setting to 60 seconds.
3. Turn off elastic-agent from services and observe agent goes inactive after 60 seconds.
4. Reset unenrollment timeout field of Default Policy to blank.
5. Re-enroll agent with default policy.
6. Turn off elastic-agent from services and observe agent still gets unenrolled from Fleet UI.

**Expected Result:**
After setting Unenrollment timeout field back to blank, installed agent should not go to inactive state after getting offline.

**Suggestion:**
Checkbox option should be there to use unenrollment timeout field or not.

**Screenshots:**

michel-laterman:

If the unenroll timeout is set, then it applies to instances even after the timeout is cleared (in the fleet-ui).

Track the canceller functions for the unenroller method separately from the policy struct.
This ensures that any running unenroller for a policy is cancelled before a new one starts.
Additionally add an `unenroller_uuid` attribute to (related) loggers to help track the issue. 

1. enroll an agent with the default policy
2. set timeout to 120 (through ui)
3. stop agent and observe that it goes to inactive after the timeout value
4. clear timeout value
5. re-enroll agent, and see that it is healthy in the ui
6. stop agent

Without the fix after the final stop the agent is set to inactive after the old timeout.
With the fix it should appear as offline.

## Checklist

- [] ~~I have made corresponding changes to the documentation~~
- [] ~~I have made corresponding change to the default configuration files~~
- [x] I have added tests that prove my fix is effective or that my feature works **tested manually**
- [] ~~I have added an entry in `CHANGELOG.next.asciidoc` or `CHANGELOG-developer.next.asciidoc`.~~

## Related issues

- Closes #735


internal/pkg/coordinator/monitor.go
```diff
@@ -393,21 +396,26 @@ func (m *monitorT) getIPs() ([]string, error) {
 }
 
 func (m *monitorT) rescheduleUnenroller(ctx context.Context, pt *policyT, p *model.Policy) {
-	l := m.log.With().Str(dl.FieldPolicyId, pt.id).Logger()
+	u := uuid.Must(uuid.NewV4())
+	l := m.log.With().Str(dl.FieldPolicyId, pt.id).Str("unenroller_uuid", u.String()).Logger()
 	unenrollTimeout := time.Duration(p.UnenrollTimeout) * time.Second
 	if unenrollTimeout != pt.unenrollTimeout {
 		// unenroll timeout changed
-		if pt.unenrollCanceller != nil {
```
michel-laterman, line 400: The issue was that this was evauating to nil under one of the initial sets.
I think it's because `ensureLeadership` does not check if a key is valid which may lead to `rescheduleUnenroller` being called with a different `policyT` then is used in `handlePolicies`.
Tracking the canceller functions in a separate map resolves this issue

lykkin: this makes sense to me, but i'm curious if there's a functional reason to have more than one monitor running? would it make more sense to send the `policyT`s in through a channel to handle mutex?

ph:
Is there a way to unit test this change?

michel-laterman:
I'll take a look at the monitor_integration_test file to see if I can add one

michel-laterman:
I could not find out a testcase that fails on master but would succeed with my changes, but i've added some validation for the number of policies and cancellers that are running

ph (approved):

ph:
This look good, can anyone in the review list take another look I am just starting to know part of this code.

internal/pkg/coordinator/monitor.go
```diff
@@ -393,21 +396,24 @@ func (m *monitorT) getIPs() ([]string, error) {
...
-			pt.unenrollCanceller()
-			pt.unenrollCanceller = nil
+		if c, ok := m.policiesCanceller[pt.id]; ok {
+			l.Debug().Msg("cancelling unenrollment monitor")
```
blakerouse, line 405: It would also be created to add more detail on why it was cancelled, say like because timeout changed or because timeout was unset.

internal/pkg/coordinator/monitor.go
```diff
@@ -393,21 +396,24 @@ func (m *monitorT) getIPs() ([]string, error) {
 }
 
 func (m *monitorT) rescheduleUnenroller(ctx context.Context, pt *policyT, p *model.Policy) {
-	l := m.log.With().Str(dl.FieldPolicyId, pt.id).Logger()
+	u := uuid.Must(uuid.NewV4())
+	l := m.log.With().Str(dl.FieldPolicyId, pt.id).Str("unenroller_uuid", u.String()).Logger()
```
blakerouse, line 400: Why the uuid here? Doesn't seem to be used again in the function. The policy ID is already unique.

internal/pkg/coordinator/monitor.go
blakerouse, line 405: It would also be created to add more detail on why it was cancelled, say like because timeout changed or because timeout was unset.

blakerouse (approved):
Looks good, just 2 comments.

The added test looks to add more coverage which is great!

internal/pkg/coordinator/monitor.go
michel-laterman: This UUID was used to track goroutines (for debugging the issue).

blakerouse: Okay. Still don't know if it's best to keep it, as it's just some random ID. But we can keep it.

PR #1101, opened 2022-01-20
Remove `HTTP Done` from the log on every request.

### LINKED ISSUE #1083: Http Done log messages
The http handler in fleet-server logs a "HTTP Done" log message every time a http request was handled: https://github.com/elastic/fleet-server/blob/8a4855bc5687cbc859dc5e82278366509ecf677e/internal/pkg/logger/http.go#L221 It logs additional meta information around response code, duration which is convenient for debugging purpose in case issues happen with fleet-server. At the same time it creates quite a bit of "ground noise" of logging events.

The question I would like to answer in this thread is if we should always log these events by default on the info level or potential reduce it through:

* Only log on the debug level
* Only log on info level if the response code is not 200
* Log a summary / counter of the check ins every x seconds
* Other ideas

Screenshot 2022-01-12 at 10 13 45

ph:
Instead of writing `HTTP Done` on every request successful or not we
will only log when request codes are non 2xx.

Closes #1083

Reduce the cognitive load of dealing with a "HTTP Done log" statement, they were recorded on every request.

## Checklist

### DIFF
--- a/internal/pkg/logger/http.go
+++ b/internal/pkg/logger/http.go
@@ -184,7 +184,7 @@ func httpDebug(r *http.Request, e *zerolog.Event) {
 	}
 }
 
-// HttpHandler returns an httprouter' Handle that wrap request with an ECS logger and
+// HttpHandler returns an httprouter.Handle that wraps the request with an ECS logger and
 // captures metrics for the current request.
 func HttpHandler(next httprouter.Handle) httprouter.Handle {
 	return func(w http.ResponseWriter, r *http.Request, p httprouter.Params) {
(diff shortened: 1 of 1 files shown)


ph:
@AndersonQ and @michel-laterman Could you take a look?

internal/pkg/logger/http.go
```diff
@@ -183,9 +184,9 @@ func httpDebug(r *http.Request, e *zerolog.Event) {
 	}
 }
 
-// ECS HTTP log wrapper
+// HttpHandler returns an httprouter' Handle that wrap request with an ECS logger and
+// captures metrics for the current request.
```
AndersonQ, line 188: typo?

```suggestion
// HttpHandler returns an httprouter.Handle that wraps the request with an ECS logger and
// captures metrics for the current request.
```

AndersonQ (approved):
LGTM

AndersonQ (approved):

Issue #1108, opened 2022-01-26
Add FleetServer version to Status API


thbkrkr:
#### Proposal

Add the current version of FleetServer in the status API as it is for other Elastic products.

Examples:
```
# Elasticsearch
> curl -sku $AUTH https://localhost:9200/ | jq .version
{
  "number": "7.16.1",
  "build_flavor": "default",
  "build_type": "docker",
  "build_hash": "5b38441b16b1ebb16a27c107a4c3865776e20c53",
  "build_date": "2021-12-11T00:29:38.865893768Z",
  "build_snapshot": false,
  "lucene_version": "8.10.1",
  "minimum_wire_compatibility_version": "6.8.0",
  "minimum_index_compatibility_version": "6.0.0-beta1"
}

# Kibana
> curl -sku $AUTH https://localhost:5601/api/status | jq .version
{
  "number": "7.16.1",
  "build_hash": "8751518b4a507d6659d93735b6f90d164514327a",
  "build_number": 46258,
  "build_snapshot": false
}

# EnterpriseSearch
> curl -sku $AUTH https://localhost:3002/api/ent/v1/internal/version | jq .
{
  "number": "7.16.2",
  "build_hash": "0b1d8ae4de8888ad38c18eff255acbacf5eae306",
  "build_date": "2021-12-18T19:56:48Z"
}
```

#### Use-case

Provide a way for ECK to verify the compatibiilty between the version of an Agent managed by ECK and the version of an associated FleetServer (see [ECK documentation to connect an Agent to a FleetServer]( https://www.elastic.co/guide/en/cloud-on-k8s/master/k8s-elastic-agent-fleet-configuration.html#k8s-elastic-agent-fleet-configuration-setting-referenced-resources)) when the FleetServer is not managed by ECK. In this case, ECK only knows the URL, the username, the password and maybe the CA of the FleetServer and could get its version by requesting `/api/status`.

scunningham:
We should not expose this data unauthenticated.  It servers as a fingerprint for an attacker.  If the requests supplies a valid api key, we can append the additional versioning data.

thbkrkr:
Yes, makes sense and not a problem for my use-case.

aleksmaus:
will update the PR with suggested change

ph:
@scunningham Would be good to have a uniform strategy with Elasticsearch.

scunningham:
@ph I believe you need creds in elastic if auth is enabled.  Not sure about kibana.

[...]

PR #1114, opened 2022-01-27
Extend the existing status API with build info

aleksmaus:

Adds additional build info to the status API response,
addresses issue https://github.com/elastic/fleet-server/issues/1108

Adds additional build info to the status API response.

## Checklist

## Related issues
- Closes https://github.com/elastic/fleet-server/issues/1108

## Screenshot

Screen Shot 2022-01-27 at 5 11 46 PM

### DIFF
--- a/cmd/fleet/handleStatus.go
+++ b/cmd/fleet/handleStatus.go
@@ -67,6 +67,10 @@ func withAuthFunc(authfn AuthFunc) OptFunc {
 }
 
 func (st StatusT) authenticate(r *http.Request) (*apikey.ApiKey, error) {
+	// This authenticates that the provided API key exists and is enabled.
+	// WARNING: This does not validate that the api key is valid for the Fleet Domain.
+	// An additional check must be executed to validate it is not a random api key.
+	// This check is sufficient for the purposes of this API
 	return authApiKey(r, st.bulk, st.cache)
 }
 
--- a/cmd/fleet/handleStatus.go
+++ b/cmd/fleet/handleStatus.go
@@ -72,7 +72,7 @@ func (st StatusT) authenticate(r *http.Request) (*apikey.ApiKey, error) {
 
 func (st StatusT) handleStatus(zlog *zerolog.Logger, r *http.Request, rt *Router) (resp StatusResponse, status proto.StateObserved_Status, err error) {
 	limitF, err := st.limit.Acquire()
-	// If failed to aquire limiter send error response
+	// When failing to acquire a limiter send an error response.
 	if err != nil {
 		return
 	}
--- a/cmd/fleet/handleStatus.go
+++ b/cmd/fleet/handleStatus.go
@@ -8,6 +8,7 @@ import (
 	"context"
 	"encoding/json"
 	"net/http"
+	"time"
 
 	"github.com/elastic/fleet-server/v7/internal/pkg/logger"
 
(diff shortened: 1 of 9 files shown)


thbkrkr:
Looks good!

Maybe move `number`, `build_hash` and `build_time` under a `version` field to be more consistent with other Elastic products?
```json
{
  "name": "fleet-server",
  "status": "STARTING",
  "version": {
    "number": "8.1.0",
    "build_hash": "6c4d2ee",
    "build_time": "2022-01-27T22:04:04Z"
  }
}
```

See also this last minute comment https://github.com/elastic/fleet-server/issues/1108#issuecomment-1024240061 to expose this information only when authenticated.

cmd/fleet/server_test.go
```diff
@@ -46,7 +47,7 @@ func TestRunServer(t *testing.T) {
 	et, err := NewEnrollerT(verCon, cfg, nil, c)
 	require.NoError(t, err)
 
-	router := NewRouter(ctx, bulker, ct, et, nil, nil, nil, nil)
+	router := NewRouter(ctx, bulker, ct, et, nil, nil, nil, nil, fbuild.Info{})
 	errCh := make(chan error)
 
 	var wg sync.WaitGroup
```
ph, line 53: @aleksmaus Can you add a test to assert format response, this would prevent us future breaking changes.

ph (changes requested):

ph:
@aleksmaus Ping me when you add the authentication part as requested by @scunningham

aleksmaus:
@scunningham @ph added the the api key check on ```status``` API, output extended build info only for authenticated keys, otherwise only short status info.

Example of requests/responses with and without auth
Screen Shot 2022-01-28 at 1 32 04 PM

Status handler unit test coverage:
Screen Shot 2022-01-28 at 1 31 11 PM

scunningham (approved):
looks ok

cmd/fleet/handleStatus.go
```diff
@@ -8,31 +8,135 @@ import (
...
+	}
+}
+
+func (st StatusT) authenticate(r *http.Request) (*apikey.ApiKey, error) {
+	return authApiKey(r, st.bulk, st.cache)
+}
+
+func (st StatusT) handleStatus(zlog *zerolog.Logger, r *http.Request, rt *Router) (resp StatusResponse, status proto.StateObserved_Status, err error) {
+	limitF, err := st.limit.Acquire()
+	// If failed to aquire limiter send error response
```
ph, line 75:
```suggestion
	// When failing to acquire a limiter send an error response.

File: cmd/fleet/handleStatus.go [hunk 3]
```diff
@@ -8,31 +8,135 @@ import (
...
+func WithAuthFunc(authfn AuthFunc) OptFunc {
+	return func(st *StatusT) {
+		if authfn != nil {
+			st.authfn = authfn
+		}
+	}
+}
+
+func (st StatusT) authenticate(r *http.Request) (*apikey.ApiKey, error) {
+	return authApiKey(r, st.bulk, st.cache)
```
ph, line 70: @aleksmaus I've looked at the implementation of `authApiKey`, this only verifies that the key is active and valid, we do not do any additional checks. I don't think the current information that the Key is protecting is crucial so in that case "it's okay". 

But it would be good to add some godoc here, so if anyone wants to use that authentication methods it knows that it has its limit.

ph (changes requested):
This looks good to me, thanks for adding the testing coverage.

@aleksmaus I've made a few comments can you address them? FYI we will soon add more strict linters rules over this repository and it will mark _Public method_ without any Godoc can you verify if they all need to be public and/or add Godoc statement to them.

cmd/fleet/handleStatus.go
aleksmaus: it's already there
https://github.com/elastic/fleet-server/blob/130056a170a3ca0b3b8ce4df12741ac40ae70c91/cmd/fleet/auth.go#L28

ph (approved):

PR #1164, opened 2022-02-23
Populate action_input_type for the .fleet-actions-results

aleksmaus:

Populates input_type for the .fleet-actions-results. This allows to use kibana transformations on .fleet-actions-results for specific integration without a need to correlate them to the original .fleet-actions (via the action_id)

Copies ```input_type``` from the fleet action to the fleet action response document.

## Checklist

## Related PRs

- Requires https://github.com/elastic/beats/pull/30562

## Screenshot

Screen Shot 2022-02-24 at 1 52 00 PM

### DIFF
--- a/model/schema.json
+++ b/model/schema.json
@@ -82,6 +82,10 @@
           "description": "The action id.",
           "type": "string"
         },
+        "input_type": {
+          "description": "The input type of the original action.",
+          "type": "string"
+        },
         "started_at": {
           "description": "Date/time the action was started",
           "type": "string",
--- a/model/schema.json
+++ b/model/schema.json
@@ -82,7 +82,7 @@
           "description": "The action id.",
           "type": "string"
         },
-        "input_type": {
+        "action_input_type": {
           "description": "The input type of the original action.",
           "type": "string"
         },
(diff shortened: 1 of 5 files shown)


scunningham (approved):

model/schema.json
```diff
@@ -82,6 +82,10 @@
           "description": "The action id.",
           "type": "string"
         },
+        "input_type": {
+          "description": "The input type of the original action.",
+          "type": "string"
+        },
```
ph, line 88: The `input_type` name is unclear to me, without reading the description I thought the `action_result` could have an actual input_type. But this is actually an `ActionInputType`, Is there hard requirement to keep that name?

ph (changes requested):
Naming question

model/schema.json
aleksmaus: no hard requirements, it's just a copy of the `input_type` field from the .fleet-actions, thus it has the same name.
If you have better suggestions for the name let me know, can adjust to whatever name feels better.

ph: `ActionInputType` looks ok to me, I just want to make sure we don't forget it.
Is there any test we could add to protect that feature from any change?

aleksmaus: sounds good, will update to the ```action_input_type```, will check on the test coverage

ph: Thanks @aleksmaus

ph (approved):

PR #1165, opened 2022-02-23
Add Golangci-lint as a linter using the github-action

### LINKED ISSUE #1161: Add linting based on the elastic-agent-libs for the Fleet Server
Now that a base configuration for [golangci-lint](https://golangci-lint.run/) has been created and setup for [elastic-agent-libs](https://github.com/elastic/elastic-agent-libs) we should do the same for this repository.

See https://github.com/elastic/beats/issues/30150 for the related PRs.

ph:
Add Golangci-lint as the linter that will be run on every pull requests,
It uses the github action workflow instead of a Makefile task, this allow [inline
 comments ](https://golangci-lint.run/usage/install/#ci-installation)in the pull-request. The system is configured to only flag new changes instead of
going through the old code.

Limitation, this only run on Linux system. I think this is fine for now, Fleet server doesn't have any platform specific code.

# Follow-up

- Refactor log make check to use golangci-lint.
- Remove make check

## Checklist

## Related issues

- Closes #1161


.github/workflows/golangci-lint.yml
```diff
@@ -0,0 +1,45 @@
...
+      - uses: actions/checkout@v2
+
+      # Uses Go version from the repository.
+      - name: Read .go-version file
+        id: goversion
+        run: echo "::set-output name=version::$(cat .go-version)"
+
+      - uses: actions/setup-go@v2
+        with:
+          go-version: "${{ steps.goversion.outputs.version }}"
```
v1v, line 30: What do you think with this other approach?

https://github.com/elastic/beats/blob/0def363871565f6024f034965fa2a26fa769e51e/.github/workflows/macos-auditbeat.yml#L21-L25

.github/workflows/golangci-lint.yml
```diff
@@ -0,0 +1,45 @@
+name: golangci-lint
+on:
+  push:
+    tags:
+      - v*
```
v1v, line 5: Maybe it's not needed for tags as it happens in branches and pull requests?

.github/workflows/golangci-lint.yml
```diff
@@ -0,0 +1,45 @@
...
+    branches:
+      - main
+      - 8.0
+      - 8.1
+      - 7.17
```
v1v, line 10: A more generic approach? like https://github.com/elastic/beats/blob/0def363871565f6024f034965fa2a26fa769e51e/.github/workflows/macos-auditbeat.yml#L8-L11

```suggestion
    branches:
      - main
      - 8.*
      - 7.17
```

v1v (approved):
Minor comments.

BTW, I've been playing a bit with the github-actions to delegate maybe tiny validations on a PR basis, to help with the non-elastician contributors. See some PRs in the `elastic/elastic-agent.git` as a reference. It's still a PoC and whether we can speed up somehow the feedback by having something hybrid in the CI.

You read my mind, didn't you?

ph:
:) I think we actually read our mind.. I just comment on the other PR and saw this.

ph:
@v1v I think it's a good idea to explore.

.github/workflows/golangci-lint.yml
ph: I think you are right. I will change it.

.github/workflows/golangci-lint.yml
ph: Yes +1 on this, I am still new to action so every comment are welcome. I will make the change.

.golangci.yml
```diff
@@ -0,0 +1,153 @@
...
+    - govet # Vet examines Go source code and reports suspicious constructs, such as Printf calls whose arguments do not align with the format string
+    - ineffassign # detects when assignments to existing variables are not used
+    - structcheck # finds unused struct fields
+    - typecheck # Like the front-end of a Go compiler, parses and type-checks Go code
+    - varcheck # Finds unused global variables and constants
+    - asciicheck # simple linter to check that your code does not contain non-ASCII identifiers
+    - bodyclose # checks whether HTTP response body is closed successfully
+    - durationcheck # check for two durations multiplied together
+    - exportloopref # checks for pointers to enclosing loop variables
+    - goimports # Goimports does everything that gofmt does. Additionally it checks unused imports
```
AndersonQ, line 56: [Suggestion]

Add the `-local` flag to goimports to organise the imports on 3 blocks, std-lib, 3rh party libs, local imports.

So far what I saw seems to kinda follow this approach, but not 100%. Personally I think it's a good way to organise the imports

.golangci.yml
```diff
@@ -0,0 +1,153 @@
...
+
+  nolintlint:
+    # Enable to ensure that nolint directives are all used. Default is true.
+    allow-unused: false
+    # Disable to ensure that nolint directives don't have a leading space. Default is true.
+    allow-leading-space: true
+    # Exclude following linters from requiring an explanation.  Default is [].
+    allow-no-explanation: []
+    # Enable to require an explanation of nonzero length after each nolint directive. Default is false.
+    require-explanation: true
```
AndersonQ, line 132: +1

.golangci.yml
```diff
@@ -0,0 +1,153 @@
...
+    # Enable to require nolint directives to mention the specific linter being suppressed. Default is false.
+    require-specific: true
```
AndersonQ, line 134: +1

AndersonQ (approved):

.golangci.yml
ph: I will not change the format of the linter, I prefer we align with the data plane team, we can create an issue and suggest a change.

AndersonQ (approved):

.github/workflows/golangci-lint.yml
ph: I prefer this way, it feels more github action way.

PR #1166, opened 2022-02-23
Refactor make check

ph:

This code refactor the `make check` task, it split the task into two different concerns:

-  One is `make check-ci` which is run as part of the Jenkins files, this task deals with code generation, headers check, notice generation and make sure all the changes are checked in the VCS.
- The other is `make check` that run all the checks above and also golangci-lint. The pull request linter is doing using a github action to receive all the inline comments.

## Checklist

## Related issues

- Requires #1165
- Relates #1161

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -59,7 +59,7 @@ generate: ## - Generate schema models
 	env PATH="${GOBIN}:${PATH}" go generate ./...
 
 .PHONY: check-ci
-check-ci: ## - Run all check of the ci without linting, the linter is run through github action to have comments in the pull-request.
+check-ci: ## - Run all checks of the ci without linting, the linter is run through github action to have comments in the pull-request.
 	@$(MAKE) generate
 	@$(MAKE) defaults
 	@$(MAKE) check-headers
--- a/Makefile
+++ b/Makefile
@@ -78,7 +78,7 @@ check-headers:  ## - Check copyright headers
 
 .PHONY: check-go
 check-go: ## - Run golangci-lint
-	@curl -sSfL https://raw.githubusercontent.com/golangci/golangci-lint/master/install.sh | sh -s v1.44.2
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -1,4 +1,4 @@
-SHELL=/bin/bash
+SHELL=env bash
```
ph: This is need locally here on my machine, where isolation of packages breaking `/bin/bash`. 
I think this line is not required, but I've used the env` strategy to find where is bash.

Makefile
```diff
@@ -58,25 +58,28 @@ generate: ## - Generate schema models
...
 	@env GOBIN=${GOBIN} go install github.com/elastic/go-licenser@latest
 	@env PATH="${GOBIN}:${PATH}" go-licenser -license Elastic
 
 .PHONY: check-go
-check-go: ## - Run go fmt, go vet, go mod tidy
-	@go fmt ./...
-	@go vet ./...
-	@go mod tidy
+check-go: ## - Run golangci-lint
+	@curl -sSfL https://raw.githubusercontent.com/golangci/golangci-lint/master/install.sh | sh -s v1.44.2
```
v1v, line 81: I wonder if we can use a fixed version instead,for instance the last commit for https://github.com/golangci/golangci-lint/releases/tag/v1.44.2, https://github.com/golangci/golangci-lint/commit/d58dbde584c801091e74a00940e11ff18c6c68bd

```suggestion
	@curl -sSfL https://raw.githubusercontent.com/golangci/golangci-lint/d58dbde584c801091e74a00940e11ff18c6c68bd/install.sh | sh -s v1.44.2
```

v1v (approved):
Only a minor comment regarding the `sha` for the golang-ci-lint tool

Makefile
ph: I was using their recommended way in their documentation to integrate with ci tool. I mean we could, but using the tag seems like solid approach.

Issue #1185, opened 2022-02-28
Fleet server currently requires an elasticsearch output defined on agent policies.


lykkin:
Fleet server should be able to handle output sets that don't include an ES output. Currently, fleet-server will sift through the agent policies while parsing them in order to designate a default policy. This default must be of type elasticsearch.

As part of supporting more output types, it's realistic that a policy won't have an ES output on it, and this behavior should be rectified.

PR #1188, opened 2022-03-01
Logstash output

lykkin:

Issue defined in https://github.com/elastic/fleet-server/issues/1185.

Fleet server currently assumes its default output will have an ES type and uses that information to update the output's API key. This PR separates the output preparation (e.g. API key updating) into its own abstraction that makes handling multiple outputs/different types easier.

Encapsulates the policy output preparation, which discerns its actions based on the type of the output.

More tests to come soon.

## Checklist

## Related issues
- Closes https://github.com/elastic/fleet-server/issues/1185


defensivedepth:
Hello there! 

Will this functionality be available for 7.x?

lykkin:
@defensivedepth Heya, thanks for the interest!

It's not currently scoped to be ported back to 7.x to my knowledge, though @nimarezainia might know more about the more distant horizon plans for backporting.

defensivedepth:
thanks, yes, it would be very helpful to know if the longer-term plan is to backport this to 7.

Thanks!

nimarezainia:
> thanks, yes, it would be very helpful to know if the longer-term plan is to backport this to 7.
> 
> Thanks!

No current plans to backport. Please let us know why you would need this in 7.x - i am currently not sure of dependancies.

defensivedepth:
> > thanks, yes, it would be very helpful to know if the longer-term plan is to backport this to 7.
> > Thanks!
> 
> No current plans to backport. Please let us know why you would need this in 7.x - i am currently not sure of dependancies.

Just needed this information for planning, thanks!

internal/pkg/policy/policy_output_test.go
```diff
@@ -0,0 +1,175 @@
...
+		Type: OutputTypeLogstash,
+		Name: "test output",
+		Role: &RoleT{
+			Sha2: "fake sha",
+			Raw:  TestPayload,
+		},
+	}
+
+	err := po.Prepare(context.Background(), zerolog.Logger{}, bulker, &model.Agent{}, smap.Map{}, false)
+	if err != nil {
```
ph, line 35: I think we have the required module in fleet-server to do `require.NoNit(t, err, "expect....")`, also what is expected here, we check we aren't returning an error but we do not check for any transformation result?

internal/pkg/policy/policy_output_test.go
```diff
@@ -0,0 +1,175 @@
...
+		Name: "test output",
+		Role: nil,
+	}
+
+	err := po.Prepare(context.Background(), zerolog.Logger{}, bulker, &model.Agent{}, smap.Map{}, false)
+	// No permissions are required by logstash currently
+	if err != nil {
+		t.Error("expected prepare to pass")
+	}
+}
```
ph, line 55: What is the expected result from this test?

internal/pkg/policy/policy_output.go
```diff
@@ -0,0 +1,179 @@
...
+					return err
+				}
+				agent.DefaultApiKey = outputAPIKey.Agent()
+			}
+		}
+	case OutputTypeLogstash:
+		zlog.Debug().Msg("preparing logstash output")
+		zlog.Info().Msg("no actions required for logstash output preparation")
+	default:
+		zlog.Debug().Msgf("unknown output type: %s; skipping preparation", p.Type)
```
ph, line 121: Not sure what was the previous behavior, but this looks like we should _return an error_ and log it as an error. I supposed the only way to get into that section of the code is that the format received from Fleet is incorrect.

internal/pkg/policy/policy_output.go
```diff
@@ -0,0 +1,179 @@
...
+			zlog.Error().Str("name", p.Name).Msg("policy does not contain required output permission section")
+			return ErrNoOutputPerms
+		}
+
+		// Determine whether we need to generate an output ApiKey.
+		// This is accomplished by comparing the sha2 hash stored in the agent
+		// record with the precalculated sha2 hash of the role.
+
+		// Note: This will need to be updated when doing multi-cluster elasticsearch support
+		// There will need to be multiple api keys tracked on the agent model
```
ph, line 56: This sentence sounded unclear to me.

internal/pkg/policy/policy_output.go
```diff
@@ -0,0 +1,179 @@
...
+				}
+				if agent.DefaultApiKeyId != "" {
+					fields[dl.FieldDefaultApiKeyHistory] = model.DefaultApiKeyHistoryItems{
+						Id:        agent.DefaultApiKeyId,
+						RetiredAt: time.Now().UTC().Format(time.RFC3339),
+					}
+				}
+				body, err := renderUpdatePainlessScript(fields)
+
+				// Using painless script to append the old keys to the history
```
ph, line 104: We might want to move this before line 102?

internal/pkg/policy/parsed_policy_test.go
```diff
@@ -514,13 +506,462 @@ func TestNewParsedPolicy(t *testing.T) {
...
 		expectedSha2 := "d4d0840fe28ca4900129a749b56cee729562c0a88c935192c659252b5b0d762a"
-		if pp.Default.Role.Sha2 != expectedSha2 {
-			t.Fatal(fmt.Sprintf("Expected sha2: '%s', got '%s'.", expectedSha2, pp.Default.Role.Sha2))
+		if defaultOutput.Role.Sha2 != expectedSha2 {
+			t.Fatal(fmt.Sprintf("Expected sha2: '%s', got '%s'.", expectedSha2, defaultOutput.Role.Sha2))
 		}
 	}
 }
+
+const noESOutputPolicy = `
```
ph, line 521: Can we move that into an external file, This would make json linting and updating it easier?

ph (changes requested):
I've did an initial pass in this PR.

Thanks for fixing some of the linting issues with the capitalization. Since you move code between files it makes it harder for the reviewer to see what logic has changed. From what I see I think this is correct. 

I've looked at the test for logstash (on the prepare operation) we seems to only check for error, should also either look at what should have changed or make sure calls are not executed.

One question concerning the behavior of fleet, when Logstash is used what name would be used in the output block? default, logstash or something else? I am talking about this section of the YAML:


```yaml
output:
  myls: # this part.
    type: logstash
```

internal/pkg/policy/policy_output_test.go
```diff
@@ -0,0 +1,175 @@
...
+		t.Error("expected prepare to pass", err)
+	}
+
+	updatedKey, ok := policyMap.GetMap("test output")["api_key"].(*bulk.ApiKey)
+
+	if !ok {
+		t.Errorf("unable to cast api key")
+	}
+
+	if updatedKey.Key != bulker.MockedAPIKey.Key {
```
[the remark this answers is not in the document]
lykkin: ah, it's in `require`, i thought it was a bit weird that there were no assertions :sweat_smile:

internal/pkg/policy/policy_output_test.go
```diff
@@ -0,0 +1,175 @@
...
+		"test output": map[string]interface{}{},
+	}
+	testAgent := &model.Agent{}
+	err := po.Prepare(context.Background(), zerolog.Logger{}, bulker, testAgent, policyMap, true)
+	if err != nil {
+		t.Error("expected prepare to pass", err)
+	}
+
+	updatedKey, ok := policyMap.GetMap("test output")["api_key"].(*bulk.ApiKey)
+
```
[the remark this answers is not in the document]
lykkin: ignoring causes linting errors, so might as well

internal/pkg/policy/policy_output_test.go
lykkin: similar to the one above, it's just a smoke test for the logstash output with no permissions.

all logstash tests will be smoke tests, as the prepare step doesn't do anything tangible.

lykkin:
> I've did an initial pass in this PR.
> 
> Thanks for fixing some of the linting issues with the capitalization. Since you move code between files it makes it harder for the reviewer to see what logic has changed. From what I see I think this is correct.
> 
> I've looked at the test for logstash (on the prepare operation) we seems to only check for error, should also either look at what should have changed or make sure calls are not executed.
> 
> One question concerning the behavior of fleet, when Logstash is used what name would be used in the output block? default, logstash or something else? I am talking about this section of the YAML:
> 
> ```yaml
> output:
>   myls: # this part.
>     type: logstash
> ```

Yeah, currently all the logstash output tests are simple smoke tests, I could go either way on adding more expectations since they are no ops. I generally dislike looking for negative signals (i.e. something DIDN'T happen) unless there are configurations where they are expected to happen (e.g. update being called in the ES case).

For the naming, it will be `default` for now, but only because that's what the UI supports. There are no restrictions incurred by this code on the name of the output.

ph (changes requested):
LGTM, only missing a changelog!

ph (approved):

PR #1208, opened 2022-03-07
Change ack response, to allow returning errors for individual events. This allows better handling for partial acks.

aleksmaus:

Presently when any of the items in the ack request fail the whole request fails and there is no way for the agent to detect which acks succeeded, which are  permanently failed (400) and which one can be retried (404, 503).

This potential issue was exposed as a part of SDH investigation: https://github.com/elastic/sdh-beats/issues/1770
In this particular case the issue was likely with the action document not being indexed on a secondary replica shard. 
Instead of being able to return "404"/"not found" error for specific action ack and retry it later, the whole request is currently fails with http status code of 400 and the agent retries to resend all acks the next time (potentially including those that succeeded previously already).

There are few issues with the acking processing on the agent side that need to be addressed after this change:
1. There is no logic to recognize the individual events failure and retry them.
2. The actions acks are not persisted, so if the agent is restarted the pending actions ack can get lost and will never retry.
3. One bad action ack that failed to process can "clog" the acking of any other actions, because the in-memory acks queue is not cleaned until the agent is rebooted.

This PR extend the ack response with an array of individual item ack results, the format is consistent with the elasticsearch bulk response, where the ```errors``` boolean flag is set if any of the ack items failed.

```
type AckResponse struct {
	Action string            `json:"action"`
	Errors bool              `json:"errors,omitempty"` // indicates that some of the events in the ack request failed
	Items  []AckResponseItem `json:"items,omitempty"`
}
```

```
type AckResponseItem struct {
	Status  int    `json:"status"`
	Message string `json:"message,omitempty"`
}
```

Each individual ack item contains the code and the message, which is effectively http status codes and the http status code messages (we probably can omit the message fields. opinions?) 
Having the http code is convenient and consistent for the cases when we get the error code from elasticsearch operation so we can just propagate it back as is.

## Checklist

## Related issues

- Relates https://github.com/elastic/sdh-beats/issues/1770

### DIFF
--- a/cmd/fleet/handleAck.go
+++ b/cmd/fleet/handleAck.go
@@ -29,8 +29,6 @@ import (
 	"github.com/rs/zerolog/log"
 )
 
-var ErrEventAgentIdMismatch = errors.New("event agentId mismatch")
-
 type AckT struct {
 	cfg   *config.Server
 	limit *limit.Limiter
@@ -130,11 +128,7 @@ func (ack *AckT) processRequest(zlog zerolog.Logger, w http.ResponseWriter, r *h
 
 	zlog = zlog.With().Int("nEvents", len(req.Events)).Logger()
 
-	if err = ack.handleAckEvents(r.Context(), zlog, agent, req.Events); err != nil {
-		return err
-	}
-
-	resp := AckResponse{"acks"}
+	resp := ack.handleAckEvents(r.Context(), zlog, agent, req.Events)
 
 	data, err := json.Marshal(&resp)
 	if err != nil {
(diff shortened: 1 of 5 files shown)


ph:
@blakerouse Can you take a look at this one, I am not completely familiar yet with the Action/Response?

cmd/fleet/handleAck.go
```diff
@@ -130,11 +128,7 @@ func (ack *AckT) processRequest(zlog zerolog.Logger, w http.ResponseWriter, r *h
 
 	zlog = zlog.With().Int("nEvents", len(req.Events)).Logger()
 
-	if err = ack.handleAckEvents(r.Context(), zlog, agent, req.Events); err != nil {
-		return err
-	}
-
-	resp := AckResponse{"acks"}
+	resp := ack.handleAckEvents(r.Context(), zlog, agent, req.Events)
```
blakerouse, line 131: This change seems to be breaking to me. Being that an error will no longer be returned from `handleAck` doesn't that mean the status code for the whole response will always be a 200, where previously it could have been a 400 or something.

Say an older not updated Elastic Agent goes to Ack a policy and that fails, this change will have it return a 200 and the Elastic Agent will not try to do the action again.

I think this flow needs to be changed to compute the worst HTTP status code from the responses and return that value as the response code for the Ack.

blakerouse (changes requested):

cmd/fleet/handleAck.go
blakerouse: I like #1. That is because a newer Elastic Agent will know to read the body and do more with it, where the older one will not need to it will just keep working how it always has.

Even in the case that a newer Elastic Agent talks to an older Fleet Server (which we don't support), it could be smart and ignore an empty response body.

blakerouse (approved):
Thanks for the changes, looks good.

PR #1259, opened 2022-03-24
Align golangci-lint with elastic-agent/beats

ph:

This run linter per operating system if we ever had specific OS code
defined with different golang build tags.

This also use the `--whole-files` arguments since from our testing
using only the patch strategy was making the linting unreliable.

See https://github.com/elastic/elastic-agent-libs/pull/30 for tests.

## Checklist


michel-laterman (approved):
lgtm

Issue #1256, opened 2022-03-23
Issues with Logstash output on cloud.


ph:
Recent changes to support Logstash (https://github.com/elastic/fleet-server/pull/1188) cause some problems, we have reverted it in https://github.com/elastic/fleet-server/pull/1252
--
Kibana version: 8.2.0-SNAPSHOT

Steps to reproduce:

Create 8.2.0-SNAPSHOT deployment on cloud staging
Ingest some data to APM Server, for example either use the apm-integration-testing repository, or post an [example payload](https://www.elastic.co/guide/en/apm/guide/master/api-event-example.html) via curl curl -i -k -H "Content-type: application/x-ndjson" -H "Authorization: Bearer ${ELASTIC_APM_SECRET_TOKEN}" --data-binary @events.ndjson ${ELASTIC_APM_SERVER_URL}/intake/v2/events
Navigate to https://admin.staging.foundit.no/, find your deployment and open the Agent Logs tab and observe errors like
Failed to connect to backoff(elasticsearch(http://localhost:9200)): Get "http://localhost:9200": dial tcp [::1]:9200: connect: cannot assign requested address
It looks like the APM Server, which is receiving the ES config from Fleet and Elastic Agent is trying to connect to ES via http://localhost:9200.

It also seems that the output handling in Fleet changed in 8.2. Navigating to Fleet/Elastic Agent on Cloud policy/Settings let users choose a different Elasticsearch output. Aside from not being convinced that this should be allowed on cloud, chosing a different output than configured breaks the Integrations Server setup, and a complete restart is required.

lykkin:
after doing more digging, i think the root cause is an inaccuracy in the way the outputs are named. here's a snippet of a config request from the agent before it gets sent down:
```
{AST:{agent:{monitoring:{enabled:true},{logs:true},{metrics:true},{namespace:default},{use_output:default}}},{monitoring_checksum:101498a889672bae78fc69dac1429a0f},{output:{elasticsearch:{api_key:{Id:Bul_un8BfGcRBobzJMTK},{Key:kfOqBxEQQcCR3Fvd5IYjTw}},{hosts:[http://rg.twitter.com:9222]},{type:elasticsearch}}},{programs:[filebeat,fleet-server,metricbeat]}}
```

Note that the `use_output` value is `default`, but the output name is `elasticsearch`.

nchaulet:
[...]

PR #1260, opened 2022-03-24
Logstash output fix

lykkin:
Fixes https://github.com/elastic/fleet-server/issues/1256

This PR reintroduces the policy-output code and fixes a regression in the ES API key assignment to agent policies.

`api_key` should be a string, and instead an object was being assigned instead. The original code path reflects the code changes in the fix commit here:
https://github.com/elastic/fleet-server/blob/7d73399bce26498e99328624271a0e65a68ee323/cmd/fleet/handleCheckin.go#L557 
↓
https://github.com/elastic/fleet-server/blob/7d73399bce26498e99328624271a0e65a68ee323/cmd/fleet/handleCheckin.go#L544
↓
https://github.com/elastic/fleet-server/blob/7d73399bce26498e99328624271a0e65a68ee323/cmd/fleet/handleCheckin.go#L502 
↓
https://github.com/elastic/fleet-server/blob/7d73399bce26498e99328624271a0e65a68ee323/cmd/fleet/handleCheckin.go#L499
↓
https://github.com/elastic/fleet-server/blob/7d73399bce26498e99328624271a0e65a68ee323/cmd/fleet/handleCheckin.go#L465

As a follow up, we should consider some sort of schema validators for the policies.


internal/pkg/policy/parsed_policy_test_data.go
```diff
@@ -0,0 +1,856 @@
...
+
+const noESOutputPolicy = `
```
ruflin, line 449: It seems this test is to check for not having an Elasticsearch output. Why do we add the full list of inputs to it? It seems to make the test example unnecessary complex / long?

ruflin (commented):
Would any of the tests we have in place now have failed with the previous change that had a bug?

lykkin:
unfortunately, there weren't any existing tests to catch this. I believe this specific case would have been caught by the compiler via the function that was responsible for placing the value in the map.

I would like to see more schema tests added around the policy transformations in the future.

internal/pkg/policy/parsed_policy_test_data.go
lykkin: I wanted to keep the additional values on the policy to make sure the parsing code respected them without ES on the policy. This should serve as smoke test when adjusting the logstash transformation logic in the future.

internal/pkg/policy/policy_output.go
```diff
@@ -79,7 +79,7 @@ func (p *PolicyOutput) Prepare(ctx context.Context, zlog zerolog.Logger, bulker
 				return err
 			}
 
-			if ok := setMapObj(outputMap, outputAPIKey, p.Name, "api_key"); !ok {
+			if ok := setMapObj(outputMap, outputAPIKey.Agent(), p.Name, "api_key"); !ok {
 				return ErrFailInjectAPIKey
```
ph, line 83: I am looking at the implementation of the `#Prepare` function, looking at the signature it seems that it should be possible to inject the related dependencies to do the operation. Looking at the problems of using the `object` vs  `.Agent()`  a unit test of the behavior would have caught that. 

This seems to be a must to start adding tests that ensures the behavior of that method since this is a critical path of the Server.

Note this doesn't remove the need to have a stricter checks on the structure of the agent policy.

ph (changes requested):
@lykkin I did a review on this PR, we really need a test on `prepare`, without that everything that impacts playing with the output would be fragile and we cause us to be reactive. 

I am doing a smoke test to on this pr.

internal/pkg/policy/parsed_policy_test_data.go
ruflin: I understand if we keep a few lines for sanity check, but do we need all 300 additional ones?

internal/pkg/policy/policy_output.go
lykkin: There are some tests already in internal/pkg/policy/policy_output_test.go - the issue wasn't caught because the tests were expecting the incorrect value.

internal/pkg/policy/parsed_policy_test_data.go
lykkin: I axed two of the inputs, should be more reasonable now.

ph (approved):
Change LGTM, @lykkin Have you done a complete e2e testing of this?
I am having local docker issues to set up everything I am trying to resovle that.

lykkin:
> Change LGTM, @lykkin Have you done a complete e2e testing of this? I am having local docker issues to set up everything I am trying to resovle that.

yeah, while I was troubleshooting i was testing against an e2e set up. though to be sure, i'll go through the set up procedure again.

lykkin:
tested locally against both LS and ES with no issues.

PR #1275, opened 2022-03-30
Update main to 8.3.0

ph:

Update main to 8.3.0


narph (commented):
> do we automatically update the version in https://github.com/elastic/elastic-agent/blob/main/version/version.go through a different process or should this be done manually?

Nevermind, was looking at a different repo

narph (approved):

PR #1299, opened 2022-04-04
Always add `api_key` to Elasticsearch output

ph:

This fixes a few issues:

- When the hash on the permission and the hash on the agent were the
same the key was not added to the output. This was causing the agent to
not be able to connect to Elasticsearch. If you change an integration
policy field, as the `host` field,  and the data stream are the same the permission hash will
stay the same.

- Fix a possible issue when the new key was not saved on the Agent model, the
the key was only saved when we were targetting a default agent policy.
Currently, we only support a single ES output so this should always be saved.

- Reorganize the test to ensure that all the scenarios from the `case switch` are tested

- Wait for the Agent's status to be Healthy
- Change a integration field value.

## Checklist

~~- [ ] I have made corresponding changes to the documentation~~
~~- [ ] I have made corresponding change to the default configuration files~~
~~- [] I have added an entry in `CHANGELOG.next.asciidoc` or `CHANGELOG-developer.next.asciidoc`.~~

## Related issues

Closes: https://github.com/elastic/elastic-agent/issues/285

### DIFF
--- a/internal/pkg/policy/policy_output.go
+++ b/internal/pkg/policy/policy_output.go
@@ -121,8 +121,8 @@ func (p *PolicyOutput) Prepare(ctx context.Context, zlog zerolog.Logger, bulker 
 		// to Elasticsearch.
 		//
 		// TODO(ph) Investigate allocation with the new LS output, we had optimization
-		// in place to reduce number of agent policy allocation when distribution the updated
-		// keys to multiples agents.
+		// in place to reduce number of agent policy allocation when sending the updated
+		// agent policy to multiple agents.
 		if ok := setMapObj(outputMap, agent.DefaultApiKey, p.Name, "api_key"); !ok {
 			return ErrFailInjectAPIKey
 		}
--- a/internal/pkg/policy/policy_output.go
+++ b/internal/pkg/policy/policy_output.go
@@ -12,13 +12,14 @@ import (
 	"strings"
 	"time"
 
+	"github.com/rs/zerolog"
+
 	"github.com/elastic/fleet-server/v7/internal/pkg/apikey"
 	"github.com/elastic/fleet-server/v7/internal/pkg/bulk"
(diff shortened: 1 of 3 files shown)


internal/pkg/policy/policy_output.go
```diff
@@ -79,42 +80,53 @@ func (p *PolicyOutput) Prepare(ctx context.Context, zlog zerolog.Logger, bulker
 				return err
 			}
 
-			if ok := setMapObj(outputMap, outputAPIKey.Agent(), p.Name, "api_key"); !ok {
-				return ErrFailInjectAPIKey
-			}
+			agent.DefaultApiKey = outputAPIKey.Agent()
 
-			if isDefault {
```
lykkin, line 86: this parameter is no longer referenced here, we should be able to remove it from the signature.

internal/pkg/policy/policy_output.go
```diff
@@ -79,42 +80,53 @@ func (p *PolicyOutput) Prepare(ctx context.Context, zlog zerolog.Logger, bulker
...
-				}
-				agent.DefaultApiKey = outputAPIKey.Agent()
+			if err = bulker.Update(ctx, dl.FleetAgents, agent.Id, body); err != nil {
+				zlog.Error().Err(err).Msg("fail update agent record")
+				return err
 			}
 		}
+
+		// Always insert the `api_key` as part of the output block, this is required
+		// because only fleet server know the api key for the specific agent, if we don't
```
lykkin, line 119:
```suggestion
		// because only fleet server knows the api key for the specific agent, if we don't
```

lykkin (changes requested):
Overall looks good, just some minor clean up.

ph:
@lykkin I've made the changes, I am confident in the unit test, but I doing a smoke test on my side, can you try that too?

lykkin (approved):
tested this locally, does what it says on the the label.

:tada: :rocket: :shipit:

PR #1319, opened 2022-04-06
Refactor: move files from cmd/fleet to pkg/api

michel-laterman:

Start refactoring cmd so api server code is in internal/pkg/api.

This makes it a bit clearer where to find API related code when developing the fleet-server.

## Checklist

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -60,7 +60,6 @@ linters:
     - noctx # noctx finds sending http request without context.Context
     - unconvert # Remove unnecessary type conversions
     - wastedassign # wastedassign finds wasted assignment statements.
-    - godox # tool for detection of FIXME, TODO and other comment keywords
 
 # all available settings of specific linters
 linters-settings:
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -60,6 +60,8 @@ linters:
     - noctx # noctx finds sending http request without context.Context
     - unconvert # Remove unnecessary type conversions
     - wastedassign # wastedassign finds wasted assignment statements.
+    #- godox # tool for detection of FIXME, TODO and other comment keywords
+
 
 # all available settings of specific linters
 linters-settings:
@@ -139,6 +141,8 @@ linters-settings:
   stylecheck:
     # Select the Go version to target. The default is '1.13'.
     go: "1.17"
+    initialisms: ["ACL", "API", "ASCII", "CPU", "CSS", "DNS", "EOF", "GUID", "HTML", "HTTP", "HTTPS", "ID", "IP", "JSON", "QPS", "RAM", "RPC", "SLA", "SMTP", "SQL", "SSH", "TCP", "TLS", "TTL", "UDP", "UI", "GID", "UID", "UUID", "URI", "URL", "UTF8", "VM", "XML", "XMPP", "XSRF", "XSS", "ECS"]
+
 
   unparam:
     # Inspect exported functions, default is false. Set to true if no external program/library imports your code.
--- a/cmd/fleet/main.go
+++ b/cmd/fleet/main.go
@@ -2,10 +2,12 @@
 // or more contributor license agreements. Licensed under the Elastic License;
 // you may not use this file except in compliance with the Elastic License.
 
+// Package fleet is the main entry point for fleet-server.
 package fleet
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"io"
 	"net/url"
@@ -44,7 +46,6 @@ import (
 	"github.com/elastic/fleet-server/v7/internal/pkg/ver"
 
 	"github.com/hashicorp/go-version"
-	"github.com/pkg/errors"
 	"github.com/rs/zerolog"
 	"github.com/rs/zerolog/log"
 	"github.com/spf13/cobra"
@@ -112,7 +113,7 @@ func initLogger(cfg *config.Config, version, commit string) (*logger.Logger, err
 
 func getRunCommand(bi build.Info) func(cmd *cobra.Command, args []string) error {
 	return func(cmd *cobra.Command, args []string) error {
-		cfgObject := cmd.Flags().Lookup("E").Value.(*config.Flag)
+		cfgObject, _ := cmd.Flags().Lookup("E").Value.(*config.Flag)
 		cliCfg := cfgObject.Config()
 
 		agentMode, err := cmd.Flags().GetBool(kAgentMode)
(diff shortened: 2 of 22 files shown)


michel-laterman:
One more thing we should consider with this cleanup: should we change the scope of some of the exported symbols in `api` to be private? I

ph:
@michel-laterman I think we should limit what we export, but I think we should do it in a separate PR.

.golangci.yml
```diff
@@ -60,7 +60,6 @@ linters:
     - noctx # noctx finds sending http request without context.Context
     - unconvert # Remove unnecessary type conversions
     - wastedassign # wastedassign finds wasted assignment statements.
-    - godox # tool for detection of FIXME, TODO and other comment keywords
```
ph, line 63: Disable it with a comment instead of removing the line.

cmd/fleet/main.go
```diff
@@ -478,10 +481,10 @@ LOOP:
 		ech := make(chan error, 2)
 
 		if started {
-			f.reporter.Status(proto.StateObserved_CONFIGURING, "Re-configuring", nil)
+			_ = f.reporter.Status(proto.StateObserved_CONFIGURING, "Re-configuring", nil)
```
ph, line 484: I would prefer we disable the linter using comments (`nolint` instead of using the `_` to not handle the error.
Because we can search for `nolint` and do the right change here.

If we were handling the error here what would be the course of action?
- Fails?
- return error?
- Should we log it? and if yes at what log level? if it's not in debug is this situation actionable by the user?

cmd/fleet/main.go
```diff
@@ -709,12 +713,14 @@ func (f *FleetServer) runServer(ctx context.Context, cfg *config.Config) (err er
...
 	// The metricsServer is only enabled if http.enabled is set in the config
-	metricsServer, err := f.initMetrics(ctx, cfg)
+	metricsServer, err := api.InitMetrics(ctx, cfg, f.bi)
 	switch {
 	case err != nil:
 		return err
 	case metricsServer != nil:
-		defer metricsServer.Stop()
+		defer func() {
+			_ = metricsServer.Stop()
```
ph, line 722: yeah its fine here we can't do anything since we are shutting down. 👍🏽

internal/pkg/api/auth.go
```diff
@@ -83,14 +84,14 @@ func authAgent(r *http.Request, id *string, bulker bulk.Bulk, c cache.Cache) (*m
...
-	key, err := authApiKey(r, bulker, c)
+	key, err := authAPIKey(r, bulker, c)
 	if err != nil {
 		return nil, err
 	}
 
 	w := log.With().
 		Str(LogAccessAPIKeyID, key.Id).
-		Str(EcsHttpRequestId, r.Header.Get(logger.HeaderRequestID))
+		Str(EcsHTTPRequestID, r.Header.Get(logger.HeaderRequestID))
```
ph, line 94: Should this be `ECSHTTPRequestID` since Ecs is the Elastic Common Schema?
I presume `Http` and `Api` are detected because it's coming from a well known dictionary.

ph (changes requested):
Just a few small changes. thanks for making this!

internal/pkg/api/auth.go
michel-laterman: I'll add it to the linter

ph (approved):
This LGTM, this will go in 8.3.0

internal/pkg/api/error.go
```diff
@@ -33,22 +33,24 @@ const (
 	LogAccessAPIKeyID = logger.AccessApiKeyId
 )
 
-type errResp struct {
+// ErrResp is an HTTP error response
+type ErrResp struct {
```
AndersonQ, line 37: [Suggestion]
If it's specific for HPPT, wouldn't be better to call it `ErrHTTPResp`?

internal/pkg/api/handleCheckin.go
```diff
@@ -40,7 +40,7 @@ import (
 var (
 	ErrAgentNotFound    = errors.New("agent not found")
 	ErrNoPolicyOutput   = errors.New("output section not found")
-	ErrFailInjectApiKey = errors.New("fail inject api key")
+	ErrFailInjectAPIKey = errors.New("fail inject api key")
```
AndersonQ, line 43: [NIT]
Perhaps `failed to inject api key` or `failure injecting api key`? They feel, to me, more natural to read.

internal/pkg/api/handleCheckin.go
```diff
@@ -210,7 +210,12 @@ func (ct *CheckinT) processRequest(zlog zerolog.Logger, w http.ResponseWriter, r
 	if err != nil {
 		return errors.Wrap(err, "subscribe policy monitor")
 	}
-	defer ct.pm.Unsubscribe(sub)
+	defer func() {
+		err := ct.pm.Unsubscribe(sub)
+		if err != nil {
+			zlog.Error().Err(err).Str("policy_id", agent.PolicyId).Msg("unable to unsubscribe")
```
AndersonQ, line 216: [Suggestion]
Add a bit more context to the message, so even who is unfamiliar with the agent or this error will have a more complete understand of the problem when reading the log.

```suggestion
			zlog.Error().Err(err).Str("policy_id", agent.PolicyId).Msg("unable to unsubscribe from policy")
```

internal/pkg/api/server.go
```diff
@@ -32,13 +33,14 @@ func diagConn(c net.Conn, s http.ConnState) {
...
-		cntHttpNew.Inc()
+		cntHTTPNew.Inc()
 	case http.StateClosed:
-		cntHttpClose.Inc()
+		cntHTTPClose.Inc()
 	}
 }
 
-func runServer(ctx context.Context, router http.Handler, cfg *config.Server) error {
+// Run runst the passed router with the config.
```
AndersonQ, line 42: typo?
```suggestion
// Run runs the passed router with the config.
```

AndersonQ (approved):
I added a few suggestions. but no blockers

cmd/fleet/main.go
```diff
@@ -478,10 +481,10 @@ LOOP:
 		ech := make(chan error, 2)
 
 		if started {
-			f.reporter.Status(proto.StateObserved_CONFIGURING, "Re-configuring", nil)
+			f.reporter.Status(proto.StateObserved_CONFIGURING, "Re-configuring", nil) //nolint:errcheck // unclear on what should we do if updating the status fails?
```
lykkin, line 484: maybe we log out the error on status update failure? having evidence of failure would help diagnose status desync with the daemon.

lykkin (approved):
changes LGTM!

PR #1352, opened 2022-04-14
Fleet-server package cache implementation

michel-laterman:

If too many elastic-agents upgrade they may exhaust network resources.
By having fleet-server cache artifacts, it can assist with air-gapped deployments, and control the bandwidth used to serve artifacts. Coordination of agents will be done through other means.

Allow the fleet-server to serve packages that the elastic-agent uses for
upgrades. The api endpoint is `GET /api/downloads/*path`. This endpoint
needs a valid API key. The endpoint for packages may be configured with
an associated concurrency limit and a bandwidth limit. The cache may have
an associated retention duration. When the fleet-server needs to retrieve a
package from the upstream source it will download the associated .sha512
and .asc files in order to verify the package. Upstream retrieval is a 1-1
mapping. i.e., `/api/downloads/some/artifact/package` will retrieve
`upstreamURI/downloads/some/artifact/package` validate it and store it on
the local filesystem as `cacheLocation/some/artifact/package`.

1. Start elasticsearch and kibana on the local machine
2. Create a service token and set the env var `ELASTICSEARCH_SERVICE_TOKEN`
3. Create a `fleet-server.yml` with
```yaml
output:
  elasticsearch:
    hosts: '${ELASTICSEARCH_HOSTS:localhost:9200}'

fleet:
  agent:
    id: 1e4954ce-af37-4731-9f4a-407b08e69e42
    logging:
      level: '${LOG_LEVEL:DEBUG}'

inputs:
  - type: fleet-server
      server:
        host: localhost
        port: 8220
      package_cache:
        enabled: false
        retention_period: -1
        upstreamURI: https://artifacts.elastic.co
        bandwidth_limit: -1
        concurrent_limit: -1

logging:
  to_stderr: true
http.enabled: true
```
4. Start a fleet server by running `./fleet-server -c fleet-server.yml`
5. Get an API Key from elasticsearch and it it to the env var `API_KEY`
6. Curl the fleet-server for an artifact: `curl  -H 'Authorization: ApiKey $API_KEY' localhost:8220/api/downloads/beats/elastic-agent/elastic-agent-8.1.2-linux-x86_64.tar.gz -o elastic-agent-8.1.2-linux-x86_64.tar.gz`
7. Check that the downloaded artifact is valid
8. Check the `fleet-package-cache` dir for the `.asc` and `.sha512` files

## Checklist


michel-laterman:
Changes from the proposal so far:
1. Config is part of the fleet-server input type instead of purely defined in `fleet.yml`. This was done to simplify how config is defined within the fleet-server code base. It also has a side effect that the agent should not need to have any code changes.
2. Route path has changed from `/api/downloads/:artifact/:package` to `/api/downloads/*path`. This is to simplify (upstream) retrieval logic as now the fleet-server does not need to insert `beats` in the path. It can also cache any other download from the `artifacts.elastic.co` server.

ph:
@jlind23 @nimarezainia Adding discussion from our last weekly. We are gonna hold on to this PR, Nima is looking at the requirement and we will revisit the implementation.

jlind23:
@nimarezainia @ph any update there?

internal/pkg/api/downloadVerifier.go
```diff
@@ -0,0 +1,140 @@
...
+			continue
+		}
+
+		lineFilename := strings.TrimLeft(parts[1], "*")
+		if lineFilename != filename {
+			// Continue looking for a match.
+			continue
+		}
+
+		sum = parts[0]
```
aleksmaus, line 107: do you want to break out of the loop at this point? since the filename match is found.

internal/pkg/api/downloadVerifier_test.go
```diff
@@ -0,0 +1,126 @@
...
+	t.Helper()
+	// sign file with new key
+	e, err := openpgp.NewEntity("testing", "", "", nil)
+	if err != nil {
+		t.Fatal(err)
+	}
+	f, err := os.Open(path)
+	if err != nil {
+		t.Fatal(err)
+	}
```
aleksmaus, line 50: strictly speaking ```defer f.Close()```

internal/pkg/api/downloadVerifier_test.go
```diff
@@ -0,0 +1,126 @@
...
+	ascF, err := os.Create(path + ".asc")
+	if err != nil {
+		t.Fatal(err)
+	}
```
aleksmaus, line 54: ```defer ascF.Close()``` ?

internal/pkg/api/handleDownload.go
```diff
@@ -0,0 +1,387 @@
...
+	d.root = cfg.PackageCache.Cache
+	d.lock = &sync.RWMutex{}
+	d.retention = cfg.PackageCache.RetentionPeriod
+
+	// if there is a retention period, need to check if existing packages are too old
+	if d.retention > 0 {
+		// filepath.WalkDir is used instead of fs.WalkDir as we want to be able to access the full file path.
+		err := filepath.WalkDir(d.root, func(fPath string, entry fs.DirEntry, err error) error {
+			if err != nil {
+				log.Error().Err(err).Str("path", fPath).Msg("unable to scan")
```
aleksmaus, line 116: return after this?

internal/pkg/api/handleDownload.go
```diff
@@ -0,0 +1,387 @@
...
+			if rErr := d.removeFile(filePath); rErr != nil {
+				log.Error().Err(rErr).Msg("unable to remove file")
+			}
+			return fmt.Errorf("package verification failed: %w", err)
+		}
+	}
+
+	// delete package later
+	if d.retention > 0 {
+		go d.startDeletionTimer(d.retention, filePath)
```
aleksmaus, line 296: this would cause multiple go routines scheduled to do deletions depending on timing, right? potentially as many as the getPackage calls?
maybe the better approach would be to expire the files on the next access, check the mod time?
and possible one go routine periodically managing expiration cleanups?

internal/pkg/api/handleDownload.go
```diff
@@ -0,0 +1,387 @@
...
+func (d *Downloader) startDeletionTimer(dur time.Duration, path string) {
+	t := time.NewTimer(dur)
+	select {
+	case <-t.C:
+		if err := d.removeFile(path); err != nil {
+			log.Error().Err(err).Msg("unable to remove file")
+		}
+	case <-d.ctx.Done():
+		t.Stop()
+	}
```
aleksmaus, line 312: nit: maybe?
```
	t := time.NewTimer(dur)
        defer t.Stop()
	select {
	case <-t.C:
		if err := d.removeFile(path); err != nil {
			log.Error().Err(err).Msg("unable to remove file")
		}
	case <-d.ctx.Done():
                break
	}
```

internal/pkg/api/handleDownload.go
```diff
@@ -0,0 +1,387 @@
...
+	return nil
+}
+
+// getFile retrieves the file at url and writes the body to path.
+// If path contains subdirs they are created as well.
+// The file is opened with create/write/truncate flags.
+func (d *Downloader) getFile(url, path string) error {
+	d.lock.Lock()
+	defer d.lock.Unlock()
```
aleksmaus, line 345: nit: maybe can be improved to allow multiple downloads at the time, lock per path?

internal/pkg/api/handleDownload.go
```diff
@@ -0,0 +1,387 @@
...
+	rErr := &RetrievalError{url: url}
+
+	// TODO maybe check if the path exists?
+	err := os.MkdirAll(filepath.Dir(path), 0750)
+	if err != nil && !errors.Is(err, fs.ErrExist) {
```
aleksmaus, line 350: not sure if MkdirAll will ever return fs.ErrExist, maybe can remove ```&& !errors.Is(err, fs.ErrExist)```


```If path is already a directory, MkdirAll does nothing and returns nil.```
https://cs.opensource.google/go/go/+/refs/tags/go1.18.2:src/os/path.go;l=18

File: internal/pkg/api/handleDownload.go [hunk 9]
```diff
@@ -0,0 +1,387 @@
...
+		rErr.err = fmt.Errorf("failed to create directories: %w", err)
+		return rErr
+	}
+
+	// TODO build an http client with httpcommon for proxy/tls support?
+	client := &http.Client{}
```
aleksmaus, line 356: timeout is not set?

internal/pkg/api/handleDownload.go
```diff
@@ -0,0 +1,387 @@
...
+		rErr.err = fmt.Errorf("unable to create request: %w", err)
+		return rErr
+	}
+	resp, err := client.Do(req)
+	if err != nil {
+		rErr.err = err
+		return rErr
+	}
+	defer func() {
+		_ = resp.Body.Close()
```
aleksmaus, line 368: nit: quiet the linter instead here maybe?

AndersonQ:
@michel-laterman is this PR still valid? if not could we either close it or make it a draft?

michel-laterman:
We'll close it; i don't think we actually reached a decision on how this will be implemented

PR #1381, opened 2022-04-22
Add start_time and minimum_execution_duration fields to actions

### LINKED ISSUE #1380: Fleet server should be able to schedule future actions
**Describe the enhancement:**

Fleet server actions should have start_time and minimum_execution_duration attributes that can be used to schedule an action in the future.

Scheduling future actions should spread out start times within the start:expiration range when given multiple agent IDs

In order to support agent upgrades we need the ability to control when upgrades occur, and also to roll out upgrades over potentially thousands of agents without exhausting network resources.

michel-laterman:

Fleet-server should be able to inform agents of actions that are scheduled to start at a later time.
Fleet-server should also be able to spread the start time that different agents within the window of start:expiration.

This should help with ensuring that hundreds (or thousands) of agents don't exhaust the network when upgrading (for instance).

Add start_time and minimum_execution_duration fields to action
definition. When start_time and expiration are defined the fleet-server
will schedule agent actions within [start:expiration], if the
minimum_execution_duration s defined the scheduling range is
[start:expiration-minimum_execution_duration]. An agent's scheduled time
within the range depends on the agent id's position within the list.

See testing in https://github.com/elastic/elastic-agent/pull/419

## Checklist

## Related issues

- Closes #1380


internal/pkg/action/dispatcher.go
```diff
@@ -96,31 +97,63 @@ func (d *Dispatcher) process(ctx context.Context, hits []es.HitT) {
...
 
-func (d *Dispatcher) getSub(agentId string) (Sub, bool) {
+// offsetStartTime will return a new start time between start:end-dur based on index i and the total number of agents
+// An empty string will be returned if start or exp are empty or if there is an error parsing inputs
+// As we expect i < total  the latest return time will always be < exp-dur
+func offsetStartTime(start, exp, dur string, i, total int) string {
+	if start == "" || exp == "" {
+		return ""
+	}
+	startTS, err := time.Parse(time.RFC3339, start) // TODO what format does a date-time string use?
```
AndersonQ, line 123: Are you asking which is the format used when converting a time.Time to JSON? Or that it needed to double check with the agent to ensure the're using the same format?

AndersonQ (changes requested):
I left a question. But LGTM

internal/pkg/action/dispatcher.go
michel-laterman: I'm not sure what format Elasticsearch serializes date attributes as, that's what the TODO was for.

ph, line 123: Shouldn't we move the validation to the original `unpack` of the action instead? If there is an `unserialization error` this means that the contract between fleet and fleet-server is broken. I don't think we should even try to process that event.

michel-laterman: To summarize a quick conversation. We should catch this on the json unpacking, however the `model.Action` is generated by `elastic/go-json-schema-generate` which does not support mapping a (json) `string/date-time` to go's `time.Time`. The upstream generator we forked our code from has this as an open issue https://github.com/a-h/generate/issues/69 so if we get around to it we should contribute it back

ph: IIRC it's RFC3339

ph (changes requested):
@michel-laterman Changes LGTM, it's there a way to test it end to end with Elasticsearch with an integration test?

AndersonQ (approved):

PR #1424, opened 2022-05-04
Linter fixes and cleanup

michel-laterman:

There's a lot of missing doc strings for fleet-server that makes learning what everything does more of a challenge.
Some of the unit tests that use mocks, don't really assert method calls or args (or return anything).
Also the linter screams at us.

- linting fixes
- provide godoc strings
- minor cleanup of testing
- get rid of pkg/testutil (used in single test file)
- get rid of pkg/wait (used in single test file)
- move pkg/rnd under pkg/testing
- use testify/mock in plg/testing/bulk.go - assert on expected calls
  that use the mock
- provide testing/log to stop log output from appearing when tests
  are ran


internal/pkg/api/handleAck_test.go
```diff
@@ -101,15 +101,6 @@ func TestEventToActionResult(t *testing.T) {
...
 }
 
-// Mock bulker
-type mockBulk struct {
-	ftesting.MockBulk
-	searchErr error
-	createErr error
-	updateErr error
-	actions   []model.Action
-}
```
michel-laterman, line 111: I changed `ftesting.MockBulk` to use `testify/mock.Mock` this allows us to remove the multiple wrappers around it in different tests, and lets us define exactly what calls we expect when interacting with `bulk.Bulk` in tests

internal/pkg/api/handleStatus.go
```diff
@@ -60,15 +60,7 @@ func NewStatusT(cfg *config.Server, bulker bulk.Bulk, cache cache.Cache, opts ..
 	return st
 }
 
-func withAuthFunc(authfn AuthFunc) OptFunc {
```
michel-laterman, line 63: moved to `handleStatus_test.go` as it was unused by the codebase

internal/pkg/apikey/auth.go
```diff
@@ -36,21 +37,23 @@ func (k ApiKey) Authenticate(ctx context.Context, es *elasticsearch.Client) (*Se
...
 	if err != nil {
-		return nil, fmt.Errorf("apikey auth request %s: %w", k.Id, err)
+		return nil, fmt.Errorf("apikey auth request %s: %w", k.ID, err)
 	}
 
 	if res.Body != nil {
-		defer res.Body.Close()
+		defer func() {
+			_ = res.Body.Close()
+		}()
```
michel-laterman, line 46: I use this pattern to get the linter to stop (mostly when defering the Close). the other option is to use `//nolint:errcheck` comments, is there a preference?

internal/pkg/bulk/doc.go
```diff
@@ -2,20 +2,5 @@
...
-func WithContext(ctx context.Context, d time.Duration) error {
-	t := time.NewTimer(d)
-	defer t.Stop()
-	select {
-	case <-ctx.Done():
-		return ctx.Err()
-	case <-t.C:
-		return nil
-	}
-}
```
michel-laterman, line 21: This method was only used in a test, I moved it into the test code base

internal/pkg/bulk/opBulk.go
```diff
@@ -78,11 +81,11 @@ func (b *Bulker) writeMget(buf *Buf, index, id string) error {
 		return err
 	}
 
-	buf.WriteString(`{"_index":"`)
-	buf.WriteString(index)
-	buf.WriteString(`","_id":"`)
-	buf.WriteString(id)
-	buf.WriteString(`"},`)
+	_, _ = buf.WriteString(`{"_index":"`)
```
michel-laterman, line 84: We do a lot of separate writes like this to the buffer (implemented in `pkg/danger`).
The buffer implementation will never return an error.

internal/pkg/bulk/setup_test.go
```diff
@@ -70,12 +73,24 @@ type testT struct {
...
 }
 
-func NewRandomSample() testT {
+// environment tracks env vars that can be used for testing
+type environment struct {
+	Username string
+	Password string
+}
 
+func getEnvironment() environment {
```
michel-laterman, line 82: this was from `testutil` and only used here

internal/pkg/checkin/bulk_test.go
```diff
@@ -14,44 +14,84 @@ import (
...
+		if ops[0].ID != c.id {
+			return false
+		}
+		if ops[0].Index != dl.FleetAgents {
+			return false
+		}
+		tb.Log("Operation match! validating details...")
+
+		// Decode and match operation
+		// NOTE putting the extra validation here seems strange, maybe we should read the args in the test body intstead?
```
michel-laterman, line 44: would welcome feedback for this, it should not be too hard to seperate.

internal/pkg/coordinator/coordinator.go
```diff
@@ -2,6 +2,7 @@
 // or more contributor license agreements. Licensed under the Elastic License;
 // you may not use this file except in compliance with the Elastic License.
 
+// Package coordinator will keep elastic-agent policies uo to date // TODO is this correct?
```
michel-laterman, line 5: i'm mostly guessing here

internal/pkg/danger/buf.go
```diff
@@ -2,6 +2,7 @@
 // or more contributor license agreements. Licensed under the Elastic License;
 // you may not use this file except in compliance with the Elastic License.
 
+// Package danger implements a bytes buffer // TODO why do we need this one instead of bytes.Buffer?
```
michel-laterman, line 5: Is there a reason why we implemented our own here?

internal/pkg/dl/action_results.go
```diff
@@ -23,7 +23,7 @@ func createActionResult(ctx context.Context, bulker bulk.Bulk, index string, acr
 	}
 	body, err := json.Marshal(acr)
 	if err != nil {
-		return "", nil
+		return "", err // TODO was there a good reason this returned nil?
```
michel-laterman, line 26: was this a previous oversight?

michel-laterman (commented):
this pr is 90% me fixing linting issues and 10% adding docs and improving tests

internal/pkg/config/output.go
```diff
@@ -180,7 +181,7 @@ func (c *Output) Validate() error {
 
 func makeURL(defaultScheme string, defaultPath string, rawURL string, defaultPort int) (string, error) {
 	if defaultScheme == "" {
-		defaultScheme = "http"
+		defaultScheme = defaultProtocol
```
aleksmaus, line 184: nit: would prefer maybe something like "schemeHTTP" constant instead "defaultProtocol", easier to to understand without scrolling up to see the defaultProtocol constant value

internal/pkg/coordinator/monitor.go
```diff
@@ -486,7 +486,8 @@ func runUnenrollerWork(ctx context.Context, bulker bulk.Bulk, policyId string, u
 
 	agentIds := make([]string, len(agents))
 
-	for i, agent := range agents {
+	for i := range agents {
+		agent := agents[i]
```
aleksmaus, line 490: good one

internal/pkg/scheduler/scheduler_test.go
```diff
@@ -62,9 +61,14 @@ func TestScheduler(t *testing.T) {
 	})
 
 	g.Go(func() error {
-		wait.WithContext(ctx, 500*time.Millisecond)
-		// return some error here to cause exit error group wait
-		return errTest
+		timer := time.NewTimer(500 * time.Millisecond)
```
aleksmaus, line 64: why replace one liner function with all this?

internal/pkg/testing/suite/suite.go
```diff
@@ -5,6 +5,7 @@
 //go:build integration
 // +build integration
 
+// Pacakge suite wraps a testify/suite.Suite
```
aleksmaus, line 8: Pacakge->Package

aleksmaus (approved):
Left few comments, please respond or address. Otherwise LGTM. Thanks!

internal/pkg/scheduler/scheduler_test.go
michel-laterman: `wait..WithContext` was only being used here. so I decided to remove the wait package to simplify the codebase a bit

aleksmaus: it was useful function, made this implementation cleaner as well, I'd leave as is.

internal/pkg/apikey/auth.go
michel-laterman: this might actually be me overreaching, when i run the linter in vim it's giving me more errors then through make

internal/pkg/scheduler/scheduler_test.go
michel-laterman: turns out `wait.WithContext` is a copy of `sleep.WithContext`, i'll replace it and keep `wait` deleted

cmd/fleet/server_integration_test.go
```diff
@@ -268,9 +266,10 @@ func TestServerInstrumentation(t *testing.T) {
 	cli := cleanhttp.DefaultClient()
 	callCheckinFunc := func() {
 		for {
-			agentId := "1e4954ce-af37-4731-9f4a-407b08e69e42"
-			cli.Post(srv.buildUrl(agentId, "checkin"), "application/json", bytes.NewBuffer([]byte("{}")))
+			agentID := "1e4954ce-af37-4731-9f4a-407b08e69e42"
+			res, err := cli.Post(srv.buildURL(agentID, "checkin"), "application/json", bytes.NewBuffer([]byte("{}"))) //nolint:noctx // test case
```
michel-laterman, line 270: found a bug with the integration tests here. the connection is refused (error wasn't checked before), is that expected @marclop? or should we be logging the error via `t.Log()`?

internal/pkg/gcheckpt/checkpoint.go
```diff
@@ -2,6 +2,8 @@
 // or more contributor license agreements. Licensed under the Elastic License;
 // you may not use this file except in compliance with the Elastic License.
 
+// Package gcheckpt handles the fleet API's global_checkpoints operations
+// TODO describe what global checkpoints are
```
michel-laterman, line 6: Can anyone chime in on this?

internal/pkg/smap/smap.go
```diff
@@ -2,6 +2,8 @@
 // or more contributor license agreements. Licensed under the Elastic License;
 // you may not use this file except in compliance with the Elastic License.
 
+// Package smap provides a string-map // TODO what is this used for?
```
michel-laterman, line 5: I'm also confused about this package

internal/pkg/bulk/doc.go
michel-laterman: `sleep` provides the same function

michel-laterman:
now passes all tests 🥳

internal/pkg/testing/port.go
```diff
@@ -6,7 +6,7 @@ package testing
 
 import "net"
 
-// FreePort find a free port on the system.
+// FreePort finds a free port on the system.
 func FreePort() (uint16, error) {
 	addr, err := net.ResolveTCPAddr("tcp", "localhost:0")
 	if err != nil {
```
ph, line 12: That seems fragile to me, everytime I've used something similar it resulted in flaky test.

ph (approved):
LGTM, I've added a few clarifications when requested and personal comment. We should merge this as soon as possible these changes can rot badly.

internal/pkg/coordinator/monitor.go
michel-laterman: The linter complained at me, the address of `agent` for `_, agent := range agents` is the same, and we pass the address below

internal/pkg/dl/constants.go
```diff
@@ -37,8 +37,8 @@ const (
 	FieldPolicyRevisionIdx           = "policy_revision_idx"
 	FieldPolicyCoordinatorIdx        = "policy_coordinator_idx"
 	FieldDefaultAPIKey               = "default_api_key"
-	FieldDefaultAPIKeyID             = "default_api_key_id"
-	FieldDefaultAPIKeyHistory        = "default_api_key_history"
+	FieldDefaultAPIKeyID             = "default_api_key_id"      //nolint:gosec // field name
+	FieldDefaultAPIKeyHistory        = "default_api_key_history" //nolint:gosec // field name
```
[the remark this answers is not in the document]
michel-laterman: yeah, a possible hardcoded credentials warning

internal/pkg/danger/buf.go
michel-laterman: https://github.com/elastic/fleet-server/issues/1431

PR #1439, opened 2022-05-10
ci: use downstream pattern

v1v:
## Motivation/summary

Use the downstream pattern to run the packaging pipeline and then pass over the Git commit to be used.

The parentstream pattern does not support the Git commit approach by default and therefore it requires further changes to accomplish the same.

## How to test these changes

In the CI.

## Further context

When the `packaging` pipeline runs there is a chance that a new commit has been pushed to the `main` branch before the `packaging` has even get triggered, therefore the `packaging` pipeline will run but it will not use the original commit.

While using the downtream pattern, then we can orchestrate what builds to run from the `main pipeline` and the pass the sha commit to those jobs. 

## Issue

Similar to https://github.com/elastic/apm-server/pull/7982

### DIFF
--- a/.ci/packaging.groovy
+++ b/.ci/packaging.groovy
@@ -13,6 +13,8 @@ pipeline {
     DOCKER_SECRET = 'secret/observability-team/ci/docker-registry/prod'
     DOCKER_REGISTRY = 'docker.elastic.co'
     DRA_OUTPUT = 'release-manager.out'
+    COMMIT = "${params?.COMMIT}"
+    JOB_GIT_CREDENTIALS = "f6c7695a-671e-4f4f-a331-acdce44ff9ba"
   }
   options {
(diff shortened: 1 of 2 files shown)


.ci/packaging.groovy
```diff
@@ -250,3 +249,18 @@ def runIfNoMainAndNoStaging(Closure body) {
     body()
   }
 }
+
+def smartGitCheckout() {
```
mdelapenya, line 253: Candidate to be added to the shared library?

mdelapenya (approved):

kuisathaverat (approved):

.ci/packaging.groovy
v1v: I was tempted initially

If for some reason I need to apply the same pattern  somewhere else then I'll refactor it  for sure

PR #1446, opened 2022-05-10
[Idea - DO NOT MERGE] Add websocket endpoint to fleet-server.

blakerouse:

During the ONWeek last week for the control plane team I took the task of looking at changing the Fleet Server to a websocket protocol versus using longpoll HTTP.

I only had 3 days to look at this which was not enough time to really get this in a state that I am happy with and something that I could even benchmark against. After playing with it both on Fleet Server side and Elastic Agent side, it requires a new protocol design between the 2, cooping the existing over the websocket does not really provide any benefits.

Because of the limitation I choose to not change to much of the way Fleet Server handles subscriptions for actions to send down to the agents and its checkin loop to elasticsearch. I think this can all be improved with the websocket design.

I think this could be improved to not spawn a go-routine per connection instead only use X amount of go-routines for all the agents. Remove the subscriptions inside the go-routines for each agent, instead store that in the websocket connection map. As long as the connection is connected a single go-routine could update all connected agents last-check-in timestamp at the same time. Same thing with the actions, when a new action needs to be sent then the Fleet Server could just send it to all connected agents with X number at a time.


jlind23:
Closing for now, will reopen later if needed.

PR #1552, opened 2022-06-13
Fix mergify for 8.3.0

ph:

Allow to backport using label to the 8.3 branch.

## Checklist


ph:
I will take another look at the test suite here, I see it failing but that current PR shouldn't impact any of the functional code, I am going to merge this and investigate in a followup pr.

PR #1597, opened 2022-06-24
Bugfix: Avoid panic when check-in setup duration is longer than poll …

scunningham:
…duration

HTTP handler panics if the setup duration, ie. the time time it takes to communicate with Elastic is longer than the poll time.  This is a very edgy case; only happens when Elastic and/or the Fleet Server is under extreme load.

Zero duration is illegal for time.NewTimer().  Replace with time.Milllisecond.


aleksmaus (approved):
👍

ph:
0 is indeed invalid for `NewTicker` not new timer.

ph (approved):

ph:
@mergifyio update

Issue #1500, opened 2022-05-24
Do not invalidate API Key right after restarting the Elastic Agent.


ph:
# Context 

When Fleet Server has issues and goes offline for a long period of time and comes back online, it will look if it needs to unenroll any agents. If the offline time is higher than the [unenrollTimeout](https://github.com/elastic/fleet-server/blob/main/internal/pkg/coordinator/monitor.go#L51), the Fleet server will unenroll any Agent matching the unenroll timeout, since all agent was not able to reach the fleet-server **most** if **not all Agents** will match and will be unenrolled from the system.

# Expectation

When Fleet Server recover from the situation it should not unenroll all the agents, it needs to give them some time to try to communicate back to the fleet server before starting the unenrollment processes.

# Solution

On restart the Fleet Server's `unenroll` monitor should wait at least the `unenrollTimeout` value before actually starting to unenroll.

PR #1605, opened 2022-06-27
On Fleet Server reboot gives a grace period

ph:
When Fleet Server is offline for a period bigger or equal to the
unenrollTimeout, on reboot the Server will start to automatically
unenroll Elastic Agent without giving them a chance to communicate with
the system. Instead on reboot we will give a grace period equivalent to
the unenrollTimeout this will give enough time to the Elastic Agent to
communicate back to Fleet Server and update their last checkin time.

Fixes: #1500

## Checklist

### DIFF
--- a/internal/pkg/coordinator/monitor.go
+++ b/internal/pkg/coordinator/monitor.go
@@ -453,20 +453,24 @@ func runCoordinatorOutput(ctx context.Context, cord Coordinator, bulker bulk.Bul
 }
 
 func runUnenroller(ctx context.Context, bulker bulk.Bulk, policyID string, unenrollTimeout time.Duration, l zerolog.Logger, checkInterval time.Duration, agentsIndex string) {
-	// When fleet-server is offline for a long period and finally recover, it means that the connected
+	// When fleet-server is offline for a long period and finally recovers, it means that the connected
 	// agent will be offline for a long period of time since their last checkin and fleet server will
 	// start unenrolling every Elastic Agent from the system. Instead on boot the Elastic Agent
-	// should wait at least the unenrolltimeout period before actively unenrolling agents in the system.
-	// This give a grace period to the Elastic Agent to connect back to the system.
+	// should wait at least the unenrollTimeout period before actively unenrolling agents in the system.
+	// This gives a grace period to the Elastic Agent to connect back to the system.
 	l.Info().
 		Dur("checkInterval", checkInterval).
 		Dur("unenrollTimeout", unenrollTimeout).
 		Msg("giving a grace period to Elastic Agent before enforcing unenrollTimeout monitor")
 
+	grace := time.NewTimer(unenrollTimeout)
+
 	select {
-	case <-time.After(unenrollTimeout):
+	case <-grace.C:
+		grace.Stop()
 		// skip to the next loop.
 	case <-ctx.Done():
+		grace.Stop()
 		return
 	}
 
--- a/internal/pkg/coordinator/monitor.go
+++ b/internal/pkg/coordinator/monitor.go
@@ -453,6 +453,23 @@ func runCoordinatorOutput(ctx context.Context, cord Coordinator, bulker bulk.Bul
 }
 
 func runUnenroller(ctx context.Context, bulker bulk.Bulk, policyID string, unenrollTimeout time.Duration, l zerolog.Logger, checkInterval time.Duration, agentsIndex string) {
+	// When fleet-server is offline for a long period and finally recover, it means that the connected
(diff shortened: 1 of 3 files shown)


ph:
@mergifyio update

internal/pkg/coordinator/monitor.go
```diff
@@ -453,6 +453,23 @@ func runCoordinatorOutput(ctx context.Context, cord Coordinator, bulker bulk.Bul
 }
 
 func runUnenroller(ctx context.Context, bulker bulk.Bulk, policyID string, unenrollTimeout time.Duration, l zerolog.Logger, checkInterval time.Duration, agentsIndex string) {
+	// When fleet-server is offline for a long period and finally recover, it means that the connected
```
aleksmaus, line 456: nit: recover -> recovers

internal/pkg/coordinator/monitor.go
```diff
@@ -453,6 +453,23 @@ func runCoordinatorOutput(ctx context.Context, cord Coordinator, bulker bulk.Bul
...
+	// start unenrolling every Elastic Agent from the system. Instead on boot the Elastic Agent
+	// should wait at least the unenrolltimeout period before actively unenrolling agents in the system.
+	// This give a grace period to the Elastic Agent to connect back to the system.
+	l.Info().
+		Dur("checkInterval", checkInterval).
+		Dur("unenrollTimeout", unenrollTimeout).
+		Msg("giving a grace period to Elastic Agent before enforcing unenrollTimeout monitor")
+
+	select {
+	case <-time.After(unenrollTimeout):
```
aleksmaus, line 467: I'd change this to NewTimer, cause this select can be exited sooner on context.Done:
"The underlying Timer is not recovered by the garbage collector until the timer fires. If efficiency is a concern, use NewTimer instead and call Timer.Stop if the timer is no longer needed."

ph: Yeah, My first shot was using a timer, I've considered this would only be called for a few time n agent policy time. But I will change it.

internal/pkg/coordinator/monitor.go
```diff
@@ -453,6 +453,27 @@ func runCoordinatorOutput(ctx context.Context, cord Coordinator, bulker bulk.Bul
...
+	l.Info().
+		Dur("checkInterval", checkInterval).
+		Dur("unenrollTimeout", unenrollTimeout).
+		Msg("giving a grace period to Elastic Agent before enforcing unenrollTimeout monitor")
+
+	grace := time.NewTimer(unenrollTimeout)
+
+	select {
+	case <-grace.C:
+		grace.Stop()
```
aleksmaus, line 470: doesn't need to be stopped, once it's fired
nit: could wrap into a func something like waitWithContext https://github.com/elastic/beats/blob/db73681d67be859f5c048d07da985eadaa780e21/x-pack/osquerybeat/internal/osqdcli/retry.go#L53

internal/pkg/coordinator/monitor.go
```diff
@@ -463,15 +463,9 @@ func runUnenroller(ctx context.Context, bulker bulk.Bulk, policyID string, unenr
...
-	select {
-	case <-grace.C:
-		grace.Stop()
-		// skip to the next loop.
-	case <-ctx.Done():
-		grace.Stop()
-		return
+	if err := waitWithContext(ctx, unenrollTimeout); err != nil {
+		l.Err(err).Dur("unenroll_timeout", unenrollTimeout).
+			Msg("failed to delay the unenroller monitor startup")
```
aleksmaus, line 468: the only error from this function would be "context cancelled", maybe not log it at the Err level

internal/pkg/coordinator/monitor.go
```diff
@@ -453,6 +453,21 @@ func runCoordinatorOutput(ctx context.Context, cord Coordinator, bulker bulk.Bul
...
+	// start unenrolling every Elastic Agent from the system. Instead on boot the Elastic Agent
+	// should wait at least the unenrollTimeout period before actively unenrolling agents in the system.
+	// This gives a grace period to the Elastic Agent to connect back to the system.
+	l.Info().
+		Dur("checkInterval", checkInterval).
+		Dur("unenrollTimeout", unenrollTimeout).
+		Msg("giving a grace period to Elastic Agent before enforcing unenrollTimeout monitor")
+
+	if err := waitWithContext(ctx, unenrollTimeout); err != nil {
+		l.Err(err).Dur("unenroll_timeout", unenrollTimeout).
```
aleksmaus, line 467: nit: maybe log "context canceled" at debug level?

PR #1671, opened 2022-07-18
Log redacted config on config changes

### LINKED ISSUE #1626: Log all of fleet server configuration at startup
When debugging an issue in cloud it is useful to know with what options the fleet-server is operating at, because some options are derived from the current container size and other can be change with the fleet-server integration.

I would be good to log a single even that we can search through the log that would container the information defined in https://github.com/elastic/fleet-server/blob/3c8cd17c1370abb07f6935021d31c8204e4d16a3/internal/pkg/config/input.go

michel-laterman:

Debugging the fleet-server can be difficult as it is not clear what config it is running.

Output redacted configuration when an updated config is received.

1. Compile a snapshot version of elastic-agent and fleet-server.
2. unzip elastic-agent snapshot.
3. Copy the fleet-server artifact and sha files to the elastic-agent `downloads` dir.
4. Install elastic-agent with a policy that only contains the `fleet-server` integration
5. Review logs to see redacted log config.

## Checklist

- [] ~~I have commented my code, particularly in hard-to-understand areas~~
- [] ~~I have made corresponding changes to the documentation~~
- [] ~~I have made corresponding change to the default configuration files~~

## Related issues

- Closes #1626


joshdover:
@michel-laterman Friendly bump on this one. @pierrehilbert should we be tracking this for 8.5?

cmd/fleet/main.go
```diff
@@ -621,6 +625,19 @@ func redactServerCfg(cfg *config.Config) config.Server {
 	return redacted
 }
 
+func redactConfig(cfg *config.Config) *config.Config {
+	redacted := &config.Config{
+		Fleet:   cfg.Fleet,
+		Output:  cfg.Output,
+		Inputs:  make([]config.Input, 1),
```
michel-laterman, line 632: having this as `cfg.Inputs` causes one of the integration tests to fail.

cmd/fleet/main.go
```diff
@@ -538,7 +538,7 @@ LOOP:
 
 		select {
 		case newCfg = <-f.cfgCh:
-			log.Info().Msg("Server configuration update")
+			log.Info().Interface("cfg", redactConfig(curCfg)).Msg("Server configuration update")
```
michel-laterman, line 541: alternatively we can use the `redactConfig` call instead of `redactServerCfg` on lines 643 & 659 in `configChangedServer` (where it logs old/new config)

ph (approved):

PR #1731, opened 2022-08-10
Update upgrade fields on version change

joshdover:

Non ideal solution to https://github.com/elastic/ingest-dev/issues/1184

We should have a more robust way of ensuring that every ack is delivered.

## Checklist

### DIFF
--- a/internal/pkg/checkin/bulk.go
+++ b/internal/pkg/checkin/bulk.go
@@ -202,6 +202,11 @@ func (bc *Bulk) flush(ctx context.Context) error {
 			if pendingData.extra.ver != "" {
 				fields[dl.FieldAgent] = map[string]interface{}{
 					dl.FieldAgentVersion: pendingData.extra.ver,
+					// Duplicates the writes done when an upgrade is ack'd by the agent.
+					// This helps eliminate the case where the agent is stuck in upgrading after it's already upgraded if the ack
+					// is never received.
+					dl.FieldUpgradeStartedAt: nil,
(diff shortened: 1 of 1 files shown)


internal/pkg/checkin/bulk.go
```diff
@@ -202,6 +202,11 @@ func (bc *Bulk) flush(ctx context.Context) error {
 			if pendingData.extra.ver != "" {
 				fields[dl.FieldAgent] = map[string]interface{}{
 					dl.FieldAgentVersion: pendingData.extra.ver,
+					// Duplicates the writes done when an upgrade is ack'd by the agent.
+					// This helps eliminate the case where the agent is stuck in upgrading after it's already upgraded if the ack
+					// is never received.
+					dl.FieldUpgradeStartedAt: nil,
+					dl.FieldUpgradedAt: nowTimestamp,
 				}
```
ph, line 210: So every agent will be upgradedAt at now() by default on any ack?

AndersonQ: besides it'll lose the `UpgradeStartedAt` as well.

joshdover: No because this logic only executes when the version has changed

juliaElastic:
Did some testing with this change and it has some unintended side effects, see here: https://github.com/elastic/fleet-server/pull/2582#discussion_r1194728597

jlind23:
Closing this for now.

PR #1747, opened 2022-08-15
Support for Elastic Agent V2 status

aleksmaus:

Adds support for Elastic Agent V2 status

Extends the current checkin logic with new payload.

Open question: What fields of any need to be indexed? Need to update the .fleet-agents index mapping with newly added fields, and need to know if any details of the components/units need to be searchable or we should not index it at all.

This needs to be tested along with the corresponding agent changes from https://github.com/elastic/elastic-agent/pull/916

## Checklist

## Related issues

- Relates https://github.com/elastic/elastic-agent/issues/100
- Requires https://github.com/elastic/elastic-agent/pull/916 (should be backwards compatible though and work for the older agent)

## Screenshots

The fragment of the .fleet-agents document with V2 data.
Screen Shot 2022-08-15 at 12 53 36 PM


ph:
@kpollich Maybe you review concerning the fields that need to be indexed to correctly report the state of integration.

I would assume that the `id`, `type` and `status` need to be indexed probably keyword would be appropriate for them.
I think I would like to answer the following question.

Which Integration are unhealthy?
What is the state of of all my endpoint integration?
Is there a problem with the filestream input?

@nimarezainia You may want to take a look at the format above if there is are telemetry that could help us?

internal/pkg/api/handleCheckin.go
```diff
@@ -555,6 +561,56 @@ func parseMeta(zlog zerolog.Logger, agent *model.Agent, req *CheckinRequest) ([]
...
+	if len(req.Components) > 0 {
+		if err := json.Unmarshal(req.Components, &reqComponents); err != nil {
+			return nil, errors.Wrap(err, "parseComponents request")
+		}
+	}
+
+	// If empty, don't step on existing data
+	if reqComponents == nil {
+		return nil, nil
+	}
```
ph, line 584: @aleksmaus We should probably assert the data we are receiving from Elastic Agent before sending that to Elasticsearch? Maybe we should also put some kind of limit, maybe a size based limit on the `payload` field to reduce the risk or document size explosion?

internal/pkg/api/handleCheckin.go
```diff
@@ -555,6 +561,56 @@ func parseMeta(zlog zerolog.Logger, agent *model.Agent, req *CheckinRequest) ([]
...
 }
 
+func parseComponents(zlog zerolog.Logger, agent *model.Agent, req *CheckinRequest) ([]byte, error) {
+
+	// Quick comparison first; compare the JSON payloads.
+	// If the data is not consistently normalized, this short-circuit will not work.
+	if bytes.Equal(req.Components, agent.Components) {
+		zlog.Trace().Msg("quick comparing agent components data is equal")
+		return nil, nil
+	}
```
ph, line 571: I always forgot the `bytes.Equal` does check the `len` before doing the bytes check, I think.

ph (changes requested):

model/schema.json
```diff
@@ -454,6 +454,15 @@
           "description": "Lst checkin status",
           "type": "string"
         },
+        "last_checkin_message": {
+          "description": "Last checkin message",
+          "type": "string"
+        },
+        "components": {
```
blakerouse, line 461: @joshdover Do we want to store the status, message, and components as they change across time in a datastream so we can track the changes of an Elastic Agent across time? I think from an alerting standpoint it would be beneficial but from a scalability it would add more data and load.

blakerouse (approved):
Looks great.

I think we need to determine what we want to do when the status, message, and components change from the Elastic Agent side. Should it disconnect from the long-poll and reconnect and send updated status, message, and components everytime it changes? At the moment the Elastic Agent only sends it based on time-interval. I think changing it to send it on every change would provide really good observability but the scale and load of that might be high.

Something to think about, but not something that is blocking this to land as it is now. Same with my inline comment to Josh.

internal/pkg/api/handleCheckin.go
aleksmaus: This approach is basically copied from parseMeta that is used currently for "local_metadata" handling which is another blob that agent can send whatever to the fleet server. 
There is already a limit check on the checking body size, default is ```defaultCheckinMaxBody  = 1024 * 1024```.
There is already the parsing of the components happens in the ```parseComponents``` which validates that is a parsable JSON. I can add a check if the components is an array.

Do we need to add field by field check? What are guarantees which fields are required vs which fields are optional?

internal/pkg/api/handleCheckin.go
aleksmaus: This approach is copied from parseMeta, works fine for nils. It seems like the compiler is optimized for this as well
https://cs.opensource.google/go/go/+/refs/tags/go1.19:src/bytes/bytes.go;l=19

internal/pkg/api/handleCheckin.go
aleksmaus: Will add fully structured parsing to the "components" property tomorrow, only things that are specified in the V2 at the moment will be stored for the components, and it will need to be extended for anything extra in the future.

ph: Yes, once everything is V2 we should be able to be more strict on any incoming data, not sure during the transition period we can do this.

ph (approved):

internal/pkg/api/handleCheckin.go
aleksmaus: Sounds good. Will leave it as is for now until we stabilize the schema, can tighten validation later.

ph (approved):

PR #1751, opened 2022-08-16
[v2] Switch to Elastic Agent v2 control protocol

blakerouse:

Add support for Elastic Agent V2 client protocol.

## Checklist


cmd/fleet/main.go
```diff
@@ -220,226 +219,371 @@ func NewAgentMode(cliCfg *ucfg.Config, reader io.Reader, bi build.Info, reloadab
...
+	if unit.Type() == client.UnitTypeOutput {
+		_, _, cfg := unit.Expected()
+		if cfg.Type != kElasticsearch {
+			// not support output type
+			_ = unit.UpdateState(client.UnitStateFailed, fmt.Sprintf("%s is an unsupported output type", cfg.Type), nil)
+			return nil
+		}
+		if a.outputUnit != nil {
+			// original output unit is being stopped; swapping in this unit as the new output unit
+			_ = a.outputUnit.UpdateState(client.UnitStateStopped, kStopped, nil)
```
cmacknz, line 340: What happens if this RPC call fails but the subsequent call for `Starting` succeeds? Is the `Stopped` nice to have or necessary?

cmacknz: In general if an input implements the state transitions incorrectly what happens? How would they find out and know to correct it?

cmacknz (approved):
Looks reasonable to me from a high level.

cmd/fleet/main.go
blakerouse: Fleet Server is a very weird input, because it's the only input that gets started before a policy is every read on first launch of an Elastic Agent with a Fleet Server bootstrap. Fleet Server is the only input that will ever have to worry about this.

The `Stopped` is just more a nice to have, not really a hard requirement.

Issue #1581, opened 2022-06-22
Update our API Key handling logic for updatable permission and bulk operation on the api endpoint


ph:
## Situation
The current implementation of the API keys come with two problematic for Fleet-Server:

- The API key permission are immutable, when an API key is created it will be stuck with the same permissions. When a new integration is added to an Agent policy it meant that all the Elastic Agent API need to be invalidated and recreated.
- The API key endpoint in Elasticsearch doesn't support bulk operation.

# Tasks
Fleet server will need to be adjusted to support the following feature.

- [ ] Elasticsearch will allow API Key permission to be updated in place
- [ ] Elasticsearch will support Bulk Operation to Update API Key.
- [ ] Verifies what happen when we migrate from non updatable to updatable api keys.

# Depends on 
- https://github.com/elastic/elasticsearch/issues/88758

ph:
Single operation route PR https://github.com/elastic/elasticsearch/issues/87870

AndersonQ:
@ph, sorry, I'm a bit unsure about what you meant when you said: _"Verifies what happen when we migrate from non updatable to updatable api keys."_. Is it to measure the performance before and after the change?

ph:
@AndersonQ let me clarify, I was more curious about upgrade in the stack. 

1. 8.4  Api keys permissions are not updatable.
2. 8.5 Api Keys permissions are updatable.

If I remember correctly from the ES proposal, you cannot update permission on keys created on 8.4, they are immutable.
So you will have to create new keys in 8.5, after that point the new keys would be updatable.

Is this more clear?

michalpristas:
checked that
- API Keys before change will be updatable
- there's no flag required to make new/old keys become updatable 

Considering this I was looking into how API keys works now in the code for quick summary:(if i missed something please correct me)

1. User changes ES policy, change is reflected in `.fleet_policies`
2. `fleet-server` is notified about the change comparing to last checkpoints
3. `fleet-server` loads the policy, populate `roles` definition based on policy content and updates it internally
4. on `CHECK-IN``fleet-server` prepares ES output meaning it will
  a. handle whatever ES needs to accept new items
  b. create a `new` API Key for agent to use with the current policy 
5. once `agent` handles config it will `ACK` the action
6. `fleet-server` when handling `ACK` action invalidates OLD api key based on agent history

To make use of Bulk and updatable API Keys i was thinking about 3 scenarios of process modifications:

### Option 1: Bulk Create/Update on policy change
[...]

PR #1779, opened 2022-08-25
Bulk API Keys update

michalpristas:

Instead of creating and invalidating API keys with every policy change it updates them.
On top of that this PR modifies bulker so updates are grouped by role applied. 

- When agent is enrolled api key is `created`
- with every policy change api key is `updated`
- when agent is unenrolled api key is `invalidated`

Bulker contains new queue where requests contains ID of an agent and proposed role descriptors.
On flush logic goes through the queue and composes IDs per roles. then it makes a single request to ES.
In case content size of this request is larger than predefined value. it splits IDs to make it suitable.  value is configured and is defaulting to 100MB as configured in ES.
Grouping is not that optimal as I would like but we need to keep in mind and make sure that if 2 changes for one ID are in a queue only latest is applied.

the tricky thing is to clean roles on ACK as we don't have a way of persisting desired roles. 
How this PR approaches the problem is that instead of persisting roles in agent structure it marks them stale in persisted API Key

e.g we remove network integration, we ended up with api key containing two sets of roles. set of roles is a map with guid as a key
e.g 
```json
"497ac86b-40ac-49ce-88d6-dbdb7ac449d4": {
			"applications": [],
			"cluster": [],
			"indices": [{
				"allow_restricted_indices": false,
				"names": ["logs-mongodb.log-default"],
				"privileges": ["auto_configure", "create_doc"]
			}, {
				"allow_restricted_indices": false,
				"names": ["metrics-mongodb.collstats-default"],
				"privileges": ["auto_configure", "create_doc"]
			}, {
				"allow_restricted_indices": false,
				"names": ["metrics-mongodb.dbstats-default"],
				"privileges": ["auto_configure", "create_doc"]
			}, {
				"allow_restricted_indices": false,
				"names": ["metrics-mongodb.metrics-default"],
				"privileges": ["auto_configure", "create_doc"]
			}, {
				"allow_restricted_indices": false,
				"names": ["metrics-mongodb.replstatus-default"],
				"privileges": ["auto_configure", "create_doc"]
			}, {
				"allow_restricted_indices": false,
				"names": ["metrics-mongodb.status-default"],
				"privileges": ["auto_configure", "create_doc"]
			}],
			"metadata": {},
			"run_as": [],
			"transient_metadata": {
				"enabled": true
			}
		},
```
when removing mongo we create a desired role description set and add this one with modified key by appending `-rdstale` also to avoid naming conflicts. sometimes content of this changes and we ended up with reduced descriptor set (e.g when disabling monitoring.logs but keepiing monitoring.metrics or disabling metrics for mongo integration)

then on ack we iterate through api key keys and leave out everything with `rdstale` suffix

To test this you need latest ES. We are using new Update API and modified `GET apikey` which also returns role descriptions. this allows us not to persist roles and rely on ES.

## Checklist

## Related issues

Closes https://github.com/elastic/fleet-server/issues/1581


jlind23:
@michalpristas does it relates https://github.com/elastic/fleet-server/issues/1581 or should it close it?

michalpristas:
@jlind23 this is exactly it, i also linked it as a related issue

internal/pkg/bulk/opApiKey.go
```diff
@@ -5,15 +5,37 @@
...
+	"encoding/json"
+	"math"
 
 	"github.com/elastic/fleet-server/v7/internal/pkg/apikey"
+	"github.com/elastic/fleet-server/v7/internal/pkg/es"
+	"github.com/rs/zerolog/log"
+)
+
+const (
+	expectedAPIKeySize = 64 // 64B
```
cmacknz, line 19: Where do these numbers come from? How would we know if they need to change?

cmacknz (commented):
Have you thought about how to scale test this? Are you able to measure that this is more efficient?

internal/pkg/bulk/opApiKey.go
michalpristas: i took api key and added some buffer. 
but knowing api key size i can do this dynamically, good point

michalpristas:
not sure how to approach scale tests. 
as this is happening per checkin we would notice just a small difference in checkin time.
we have nothing to compare in regards create/update as before it was split per checkin now it's bulked into a single operation. 
what we could measure is CPU load of ES when dealing with policy change or how long it takes all agents to be reported as up to date.

joshdover:
> not sure how to approach scale tests.
> as this is happening per checkin we would notice just a small difference in checkin time.
> we have nothing to compare in regards create/update as before it was split per checkin now it's bulked into a single operation.
> what we could measure is CPU load of ES when dealing with policy change or how long it takes all agents to be reported as up to date.

Yes I believe we should expect to see much lower ES CPU usage, especially during bulk policy reassignments. Unfortunately, I don't believe we yet have a way to build a custom elastic-agent image to use in a scale test before we merge (this is something I will follow up on solving).

I think we should just try to merge this and take a look at how it impacts scale tests once it's in 8.5 snapshots. If there's an issue we can revert this.

internal/pkg/es/bulk_update_api_key.go
```diff
@@ -0,0 +1,115 @@
...
+      Header:     res.Header,
+    }
+
+    return &response, nil
+}
+
+
+// WithContext sets the request context.
+//
+func (f UpdateApiKeyBulkRequest) WithContext(v context.Context) func(*UpdateApiKeyBulkRequest) {
```
AndersonQ, line 89: [Blocker]
No always right?
It's a bit odd, why there is a context in the struct but the Do method take in another context? Which one takes precedence? Besides on `Do` I did not see `r.ctx` being used

internal/pkg/es/bulk_update_api_key.go
```diff
@@ -0,0 +1,115 @@
...
+	return func(r *UpdateApiKeyBulkRequest) {
+		r.ctx = v
+	}
+}
+
+// WithBody - The search definition using the Query DSL.
+//
```
AndersonQ, line 96:
```suggestion
```

internal/pkg/es/bulk_update_api_key.go
```diff
@@ -0,0 +1,115 @@
...
+func (f UpdateApiKeyBulkRequest) WithBody(v io.Reader) func(*UpdateApiKeyBulkRequest) {
+	return func(r *UpdateApiKeyBulkRequest) {
+		r.Body = v
+	}
+}
+
+
+// WithHeader adds the headers to the HTTP request.
+//
```
AndersonQ, line 105:
```suggestion
```

internal/pkg/es/bulk_update_api_key.go
```diff
@@ -0,0 +1,115 @@
...
+      Body:       res.Body,
+      Header:     res.Header,
+    }
+
+    return &response, nil
+}
+
+
+// WithContext sets the request context.
+//
```
AndersonQ, line 88:
```suggestion
```

internal/pkg/policy/policy_output.go
```diff
@@ -56,18 +56,73 @@ func (p *PolicyOutput) Prepare(ctx context.Context, zlog zerolog.Logger, bulker
...
 			zlog.Debug().Msg("policy output permissions are the same")
 		}
 
-		if needNewKey {
+		if needUpdateKey {
+			zlog.Debug().
+				RawJSON("roles", p.Role.Raw).
+				Str("oldHash", agent.PolicyOutputPermissionsHash).
+				Str("newHash", p.Role.Sha2).
+				Msg("Generating a new API key")
```
AndersonQ, line 77: It's updating, isn't it?

internal/pkg/policy/policy_output.go
```diff
@@ -56,18 +56,73 @@ func (p *PolicyOutput) Prepare(ctx context.Context, zlog zerolog.Logger, bulker
...
+					Str("apiKeyID", agent.DefaultAPIKeyID).
+					Err(err).Msg("fail merging roles for key")
+				return err
+			}
+
+			// hash provided is only for merging request together and not persisted
+			err = bulker.APIKeyUpdate(ctx, agent.DefaultAPIKeyID, newRoles.Sha2, newRoles.Raw)
+			if err != nil {
+				zlog.Error().Err(err).Msg("fail generate output key")
+				zlog.Debug().RawJSON("roles", newRoles.Raw).Str("sha", newRoles.Sha2).Err(err).Msg("fail generate output key")
```
AndersonQ, line 101: [Suggestion]
Please, make the messages different, it'd be confusing seeing the same message twice with different levels

internal/pkg/policy/policy_output.go
```diff
@@ -56,18 +56,73 @@ func (p *PolicyOutput) Prepare(ctx context.Context, zlog zerolog.Logger, bulker
...
+			zlog.Debug().
+				Str("hash.sha256", p.Role.Sha2).
+				Str("roles", string(p.Role.Raw)).
+				Msg("Updating agent record to pick up most recent roles.")
+
+			fields := map[string]interface{}{
+				dl.FieldPolicyOutputPermissionsHash: p.Role.Sha2,
+			}
+
+			// Using painless script to append the old keys to the history
```
AndersonQ, line 114: if the only field you're setting is `dl.FieldPolicyOutputPermissionsHash`, it won't _append the old keys to the history_

internal/pkg/policy/policy_output.go
```diff
@@ -56,18 +56,73 @@ func (p *PolicyOutput) Prepare(ctx context.Context, zlog zerolog.Logger, bulker
...
+			body, err := renderUpdatePainlessScript(fields)
+			if err != nil {
+				return err
+			}
+
+			if err = bulker.Update(ctx, dl.FleetAgents, agent.Id, body); err != nil {
+				zlog.Error().Err(err).Msg("fail update agent record")
+				return err
```
AndersonQ, line 122: isn't it, and all the other returns here double logging as it logs the error and returns the same error?

internal/pkg/policy/policy_output.go
```diff
@@ -137,6 +192,104 @@ func (p *PolicyOutput) Prepare(ctx context.Context, zlog zerolog.Logger, bulker
...
+	}
+	if new == nil {
+		return old, nil
+	}
+
+	oldMap, err := smap.Parse(old.Raw)
+	if err != nil {
+		return nil, err
+	}
+	if oldMap == nil {
```
AndersonQ, line 229: can it be nil, even if `old` isn't?

internal/pkg/policy/policy_output.go
```diff
@@ -137,6 +192,104 @@ func (p *PolicyOutput) Prepare(ctx context.Context, zlog zerolog.Logger, bulker
...
+	}
+
+	newMap, err := smap.Parse(new.Raw)
+	if err != nil {
+		return nil, err
+	}
+
+	if newMap == nil {
+		return old, nil
+	}
```
AndersonQ, line 240: same as above

[Consistency]
```suggestion
	if newMap == nil {
		return old, nil
	}
```

internal/pkg/policy/policy_output.go
```diff
@@ -137,6 +192,104 @@ func (p *PolicyOutput) Prepare(ctx context.Context, zlog zerolog.Logger, bulker
...
+	// copy old
+	for k, v := range oldMap {
+		newKey := findNewKey(destMap, k)
+		if newKey == "" {
+			zlog.Warn().Msg("Failed to find a key for role assignement.")
+
+			zlog.Debug().
+				RawJSON("roles", new.Raw).
+				Str("candidate", k).
+				Msg("Failed to find a key for role assignement.")
```
AndersonQ, line 275: [Blocker]
Again, please avoid 2 different logs with the same message

internal/pkg/policy/policy_output.go
```diff
@@ -137,6 +192,104 @@ func (p *PolicyOutput) Prepare(ctx context.Context, zlog zerolog.Logger, bulker
...
+
+			continue
+		}
+		destMap[newKey] = v
+	}
```
AndersonQ, line 280: I didn't understand how it works

AndersonQ (changes requested):

internal/pkg/api/handleAck.go
```diff
@@ -351,38 +351,118 @@ func (ack *AckT) handlePolicyChange(ctx context.Context, zlog zerolog.Logger, ag
...
 		ctx,
 		dl.FleetAgents,
-		agent.Id,
+		agentID,
 		body,
 		bulk.WithRefresh(),
 		bulk.WithRetryOnConflict(3),
 	)
 
 	zlog.Info().Err(err).
```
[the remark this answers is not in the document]
michalpristas: this one is not my update actually. but i like the approach of dynamic log level

internal/pkg/apikey/apikey_integration_test.go
```diff
@@ -121,6 +156,12 @@ func TestCreateAPIKeyWithMetadata(t *testing.T) {
 			if diff != "" {
 				t.Error(diff)
 			}
+
+			// Try to get the key that doesn't exists, expect ErrApiKeyNotFound
+			_, err = Read(ctx, es, "0000000000000", false)
+			if !errors.Is(err, ErrAPIKeyNotFound) {
+				t.Errorf("Unexpected error type: %v", err)
+			}
```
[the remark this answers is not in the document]
michalpristas: this went away

internal/pkg/bulk/opApiKey.go
```diff
@@ -49,3 +70,182 @@ func (b *Bulker) APIKeyInvalidate(ctx context.Context, ids ...string) error {
...
+	_, err = b.waitBulkAction(ctx, ActionUpdateAPIKey, "", id, body)
+	return err
+}
+
+// flushUpdateAPIKey takes an update API Key queue and groups request based on roles applied
+// it needs to group agent IDs per Role Hash in order to produce more efficient request containing a list of IDs for a change(update)
+// one thing to have in mind is that in a single queue there may be change and ack request with roles. in this case
+// later occurrence wins overwriting policy change to reduced set of permissions.
+// even if the order was incorrect we end up with just a bit broader permission set, never too strict, so agent does not
+// end up with fewer permissions than it needs
```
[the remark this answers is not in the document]
michalpristas: on a next permissions set, talked about it with Josh and this is OK and safer to have broader set than missing perms.

internal/pkg/bulk/opApiKey.go
```diff
@@ -49,3 +70,182 @@ func (b *Bulker) APIKeyInvalidate(ctx context.Context, ids ...string) error {
...
+		case n.ch <- respT{
+			err: nil,
+			idx: n.idx,
+			data: &BulkIndexerResponseItem{
+				DocumentID: "",
+				Status:     res,
+			},
+		}:
+		default:
+			panic("Unexpected blocked response channel on flushRead")
```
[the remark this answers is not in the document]
michalpristas: this cannot happen, if it happens something's really wrong

internal/pkg/es/bulk_update_api_key.go
```diff
@@ -0,0 +1,115 @@
...
+
+package es
+
+import (
+	"context"
+	"io"
+	"net/http"
+  "strings"
+
+	"github.com/elastic/go-elasticsearch/v7/esapi"
```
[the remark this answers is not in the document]
michalpristas: weird, vscode showed it ok for me

internal/pkg/es/bulk_update_api_key.go
michalpristas: this is http.Request context. not used here. used internally

internal/pkg/es/bulk_update_api_key.go
michalpristas: these are actually following pattern set here in fleet-server code:check fleet_msearch.go for example

internal/pkg/policy/policy_output.go
michalpristas: added comments describing merge strategy

internal/pkg/bulk/opApiKey.go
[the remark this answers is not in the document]
AndersonQ: ok

internal/pkg/policy/policy_output.go
```diff
@@ -137,6 +192,119 @@ func (p *PolicyOutput) Prepare(ctx context.Context, zlog zerolog.Logger, bulker
...
+
+		}
+		// creates new key for permissions
+		// marks it stale so they can be removed later
+		if _, exists := m[candidate]; !exists {
+			return fmt.Sprintf("%s-0-rdstale", candidate)
+		}
+
+		// 1 should be enough, 100 is just to have some space
+		for i := 0; i < 100; i++ {
```
AndersonQ, line 271: You could let just the `for`, no need for this first try, besides the `for` is trying `%s-0-rdstale` again

AndersonQ (approved):

PR #1794, opened 2022-08-31
Action APM tracing

joshdover:
### Checkin transaction from a horde drone's perspective

We can see how the ack step failed several times due to 429 errors before eventually succeeding

Screen Shot 2022-08-31 at 17 28 17

### Correlation between bulk action in UI and corresponding processing on each agent

Using span links, we can observe and find where actions were delivered by Fleet Server, received by a horde drone, and eventually ack'd back. Filtering by agent_id allows for finding a specific problematic agent:

image

image

image

## Checklist


juliaElastic:
Opened a new pr https://github.com/elastic/fleet-server/pull/2472 with resolved conflicts

PR #1833, opened 2022-09-06
updating upgrade_status: completed

juliaElastic:

https://github.com/elastic/kibana/issues/139704

Updating `upgrade_status: completed` agent field when an upgrade is acked, this is a fix instead of kibana relying on `upgraded_at: null` field, which caused a Lucene bug.

The `upgrade_status` is set to `started` initially by kibana, when the action is started. Fleet Server then has to set it to `completed`, to signal to kibana that the action has completed.

1. Enroll fleet server, and many agents with horde
2. Bulk select agents and upgrade immediately
3. Wait for upgrade to complete, notice that agents are gradually showing up as `Healthy` and the Upgrade callout decreases in count.
4. Expect that the lucene error does not come anymore on the `agents` API call

## Checklist


internal/pkg/model/schema.go
```diff
@@ -190,6 +190,9 @@ type Agent struct {
 	// Date/time the Elastic Agent was last upgraded
 	UpgradedAt string `json:"upgraded_at,omitempty"`
 
+	// Upgrade status
+	UpgradeStatus string `json:"upgrade_status,omitempty"`
```
michel-laterman, line 194: This file is generated from `model/schema.json`; you should update that and run the json schema generator in order to update the go definition.

juliaElastic: Do you know how can I run the generator?

michel-laterman:
I think we would also need to add the new attribute to the ES schema definition: https://github.com/michel-laterman/elasticsearch/blob/master/x-pack/plugin/core/src/main/resources/fleet-agents.json

I have an example PR where I added fields to the actions schema: https://github.com/elastic/elasticsearch/pull/86167

internal/pkg/model/schema.go
juliaElastic: updated

juliaElastic:
@michel-laterman Do you know how to fix this build error?
```
14:44:12  Makefile:116: recipe for target 'check-no-changes' failed
14:44:12  make[1]: *** [check-no-changes] Error 1
14:44:12  make[1]: Leaving directory '/var/lib/jenkins/workspace/_server_fleet-server-mbp_PR-1833/src/github.com/elastic/fleet-server'
14:44:12  Makefile:74: recipe for target 'check-ci' failed
14:44:12  make: *** [check-ci] Error 2
```

juliaElastic:
@michel-laterman are you ready to approve?

aleksmaus (approved):

PR #1840, opened 2022-09-08
ci: rename dra staging for release dra release staging 

v1v:

Change stage name to reflect `DRA staging` is not available for `main`

## Why

DRA staging does not run for the `main` branch but the `release branches` (7.17, 8.4...)

image

I manually triggered the [build](https://fleet-ci.elastic.co/blue/organizations/jenkins/fleet-server%2Ffleet-server-package-mbp/detail/PR-1840/3/pipeline/) for this PR, so the change is just the stage name.

### DIFF
--- a/.ci/packaging.groovy
+++ b/.ci/packaging.groovy
@@ -133,7 +133,7 @@ pipeline {
             }
           }
         }
-        stage('DRA Staging (if no main)') {
+        stage('DRA Staging (release branches only)') {
           options { skipDefaultCheckout() }
           when {
             allOf {
--- a/.ci/packaging.groovy
+++ b/.ci/packaging.groovy
@@ -133,7 +133,7 @@ pipeline {
             }
           }
         }
-        stage('DRA Staging') {
+        stage('DRA Staging (if no main)') {
           options { skipDefaultCheckout() }
           when {
             allOf {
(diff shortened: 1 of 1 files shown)


.ci/packaging.groovy
```diff
@@ -133,7 +133,7 @@ pipeline {
             }
           }
         }
-        stage('DRA Staging') {
+        stage('DRA Staging (if no main)') {
```
cmacknz, line 136:
```suggestion
        stage('DRA Staging (release branches only)') {
```

I think this would be a bit clearer, but approving regardless. Thanks!

cmacknz (approved):

.ci/packaging.groovy
v1v: There is certain limitation in the stage name, let me commit your changes and see if it works as expected

v1v:
@cmacknz , it truncates the text a bit, what do you prefer?

image

.ci/packaging.groovy
```diff
@@ -133,7 +133,7 @@ pipeline {
             }
           }
         }
-        stage('DRA Staging') {
+        stage('DRA Staging (release branches only)') {
```
cmacknz, line 136:
```suggestion
        stage('DRA Release Staging') {
```

It is annoying that it is truncated, maybe just putting Release in the name will help and it won't be truncated?

We could also go back to the original wording, I don't have a strong preference at this point. Pick whichever of this or your first suggestion you like and merge it :)

v1v:
image

neat!

PR #1849, opened 2022-09-09
Avoid mapping conflict on debug log

joshdover:
Fixes a minor bug when debugging logs were enabled resulting in dropped logs:

```
        "error": {
          "type": "mapper_parsing_exception",
          "reason": "failed to parse field [rev] of type [long] in document with id 'PVsnrKOYhCal1hE0hJotzUFoA9s='. Preview of field's value: '{policyId=981d1020-3028-11ed-893b-67610e46ff24}'",
          "caused_by": {
            "type": "illegal_argument_exception",
            "reason": "Cannot parse object as number"
          }
        }
```

### DIFF
--- a/internal/pkg/api/handleAck.go
+++ b/internal/pkg/api/handleAck.go
@@ -333,9 +333,9 @@ func (ack *AckT) handlePolicyChange(ctx context.Context, zlog zerolog.Logger, ag
 			Str("agent.policyId", agent.PolicyID).
 			Int64("agent.revisionIdx", currRev).
 			Int64("agent.coordinatorIdx", currCoord).
-			Str("rev.policyId", rev.PolicyID).
-			Int64("rev.revisionIdx", rev.RevisionIdx).
-			Int64("rev.coordinatorIdx", rev.CoordinatorIdx).
+			Str("rev_policyId", rev.PolicyID).
+			Int64("rev_revisionIdx", rev.RevisionIdx).
+			Int64("rev_coordinatorIdx", rev.CoordinatorIdx).
 			Msg("ack policy revision")
 
 		if ok && rev.PolicyID == agent.PolicyID &&


internal/pkg/api/handleAck.go
```diff
@@ -333,9 +333,9 @@ func (ack *AckT) handlePolicyChange(ctx context.Context, zlog zerolog.Logger, ag
 			Str("agent.policyId", agent.PolicyID).
 			Int64("agent.revisionIdx", currRev).
 			Int64("agent.coordinatorIdx", currCoord).
-			Str("rev.policyId", rev.PolicyID).
-			Int64("rev.revisionIdx", rev.RevisionIdx).
-			Int64("rev.coordinatorIdx", rev.CoordinatorIdx).
+			Str("rev_policyId", rev.PolicyID).
+			Int64("rev_revisionIdx", rev.RevisionIdx).
+			Int64("rev_coordinatorIdx", rev.CoordinatorIdx).
```
AndersonQ, line 338: I don't want to hold off the fix, but isn't the convention to use `.` to separate the names in the log fields? Shouldn't we keep it, then perhaps finding another name for these fields?

AndersonQ (approved):
It isn't a blocker, so I'm approving it. However I believe we should define a standard for out log field names

internal/pkg/api/handleAck.go
```diff
@@ -333,9 +333,9 @@ func (ack *AckT) handlePolicyChange(ctx context.Context, zlog zerolog.Logger, ag
 			Str("agent.policyId", agent.PolicyID).
 			Int64("agent.revisionIdx", currRev).
 			Int64("agent.coordinatorIdx", currCoord).
-			Str("rev.policyId", rev.PolicyID).
-			Int64("rev.revisionIdx", rev.RevisionIdx).
-			Int64("rev.coordinatorIdx", rev.CoordinatorIdx).
+			Str("rev_policyId", rev.PolicyID).
```
blakerouse, line 336:
```suggestion
			Str(LogPolicyID, rev.PolicyID).
```

internal/pkg/api/handleAck.go
```diff
@@ -333,9 +333,9 @@ func (ack *AckT) handlePolicyChange(ctx context.Context, zlog zerolog.Logger, ag
...
+			Int64("rev_revisionIdx", rev.RevisionIdx).
+			Int64("rev_coordinatorIdx", rev.CoordinatorIdx).
 			Msg("ack policy revision")
```
blakerouse, line 339:
```suggestion
			Msg("searching ack for valid policy revision")
```

blakerouse (changes requested):
It would be better to update the fields to match its similar `zlog.Info()` below. An update to the message would be helpful to know this debug log is just it searching to ensure its an ACK it should even accept.

michel-laterman:
@joshdover will you have time to update/address the comments in this pr? (and also add a changelog fragment?)

joshdover:
I won't be able to get to it this week or next due to being out of office
and on PTO. If we'd like to get this in for 8.8, feel free to take it over.

juliaElastic (approved):

PR #1684, opened 2022-07-22
Allow multiple ES outputs as long as they are the same ES

AndersonQ:
## What is the problem this PR solves?

It fixes #1672. Currently fleet-server does not handle correctly the multiple ES outputs the fleet-ui might create.

## How does this PR solve the problem?

It creates a non-breaking change on the agent documents to store ES API key information per output.

## How to test this PR locally

- add a second ES output:
    - go to Fleet > Settings
    - outputs > add output
    - just copy the data from the `default` output and git it another name. Let's call it `non-default`
- create/edit a policy to use `non-default` output for integrations. Let the monitoring using the default:
    - navigate to Fleet > Agent policies > your_policy
    - go to the settings "tab"
    - select `non-default` as the output for _Output for integrations_. (You'll need a paid license or use Elastic Cloud)
- enroll a Elastic Agent on this policy with some integration

At first both outputs should have their on API keys. Use [`elastic-agent inspect`](https://www.elastic.co/guide/en/fleet/current/elastic-agent-cmd-options.html#elastic-agent-inspect-command) to check that.

Now modify the policy in a way that one of the outputs has its permissions changed, but not the other. I usually add and remove the Endpoint Security. Then both outputs should have the same API key, again, use the `inspect` command to check.

## Checklist

- [x] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation
- [ ] I have made corresponding change to the default configuration files
- [x] I have added tests that prove my fix is effective or that my feature works
- [ ] I have added an entry in `CHANGELOG.next.asciidoc` or `CHANGELOG-developer.next.asciidoc`.

## Author's checklist

- [x] Modify the agent model to support multiple ES outputs: add fields, don't change the existing ones, keep backward compatible
- [x] ensure new agent model is correctly saved and loaded (Marshal/UnMarshaled)
- [x] add migration process to internal/pkg/dl/migration.go
    - [x] add doc explaining the migration
- [x] when preparing ES output: 
    - [x] set the default API key to empty -> force all API keys to be regenerated
    - [x] set old fields to empty string/nil
    - [x] use new fields to store output API key data 
- [x] update painless script to use new agent model
- [x] ensure the `model.Agent` aren't used (except for tests checking they aren't populated anymore)
    - [x] update `handleAck` to use the new model
    - [x] update `monitor` to use the new model
- [x] on upgrade bump the coordinator index
[...]

PR #1878, opened 2022-09-19
Revert "Fix v8.5.0 migration painless script"

AndersonQ:
Reverts elastic/fleet-server#1684 and
Reverts elastic/fleet-server#1839


cmacknz (approved):

cmacknz:
Can we just revert both changes as two separate commits from this PR so it all gets merged together?

AndersonQ:
> Can we just revert both changes as two separate commits from this PR so it all gets merged together?

yes, of course. I updated the PR

cmacknz (approved):

Issue #1877, opened 2022-09-19
8.5.0-SNAPSHOT Endpoint can't connect to ES


kevinlog:
When installing the latest 8.5.0-SNAPSHOT Agent and Endpoint, it cannot connect to Elasticsearch.

The version I used in this bug report: https://snapshots.elastic.co/8.5.0-c7913db3/downloads/beats/elastic-agent/elastic-agent-8.5.0-SNAPSHOT-linux-arm64.tar.gz

There are two somewhat different scenarios depending on the protocol you use when installing the Agent.

#### Using `--insecure` flag when installing Agent
The first scenario I encountered was when I was doing local development and testing.  I used the `--insecure` flag when installing the Agent on my host to connect to my insecure local development server.

When installing, I can see in the logs that the Endpoint cannot connect to Elasticsearch.  Find a snippet below, full logs attached.

```
...
{"@timestamp":"2022-09-14T20:06:34.801859694Z","agent":{"id":"6d7207c2-1e6e-4fa8-b8b9-70548a06e232","type":"endpoint"},"ecs":{"version":"1.11.0"},"log":{"level":"notice","origin":{"file":{"line":86,"name":"BulkQueueConsumer.cpp"}}},"message":"BulkQueueConsumer.cpp:86 Elasticsearch connection is down","process":{"pid":106912,"thread":{"id":106919}}}
...
```

I can see in the policy yaml pulled from the host itself that there is ES connection information.  Using this same Policy, I can successfully connect `8.4.0` Agent/Endpoint.  Find a snipped below, full policy yml attached.

```
...
output:
  elasticsearch:
    api_key: ****
    hosts:
    - http://10.0.0.199:9200
revision: 4
...
```

[endpoint.log.zip](https://github.com/elastic/fleet-server/files/9599210/endpoint.log.zip)
[elastic-endpoint.yml.zip](https://github.com/elastic/fleet-server/files/9599212/elastic-endpoint.yml.zip)

#### Connecting to a secure cloud instance (no `--insecure` flag)
I also was unable to get the Endpoint to connect to ES when connecting to a secure cloud instance.

I tried to reproduce again on a secure cloud server to eliminate my insecure, local `http` connection and I'm still seeing rhe same end result of data not making it to my stack, although the logs have a bit more context surrounding invalidated API keys.

```
[...]

PR #1879, opened 2022-09-19
Fix and reintroduce "Allow multiple ES outputs as long as they are the same ES"

AndersonQ:
Reverts #1878, adding #1684, #1839 and fixes https://github.com/elastic/fleet-server/issues/1877


michalpristas (approved):
change looks good, also it's guarded in ack now

internal/pkg/policy/policy_output_integration_test.go
```diff
@@ -125,3 +127,74 @@ func TestRenderUpdatePainlessScript(t *testing.T) {
...
+		"test output": map[string]interface{}{},
+	}
+
+	err = output.prepareElasticsearch(
+		context.Background(), zerolog.Nop(), bulker, &agent, policyMap)
+	require.NoError(t, err)
+
+	// need to wait a bit before querying the agent again
+	// TODO: find a better way to query the updated agent
+	time.Sleep(time.Second)
```
cmacknz, line 159: You can use [require.Eventually](https://pkg.go.dev/github.com/stretchr/testify/require#Eventually) to poll for the state you need. There will still technically be a timeout, but it will allow the test to complete faster than the timeout by polling with shorter intervals.

internal/pkg/dl/constants.go
```diff
@@ -27,22 +27,23 @@ const (
...
+	FieldPolicyCoordinatorIdx          = "policy_coordinator_idx"
+	FieldPolicyID                      = "policy_id"
+	FieldPolicyOutputAPIKey            = "api_key"
+	FieldPolicyOutputAPIKeyID          = "api_key_id"
+	FieldPolicyOutputPermissionsHash   = "permissions_hash"
+	FieldPolicyOutputToRetireAPIKeyIDs = "to_retire_api_key_ids" //nolint:gosec // false positive
+	FieldPolicyRevisionIdx             = "policy_revision_idx"
+	FieldRevisionIdx                   = "revision_idx"
+	FieldUnenrolledReason              = "unenrolled_reason"
+	FiledType                          = "type"
```
cmacknz, line 46:
```suggestion
	FieldType                          = "type"
```

internal/pkg/policy/policy_output.go
```diff
@@ -32,163 +32,208 @@ var (
...
+
+		fields := map[string]interface{}{
+			dl.FieldPolicyOutputAPIKey:          outputAPIKey.Agent(),
+			dl.FieldPolicyOutputAPIKeyID:        outputAPIKey.ID,
+			dl.FieldPolicyOutputPermissionsHash: p.Role.Sha2,
+		}
 
-			if err != nil {
-				return err
+		if !foundOutput {
```
cmacknz, line 193: What was the fix for the bug? I see this if statement added in the commit that contains a fix, along with the block below being moved later, but it isn't actually clear to me why this fixes the problem.

```go
		// Now that all is done, we can update the output on the agent variable
		// Right not it's more for consistency and to ensure the in-memory agent
		// data is correct and in sync with ES, so it can be safely used after
		// this method returns.
		output.Type = OutputTypeElasticsearch
		output.APIKey = outputAPIKey.Agent()
		output.APIKeyID = outputAPIKey.ID
		output.PermissionsHash = p.Role.Sha2 // for the sake of consistency
```

internal/pkg/policy/policy_output_integration_test.go
```diff
@@ -0,0 +1,200 @@
...
+
+	got, err := dl.FindAgent(
+		context.Background(), bulker, dl.QueryAgentByID, dl.FieldID, agentID, dl.WithIndexName(index))
+	if err != nil {
+		require.NoError(t, err, "failed to find agent ID %q", agentID)
+	}
+
+	gotOutput, ok := got.Outputs[output.Name]
+	require.True(t, ok, "no '%s' output fouled on agent document", output.Name)
```
cmacknz, line 168:
```suggestion
	require.True(t, ok, "no '%s' output failed on agent document", output.Name)
```

internal/pkg/policy/policy_output_integration_test.go
```diff
@@ -0,0 +1,200 @@
...
+			require.NoError(t, err)
+
+			assert.Equal(t, agentID, gotAgent.Id)
+			assert.Len(t, gotAgent.Outputs, len(wantOutputs))
+			assert.Equal(t, wantOutputs, gotAgent.Outputs)
+		})
+	}
+}
+
+func TestPolicyOutputESPrepareRealES(t *testing.T) {
```
cmacknz, line 131: Does this new test reproduce the API key invalidation problem?

PR #1887, opened 2022-09-20
Allow upgrade action to signal retry

michel-laterman:

Allow agent to retry upgrade operations that have failed.

Allow the ack of an upgrade action to set the upgrade status to retrying.
The event associated with the upgrade action should have a non-empty `error` attribute, and a `payload` attribute that is a json object with `retry: bool` and `retry_attempt: int` attributes.

TODO

## Checklist

## Related issues

- Relates https://github.com/elastic/elastic-agent/issues/778
- Agent PR: https://github.com/elastic/elastic-agent/pull/1219


internal/pkg/api/handleAck.go
```diff
@@ -505,14 +505,30 @@ func (ack *AckT) handleUnenroll(ctx context.Context, zlog zerolog.Logger, agent
...
+		}
+		err := json.Unmarshal(event.Payload, &pl)
+		if err != nil {
+			zlog.Error().Err(err).Msg("unable to unmarshal upgrade event payload")
+		}
+
+		// if the payload indicates a retry, mark change the upgrade status to retrying.
+		if pl.Retry {
+			zlog.Info().Int("retry_attempt", pl.Attempt).Msg("marking agent upgrade as retrying")
+			doc[dl.FieldUpgradeStatus] = "retrying" // TODO should we also change FieldUpgradedAt and FieldUpgradeStated at?
```
juliaElastic, line 529: Is there a max retry count? I am wondering what happens if the action is still not successful after x retries, should we set the status to failed?

michel-laterman: changed, if there is an error with no retry then it will set to failed

juliaElastic (approved):
Code LGTM

internal/pkg/api/handleAck.go
```diff
@@ -503,13 +503,33 @@ func (ack *AckT) handleUnenroll(ctx context.Context, zlog zerolog.Logger, agent
...
+		} else {
+			zlog.Info().Int("retry_attempt", pl.Attempt).Msg("marking agent upgrade as failed")
```
michalpristas, line 529: how will we handle stale agent (older versions without retry in ack)? is it expected for them not to retry?
i'd say yes, just want to double check with your intentions.

michalpristas: `FieldUpgradeStartedAt` and `FieldUpgradedAt` is used in Fleet UI side to set status `Upgrading` 
i wonder if we could mark `FieldUpgradedAt` in case of failed retry as it was not upgraded so far.

michel-laterman: Yeah, a pre v8.6 agent that reports an error will set `FieldUpgradeStatus` to `failed`. However I'm not sure if it is able to send that ack as I don't think it could actually ack a failure.

If it sends a successful upgrade ack `FieldUpgradeStatus` will be set to nil and `FieldUpgradedAt` is set to the current time (keeping the current behaviour)

@juliaElastic should we be setting `FieldUpgradedAt` and clearing `FieldUpgradeStartedAt` in failure cases, or should we keep these attributes to their original values?

juliaElastic: Currently there is no `Failed` agent status. So the agent can either remain in `Updating` or move back to `Healthy` status. The error should be captured in action results similar to other action failures. 
So I think the agent can move back to `Healthy` state if the upgrade failed (supposing nothing changed for the agent) - this means setting `FieldUpgradeStartedAt` to nil and leaving `FieldUpgradedAt` at the previous value.

internal/pkg/api/handleAck.go
```diff
@@ -505,11 +505,7 @@ func (ack *AckT) handleUnenroll(ctx context.Context, zlog zerolog.Logger, agent
 
 func (ack *AckT) handleUpgrade(ctx context.Context, zlog zerolog.Logger, agent *model.Agent, event Event) error {
 	now := time.Now().UTC().Format(time.RFC3339)
-	doc := bulk.UpdateFields{
-		dl.FieldUpgradeStartedAt: nil,
-		dl.FieldUpgradeStatus:    nil,
-		dl.FieldUpgradedAt:       now,
-	}
+	doc := bulk.UpdateFields{}
```
juliaElastic, line 508: what does an empty UpdateFields do?

internal/pkg/api/handleAck.go
```diff
@@ -524,10 +520,20 @@ func (ack *AckT) handleUpgrade(ctx context.Context, zlog zerolog.Logger, agent *
...
-			doc[dl.FieldUpgradeStatus] = "retrying" // TODO should we also change FieldUpgradedAt and FieldUpgradeStated at?
+			doc[dl.FieldUpgradeStatus] = "retrying" // Keep FieldUpgradeStatedAt abd FieldUpgradeded at to original values
 		} else {
-			zlog.Info().Int("retry_attempt", pl.Attempt).Msg("marking agent upgrade as failed")
-			doc[dl.FieldUpgradeStatus] = "failed" // TODO should we also change FieldUpgradedAt and FieldUpgradeStated at?
+			zlog.Info().Int("retry_attempt", pl.Attempt).Msg("Agent upgrade failed, marking agent as healthy, agent logs contain failure message")
+			doc = bulk.UpdateFields{
+				dl.FieldUpgradeStartedAt: nil,
+				dl.FieldUpgradeStatus:    nil,
+				dl.FieldUpgradedAt:       now,
```
juliaElastic, line 529: I meant more like keeping the previous `upgraded_at` value, but we might not have that anymore. I think Fleet UI clears this field when starting the action.

internal/pkg/api/handleAck.go
```diff
@@ -524,10 +520,20 @@ func (ack *AckT) handleUpgrade(ctx context.Context, zlog zerolog.Logger, agent *
...
 			zlog.Info().Int("retry_attempt", pl.Attempt).Msg("marking agent upgrade as retrying")
-			doc[dl.FieldUpgradeStatus] = "retrying" // TODO should we also change FieldUpgradedAt and FieldUpgradeStated at?
+			doc[dl.FieldUpgradeStatus] = "retrying" // Keep FieldUpgradeStatedAt abd FieldUpgradeded at to original values
 		} else {
-			zlog.Info().Int("retry_attempt", pl.Attempt).Msg("marking agent upgrade as failed")
-			doc[dl.FieldUpgradeStatus] = "failed" // TODO should we also change FieldUpgradedAt and FieldUpgradeStated at?
+			zlog.Info().Int("retry_attempt", pl.Attempt).Msg("Agent upgrade failed, marking agent as healthy, agent logs contain failure message")
+			doc = bulk.UpdateFields{
+				dl.FieldUpgradeStartedAt: nil,
+				dl.FieldUpgradeStatus:    nil,
```
juliaElastic, line 528: we can set upgrade status to failed here

internal/pkg/api/handleAck.go
michel-laterman: I think it's a nop

michalpristas (approved):
Lgtm now

internal/pkg/api/handleAck.go
juliaElastic: can we remove it then?

PR #1891, opened 2022-09-21
Update pre-sets limits to avoid overlap.

### LINKED ISSUE #1858: Load limit presets should not overlap on max agents
Our presets currently overlap by 1 which ends up in having unexpected behavior if you configure max agents to a value like 30000. In this case you'll end up using the 12,500 - 30,000 presets rather than the >= 30000 preset:

```
Using system limits for 12500 to 30000 agents for a configured value of 30000 agents
```

You can see how the bug presents itself:
- 12.5k preset has a max of 30k: https://github.com/elastic/fleet-server/blob/0181e484a71fe98864a2c96866b3594970116612/internal/pkg/config/defaults/gt12500_limits.yml#L3
- \> 30k preset has a min of 30k: https://github.com/elastic/fleet-server/blob/0181e484a71fe98864a2c96866b3594970116612/internal/pkg/config/defaults/max_limits.yml#L2
- The code that iterates through these to choose a preset: https://github.com/elastic/fleet-server/blob/5b1c0df774371a15bfbbaf735de6cdeaa856c38c/internal/pkg/config/env_defaults.go#L181

This is sort of related to:
- https://github.com/elastic/fleet-server/issues/1841

jlind23:
Fixes #1858


michalpristas (approved):
looks good, for a manager

PR #1896, opened 2022-09-21
Catch error in waitBulkAction. Add bulk.WithRetryOnConflict(3) in multiple places.

### LINKED ISSUE #1894: Bulker swallows errors from elasticsearch when the error is specific to the document
## Overview

Inside of the `BulkIndexerResponseItem` there is an error field that could have possible been set but is ignored in here https://github.com/elastic/fleet-server/blob/main/internal/pkg/bulk/opBulk.go#L77.

## What should be done?

Any code path that reads a `BulkIndexerResponseItem` should check the `Error` field and ensure that an error has not occurred. If an error has occurred then it should return an error instead of continuing on.

Relates to https://github.com/elastic/elastic-agent/issues/760

blakerouse:

It solves an issue where calling an action through the bulker would result in errors on a single document being ignored. It also adds `bulk.WithRetryOnConflict(3)` in many places that update the `.fleet-agents` index. Being that the `.fleet-agents` index is written to often it is possible that conflicts will occur and elasticsearch should retry on conflicts.

This changes the `waitBulkAction` to look at the status code and error document of a bulk action response to ensure that an error should not be returned instead of the document.

Scale test runs are the only place to really test that this solves the issue.

## Checklist

## Related issues

- Closes #1894

### DIFF
--- a/CHANGELOG.next.asciidoc
+++ b/CHANGELOG.next.asciidoc
@@ -10,7 +10,7 @@
 - Remove events from agent checkin body. {issue}1774[1774]
 - Improve authc debug logging. {pull}1870[1870]
 - Add error detail to catch-all HTTP error response. {pull}1854[1854]
-- Fix issue where errors where being ignored writting to elasticsearch. {pull}1896[1896]
+- Fix issue were errors where being ignored written to elasticsearch. {pull}1896[1896]
 
 ==== New Features
 
--- a/CHANGELOG.next.asciidoc
+++ b/CHANGELOG.next.asciidoc
@@ -10,6 +10,7 @@
 - Remove events from agent checkin body. {issue}1774[1774]
 - Improve authc debug logging. {pull}1870[1870]
 - Add error detail to catch-all HTTP error response. {pull}1854[1854]
+- Fix issue where errors where being ignored writting to elasticsearch. {pull}1896[1896]
 
 ==== New Features
 
--- a/internal/pkg/policy/policy_output.go
+++ b/internal/pkg/policy/policy_output.go
@@ -157,7 +157,7 @@ func (p *Output) prepareElasticsearch(
 			return err
 		}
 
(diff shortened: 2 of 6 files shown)


CHANGELOG.next.asciidoc
```diff
@@ -10,6 +10,7 @@
 - Remove events from agent checkin body. {issue}1774[1774]
 - Improve authc debug logging. {pull}1870[1870]
 - Add error detail to catch-all HTTP error response. {pull}1854[1854]
+- Fix issue where errors where being ignored writting to elasticsearch. {pull}1896[1896]
```
cmacknz, line 13:
```suggestion
- Fix issue were errors where being ignored written to elasticsearch. {pull}1896[1896]
```

internal/pkg/policy/policy_output.go
```diff
@@ -157,7 +157,7 @@ func (p *Output) prepareElasticsearch(
 			return err
 		}
 
-		if err = bulker.Update(ctx, dl.FleetAgents, agent.Id, body); err != nil {
+		if err = bulker.Update(ctx, dl.FleetAgents, agent.Id, body, bulk.WithRefresh(), bulk.WithRetryOnConflict(3)); err != nil {
```
cmacknz, line 160: 1. What does `WithRefresh` do?
2. Is there any place where we don't want these options? Can we make the default so we can't forget the next time someone adds an Update request?

cmacknz (approved):
LGTM, minus typos in the changelog.

internal/pkg/policy/policy_output.go
blakerouse: It is not standard elasticsearch for refresh to be there, and there are places where we do not want refresh. It just happens that in most of these cases we do want refresh.

blakerouse: https://www.elastic.co/guide/en/elasticsearch/reference/current/docs-refresh.html

PR #1901, opened 2022-09-22
Refactor prepare Elasticsearch output

AndersonQ:

The current [Elasticsearch output preparation](https://github.com/elastic/fleet-server/blob/f77b97c875664beb267dd8695aba5b150da12679/internal/pkg/policy/policy_output.go#L64) has grown a way too complex and log, with repeated code and side effects that are hard to track and keep in mind when working on it. Besides it's such a critical piece of code that it's not advised to refactor it while intorducing a feature or fixing a bug.

It splits out the code to create and update a key, also it avoids side effects on these 2 new functions

Nothing should break as it only changes implementation details

## Checklist


internal/pkg/policy/policy_output.go
```diff
@@ -91,146 +89,166 @@ func (p *Output) prepareElasticsearch(
...
+	}
 
-		if err = bulker.Update(ctx, dl.FleetAgents, agent.Id, body, bulk.WithRefresh(), bulk.WithRetryOnConflict(3)); err != nil {
-			zlog.Error().Err(err).Msg("fail update agent record")
-			return err
-		}
+	// Using painless script to append the old keys to the history
+	body, err := renderUpdatePainlessScript(p.Name, fields)
+	if err != nil {
+		return nil, fmt.Errorf("could no tupdate painless script: %w", err)
```
michel-laterman, line 186:
```suggestion
		return nil, fmt.Errorf("could not update painless script: %w", err)
```

michel-laterman (approved):

michel-laterman:
@AndersonQ, can you see if this needs to be updated + merge?

PR #1902, opened 2022-09-22
File Upload Feature

pzl:

Adding the ability to upload files from integrations, through fleet server, into elasticsearch

Link to [Swagger docs for API contract](https://petstore.swagger.io/?url=https://gist.githubusercontent.com/pzl/1374c45a28d51b048a015fa71189caed/raw/270fdd19c06afb8d0383f34dc4f43205edb0c50e/fleet-server-api.json)

A followup PR is under way to add this API to the new `openapi.yml`.

Extracting a few heuristics (file size limit, timeout) into fleet server configs will also come as a followup during scale testing

Adds three HTTP routes,

- **POST** `/api/fleet/uploads`: Begins an "upload operation". Includes the full metadata about a file, with some required fields like name and size
- **PUT** `/api/fleet/uploads/<uploadID>/<chunkNum>`: Uploading a segment (chunk) of the file contents.
- **POST** `/api/fleet/uploads/<uploadID>`: Completes an "upload operation". Fleet server verifies all contents were uploaded, and initially-provided checksums match

These routes are API-Key authorized, and as well deal with file chunking and checksums. It may be difficult to test routes in isolation without a nearly-fully-operational client implementation. Both Endpoint integration and the Agent itself are developing client implementations for this release.

If one were to (for development or checking purposes) disable the API key checking, they may be able to use this gist as a [mock file upload client](https://gist.github.com/pzl/39b052f55e646cfebcf970d0478566ad).

## Checklist

- [ ] I have made corresponding changes to the documentation (update to openapi.yml spec as a followup)
- [ ] I have made corresponding change to the default configuration files (adding configs coming in followup)


internal/pkg/api/handleUpload.go
```diff
@@ -0,0 +1,381 @@
...
+	UploadDel      UploadStatus = "DELETED"
+)
+
+const (
+	// TODO: move to a config
+	maxParallelUploadOperations = 3
+	maxParallelChunks           = 4
+	maxFileSize                 = 104857600 // 100 MiB
+
+)
```
michel-laterman, line 53: I think server limits would be a good place to put this

internal/pkg/api/schema.go
```diff
@@ -175,3 +175,47 @@ type StatusResponse struct {
...
+		Hostname string `json:"hostname"`
+	} `json:"host"`
+}
+
+type FileData struct {
+	Size        int64  `json:"size"`
+	Name        string `json:"name"`
+	Extension   string `json:"ext"`
+	Mime        string `json:"mime_type"`
+	Compression string `json:"Compression"`
```
michel-laterman, line 198: (nit) Do we want a capitalized attribute here?

internal/pkg/api/schema.go
```diff
@@ -175,3 +175,47 @@ type StatusResponse struct {
...
+	Hash        struct {
+		SHA256 string `json:"sha256"`
+		MD5    string `json:"md5"`
+	}
```
michel-laterman, line 202: (nit) Here Hash will be capitalized as well

internal/pkg/model/schema.go
```diff
@@ -1,7 +1,3 @@
-// Copyright Elasticsearch B.V. and/or licensed to Elasticsearch B.V. under one
-// or more contributor license agreements. Licensed under the Elastic License;
-// you may not use this file except in compliance with the Elastic License.
-
```
michel-laterman, line 4: you'll need to re-add the headers after generating this file

internal/pkg/upload/doc.go
```diff
@@ -0,0 +1,14 @@
+// Copyright Elasticsearch B.V. and/or licensed to Elasticsearch B.V. under one
+// or more contributor license agreements. Licensed under the Elastic License;
+// you may not use this file except in compliance with the Elastic License.
+
+/*
+
+	This package concentrates the responsibility of following the
+	File Storage specification:
+	https://docs.google.com/document/d/1BlCQYxHwgGZMxysc0BiWxT9ZCLKwgryhkD9ryDNDIuI/edit?usp=sharing
```
michel-laterman, line 9: This file is not publicly accessible, please summarize what the upload package is for instead.

Does this module upload files to ES?

model/schema.json
```diff
@@ -607,6 +607,202 @@
...
+          "type": "string"
+        },
+        "file": { "$ref": "#/definitions/file-data" },
+        "contents": {
+          "description": "Description of the contents of an archive, when the file property describes an archive (e.g. zip)",
+          "type": "array",
+          "items": {
+            "ref": "#/definitions/file-data"
+          }
+        },
```
michel-laterman, line 666: I have yet to read the proposal so I may be lacking in knowledge here. But is `contents` required for an archive? or just recommended?

To give some context: my use case for these endpoints will be having the `elastic-agent` upload a diagnostics bundle (zip file).

internal/pkg/upload/upload.go
```diff
@@ -0,0 +1,340 @@
...
+		parallelChunkLimit: chunkLimit,
+		sizeLimit:          sizeLimit,
+		opThrottle:         throttle.NewThrottle(opLimit),
+		current:            make(map[string]upload, opLimit),
+	}
+}
+
+// Start an upload operation, as long as the max concurrent has not been reached
+// returns the upload ID
+func (u *Uploader) Begin(size int64, docID string, source string, hashsum string, hasher hash.Hash) (Info, error) {
```
michel-laterman, line 92: Should we consider adding a context as well?

internal/pkg/upload/upload.go
```diff
@@ -0,0 +1,340 @@
...
+		Total:     size,
+		Hasher:    hasher,
+		HashSum:   hashsum,
+	}
+	cnt := info.Total / info.ChunkSize
+	if info.Total%info.ChunkSize > 0 {
+		cnt += 1
+	}
+	info.Count = int(cnt)
+	u.current[id] = upload{
```
michel-laterman, line 167: needs mutex access

internal/pkg/upload/upload.go
```diff
@@ -0,0 +1,340 @@
...
+	opToken       *throttle.Token
+	chunkThrottle *throttle.Throttle
+	complete      chan<- struct{}
+	chunkRecv     chan<- struct{}
+	begun         bool
+	Info          Info
+}
+
+type Uploader struct {
+	current            map[string]upload
```
michel-laterman, line 55: perhaps [sync.Map](https://pkg.go.dev/sync#Map)

internal/pkg/upload/upload.go
```diff
@@ -0,0 +1,340 @@
...
+	mu                 sync.Mutex
```
michel-laterman, line 56: Is this mutex meant to control access to the current map, or each upload?

internal/pkg/upload/upload.go
```diff
@@ -0,0 +1,340 @@
...
+		complete:      complete,
+		chunkRecv:     chunkRecv,
+		Info:          info,
+	}
+	return info, nil
+}
+
+func (u *Uploader) Chunk(uplID string, chunkID int) (ChunkInfo, error) {
+	u.mu.Lock()
+	defer u.mu.Unlock()
```
michel-laterman, line 179: With this locking, 2 two chunks can be uploaded at once, is that desired?

michel-laterman (commented):
I did a very quick 1st look here.
I can see that currently the api handers are interacting with `upload`, and `dl` to add files to ES.

However I think a cleaner implementation would be to have the ES interactions in `upload` so that the caller can just call something like `upload.Chunk(uplID, chunkID, r.Body) error`. What do you think?

internal/pkg/api/schema.go
pzl: most of these fields follow the ecs convention that lowercased fields represent an ecs standard, and initial-uppercase is reserved for custom fields not present in ECS.

Most of these fields are following the [ecs file](https://www.elastic.co/guide/en/ecs/current/ecs-file.html) standard. The uppercased are custom extensions, and are part of the [storage spec](https://docs.google.com/document/d/1BlCQYxHwgGZMxysc0BiWxT9ZCLKwgryhkD9ryDNDIuI/edit?usp=sharing). The capitalization needs to match for a Kibana App Services file plugin to read the known format

internal/pkg/api/schema.go
pzl: good catch, this one is ecs and should be lowercase

model/schema.json
pzl: contents is not required. It will be used by Endpoint since the facts surrounding the file _within_ the zip may be very relevant to an investigation. For agent, unless one wants to know the hash or details or contents of the archive, it is very much not required.

If multiple files were present within the archive, and one wanted to display that fact in a UI on kibana (e.g. `"Agent Diagnostics 2022-10-20  (3 files)"`) then it could be used for that purpose.

internal/pkg/api/schema.go
```diff
@@ -175,3 +175,47 @@ type StatusResponse struct {
...
+	Accessed    string   `json:"accessed"`
+	Attributes  []string `json:"attributes"`
+	Created     string   `json:"created"`
+	CTime       string   `json:"ctime"`
+	Device      string   `json:"device"`
+	Directory   string   `json:"directory"`
+	DriveLetter string   `json:"drive_letter"`
+	Ext         string   `json:"extension"`
```
michel-laterman, line 210: Line 196 has
```
Extension string `json:"ext"`
```
It's a a bit confusing

pzl: This is all going to get removed soon. After some back and forth, mapping will be set to dynamic so fleet server has no need to know or decode whatever the JSON structure will be. So most of these fields are getting removed, just need to decide on a nice way for us to inject one or two properties into an otherwise nearly-anything-goes payload, and check a few minimum-required fields.

internal/pkg/upload/upload.go
pzl: this function isn't performing the chunk upload, but checking that that chunk number (`chunkID`) _can_ be uploaded for that upload ID. 

Checking bounds (valid Chunk ID, and it is not an unexpectedly large ID for the given file size), and checks against the throttle for that upload.

A given upload can have a few chunks in parallel (the `chunkThrottle` below). Each upload has its own throttle.

This also returns some info so the uploading logic knows whether this is the final chunk in the sequence (which affects a parameter in the chunk document). It also is watching for the "first" chunk activity to start a total-upload timer.

It's probably more of "canChunkUploadProceed", but I expect some refactoring here anyway, as you pointed out the `dl` layer and things. I have no problem moving around _where_ the uploading takes place, or where some of these checks and responsibilities end up

internal/pkg/upload/upload.go
pzl: the initial write should be fine.  `id` will be some freshly generated random value so there is near zero chance its overwriting anything existing. Just adding a new value. And no client has been told the `id` yet from any responses, so nothing should be concurrently checking its value

kevinlog:
I did a dev build of this PR and packaged it with a dev Agent for an e2e test - it works!

The steps I took:
1. Build a dev fleet-server by checking out this PR and running: `DEV=true SNAPSHOT=true make release-linux/amd64`
2. Package it with a dev Agent and running in a Vagrant VM
3. Connect an Agent/Endpoint via the deployed Fleet server
4. Call our response action to fetch a file from the Agent/Endpoint and download it through the browser.
5. I tried fetching files of various sizes between 1 - 100 MB and all were successful.

image

from a functionality standpoint 🚢

internal/pkg/upload/es.go
```diff
@@ -0,0 +1,172 @@
...
+	"github.com/elastic/go-elasticsearch/v7"
+	"github.com/elastic/go-elasticsearch/v7/esapi"
+	"github.com/rs/zerolog/log"
+)
+
+const (
+	// @todo: neither of these should be static. But should be specific to an integration
+	// somewhat configurable, but need to follow a pattern so that Fleet Server has write access
+	FileHeaderIndexPattern = ".fleet-%s-files"
+	FileDataIndexPattern   = ".fleet-%s-file-data"
```
kevinlog, line 27: based on this conversation: https://github.com/elastic/security-team/issues/4661#issuecomment-1309075519

We should update these to be `.fleet-files-%s` and `.fleet-file-data-%s` .

internal/pkg/upload/upload.go
cmacknz: Just to confirm, concurrent access to this map is impossible? Concurrent access to a map will panic so I would recommend being defensive with a mutex if there is any possibility of concurrent access here.

pzl: Logically impossible, but I'm happy to wrap this in a lock to be safe

internal/pkg/api/handleUpload.go
```diff
@@ -0,0 +1,390 @@
...
+}
+
+func (ut *UploadT) handleUploadComplete(zlog *zerolog.Logger, w http.ResponseWriter, r *http.Request, uplID string) error {
+	info, err := ut.upl.Complete(uplID, ut.bulker)
+	if err != nil {
+		return err
+	}
+
+	if err := updateUploadStatus(r.Context(), ut.bulker, info, UploadDone); err != nil {
+		// should be 500 error probably?
```
michel-laterman, line 322: In this error case, the complete file has been uploaded and verified, but the status is stuck, a 500 would be safe, but it would be a very different error then if the `Complete` call above failed.

internal/pkg/upload/upload.go
```diff
@@ -0,0 +1,341 @@
...
+	ok, err = u.verifyChunkData(info.Info, bulker)
+	if err != nil {
+		return Info{}, err
+	}
+	if !ok {
+		return Info{}, errors.New("file contents did not pass validation")
+	}
+
+	// @todo: verify chunks
+	// verify hashes, etc
```
cmacknz, line 231: It looks like this might be taken care of by `verifyChunkData` above, and can be removed.

internal/pkg/upload/upload.go
```diff
@@ -0,0 +1,341 @@
...
+	chunkT := time.NewTimer(chunkProgressTimeout)
+	chunkRecv := make(chan struct{})
+	complete := make(chan struct{})
+	// total timer could also be achieved with context deadline and cancelling
+
+	go func() {
+		for {
+			select {
+			case <-total.C: // entire upload operation timed out
+				log.Trace().Str("uploadID", id).Msg("upload operation timed out")
```
cmacknz, line 120: Is there a way to associate an uploadID back to the agent/endpoint host it originated from? If we run into issues where uploads from particular hosts are timing out, is there a way for us to figure out what hosts those are so we can correlate this to agent/endpoint logs?

Including the agent ID in addition to the uploadID is an easy way to solve this.

internal/pkg/upload/upload.go
```diff
@@ -0,0 +1,341 @@
...
+		}
+	}
+	delete(u.current, uplID)
+}
+
+func (u *Uploader) cancel(uplID string) error {
+	u.cleanupOperation(uplID)
+
+	// @todo: delete any uploaded chunks from ES
+	// leave header doc and mark failed?
```
cmacknz, line 251: How important is this? Is it possible that orphaned chunks could accumulate in ES over and over again because uploads from an agent on a poor network connection can never complete?

internal/pkg/upload/upload.go
```diff
@@ -0,0 +1,339 @@
...
+	return nil
+}
+
+func (u *Uploader) allChunksPresent(info Info, bulker bulk.Bulk) (bool, error) {
+	hits, err := ListChunkIDs(context.TODO(), bulker, info.Source, info.DocID)
+	if err != nil {
+		log.Warn().Err(err).Msg("error listing chunks")
+		return false, err
+	}
+	if len(hits) != info.Count {
```
aleksmaus, line 267: think you might hit this place if you have more than 10 chunks. need to use info.Count as "size" in the query to get the max expected number of hits

michel-laterman (changes requested):
We also need to think about how would this API handle being deployed across multiple fleet-server instances.
The agent/client may only have 1 entry for a fleet-server however multiple servers may be deployed behind a load balancer.

Currently if Server 1 receives the start of upload request, another server is not able to handle a chunk as key details are kept in (server 1's) in-memory current map.
We may need to write this to an index on upload start and use monitors to read/update any changes across multiple instances.

internal/pkg/api/handleUpload.go
cmacknz: +1 to making these configurable, when this is released you will want a way to change them without having to wait for the next stack release.

internal/pkg/upload/es.go
```diff
@@ -0,0 +1,171 @@
...
+		Index:      fmt.Sprintf(FileDataIndexPattern, source),
+		Body:       body,
+		DocumentID: fmt.Sprintf("%s.%d", docID, chunkID),
+		Refresh:    "true",
+	}
+	// need to set the Content-Type of the request to CBOR, notes below
+	overrider := contentTypeOverrider{client}
+	resp, err := req.Do(ctx, overrider)
+	/*
+		standard approach when content-type override no longer needed
```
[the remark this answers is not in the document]
pzl: yes, I fixed in main with: https://github.com/elastic/go-elasticsearch/pull/536

when we update to `v8` eventually, we should pick up the fix and then setting the header the normal way will suffice

internal/pkg/upload/upload.go
cmacknz: I see the agent ID included in the request handlers, but given this is asynchronous it looks like we'd have to manually try to map uploadIDs back to log statements for requests which seems tedious.

internal/pkg/upload/cbor/chunk.go
```diff
@@ -0,0 +1,146 @@
...
+	for _, c := range baseID { // now write the document baseID
+		preamble[i] = byte(c)
+		i++
+	}
+	preamble[i] = 0x64 // string with 4 chars
+	preamble[i+1] = 'd'
+	preamble[i+2] = 'a'
+	preamble[i+3] = 't'
+	preamble[i+4] = 'a'
+	i += 5
```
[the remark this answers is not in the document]
pzl: `i` is not actually being incremented for the above few lines. In C, `preamble[i++] = 'd'` would cover this for each line.

This is "catching-up" `i` with the values just written (0x64 + "data")

cmacknz (commented):
Did a high level review, LGTM besides a few questions and nits.

I'll delegate final approval to @michel-laterman since has been following this much closer than I have. 

The biggest missing piece to me would be making the uploader limits configurable instead of hard coded, but that could be done as a follow up change.

Edit: Asked michel to highlight blocking changes, but he posted them 9 minutes before I made this comment :)

aleksmaus (changes requested):
The implementation needs to change to make the upload code stateless.

internal/pkg/upload/upload.go
```diff
@@ -0,0 +1,341 @@
...
+	"github.com/rs/zerolog/log"
+)
+
+const (
+	//these should be configs probably
+	uploadRequestTimeout = time.Hour
+	chunkProgressTimeout = time.Hour / 4
+
+	// specification-designated maximum
+	MaxChunkSize = 4194304 // 4 MiB
```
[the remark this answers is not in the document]
pzl: In the future we could add a configurable option to this, for anything lower. But 4MiB is the upper bound maximum to maintain compatibility with the Kibana File plugin.

internal/pkg/api/handleUpload.go
```diff
@@ -0,0 +1,390 @@
...
+	upl         *upload.Uploader
+}
+
+func NewUploadT(cfg *config.Server, bulker bulk.Bulk, chunkClient *elasticsearch.Client, cache cache.Cache) *UploadT {
+	log.Info().
+		Interface("limits", cfg.Limits.ArtifactLimit).
+		Int64("maxFileSize", maxFileSize).
+		Int("maxParallelOps", maxParallelUploadOperations).
+		Int("maxParallelChunks", maxParallelChunks).
+		Msg("Artifact install limits")
```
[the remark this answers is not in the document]
pzl: I think this was just copy-pasted from elsewhere and I didn't change the message. Since this is definitely not Artifacts. I guess you've got a capital over there in `handleArtifacts.go`

internal/pkg/api/handleUpload.go
```diff
@@ -0,0 +1,390 @@
+// Copyright Elasticsearch B.V. and/or licensed to Elasticsearch B.V. under one
+// or more contributor license agreements. Licensed under the Elastic License;
+// you may not use this file except in compliance with the Elastic License.
+
+package api
+
+import (
+	"context"
+	"crypto/md5"
+	"crypto/sha256"
```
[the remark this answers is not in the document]
pzl: md5 and sha256 were decided as reasonable primitives that the client implementations (agent and endpoint) would always have access to

This was coordinated with the clients

internal/pkg/api/handleUpload.go
```diff
@@ -0,0 +1,390 @@
...
+		bulker:      bulker,
+		cache:       cache,
+		upl:         upload.New(maxFileSize, maxParallelChunks, maxParallelChunks),
+	}
+}
+
+func (ut *UploadT) handleUploadStart(zlog *zerolog.Logger, w http.ResponseWriter, r *http.Request) error {
+
+	// store raw body since we will json-decode twice
+	// 2MB is a reasonable json payload size. Any more might be an indication of garbage
```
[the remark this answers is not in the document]
pzl: possibly. The error checking for nested fields became a bit cumbersome.

```go
fileInfo, ok := payload['file'].(map[string]interface{})
if !ok {
  return badrequest
}

if mime, ok := fileInfo["mime_type"].(string); !ok {
  return badrequest
} else if strings.TrimSpace(mime) == "" {
  return badrequest
}

// ...
```

internal/pkg/upload/upload.go
```diff
@@ -0,0 +1,341 @@
...
+		return Info{}, ErrFileSizeTooLarge
+	}
+
+	uid, err := uuid.NewV4()
+	if err != nil {
+		return Info{}, fmt.Errorf("unable to generate upload operation ID: %w", err)
+	}
+	id := uid.String()
+
+	token := u.opThrottle.Acquire(id, 300*time.Hour)
```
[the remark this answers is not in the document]
pzl: good point. Random upper bound above any reasonable upload limit that would stop all this first. Time parameter was mandatory but specifically didn't want it in this use case

internal/pkg/upload/upload.go
pzl: less important now that these indices have an ILM policy that cleans up and deletes documents on a somewhat short timescale (XX days)

cmacknz:
Now that it has been pointed out, I agree that this implementation will have problems if there are multiple fleet servers behind a load balancer. There is no guarantee that each request is going to land with the same server. Sorry we didn't catch this earlier.

I suspect that resolving this before the 8.6 feature freeze will be challenging. Likely you could likely proceed by documenting the explicit limitation that this will only work reliably with a single fleet server instance until that problem is resolved.

cmacknz:
I don't want to dictate a particular solution to the multiple Fleet server problem, but one alternative to moving all of the upload state to Elasticsearch would be to do the upload using a WebSocket to create an affinity to a particular server instance while the upload is in progress. That would likely require less changes to this implementation.

joshdover:
> upload using a WebSocket to create an affinity to a particular server instance while the upload is in progress.

I don't think this is possible today on our Cloud service. It's also a bit of a grey area whether or not we really require this for on-prem deployments. It's not really a breaking change since it's a new feature, but at the very least we'd need some way to detect that Agent/Endpoint can't reach Fleet Server via websockets and somehow surface this to the user (Agent logs at minimum, UI ideally).

internal/pkg/api/handleUpload.go
scunningham: configurable or not "maxParallelUploadOperations = 3" is way too low.  there can be up to 125k endpoints attached to one fleet server.  We have to be able to handle much higher rates.

pzl: > We have to be able to handle much higher rates

I think for the first release, we are specifically not looking to handle high throughput. When we have better purpose-built file storage solutions for this to flow into (native ES file storage, or being able to use an object store directly), we will certainly look to crank this up.

The UI is also built around this being a manual single-action tool. This is not designed to "fetch the same file from 50 endpoints". The user is fetching from one machine at a time. 

the client implementations were designed to buffer files for as long as desired, backoff, and retry if hitting limits

internal/pkg/api/handleUpload.go
```diff
@@ -0,0 +1,388 @@
...
+		Str(LogAgentID, id).
+		Str(ECSHTTPRequestID, reqID).
+		Logger()
+
+	// simpler authentication check, since chunk checksum must
+	// ultimately match the initial hash provided with the stricter key check
+	if _, err := authAPIKey(r, rt.bulker, rt.ut.cache); err != nil {
+		cntUpload.IncError(err)
+		resp := NewHTTPErrResp(err)
+		if err := resp.Write(w); err != nil {
```
scunningham, line 107: Typically you write the http response at the top of the stack.  Not buried down.  Better to return an error and have one block of code that writes out the  http response.

internal/pkg/api/handleUpload.go
scunningham: ;) If you build it, they will come.

michel-laterman:
@pzl, I've added an openapi spec for fleet-server - https://github.com/elastic/fleet-server/blob/main/model/openapi.yml
Can you please add the file upload endpoints as part of the PR?
I may work on using a codegen tool to ensure that the api is treated as the source of truth in the future

kevinlog:
I checked out the latest code and tested it against a real Agent/Endpoint with the [Endpoint changes](https://github.com/elastic/endpoint-dev/pull/12270) it all works e2e!

image

We'll need to make an update the mappings in ES, I opened that PR here: https://github.com/elastic/elasticsearch/pull/92591. I verified these changes running locally against this Fleet Server change and the Endpoint changes.

cc @pzl @aleksmaus @michel-laterman @hop-dev @kpollich

aleksmaus (commented):
left few comments, will do another pass later today

internal/pkg/upload/upload.go
```diff
@@ -0,0 +1,286 @@
...
+	Pos  int  // Ordered chunk position in file
+	Last bool // Is this the final chunk in the file
+	SHA2 string
+	Size int
+	BID  string // base id, matches metadata doc's _id
+	//FirstReceived bool
+}
+
+type Uploader struct {
+	metaCache map[string]Info // cache of file metadata doc info
```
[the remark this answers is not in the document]
pzl: yes, noted in the todo comment below it:

`@todo: cache eviction so it's not unbounded growth`

looks like we're already importing/using ristretto elsewhere, happy to go with that

internal/pkg/api/handleUpload.go
```diff
@@ -0,0 +1,269 @@
...
+	if err := upload.IndexChunk(r.Context(), ut.chunkClient, ce, upinfo.Source, chunkInfo.BID, chunkInfo.Pos); err != nil {
+		return err
+	}
+
+	hashsum := hex.EncodeToString(hash.Sum(nil))
+
+	if !strings.EqualFold(chunkHash, hashsum) {
+		// delete document, since we wrote it, but the hash was invalid
+		// context scoped to allow this operation to finish even if client disconnects
+		if err := upload.DeleteChunk(context.Background(), ut.bulker, upinfo.Source, chunkInfo.BID, chunkInfo.Pos); err != nil {
```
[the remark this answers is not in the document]
scunningham: We should assume the hash is ok as it will be 99.9% of the time.  Writing to a temporary index implies a chunk rewrite on success, which we should avoid.

internal/pkg/upload/finalize.go
```diff
@@ -0,0 +1,145 @@
...
+		return info, err
+	}
+
+	/*
+		Verify Upload
+	*/
+
+	// if already done, failed or deleted, exit
+	if !info.StatusCanUpload() {
+		return info, err
```
aleksmaus, line 32: if intention to return no error here then 
```
return info, nil
```
otherwise define error if can't upload?

internal/pkg/upload/finalize.go
```diff
@@ -0,0 +1,145 @@
...
+		return info, ErrMissingChunks
+	}
+	if !u.verifyChunkInfo(info, chunks, transitHash) {
+		if err := SetStatus(ctx, u.bulker, info, StatusFail); err != nil {
+			log.Warn().Err(err).Str("fileID", info.DocID).Str("uploadID", info.ID).Msg("file upload failed chunk validation, but encountered an error setting the upload status to failure")
+		}
+		if err := DeleteChunksByQuery(ctx, u.bulker, info.Source, info.DocID); err != nil {
+			log.Warn().Err(err).Str("fileID", info.DocID).Str("uploadID", info.ID).Msg("file upload failed chunk validation, but encountered an error deleting left-behind chunk data")
+		}
+		return info, errors.New("file contents did not pass validation")
```
aleksmaus, line 49: nit: ```"file contents failed validation"```

aleksmaus: nit: if you want to have a unit test coverage for specific error cases, then it's worth defining errors as vars, so you can compare in the utz

internal/pkg/upload/finalize.go
```diff
@@ -0,0 +1,145 @@
...
+		if c.Pos != i {
+			log.Warn().Int("expected", i).Interface("chunk", c).Msg("chunk position doesn't match. May be a gap in uploaded file")
+			return false
+		}
+	}
+
+	return true
+}
+
+func (u *Uploader) verifyChunkInfo(info Info, chunks []ChunkInfo, transitHash string) bool {
```
aleksmaus, line 85: nit: might return ```error``` instead of ```bool```, allows you to write better unit tests, testing different error cases, verifying the expected error

internal/pkg/upload/upload.go
```diff
@@ -0,0 +1,286 @@
...
+
+	// make sure all required fields are present and non-empty
+	if err := validateUploadPayload(data); err != nil {
+		return Info{}, err
+	}
+
+	size, _ := data.Int64("file", "size")
+	if size > u.sizeLimit {
+		return Info{}, ErrFileSizeTooLarge
+	}
```
aleksmaus, line 147: need to check if size is <= 0 ?

internal/pkg/upload/upload.go
```diff
@@ -0,0 +1,286 @@
...
+	u.mu.RUnlock() // not deferred since this must be clear before we gain a write lock below
+	if exist {
+		return info, nil
+	}
+
+	// not found in cache, try fetching
+	info, err := FetchUploadInfo(ctx, u.bulker, uploadID)
+	if err != nil {
+		return Info{}, fmt.Errorf("unable to retrieve upload info: %w", err)
+	}
```
aleksmaus, line 281: just to double check: this would allows for multiple fetches for the same upload id, with misbehaving client, correct?

aleksmaus (commented):
left some more comments and nits

internal/pkg/api/handleUpload.go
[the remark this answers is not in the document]
pzl: the reasoning for _not_ using the request context is just above:

this is the cleanup step of a (rare) failure. It didn't make sense to abort cleanup if the client disconnects early

internal/pkg/upload/upload.go
pzl: `<= 0` would be an input error ( illogical for negative file size, useless upload of 0 size) so these cases are filtered out during input validation in `upload.go:260`, `func validateUploadPayload`.

Anything past that is valid input (i.e. a positive non-zero file size). And we check against server limits later (here)

internal/pkg/api/handleUpload.go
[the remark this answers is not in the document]
aleksmaus: how long this request would be left running blocking request handler completion in the worst case scenario?

internal/pkg/upload/es.go
```diff
@@ -0,0 +1,171 @@
+// Copyright Elasticsearch B.V. and/or licensed to Elasticsearch B.V. under one
```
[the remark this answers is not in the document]
pzl: almost all the files deal with chunks to this same level,

this one has a lot of methods at the top with chunk in the name, but this file is only preparing `dsl` queries that get sent to elasticsearch, and sending requests to elasticsearch, lightly parsing responses.

No business logic here, just formatting request body and send

internal/pkg/upload/upload.go
```diff
@@ -0,0 +1,288 @@
...
+// convenience functions for computing current "Status" based on the fields
+func (i Info) Expired(timeout time.Duration) bool { return time.Now().After(i.Start.Add(timeout)) }
+func (i Info) StatusCanUpload() bool { // returns true if more chunks can be uploaded. False if the upload process has completed (with or without error)
+	return !(i.Status == StatusFail || i.Status == StatusDone || i.Status == StatusDel)
+}
+
+type FileData struct {
+	Size      int64  `json:"size"`
+	ChunkSize int64  `json:"ChunkSize"`
+	Status    string `json:"Status"`
```
aleksmaus, line 69: this is deviating from the usual json naming convention where the rest of the fields user lower case letter and underscore "_" delimiter

internal/pkg/api/handleUpload.go
```diff
@@ -0,0 +1,261 @@
...
+
+	return &UploadT{
+		chunkClient: chunkClient,
+		bulker:      bulker,
+		cache:       cache,
+		uploader:    upload.New(chunkClient, bulker, maxFileSize, maxUploadTimer),
+	}
+}
+
+func (ut *UploadT) handleUploadStart(zlog *zerolog.Logger, w http.ResponseWriter, r *http.Request) error { //nolint:unparam // log is standard first arg for the handlers
```
aleksmaus, line 126: another way is to do something like this 
```
func (ut *UploadT) handleUploadStart(_ *zerolog.Logger, w http.ResponseWriter, r *http.Request) error
```

internal/pkg/cache/cache.go
```diff
@@ -270,3 +274,37 @@ func (c *CacheT) SetArtifact(artifact model.Artifact) {
 		Dur("ttl", ttl).
 		Msg("Artifact cache SET")
 }
+
+func (c *CacheT) SetUpload(id string, info upload.Info) {
+	c.mut.RLock()
+	defer c.mut.RUnlock()
+
+	scopedKey := "upload:" + id
+	ttl := time.Hour / 2 // @todo: add to configurable
```
aleksmaus, line 283: it's usually
```
ttl := 30 * time.Minute
```

internal/pkg/cache/cache.go
```diff
@@ -270,3 +274,37 @@ func (c *CacheT) SetArtifact(artifact model.Artifact) {
...
+	cost := int64(len(info.ID) + len(info.DocID) + len(info.ActionID) + len(info.AgentID) + len(info.Source) + len(info.Status) + 8*4)
```
aleksmaus, line 284: maybe add a comment how the cost was chosen here and why, for the future devs.

internal/pkg/cache/cache.go
```diff
@@ -270,3 +274,37 @@ func (c *CacheT) SetArtifact(artifact model.Artifact) {
...
+		Str("id", id).
+		Int64("cost", cost).
+		Dur("ttl", ttl).
+		Msg("Upload info cache SET")
+}
+func (c *CacheT) GetUpload(id string) (upload.Info, bool) { //nolint:dupl // a little repetition to support strong typing
+	c.mut.RLock()
+	defer c.mut.RUnlock()
+
+	scopedKey := "upload:" + id
```
aleksmaus, line 297: nit: maybe use a function that makes the key string to avoid doing this in two places, easier to make mistake if anything changes.

internal/pkg/uploader/finalize.go
```diff
@@ -0,0 +1,151 @@
...
+
+func (u *Uploader) allChunksPresent(info upload.Info, chunks []ChunkInfo) bool {
+	// check overall count
+	if len(chunks) != info.Count {
+		log.Warn().Int("expectedCount", info.Count).Int("received", len(chunks)).Interface("chunks", chunks).Msg("mismatch number of chunks")
+		return false
+	}
+
+	// now ensure all positions are accounted for, no gaps, etc
+	sort.Slice(chunks, func(i, j int) bool {
```
aleksmaus, line 77: question/nit: could you avoid sorting if the chunks are always selected "ordered by pos" from elasticsearch?

internal/pkg/uploader/finalize.go
```diff
@@ -0,0 +1,151 @@
...
+
+	for i, chunk := range chunks {
+		if i < info.Count-1 {
+			// all chunks except last must have last:false
+			// and be PRECISELY info.ChunkSize bytes long
+			if chunk.Last {
+				log.Debug().Int("chunkID", i).Msg("non-final chunk was incorrectly marked last")
+				return false
+			}
+			if chunk.Size != int(info.ChunkSize) {
```
aleksmaus, line 107: should we just have the chunk size defined as int64(or uint64) for both structs and avoid casting? the int is int64 anyways, the fleet server is built for 64bit only atm

internal/pkg/uploader/upload.go
```diff
@@ -0,0 +1,251 @@
...
+	return nil
+}
+
+type ChunkInfo struct {
+	Pos  int  // Ordered chunk position in file
+	Last bool // Is this the final chunk in the file
+	SHA2 string
+	Size int
+	BID  string // base id, matches metadata doc's _id
+	//FirstReceived bool
```
aleksmaus, line 75: nit: remove

aleksmaus (approved):
lgtm, assuming the @todos will be addressed before merging.
left few comments/nits, please respond or address.

aleksmaus:
@cmacknz @michel-laterman would like to have somebody from the agent/fleet team to review this as well.

internal/pkg/uploader/finalize.go
pzl: I'm not sure I can guarantee that they are always sorted correctly by ES. The positional information is not in a sortable field, it is in just a portion of the document's `_id`. Considering this is a bounded size slice (limited maximum file size), sorting was not out of the question

internal/pkg/api/handleUpload_test.go
```diff
@@ -0,0 +1,8 @@
+// Copyright Elasticsearch B.V. and/or licensed to Elasticsearch B.V. under one
+// or more contributor license agreements. Licensed under the Elastic License;
+// you may not use this file except in compliance with the Elastic License.
+
+//go:build !integration
+// +build !integration
+
+package api
```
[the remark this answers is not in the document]
juliaElastic: Are unit tests planned for this pr?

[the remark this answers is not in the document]
pzl: Yes, will add tests for the handlers. The handlers do very little aside from calling their respective operation in `uploader` module, so tests were added there first.

internal/pkg/uploader/upload_test.go
```diff
@@ -0,0 +1,273 @@
...
+				assert.ErrorIs(t, err, ErrFileSizeTooLarge)
+			} else {
+				assert.NoError(t, err)
+			}
+		})
+	}
+}
+
+/*
+func TestUploadRejectsMissingRequiredFields(t *testing.T) {
```
juliaElastic, line 223: are you planning to fix these tests or remove the code?

juliaElastic (approved):
LGTM, other than CHANGELOG missing, and few TODOs in the code.

internal/pkg/uploader/upload_test.go
pzl: these are staying in, just need updating

internal/pkg/uploader/upload.go
```diff
@@ -0,0 +1,247 @@
...
+		return errors.New("file.size is required")
+	} else if size <= 0 {
+		return fmt.Errorf("invalid file.size: %d", size)
+	}
+	return nil
+}
+
+// Searches for Upload Metadata document in local memory cache if available
+// otherwise, fetches from elasticsearch and caches for next use
+func (u *Uploader) GetUploadInfo(ctx context.Context, uploadID string) (upload.Info, error) {
```
AndersonQ, line 233: [Suggestion | Go convention]
```suggestion
// GetUploadInfo searches for Upload Metadata document in local memory cache if available
// otherwise, fetches from elasticsearch and caches for next use
func (u *Uploader) GetUploadInfo(ctx context.Context, uploadID string) (upload.Info, error) {
```

internal/pkg/uploader/upload/info.go
```diff
@@ -0,0 +1,50 @@
...
+package upload
+
+import (
+	"time"
+
+	"github.com/elastic/fleet-server/v7/internal/pkg/model"
+)
+
+// the only valid values of upload status according to storage spec
+type Status string
```
AndersonQ, line 14: [Suggestion | Go convention]
```suggestion
// Status represents the only valid values of upload status according to storage spec
type Status string
```

AndersonQ (approved):

kevinlog:
Checked out the latest and tested again e2e, LGTM!

I'm able to get files from the Endpoint host:

image

This also includes testing with Fleet installation of the relevant indices.  This includes [running open my PR](https://github.com/elastic/elasticsearch/pull/92591) locally to update the mappings needed by the new fleet server implementation.  I'll merge this PR soon.

image

kevinlog (approved):

PR #1904, opened 2022-09-22
Unique limiters for each API listener

### LINKED ISSUE #1859: Load limit on fleet server also limits its own Elastic Agent
Apparently, the connection limit is applied to Fleet Server own Elastic Agent.
This mean that under an heavy load, the Fleet Server Elastic Agent can be prevented from checkin in and as a result FS is shown as unhealthy.

This issue will act as a placeholder for any further discoveries regarding Fleet Server config.

michel-laterman:

Due to the internal and external listenser using the same limiter, the fleet-server may send a `429` response to the elastic-agent that it is running under when under load. This can cause the agent and fleet-server instance to be flagged as unhealthy and shut down.

Refactor the limit.Limiter so it can wrap the separate API httprouter endpoints. Limiter.WrapX() calls take the handler and stats incrementer for metrics/error counting. api.Run() replaced with Router.Run(), which will generate an httprouter for each listener in order to be able to associate the httprouter with a unique Limiter.

## How to test

Start an elastic-agent that uses the `fleet-server` integration and view the fleet-server logs:

```
{"log.level":"info","ecs.version":"1.6.0","service.name":"fleet-server","addr":"0.0.0.0:8220","limits":{"MaxAgents":0,"PolicyThrottle":500000,"MaxHeaderByteSize":8192,"MaxConnections":52428800,"CheckinLimit":{"Interval":1000000,"Burst":1000,"Max":0,"MaxBody":1048576},"ArtifactLimit":{"Interval":5000000,"Burst":25,"Max":50,"MaxBody":0},"EnrollLimit":{"Interval":10000000,"Burst":100,"Max":50,"MaxBody":524288},"AckLimit":{"Interval":10000000,"Burst":100,"Max":50,"MaxBody":2097152},"StatusLimit":{"Interval":5000000,"Burst":25,"Max":50,"MaxBody":0}},"@timestamp":"2022-09-28T21:14:25.355Z","message":"fleet-server creating new limiter"}
...
{"log.level":"info","ecs.version":"1.6.0","service.name":"fleet-server","addr":"localhost:8221","limits":{"MaxAgents":0,"PolicyThrottle":500000,"MaxHeaderByteSize":8192,"MaxConnections":52428800,"CheckinLimit":{"Interval":1000000,"Burst":1000,"Max":0,"MaxBody":1048576},"ArtifactLimit":{"Interval":5000000,"Burst":25,"Max":50,"MaxBody":0},"EnrollLimit":{"Interval":10000000,"Burst":100,"Max":50,"MaxBody":524288},"AckLimit":{"Interval":10000000,"Burst":100,"Max":50,"MaxBody":2097152},"StatusLimit":{"Interval":5000000,"Burst":25,"Max":50,"MaxBody":0}},"@timestamp":"2022-09-28T21:14:25.385Z","message":"fleet-server creating new limiter"}
...
{"log.level":"info","ecs.version":"1.6.0","service.name":"fleet-server","addr":"localhost:8221","@timestamp":"2022-09-28T21:14:25.385Z","message":"fleet-server routes set up"}
```

## Checklist

## Related issues

- Closes #1859


internal/pkg/limit/limiter.go
```diff
@@ -2,38 +2,88 @@
...
+}
+
+type releaseFunc func()
+
+type limiter struct {
+	rateLimit *rate.Limiter
+	maxLimit  *semaphore.Weighted
+}
 
+func newLimiter(cfg *config.Limit) *limiter {
```
scunningham, line 81: I don't love that the higher level app specific limiter is mixed in with the lower level implementation.  Makes it harder to understand.  Perhaps two separate files.

internal/pkg/limit/limiter.go
```diff
@@ -63,11 +113,30 @@ func (l *Limiter) Acquire() (ReleaseFunc, error) {
...
-func (l *Limiter) release() {
+func (l *limiter) release() {
 	if l.maxLimit != nil {
 		l.maxLimit.Release(1)
 	}
 }
 
+func (l *limiter) wrap(logger zerolog.Logger, level zerolog.Level, h httprouter.Handle, i StatIncer) httprouter.Handle {
+	return func(w http.ResponseWriter, r *http.Request, p httprouter.Params) {
+		dfunc := i.IncStart()
```
scunningham, line 124: For better or worse, this logic is different than the original implementation.  In the original implementation, the counter was not incremented until:

1) The limiter semaphore was acquired
2) The transaction was authorized

internal/pkg/api/router.go
```diff
@@ -27,30 +33,41 @@ const (
 )
 
 type Router struct {
-	ctx    context.Context
+	ctx    context.Context // used only by handleEnroll
```
scunningham, line 36: Maybe a comment as to where/why this is intialized.

internal/pkg/api/router.go
```diff
@@ -104,8 +121,120 @@ func NewRouter(ctx context.Context, bulker bulk.Bulk, ct *CheckinT, et *Enroller
...
+	listeners := rt.cfg.BindEndpoints()
+	rdto := rt.cfg.Timeouts.Read
+	wrto := rt.cfg.Timeouts.Write
+	idle := rt.cfg.Timeouts.Idle
+	rdhr := rt.cfg.Timeouts.ReadHeader
+	mhbz := rt.cfg.Limits.MaxHeaderByteSize
+	bctx := func(net.Listener) context.Context { return ctx }
+
+	errChan := make(chan error)
+	cancelCtx, cancel := context.WithCancel(ctx) // TODO should we set rt.ctx = cancelCtx?
```
scunningham, line 142: No.  The cancel context wraps the lifetime of the server.  We are using the global context here in handleEnroll to do a rollback.  Perhaps clearer to rename ctx in the router to "baseCtx" or "appCtx".

scunningham (approved):
Looks ok.

internal/pkg/limit/limiter.go
michel-laterman: Yeah, i made a comment about that, we can definitely restore that behaviour if we need to.
But with it tracking every attempt it would mean that the error count associated with a route can't be higher than the total calls

scunningham: Just metrics; doesn't affect logic so doesn't matter much.

AndersonQ (changes requested):

AndersonQ (approved):

PR #1912, opened 2022-09-26
LoadLimits does not override existing values

### LINKED ISSUE #1841: custom server.limit values are not used
Loading agent limits (introduced in https://github.com/elastic/fleet-server/pull/1014) ignores any custom `server.limits` settings that are set in the fleet-ui and propagated to the fleet-server instance.

The fleet-server instance will receive the changed config and emit the "Re-configuring" message, however the `configServerChagned` comparison returns false

This is due to the [LoadServerLimits()](https://github.com/elastic/fleet-server/blob/23be42ad5da92250a4f62e762b4971c357e7ae15/internal/pkg/config/config.go#L69) function overriding any `server.limit.*` config value is it uses the config blob returned by [loadLimits()](https://github.com/elastic/fleet-server/blob/23be42ad5da92250a4f62e762b4971c357e7ae15/internal/pkg/config/env_defaults.go#L174).

michel-laterman:

Custom user specified values for cache and server limits are silently discarded when `LoadLimits` is ran.

Fleet-server will use any specified cache or server limit values over whatever is returned by the default/agent number loader. For example, if A max body size is specifically set to a value such as 5MB, and the default returned by the LoadLimits is 1MB, the 5MB value is used.

## Checklist

## Related issues

- Closes #1841

### DIFF
--- a/CHANGELOG.next.asciidoc
+++ b/CHANGELOG.next.asciidoc
@@ -12,6 +12,7 @@
 - Add error detail to catch-all HTTP error response. {pull}1854[1854]
 - Fix issue were errors where being ignored written to elasticsearch. {pull}1896[1896]
 - Update apikey.cache_hit log field name to match convention. {pull}1900[1900]
+- LoadServerLimits will not overwrite specified limits when loading default/agent number specified values. {issue}1841[1841] {pull}1912[1912]
 
 ==== New Features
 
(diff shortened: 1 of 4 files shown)


blakerouse (approved):
Looks good.

Only downside too doing it this way is it does mean that they can never be overridden to 0. I don't think we want that anyway, but just something to note.

joshdover:
Thanks for the quick fix, @michel-laterman 🚀

PR #1921, opened 2022-09-27
Add extra protection against accessing null fields to 8.5 migration

AndersonQ:

## Checklist


blakerouse (approved):
+1

Weird that its needed, but hopefully it fixes the issue.

cmacknz:
@AndersonQ we need to forward port this to main, correct?

jlind23:
@cmacknz We do!

AndersonQ:
Yes, we do

AndersonQ:
forwardport PR created https://github.com/elastic/fleet-server/pull/1926

PR #1926, opened 2022-09-28
[8.6](forwardport) Add extra protection against accessing null fields to 8.5 migration (#1921)

AndersonQ:
This is a manual forwardport of pull request https://github.com/elastic/fleet-server/pull/1921 and a wee improvement on the logs.

PR #1935, opened 2022-09-29
Improve migration logs on 8.5 branch

AndersonQ:

The forward port of https://github.com/elastic/fleet-server/pull/1921, the https://github.com/elastic/fleet-server/pull/1926 PR, includes a few improvements on the log messages, this PR backports those improvements.

By backporting the extra changes on the forwardport https://github.com/elastic/fleet-server/pull/1926

## Checklist

### DIFF
--- a/internal/pkg/dl/migration.go
+++ b/internal/pkg/dl/migration.go
@@ -61,11 +61,16 @@ func migrate(ctx context.Context, bulker bulk.Bulk, fn migrationBodyFn) (int, er
 	for {
 		name, index, body, err := fn()
 		if err != nil {
-			return updatedDocs, fmt.Errorf(": %w", err)
+			return updatedDocs,
+				fmt.Errorf("failed to prepare request for migration %s: %w",
+					name, err)
 		}
 
 		resp, err := applyMigration(ctx, name, index, bulker, body)
 		if err != nil {
+			log.Err(err).
+				Bytes("http.request.body.content", body).
+				Msgf("migration %s failed: %w", name, err)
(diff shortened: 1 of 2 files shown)


internal/pkg/dl/migration.go
```diff
@@ -61,11 +61,16 @@ func migrate(ctx context.Context, bulker bulk.Bulk, fn migrationBodyFn) (int, er
...
-			return updatedDocs, fmt.Errorf(": %w", err)
+			return updatedDocs,
+				fmt.Errorf("failed to prepare request for migration %s: %w",
+					name, err)
 		}
 
 		resp, err := applyMigration(ctx, name, index, bulker, body)
 		if err != nil {
+			log.Err(err).
+				Bytes("http.request.body.content", body).
```
cmacknz, line 72: Are you concerned about how large body could be here? Or that it might contain information we want to sanitize out of the logs? I am not sure what exactly will be logged here so perhaps this isn't a concern.

Besides this small concern it looks good.

cmacknz (approved):

internal/pkg/dl/migration.go
AndersonQ: I'm a bit concerned about the size, but I think it'd be fine. The largest it might log is that [script/request body](https://github.com/elastic/fleet-server/blob/16ee12afe3ba8740b3bfa568d079a493e7664b9a/internal/pkg/dl/migration.go#L230-L264)

AndersonQ: worst case, AFAIK, ES wont take the document in or truncate it, so not a big issue

PR #1946, opened 2022-10-04
Conditional log level for api key read

michalpristas:

The issue is actually a race issue and this is not a proper fix.
Proper fix would require introducing separate processing pipeline shared across kibana/fleet-server which would be capable of handling out of order requests.

What's happening is that we receive `API Key not found` response from ES during ACK in case when policy is updated in a way new apikey needs to be generated and unenrolling agent quickly after that.

this is resulting in out of order/race behavior on ACK handler where:
- agent is authenticated and agent instance fetched from ES, this agent is active and API keys are fine
- kibana side invalidates API key on force unenroll
- ACK handler tries to fetch API key for the agent but the API Key is already invalidated and 404 is returned

As this is valid behavior and we don't have a way to prevent that in case 404 API Key response is returned we fetch new instance of agent and if agent is not marked `Active` we decrease log level to `Info`

Described above, change policy, force unenroll, it will get hit

## Checklist

## Related issues

No related issue


internal/pkg/api/handleAck.go
```diff
@@ -364,10 +365,18 @@ func (ack *AckT) updateAPIKey(ctx context.Context,
 	if apiKeyID != "" {
 		res, err := ack.bulk.APIKeyRead(ctx, apiKeyID, true)
 		if err != nil {
-			zlog.Error().
-				Err(err).
-				Str(LogAPIKeyID, apiKeyID).
-				Msg("Failed to read API Key roles")
+			if ack.isAPIKeyReadError(ctx, zlog, agentID, err) {
```
aleksmaus, line 368: a possible concern here is that, now for each agent, for each error, for each output in the policy
https://github.com/elastic/fleet-server/blob/c99ccd81baa26b75e55a3b71b7d65be4111d555c/internal/pkg/api/handleAck.go#L337
this could do an additional request to elasticsearch (waiting on bulk flushes)?
is this a correct understanding?

michalpristas: yes that's a good point.

internal/pkg/api/handleAck.go
```diff
@@ -356,18 +360,29 @@ func (ack *AckT) handlePolicyChange(ctx context.Context, zlog zerolog.Logger, ag
...
+					Msg("Failed to read API Key roles")
+			} else {
+				// not an error, race when API key was invalidated before acking
+				zlog.Info().
+					Err(err).
+					Str(LogAPIKeyID, apiKeyID).
+					Msg("Failed to read invalidated API Key roles")
+
+				// update to inactive for future checks
+				agent.Active = false
```
scunningham, line 384: This is still an error.  If the ApiKey is gone, the ACK should fail.  We should still output an error, but it should be an info level log message.  Ie. this is an error, but it is expected.

It is also unexpected to have this call produce a side effect of changing the agent record.

internal/pkg/api/handleAck.go
```diff
@@ -513,6 +528,18 @@ func (ack *AckT) handleUpgrade(ctx context.Context, zlog zerolog.Logger, agent *
 	return nil
 }
 
+func isAgentActive(ctx context.Context, zlog zerolog.Logger, bulk bulk.Bulk, agentID string) bool {
+	agent, err := dl.FindAgent(ctx, bulk, dl.QueryAgentByID, dl.FieldID, agentID)
```
scunningham, line 532: This does not  do a wait for refresh.  I'm ok with skipping that, but should comment in here that we have a potential race.

scunningham (changes requested):

internal/pkg/api/handleAck.go
michalpristas: discussed with @scunningham offline, approach agreed on is: returning error and propagating it to the client while decreasing log level on the server side inside handler

scunningham (approved):

internal/pkg/api/handleAck.go
```diff
@@ -513,6 +528,18 @@ func (ack *AckT) handleUpgrade(ctx context.Context, zlog zerolog.Logger, agent *
...
+	if err != nil {
+		zlog.Warn().
```
AndersonQ, line 534: Why just warn? isn't it a real error?

AndersonQ (approved):

PR #1963, opened 2022-10-06
Don't send POLICY_CHANGE actions retrieved from index to agent.

### LINKED ISSUE #1773: filter out POLICY_CHANGE actions from .fleet-actions
The fleet-server will blindly serve all actions from the `.fleet-actions` index.
We should filter out certain actions - specifically the `POLICY_CHANGE` actions as they are generated by fleet-server when an agent checks in.
Allowing them to be sourced from the `.fleet-actions` index makes it possible to have an agent that is not running the correct policy.

michel-laterman:

The fleet-server should not send any policy change actions that are written to the actions index to an agent on checkin.
The current behaviour is that policy change actions are dynamically generated by the fleet-server when it detects that the agent is not running an up to date version of the policy.

The server will remove these actions in the convert method and emit a warning message.
The ack token that is used is not altered in this case. 

## Checklist

## Related issues

- Closes #1773


juliaElastic (approved):
Code LGTM.
One question on the behavior: how does the agent get notified about he policy change if Fleet Server doesn't deliver the POLICY_CHANGE action?

michalpristas:
the question Julia asked seems valid. agent dispatches policy change on checkin, if it won't receive it how will it keep ability to respond to policy changes

michalpristas:
testing this today

michalpristas:
i see it's coming from a different channel

```go
case policy := <-sub.Output():
		actionResp, err := processPolicy(ctx, zlog, ct.bulker, agent.Id, policy)
		if err != nil {
			return errors.Wrap(err, "processPolicy")
		}
		actions = append(actions, *actionResp)
		break LOOP
```

tested ok

internal/pkg/api/handleCheckin.go
```diff
@@ -376,6 +376,10 @@ func convertActions(agentID string, actions []model.Action) ([]ActionResp, strin
 
 	respList := make([]ActionResp, 0, sz)
 	for _, action := range actions {
+		if action.Type == "POLICY_CHANGE" {
```
michalpristas, line 379: please use `const`

internal/pkg/api/handleCheckin.go
```diff
@@ -376,6 +376,10 @@ func convertActions(agentID string, actions []model.Action) ([]ActionResp, strin
...
+			log.Warn().Str("agent_id", agentID).Str("action_id", action.ActionID).Msg("Removing POLICY_CHANGE action found in index from check in response")
```
michalpristas, line 380: is this a warning or should we stick with INFO. is it something unexpected we should care about.

michalpristas (approved):

internal/pkg/api/handleCheckin.go
jsoriano, line 379: Nit. Consider adding a comment mentioning why these actions are removed here.

jsoriano, line 379: Nit. This is adding an additional responsibility to `convertActions`, it doesn't only convert anymore, but it also filters out certain actions. This may lead to issues in the future if we reuse this method to purely convert actions.

Consider adding something like a `filterActions([]model.Action) []model.Action` method to filter out the actions that are not relevant here, and explicitly filter out when needed.

internal/pkg/api/handleCheckin.go
jsoriano, line 380: Question: When are these policy change actions an issue on check in?

Looking at the code, it seems that `TypePolicyChange` actions are only created on checkin code, in the `processPolicy` method, so it doesn't seem to be always an issue on checkin.

Here a filter is being added to `convertAction`, that is used when:
* There are pending actions.
* There are not pending actions, but something is received from the actions dispatcher.

Are these policy change actions an issue on both cases?
Are they generated in some other place too?

If this checkin handler is the only source of these actions, there would be some way to avoid generating them instead of having to filter them later?

jsoriano (commented):
Some nitpicking and a question.

internal/pkg/api/handleCheckin.go
michel-laterman: We only want to filter out policy_change actions if the actions are read from the index, not if fleet-server generates them by detecting a different policy

jsoriano: Thanks for the explanations!

> the actions index 

> It's possible for a user to insert a policy change action this way.

How can actions be added to the actions index? Or how can users insert policy changes?

Issue #1289, opened 2022-04-01
Flaky test   enrollment_api_key_integration_test.go:113


ph:
## Flaky Test

* **Notes:**
  Looking at the `diff` and the code it's one of the following:
 - The return of the values from Elastic is not deterministic, maybe not sorting
 - Do the code require the values to be always return in the right order?

Found in https://github.com/elastic/fleet-server/pull/1287

Where unrelated code was touched, rekicking the ci and the test was green again.

### Stack Trace

```
[2022-04-01T13:17:50.735Z]     enrollment_api_key_integration_test.go:113:   []model.EnrollmentApiKey{
[2022-04-01T13:17:50.735Z]           	{
[2022-04-01T13:17:50.735Z]           		ESDocument: model.ESDocument{
[2022-04-01T13:17:50.735Z]         - 			Id:      "c93fmvc2jgjfauojood0",
[2022-04-01T13:17:50.735Z]         + 			Id:      "c93fmvc2jgjfauojooe0",
[2022-04-01T13:17:50.735Z]           			Version: 0,
[2022-04-01T13:17:50.735Z]           			SeqNo:   0,
[2022-04-01T13:17:50.735Z]           		},
[2022-04-01T13:17:50.735Z]           		Active:    true,
[2022-04-01T13:17:50.735Z]           		ApiKey:    "d2JndlFIWUJJUVVxWDVia2NJTV86X0d6ZmljZGNTc1d4R1otbklrZFFRZw==",
[2022-04-01T13:17:50.735Z]         - 		ApiKeyId:  "c93fmvc2jgjfauojoodg",
[2022-04-01T13:17:50.735Z]         + 		ApiKeyId:  "c93fmvc2jgjfauojooeg",
[2022-04-01T13:17:50.735Z]           		CreatedAt: "2022-04-01T13:17:49Z",
[2022-04-01T13:17:50.735Z]           		ExpireAt:  "",
[2022-04-01T13:17:50.735Z]           		... // 3 identical fields
[2022-04-01T13:17:50.735Z]           	},
[2022-04-01T13:17:50.735Z]           	{
[2022-04-01T13:17:50.735Z]           		ESDocument: model.ESDocument{
[2022-04-01T13:17:50.735Z]         - 			Id:      "c93fmvc2jgjfauojooe0",
[2022-04-01T13:17:50.735Z]         + 			Id:      "c93fmvc2jgjfauojood0",
[2022-04-01T13:17:50.735Z]           			Version: 0,
[...]

PR #2044, opened 2022-10-31
Add active filter for enrollment key queries.

### LINKED ISSUE #2029: [Fleet] Unable to enroll a Fleet server in a policy with more than 10 inactive enrollment tokens
It's not possible to install a Fleet server in a policy with more than 10 inactive enrollment tokens.

let say you have a `fleet-server-policy` with 10 inactive enrollment tokens and 1 active installing Fleet server results in Fleet stuck in the following state

```
{"log.level":"info","@timestamp":"2022-10-26T14:14:32.693Z","log.origin":{"file.name":"cmd/enroll_cmd.go","file.line":792},"message":"Fleet Server - Waiting on active enrollment keys to be created in policy with Fleet Server integration: fleet-server-policy","ecs.version":"1.6.0"}
```

I think the issue is because we fetch all enrollment token (active or not) in a policy without size parameter, by default ES search return only the 10 first document matching the search, so the active token could not be returned.

## Potential solutions

We should probably pass a size paramter when fetching enrollment tokens, we could also filter on active tokens in the query instead of doing the filtering on Fleet sever here https://github.com/elastic/fleet-server/blob/df33e2f32a10de630aec6aa0a9bf056d67a41fd3/internal/pkg/policy/self.go#L231

michel-laterman:

Fleet-server cannot be added to a policy that has 10+ inactive keys.

Add an `active: true` filter to enrolment key queries.
This allows fleet-server to handle cases where there may be 10+ inactive keys associated with a policy.

## Checklist

## Related issues

- Closes #2029


internal/pkg/dl/enrollment_api_key_integration_test.go
```diff
@@ -114,3 +114,32 @@ func TestSearchEnrollmentAPIKeyByPolicyID(t *testing.T) {
 		t.Fatal(diff)
 	}
 }
+
+func TestSearchEnrollmentAPIKeyByPolicyIDWithInactiveIDs(t *testing.T) {
```
michel-laterman, line 118: The previous test is deemed flakey (#1289), however it may be because that validates multiple returns; this test only does one so it should be OK.

michalpristas (approved):
what a funny bug
in `internal/pkg/policy/self.go` Line 231 there's `tokens = filterActiveTokens(tokens)` this is not needed anymore we can do len check right away

AndersonQ (approved):

PR #2062, opened 2022-11-07
[v2] Merge main as of Nov 7

blakerouse:

Merges `main` branch into `feature-arch-v2` as of Nov. 7th

Prepare for merging `feature-arch-v2` into `main`.


blakerouse:
Seems there is a data race in the integration tests unrelated to the merge. Looking into it.

cmacknz (approved):
Approving. I did not look closely at all the agents, but I doubt anyone really could given how many there are.

PR #2079, opened 2022-11-14
Fix error when raw error is retrieved

michalpristas:

Enrollment or any other operation fails when error returned by elastic is a raw string instead of json

This PR adds ability to parse error either as a ErrorT type or string as a secondary option.

For string it tries to guess error type based on some well known occurrences of Exception names or error messages in retrieved error message. 

Disable `detailed_errors` in ES and try everything works.

## Checklist

## Related issues

Fixes: https://github.com/elastic/elastic-agent/issues/931


internal/pkg/es/error.go
```diff
@@ -66,12 +74,77 @@ var (
...
+
+	if len(rawError) == 0 {
+		// error was omitted
+		return &ErrElastic{
+			Status: status,
+		}
+	}
+
+	// try decoding detailed error by default
+	detailedError := &ErrorT{}
```
aleksmaus, line 112: nit: maybe?
```
var detailedError ErrorT
```

internal/pkg/es/error.go
```diff
@@ -66,12 +74,77 @@ var (
 	ErrIndexNotFound          = errors.New("index not found")
 	ErrTimeout                = errors.New("timeout")
 	ErrNotFound               = errors.New("not found")
+
+	knownErrorTypes = [3]string{
```
aleksmaus, line 78: nit: could omit the size (easier to add more elements later without updating the size)
```
knownErrorTypes = [...]string{
```

internal/pkg/es/error_test.go
```diff
@@ -0,0 +1,216 @@
...
+package es
+
+import (
+	"encoding/json"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+)
+
+func TestErrorTranslation(t *testing.T) {
```
aleksmaus, line 17: 👍

aleksmaus (approved):
lgtm

Issue #2063, opened 2022-11-08
[Fleet] Remove automatic unenrollment of agents


kpollich:
Relates to https://github.com/elastic/kibana/issues/143455

Currently, after agents haven't checked in in a time period greater than the configure "unenrollment timeout," they are automatically "unenrolled" and their API keys are decommissioned. With the work being done in https://github.com/elastic/kibana/issues/143455, this process will no longer be necessary as the filtering of "inactive" agents will become a UI-only concern. We're removing the destructive nature of unenrollment to avoid a broad set of issues that have come up in support cases and elsewhere.

jlind23:
After discussing with @michel-laterman this one is a pretty light one and @hop-dev should be able to take this one too.

Issue #123, opened 2021-03-04
Unenroll ACK needs to invalidate the API keys for the Agent


blakerouse:
At the moment the ACK of an unenroll action does not invalidate the API keys for the Agent. It needs to invalidate both the API key used to communicate with Fleet Server as well as its output API key used for communication to Elasticsearch.

PR #2096, opened 2022-11-18
Remove automatic agent unenrollment

hop-dev:

https://github.com/elastic/fleet-server/issues/2063 Remove the unenrollment of agents after timeout. The `unenroll_timeout` field on agent policies will now be ignored, this has been replaced by a configurable inactivity timeout which does not unenroll the agent and is not managed by fleet server. 

I have added this as a breaking change as it removes functionality is that correct? 

Removes any code and tests related to agent unenrollment

Enroll an agent using fleet server, note that the agent does not get unenrolled after `unenrollment_timeout`

## Design Checklist

- [ ] (❌ I dont believe this is needed) I have or intend to scale test my changes, ensuring it will work reliably with 100K+ agents connected.
- [ ] (❌ I dont believe this is needed) I have included fail safe mechanisms to limit the load on fleet-server: rate limiting, circuit breakers, caching, load shedding, etc. 

## Checklist

- [ ] (❌ I dont believe this is needed)I have made corresponding change to the default configuration files
- [ ] (❌ I dont believe this is needed)I have added tests that prove my fix is effective or that my feature works

## Related issues
- Closes [#123](https://github.com/elastic/fleet-server/issues/2063)


internal/pkg/dl/agent.go
```diff
@@ -67,28 +66,3 @@ func FindAgent(ctx context.Context, bulker bulk.Bulk, tmpl *dsl.Tmpl, name strin
 
 	return agent, nil
 }
-
-func FindOfflineAgents(ctx context.Context, bulker bulk.Bulk, policyID string, unenrollTimeout time.Duration, opt ...Option) ([]model.Agent, error) {
-	o := newOption(FleetAgents, opt...)
-	past := time.Now().UTC().Add(-unenrollTimeout).Format(time.RFC3339)
```
michel-laterman, line 73: Does anything else use the `QueryOfflineAgentsByPolicyID` var/query?

internal/pkg/dl/agent.go
```diff
@@ -67,28 +66,3 @@ func FindAgent(ctx context.Context, bulker bulk.Bulk, tmpl *dsl.Tmpl, name strin
...
-	res, err := Search(ctx, bulker, QueryOfflineAgentsByPolicyID, o.indexName, map[string]interface{}{
```
michel-laterman, line 74: Does anything else use the `prepareOfflineAgentsByPolicyID` var/query?

hop-dev: Good spot, I've removed it 👍

michel-laterman (approved):

PR #2101, opened 2022-11-21
[7.17] Update to go v1.18.8 Update deps to latest patch

michel-laterman:

Update go version to v1.18.8 and all deps excluding `elastic-agent-client`


cmacknz:
The snyk error can likely be fixed by upgrading the `github.com/spf13/cobra` package. If not there is a transitive dependency through cobra that we will have to manually pin the version of.

cmacknz:
@michel-laterman feature freeze for 7.17.8 is tomorrow (Nov 29th). We'll want to merge this to ensure fleet-server has the latest security updates. It doesn't look like there is anything major outstanding here.

cmacknz (approved):

cmacknz (approved):

PR #2118, opened 2022-11-24
Respect cli overrrides

michalpristas:

Cli overrides provided with `-E` flag are ignored as soon as new config arrives.
This prevents querying `stats` when agent configures exposing them through HTTP endpoint.

This PR merges new config into cli overrides so we respect them

## Checklist

## Related issues

https://github.com/elastic/elastic-agent/pull/1787


internal/pkg/server/agent.go
```diff
@@ -385,5 +381,11 @@ func (a *Agent) configFromUnits() (*config.Config, error) {
 	if err != nil {
 		return nil, err
 	}
-	return config.FromConfig(cfgData)
+
+	cliCfg := ucfg.MustNewFrom(a.cliCfg, config.DefaultOptions...)
+	err = cliCfg.Merge(cfgData, config.DefaultOptions...)
```
cmacknz, line 386: I think we can probably unit test this to avoid future regressions here.

Was this problem introduced in v2 or was it always here?

michalpristas: i think it's introduced but i cannot tell for sure.

michalpristas:
failures not related

cmacknz (approved):

internal/pkg/server/agent_test.go
```diff
@@ -0,0 +1,118 @@
...
+	require.Equal(t, httpEnabledExpected, generatedCfg.HTTP.Enabled)
+	require.Equal(t, httpHostExpected, generatedCfg.HTTP.Host)
+	require.Equal(t, loggingFilesNameExpected, generatedCfg.Logging.Files.Name)
+}
+
+type testClientV2 struct {
+	errorChan       chan error
+	unitChangedChan chan client.UnitChanged
+	agentInfo       *client.AgentInfo
+}
```
michel-laterman, line 78: Personally I like to see test interfaces implemented with `testify/mock`.

michel-laterman (commented):
lgtm.

What's the expected behaviour if the policy and command line specify the same variable? Can we add a test to make sure that behaviour works as expected?

michel-laterman (approved):

PR #2120, opened 2022-11-24
Fix check-go linter version

nchaulet:
## Description

I was not able to run the linter locally running `make check-go` failed with the following message 
Screen Shot 2022-11-24 at 12 33 20 PM

By updating the version in the Makefile it seems to works for me.


Makefile
```diff
@@ -89,7 +89,7 @@ check-headers:  ## - Check copyright headers
 
 .PHONY: check-go
 check-go: ## - Run golangci-lint
-	@curl -sSfL https://raw.githubusercontent.com/golangci/golangci-lint/d58dbde584c801091e74a00940e11ff18c6c68bd/install.sh | sh -s v1.44.2
+	@curl -sSfL https://raw.githubusercontent.com/golangci/golangci-lint/d58dbde584c801091e74a00940e11ff18c6c68bd/install.sh | sh -s v1.45.2
```
cmacknz, line 92: https://github.com/elastic/beats/blob/a1a6bd8cd569182aeb0082e252ecbabad338a797/.github/workflows/golangci-lint.yml#L47

1.47.2 is the version beats and agent use, you can upgrade to an even higher version than this.

nchaulet: Just updated here and in the github action doing the linting too.

cmacknz (approved):

nchaulet:
@elasticmachine merge upstream

michel-laterman (approved):
lgtm

PR #2126, opened 2022-11-28
Update module dependencies

michel-laterman:

The elastic-agent client has a bad commit that contains a dependency to the elastic-agent: https://github.com/elastic/elastic-agent-client/pull/48

Update the dependencies with `go get -u=patch`


juliaElastic (approved):

juliaElastic:
@michel-laterman can we backport this to 8.6? I have the same build error there: https://github.com/elastic/fleet-server/pull/2134

PR #2154, opened 2022-12-02
Use the standard embed library to handle embedding defaults

michel-laterman:
Remove the buildlimits dev-tool and use the standard library's embed to read all `default/*.yml` files at compile time instead of using `elastic-agent/pkg/packer`.
This should get rid of the elastic-agent dependency.


internal/pkg/config/env_defaults.go
```diff
@@ -107,8 +107,8 @@ type serverLimitDefaults struct {
 
 func defaultserverLimitDefaults() *serverLimitDefaults {
 	return &serverLimitDefaults{
-		PolicyThrottle: defaultCacheNumCounters,
-		MaxConnections: defaultCacheMaxCost,
+		PolicyThrottle: defaultPolicyThrottle,
+		MaxConnections: defaultMaxConnections,
```
michel-laterman, line 111: Looks like we had a bug in the buildlimits template

AndersonQ (approved):

PR #2173, opened 2022-12-08
Update all modules

michel-laterman:
Update all modules listed in go.mod to the latest versions.

Major version changes include:
* updating `github.com/hashicorp/golang-lru` to v2.0.1 which uses generics and requires updating fleet-server's minimum required go version to 1.18 as a result.
* updating `github.com/elastic/go-elasticsearch` to v8.5.0 which includes functionality for the global_checkpoints and fleet_search calls, so we can subsequently remove our custom implementations.


internal/pkg/action/token_resolver.go
```diff
@@ -22,12 +21,12 @@ const cacheSize = 5000
 // by fleet-server to send state information to the agent.
 type TokenResolver struct {
 	bulker bulk.Bulk
-	cache  *lru.Cache
+	cache  *lru.Cache[string, int64]
```
aleksmaus, line 24: nice! 👍

aleksmaus (approved):

PR #2180, opened 2022-12-12
Add OpenAPI spec for fleet-server endpoints.

michel-laterman:
Add an OpenAPI spec for the fleet-server endpoints.

Currently the spec is just a description of how the fleet-server and elastic-agent communicate.
Code generation may be used in the future. 
If using the [openapi-generator](https://openapi-generator.tech/) the `--type-mappings=string+application\/json=json.RawMessage` flag should be used.


michel-laterman:
This was my on-week project. There's some FIXME comments that detail some oddities we may want to address

michalpristas (commented):
would it make sense to add `make generate-open-api` 
also you can add a enrty into README with a link to e.g: https://petstore.swagger.io/?url=https://raw.githubusercontent.com/elastic/fleet-server/cd8b9c207ef6eaa707b1d085be7e28f1c7ff1b06/model/openapi.yml (replace with final path)

michalpristas (approved):

cmacknz:
Is there an automated way we can enforce that this stays up to date? Could we generate a client from it and write a test to ensure it can talk to fleet-server for example?

That is probably enough work to be done as a followup change though.

michel-laterman:
@cmacknz, i think the next steps  would be to use a codegen tool to generate server side boilerplate for the `api` package.

I briefly experimented with https://github.com/deepmap/oapi-codegen and https://openapi-generator.tech/ as generators and found that we may want to change the router we are using in the server from https://github.com/julienschmidt/httprouter to another in order to better support the change

cmacknz (approved):

PR #2192, opened 2022-12-16
Remove custom elasticsearch api methods.

### LINKED ISSUE #2187: remove custom es methods
We've update the go-elasticsearch library to v8 with https://github.com/elastic/fleet-server/pull/2173 and as a result we may be able to remove the custom methods we used for some calls; including fleet_search and global checkpoints, and use the ones provided in the official library.

Generally the custom calls are in `internal/pkg/es`, but we may want to verify our usage in other locations as well.

michel-laterman:
Use the official client methods for BulkUpdateApiKeys, FleetGlobalCheckpoint, FleetMSearch, and FleetSearch instead of the custom methods that were needed before the official client supported these calls.

## Related issues

- Closes #2187


jsoriano (approved):

aleksmaus (approved):
👍 practically drop-in replacement. thank you!

Issue #2205, opened 2022-12-21
Fleet server actions monitor goes into the tight loop when internally tracked checkpoint is out of sync with the index checkpoint


aleksmaus:
**Description**

The report from the field with 8.4.1:
"ESS deployment being flooded by requests to /.fleet-actions/_fleet/_fleet_search. 1.3M requests over last 2 hours when customer workload is mainly Logstash indexing."

![image](https://user-images.githubusercontent.com/872351/208974900-6d3607db-6a72-41a4-a791-9a8593849a52.png)
![image](https://user-images.githubusercontent.com/872351/208974923-7de9c387-d6c3-4dd4-a15e-326fbbb3fcdb.png)

**Investigation**

According to the logs it looks like the "wait for checkpoint advance" request returns immediately, followed by the fetching of the documents from the index and continues on the same loop:
https://github.com/elastic/fleet-server/blob/9c8c9ee7519910276cced8d001ac5e8ab2f244e0/internal/pkg/monitor/monitor.go#L225

The ```.fleet-actions``` index has only 28 documents.
The last document _seq_no is 107. The index checkpoint is 109. The documents with _seq_no of 108 and 109 are missing.

The way to reproduce the problem is when the fleet-server internally tracked checkpoint value is lagging behind the index checkpoint.

1. Assuming the  fleet server monitor internally tracked checkpoint value is 107 or less than the index checkpoint value 109 in this case. The long poll request for checkpoint at: 
https://github.com/elastic/fleet-server/blob/9c8c9ee7519910276cced8d001ac5e8ab2f244e0/internal/pkg/monitor/monitor.go#L225

```
.fleet-actions/_fleet/global_checkpoints?wait_for_advance=true&wait_for_index=true&checkpoints=107&timeout=5000ms
```

returns immediately with response

```
{
  "timed_out": false,
  "global_checkpoints": [
    109
  ]
}
```

2. The fetch for documents between the last known checkpoint value and the new checkpoint value, in this case between 107 and 109 returns 0 documents at:
https://github.com/elastic/fleet-server/blob/9c8c9ee7519910276cced8d001ac5e8ab2f244e0/internal/pkg/monitor/monitor.go#L259

3. The current implementation advances the checkpoint values based on the max(_seq_no) from the selected documents:
https://github.com/elastic/fleet-server/blob/9c8c9ee7519910276cced8d001ac5e8ab2f244e0/internal/pkg/monitor/monitor.go#L282
[...]

PR #2206, opened 2022-12-21
Fix the index monitor checkpoint tracking for when the documents are deleted

aleksmaus:

This fixes the bug with the fleet server actions monitor that doesn't account correctly on the document sequence number increment when the expired action is deleted. This leads to the monitor checkpoint getting out of sync (lagging behind) with the index checkpoint, and causes the monitor to go into the tight loop of awaiting for checkpoint advance (which returns immediately and fetching the actions documents.

"ESS deployment being flooded by requests to /.fleet-actions/_fleet/_fleet_search. 1.3M requests over last 2 hours when customer workload is mainly Logstash indexing."

The current workaround for this would be either restart the fleet-server or push the new action, like change the logging level for example.

Update monitor checkpoint value from the "wait advance" response upon the last fetch. This ensures that the checkpoint stays in sync with the index checkpoint even if the checkpoint advance happened due to the documents deleted from the index.

1. Run the stack with the fleet server and the agent.
2. Place some actions, like for example changing the agent logging level, or osquery live query.
3. Delete the action document.
4. Check the fleet server logs. The fleet server should not go into the tight loop after the deletion.

## Design Checklist

## Checklist

## Related issues

- Closes https://github.com/elastic/fleet-server/issues/2205


aleksmaus:
hmmm
```
[2022-12-22T16:11:17.761Z] error obtaining VCS status: exit status 128
[2022-12-22T16:11:17.761Z] 	Use -buildvcs=false to disable VCS stamping.
[2022-12-22T16:11:17.761Z] make: *** [Makefile:150: release-darwin/amd64] Error 1
[2022-12-22T16:11:18.865Z] Makefile:173: recipe for target 'docker-release' failed
[2022-12-22T16:11:18.865Z] make: *** [docker-release] Error 2
```

internal/pkg/monitor/monitor_integration_test.go
```diff
@@ -9,16 +9,22 @@ package monitor
 
 import (
 	"context"
-	"errors"
 	"sync"
 	"testing"
+	"time"
 
-	"github.com/google/go-cmp/cmp"
+	"github.com/rs/zerolog/log"
```
michel-laterman, line 16: nit: we have https://github.com/elastic/fleet-server/blob/main/internal/pkg/testing/log/log.go to redirect zerolog from what's tested to the `t.Log` stream

aleksmaus: Updated

aleksmaus:
Tested end to end. Here are the steps performed to test this feature.

Used the local environment running 8.7.0-SNAPSHOT Elasticsearch and Kibana.

Running locally build fleet-server is a bit tricky: I downloaded the latest elastic-agent 8.7.0-SNAPSHOT docker image, run the image, copied the locally built the fleet-server binary into the container, created the new image from the container, used run that against the local elasticsearch&kibana.

Then enrolled the regular agent with osquery enabled, placed few queries, made sure the .fleet-actions documents are created and the checkpoint is incremented and tracked by the fleet-server normally.

Screen Shot 2023-01-03 at 8 48 51 AM

Screen Shot 2023-01-03 at 8 49 03 AM

Then deleted the last action, used curl with the Authroization header with the fleet server token, the elastic user doesn't have permissions to delete from .fleet-actions index.

After delete observed that:
1. The checkpoint is incremented and has value 3 in this case

Screen Shot 2023-01-03 at 8 56 55 AM

2. While the last .fleet-actions document now with the _seq_no of 1

3. (IMPORTANT!) The fleet server continues to run and the .fleet-actions monitor doesn't go into the tight loop of checking the checkpoint and fetching 0 documents. This is confirmed from the logs.

Now issue another live query. 
1. The new action document is added normally
2. The fleet-server continues to track the checkpoint (monitor works) and dispatches it to the agent
3. The query result is sent back normally 

Screen Shot 2023-01-03 at 8 57 51 AM

4. The global checkpoint is incremented as expected along with the new action document added in the index
Screen Shot 2023-01-03 at 8 58 28 AM

Screen Shot 2023-01-03 at 8 58 19 AM

aleksmaus:
Tested end to end and regression tested this actions monitor change. Should be good to merge upon approval.

cmacknz (approved):
Looks good to me, with the disclaimer that I did not manually test this myself. I appreciated all the comments it made it easy to follow.

It would be nice to see this used in our scale tests before release, but it seems low risk considering that this situation isn't that common and the primary code path is unmodified.

You should likely add a change log entry for this as well.

PR #2211, opened 2022-12-22
Correct build using docker image

amannocci:
Signed-off-by: Adrien Mannocci <adrien.mannocci@elastic.co>

* This PR solves CI build issues.
* Ref: https://fleet-ci.elastic.co//blue/rest/organizations/jenkins/pipelines/fleet-server/pipelines/fleet-server-mbp/pipelines/PR-2180/runs/4/steps/290/log/?start=0

* Create a cache directory within the docker image at `/.cache`
* Forward local uid and gid when we run the docker image.

* Run the following make target `make docker-release`.

### DIFF
--- a/Dockerfile.build
+++ b/Dockerfile.build
@@ -1,7 +1,11 @@
 ARG GO_VERSION
 FROM golang:${GO_VERSION}-buster
 
-RUN apt-get update && apt-get install -y zip
+RUN \
+  apt-get update \
+  && apt-get install -y zip \
+  && mkdir /.cache \
+  && chmod 777 /.cache
 
 WORKDIR /go/src/github.com/elastic/fleet-server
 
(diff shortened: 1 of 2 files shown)


kuisathaverat (approved):

Dockerfile.build
```diff
@@ -1,7 +1,11 @@
 ARG GO_VERSION
 FROM golang:${GO_VERSION}-buster
 
-RUN apt-get update && apt-get install -y zip
+RUN \
+  apt-get update \
+  && apt-get install -y zip \
+  && mkdir /.cache \
```
michel-laterman, line 7: What is this dir used by?

michel-laterman (approved):
Thanks for the fix!

Dockerfile.build
amannocci: The go-build tool seems to use it to cache some data.

PR #2212, opened 2022-12-23
Allow dev stand-alone execution

### LINKED ISSUE #1423: Document how to locally run fleet server within Go and add it to our development guide

michel-laterman:

Allow developers to more easily run fleet-server without the elastic-agent.

Updates https://github.com/elastic/fleet-server/pull/1790.

Allow the fleet-server to run independently of the elastic-agent when built with `DEV=true`.
Specifically if a dev build runs without the `--agent-mode` flag it will will create an agent record for itself in the agents index with the ID specified in the fleet-server.yml file.
After regular startup the agent will also send checkins to its API handler in order to appear as a healthy agent in the UI. All config is read from the fleet-server.yml file. Actions and policy changes are ignored. action seq numbers are not persisted.

When compiling a release artifact, running without the `--agent-mode` is not supported.
However the `policy.SelfMonitor` will run in the main goroutine and the checkin method will be a nop.

TODO
- [ ] decide if policy should be read from config or use policy retrieved from ES
- [ ] reject actions?

1. Run + setup ES/Kibana
2. Build test binary
```shell
DEV=true SNAPSHOT=true make release-darwin/amd64
```
3. Create TLS CA + cert/key
```shell
../elasticsearch/bin/elasticsearch-certutil ca --pem --out stack.zip
unzip stack.zip
../elasticsearch/bin/elasticsearch-certutil cert --pem --ca-cert ca/ca.crt --ca-key ca/ca.key --ip $HOST_IP_ADDR --out cert.zip
unzip cert.zip
```
4. Run fleet-server
```shell
ELASTICSEARCH_CA_TRUSTED_FINGERPRINT="es-fingerprint" \
ELASTICSEARCH_SERVICE_TOKEN="service-token" \
FLEET_SERVER_POLICY_ID="policy-id" \
./build/binaries/fleet-server-8.7.0-SNAPSHOT-darwin-x86_64/fleet-server -c fleet-server.yml
```
5. enroll other agents with CA

## Design Checklist

Change is meant to assist running the fleet-server for local testing/development; not for production deployments.

## Checklist

## Related issues

- Closes #1423


internal/pkg/server/standAlone_dev.go
```diff
@@ -0,0 +1,138 @@
...
+)
+
+func (f *Fleet) standAloneSetup(ctx context.Context, bulker bulk.Bulk, sm policy.SelfMonitor, policyID, agentID string) (*model.Agent, error) {
+	err := sm.Run(ctx)
+	if err != nil {
+		return nil, fmt.Errorf("unable to run self monitor for stand alone setup: %w", err)
+	}
+	policy := sm.Policy()
+	log.Debug().Msgf("Policy Data: %s", string(policy.Data))
+	// TODO use policy from self monitor
```
michel-laterman, line 36: We should make a decision about the source of truth for the fleet-server in stand alone.
Do we want to :
1. feed the retrieved policy back into the config that get's executed
2. run only what's in the local `fleet-server.yml` config file

Option 1 will more closely mimic how the fleet-server runs under the agent. But may need more bootstrapping
Option 2 is much simpler, and we can easily fake any acks to policy_change or reassign actions.

fleet-server.yml
```diff
@@ -1,62 +1,30 @@
 output:
   elasticsearch:
-    hosts: '${ELASTICSEARCH_HOSTS:localhost:9200}'
+    hosts: '${ELASTICSEARCH_HOSTS:https://localhost:9200}'
     service_token: '${ELASTICSEARCH_SERVICE_TOKEN}'
+    ssl.ca_trusted_fingerprint: '${ELASTICSEARCH_CA_TRUSTED_FINGERPRINT}'
```
michel-laterman, line 5: this is currently breaking some tests; i'll look into it

michel-laterman:
Currently the stand-alone server gets into a stuck state on initial startup where the policy coordinator does not succeed:
```
14:05:02.443 DBG ensuring leadership of policies ctx="policy leader manager" ecs.version=1.6.0 service.name=fleet-server
14:05:02.697 ERR Fail BulkRequest result ecs.version=1.6.0 err="[400 Bad Request] {\"error\":{\"root_cause\":[{\"type\":\"action_request_validation_exception\",\"reason\":\"Validation Failed: 1: id is missing;\"}],\"type\":\"action_request_validation_exception\",\"reason\":\"Validation Failed: 1: id is missing;\"},\"status\":400}" mod=bulk service.name=fleet-server
14:05:02.697 ERR Eat bulk checkin error; Keep on truckin' error.message="elastic fail 400: action_request_validation_exception: Validation Failed: 1: id is missing;" ecs.version=1.6.0 service.name=fleet-server
```
However, restarting the server is enough to make it appear healthy in the UI

internal/pkg/bulk/opMulti_test.go
```diff
@@ -10,13 +10,14 @@ import (
...
 	testlog "github.com/elastic/fleet-server/v7/internal/pkg/testing/log"
+	"github.com/rs/zerolog/log"
 )
 
 const payload = `{"_id" : "1", "_index" : "test"}`
 
 // Test throughput of creating multiOps
 func BenchmarkMultiUpdateMock(b *testing.B) {
-	_ = testlog.SetLogger(b)
+	log.Logger = testlog.SetLogger(b)
```
michel-laterman, line 20: I've had to make a few changes like this; the `testlog.SetLogger` method creates a zerolog logger that will write to testing output (so integration tests look cleaner) and had a side effect of setting `log.Logger` to it. However the side effect was causing race issues in tests

fleet-server.yml
```diff
@@ -1,62 +1,31 @@
...
-#          certificate: /creds/cert.pem
-#          key: /creds/key.pem
-#    cache:
-#      num_counters: 500000  # 10x times expected count
-#      max_cost: 50 * 1024 * 1024  # 50MiB cache size
+#      ssl:
+#        enabled: true
+#        certificate: /creds/cert.pem
+#        key: /creds/key.pem
+#        key_passphrase: /creds/key.pem
```
michel-laterman, line 20: I wasn't able to get an external agent operating correctly if the fleet-server does not have a cert (in http mode).
The agent could enroll but would be stuck in the updating state.

I think that was because the fleet UI requires an https url.
I can use and http url and the `--insecure` flag to enroll, the policy it retrieves would differ.
I don't think we really support fleet-server running as an http server and may want to emit a warning (mark as deprecated?) if we try to do so

internal/pkg/server/standAlone_dev.go
michel-laterman: After giving this more thought, i'm leaning towards option 2 and just using config when the agent is in stand-alone mode.
The main purpose of option 1 would be to mimic running under the agent, but that scenario should be tested anyways and I don't think a stand-alone fleet-server gains much value if it needs to bootstrap in order to get remote config.
Also if we want to make a fleet service the config should not be exposed through kibana

internal/pkg/config/env_defaults.go
```diff
@@ -185,6 +185,9 @@ func loadLimits(agentLimit int) *envLimits {
 }
 
 func loadLimitsForAgents(agentLimit int) *envLimits {
+	if agentLimit == 0 {
+		return defaultEnvLimits()
+	}
```
michel-laterman, line 190: This change resolves a nitpick I have with the startup process.

Without the change, the fleet-server proc outputs:
```
{"level":"info","time":"2023-01-05T10:30:28-08:00","message":"No applicable limit for 0 agents, using default."}
{"level":"info","time":"2023-01-05T10:30:28-08:00","message":"No applicable limit for 0 agents, using default."}
{"level":"info","time":"2023-01-05T10:30:28-08:00","message":"No applicable limit for 0 agents, using default."}
{"level":"info","time":"2023-01-05T10:30:28-08:00","message":"No applicable limit for 0 agents, using default."}
10:30:28.650 INF Boot fleet-server args=["-c","fleet-server.yml"] commit=603c9df ecs.version=1.6.0 exe=./build/binaries/fleet-server-8.7.0-SNAPSHOT-darwin-x86_64/fleet-server pid=77997 ppid=69589 service.name=fleet-server version=8.7.0-SNAPSHOT
```

Where the first 4 lines are output when the go-ufcg and config.Config structs are merging as go-ufcg calls `InitDefaults` on any structs it unpacks, and may of these calls contain a call to `loadLimits(0)` which would result in the log call on line 202 executing before the logger is setup.
Adding the if statement silences log output when it's called with a 0 so that the fleet-server is able to setup its logger with the config before anything is written

fleet-server.yml
juliaElastic: It would be nice though to have an easier way to start Fleet Server + Agent without a need to manage certs for development.

juliaElastic (approved):
I tested this locally and it works nicely.

jlind23:
@juliaElastic Shall we merge it on Michel's behalf?

juliaElastic:
@jlind23 It would be great to have another review by someone who is more familiar with this code. @cmacknz who could help with that?

internal/pkg/server/fleet.go
```diff
@@ -47,26 +48,28 @@ const kUAFleetServer = "Fleet-Server"
...
+	verCon     version.Constraints
 
 	cfgCh    chan *config.Config
 	cache    cache.Cache
 	reporter state.Reporter
 }
 
 // NewFleet creates the actual fleet server service.
-func NewFleet(bi build.Info, reporter state.Reporter) (*Fleet, error) {
+func NewFleet(bi build.Info, reporter state.Reporter, standAlone bool) (*Fleet, error) {
```
michalpristas, line 61: can we bind this to a flag rather than hardcoding passed values

internal/pkg/server/standAlone_dev.go
```diff
@@ -0,0 +1,148 @@
...
+
+// standAloneSetup will ensure that the agent is enrolled.
+func (f *Fleet) standAloneSetup(ctx context.Context, bulker bulk.Bulk, sm policy.SelfMonitor, policyID, agentID string) (*model.Agent, error) {
+	log.Debug().Str("agent_id", agentID).Msg("stand-alone dev setup starting")
+	err := sm.Run(ctx)
+	if err != nil {
+		return nil, fmt.Errorf("unable to run self monitor for stand alone setup: %w", err)
+	}
+	policy := sm.Policy()
+	//log.Debug().Str("policy", string(policy.Data)).Msg("Found policy")
```
michalpristas, line 37: remove code or comment

README.md
```diff
@@ -1,166 +1,151 @@
...
-server.host: 0.0.0.0
-```
+#### fleet-server+agent on a Vagrant VM
 
-This makes sure, Kibana is accessible from the container. Start Kibana as before but for Elasticsearch, run the following command:
-
-```
-yarn es snapshot -E xpack.security.authc.api_key.enabled=true -E http.host=0.0.0.0
+The development Vagrant machine assumes the `elastic-agent`, `beats`, and `fleet-server` repos are in the same folder.
+Thus, it mounts `../` to `/vagrant` on the Vagrant machine.
```
AndersonQ, line 117: [Suggestion]
I'd keep the IP as it's hard coded in the Vagrant file, it won't change and is useful is someone is using the vagrant setup.

```suggestion
The development Vagrant machine assumes the `elastic-agent`, `beats`, and `fleet-server` repos are in the same folder.
Thus, it mounts `../` to `/vagrant` on the Vagrant machine. The vagrant machine IP address is `192.168.56.43`.
Use `https://192.168.56.43:8220` as fleet-server host.
```

AndersonQ (approved):

internal/pkg/server/standAlone_release.go
```diff
@@ -0,0 +1,28 @@
...
+func (f *Fleet) standAloneSetup(ctx context.Context, _ bulk.Bulk, sm policy.SelfMonitor, _, _ string) (*model.Agent, error) {
+	return nil, sm.Run(ctx)
+}
+
+// nop
+func (f *Fleet) standAloneCheckin(_ *model.Agent, _ *api.CheckinT) runFunc {
+	return func(_ context.Context) error {
+		return nil
+	}
+}
```
AndersonQ, line 28: I believe it's not allowed at all on production, right?
So I'd suggest to panic, just in case it's ever called on a production build. That way we'd catch it immediately and now exactly what is happening.

internal/pkg/server/fleet.go
michalpristas: this is what i had in mind, keep parameter for testing but use a Fn implementing `return true/false` based on DEV flag as input to NewFleet.

internal/pkg/server/standAlone_release.go
```diff
@@ -0,0 +1,28 @@
...
+	"github.com/elastic/fleet-server/v7/internal/pkg/api"
+	"github.com/elastic/fleet-server/v7/internal/pkg/bulk"
+	"github.com/elastic/fleet-server/v7/internal/pkg/model"
+	"github.com/elastic/fleet-server/v7/internal/pkg/policy"
+)
+
+// Run the SelfMonitor synchronously
+func (f *Fleet) standAloneSetup(ctx context.Context, _ bulk.Bulk, sm policy.SelfMonitor, _, _ string) (*model.Agent, error) {
+	return nil, sm.Run(ctx)
+}
```
AndersonQ, line 21:
```suggestion
func (f *Fleet) standAloneSetup(ctx context.Context, _ bulk.Bulk, sm policy.SelfMonitor, _, _ string) (*model.Agent, error) {
	panic(" fleet-server stand alone mode is only allowed for dev builds")
}
```

internal/pkg/server/standAlone_release.go
AndersonQ, line 28:
```suggestion
func (f *Fleet) standAloneCheckin(_ *model.Agent, _ *api.CheckinT) runFunc {
	panic(" fleet-server stand alone mode is only allowed for dev builds")
}
```

juliaElastic:
Capturing the steps based on this issue's description that worked for me. I've used this to test a change locally, unfortunately didn't find a way to enroll more than one agents.

# Steps to start Fleet Server locally, it only works with HTTPS

Steps based on: https://github.com/elastic/fleet-server/pull/2212 

- Set up certs 

```
cd /Users/juliabardi/Downloads/elasticsearch-8.5.3
bin/elasticsearch-certutil ca --pem --out stack.zip
unzip stack.zip
bin/elasticsearch-certutil cert --pem --ca-cert ca/ca.crt --ca-key ca/ca.key --ip 192.168.178.31 --out cert.zip
unzip cert.zip
./bin/elasticsearch-certutil cert --name fleet-server --ca-cert ca/ca.crt --ca-key ca/ca.key --ip 192.168.178.31 --pem
```

- Check out Fleet Server repo
- Uncomment ssl settings in fleet-server.yml (copy certs to creds directory or update path, make it accessible to root)

```
inputs:
  - type: fleet-server
    policy.id: '${FLEET_SERVER_POLICY_ID:fleet-server-policy}'
    server:
      ssl:
        enabled: true
        certificate: creds/fleet-server.crt
        key: creds/fleet-server.key
```        

- Build Fleet Server 

```
DEV=true SNAPSHOT=true make release-darwin/amd64
```

- Start kibana
- Create a Fleet Server policy 
- Copy the service token to env var below

- take es-fingerprint
```
openssl x509 -fingerprint -sha256 -in fleet-server/fleet-server.crt

#output contains fingerprint:

SHA256 Fingerprint=43:F9:E4:86:D7:06:CC:66:E1:8D:D3:DB:EC:B4:76:3A:25:20:C1:44:DA:41:6A:04:F1:63:63:D4:FC:28:C9:0F
-----BEGIN CERTIFICATE-----
...
```

- Start Fleet Server

```
ELASTICSEARCH_CA_TRUSTED_FINGERPRINT="es-fingerprint" \
ELASTICSEARCH_SERVICE_TOKEN="service-token" \
FLEET_SERVER_POLICY_ID="fleet-server-policy" \
./build/binaries/fleet-server-8.7.0-SNAPSHOT-darwin-x86_64/fleet-server -c fleet-server.yml
```

- enroll other agents with CA (--certificate-authorities) pointing to the same fleet-server.crt and a valid enrollment token

```
sudo ./elastic-agent install --url=https://192.168.178.31:8220 --enrollment-token=WFNmWm1vVUJtNExuOVlQQ3FYZGI6cFZ4UnQzUDhSYWkxRnJCeDF1S2FEUQ== --certificate-authorities=/Users/juliabardi/fleet-server/creds/fleet-server.crt
```

- Tried similarly to enroll an agent in docker with CA, but no luck. Tried with the available CA envvars listed here, nothing worked https://www.elastic.co/guide/en/fleet/master/agent-environment-variables.html

```
docker run  \
  --env FLEET_ENROLL=1 \
  --env FLEET_URL=https://192.168.178.31:8220 \
  --env FLEET_ENROLLMENT_TOKEN=ZTl5TGg0VUJuZ3lnMll5RHZjTUs6N2lncGhUNjFUc3VfSUt5aHZ4SkZfZw==  \
  --env FLEET_CA="/Users/juliabardi/fleet-server/creds/fleet-server.crt" \
  docker.elastic.co/beats/elastic-agent:8.5.0-SNAPSHOT
```

Error:
```
{"log.level":"error","@timestamp":"2023-01-12T14:22:49.183Z","log.logger":"tls","log.origin":{"file.name":"tlscommon/tls.go","file.line":154},"message":"Failed reading CA certificate: open /Users/juliabardi/fleet-server/creds/fleet-server.crt: no such file or directory","ecs.version":"1.6.0"}
Error: Error: 1 error: open /Users/juliabardi/fleet-server/creds/fleet-server.crt: no such file or directory reading <nil>
For help, please see our troubleshooting guide at https://www.elastic.co/guide/en/fleet/8.5/fleet-troubleshooting.html
Error: enrollment failed: exit status 1
```

- Not aware of a way to enroll horde agents with CA

joshdover:
In our last discussion of this PR, we discussed removing some of the parts that emulate an Agent in the UI. This is how we expect to run Fleet Server in our managed offering. @michel-laterman how do you think we should address this? Should we make the agent emulation on `.fleet-agents` optional or just remove it?

PR #2243, opened 2023-01-03
[Fleet] schedule upgrade no longer using expiration

juliaElastic:
- Enhancement

Fleet UI removed expiration (using default 1M) from Upgrade actions. This needs a corresponding change in Fleet Server not to use expiration when calculating start time offset per agents.

Fleet Server uses `StartTime:StartTime+Duration` range to schedule upgrade.

## Checklist

## Related issues

- Relates https://github.com/elastic/kibana/issues/146018


internal/pkg/model/schema.go
```diff
@@ -48,8 +48,8 @@ type Action struct {
 	// The input type the actions should be routed to.
 	InputType string `json:"input_type,omitempty"`
 
-	// The minimum time (in seconds) provided for an action execution when scheduled by fleet-server.
-	MinimumExecutionDuration int64 `json:"minimum_execution_duration,omitempty"`
+	// The rollout duration (in seconds) provided for an action execution when scheduled by fleet-server.
+	RolloutDurationSeconds int64 `json:"rollout_duration_seconds,omitempty"`
```
michel-laterman, line 52: This file is generated from `model/schema.json`. you need to update that file then run `make update` to update this file.

michel-laterman (commented):
This change may need a change in the elastic-agent, the current de-queuing logic assumes that an expiration is returned:
https://github.com/elastic/elastic-agent/blob/ee1b43f4389b763342762e5f2706b4f64a6aa8bf/internal/pkg/agent/application/dispatcher/dispatcher.go#L221-L231
However, if we are getting rid of expirations we will want to check for that error
https://github.com/elastic/elastic-agent/blob/ee1b43f4389b763342762e5f2706b4f64a6aa8bf/internal/pkg/fleetapi/action.go#L235

We may also want to more throughly test and document the upgrade scenario for upgrading the fleet-server/agent from `X-1` to X. 
We need to make sure that the fleet-server is upgraded before any other agents are scheduled so the schema changes will be respected

juliaElastic:
@michel-laterman @kpollich 
I wonder if we should make this change backward compatible e.g. use expiration rollout period if it is present, otherwise fall back to old logic with expiration.
What do we usually do we these kind of changes, should it be a guideline?

juliaElastic:
@michel-laterman what I don't understand here, how does the change in elastic-agent help? Agents won't have the latest version, that is why users would upgrade them, so adding a change to agent code won't help for upgradeable agents with older versions.
Does this mean we shouldn't remove expiration, as it will throw an error in agent dequeue logic?

michalpristas:
the way i see it, it is a breaking change and we should plan on adding this on next major. 
we cannot guarantee versions deployed so there still can be stale agent which loses capability of being upgraded. 
agent change should still be performed though. is it possible for you to have a conditional behavior/separate strategies for older agents?

@joshdover how deep we should support agents? 
we still can provide a update guide that upgrade to 8.8 is only possible through 8.7 but i'd like to avoid it.

joshdover:
We have UI enforcement to ensure that Fleet Server agents are upgraded before all other agents, not sure if that helps here.

Would we be able to do this in a non-breaking way by having Kibana keep including the expiration time on upgrade actions sent to Fleet Server agents < 8.7? 

I agree we can't require changes to Agent or else that will be a breaking change. We need to be able to upgrade any version of Elastic Agent from the latest version of Kibana and Fleet Server.

juliaElastic:
> Would we be able to do this in a non-breaking way by having Kibana keep including the expiration time on upgrade actions sent to Fleet Server agents < 8.7?

To make this backward compatible, we could leave the same expiration as before, and only add the new rollout period field. We can then remove the expiration on the next major. The new Fleet Server version can disregard the existing expiration value, though it could be confusing to have an expiration that doesn't work the same way as in other actions.

joshdover:
This makes sense to me, let's just make sure there's good comments. I do think expiration time is distinct from a rollout period so not overloading that field makes sense.

As an aside, @kpollich we should start tracking an issue with a list of things we want to remove in the next major version.

kpollich:
> As an aside, @kpollich we should start tracking an issue with a list of things we want to remove in the next major version.

Good idea. I created https://github.com/elastic/ingest-dev/issues/1490 to start tracking these.

juliaElastic:
Added back existing expiration logic to kibana [here](https://github.com/elastic/kibana/pull/148240/commits/6eb9a13caa25b6573eff20358eb8c2c7bc030bca). 
Are we good if the `fleet-server` changes stay as-is (only using rollout duration in the new version)?
The only edge case that can happen is to have agent documents in ES that were created with the old kibana version (without rollout period), and the new FS would take those as immediate upgrade actions.

kpollich:
> Are we good if the fleet-server changes stay as-is (only using rollout duration in the new version)?
>
> The only edge case that can happen is to have agent documents in ES that were created with the old kibana version (without rollout period), and the new FS would take those as immediate upgrade actions.

I think this is ok? I can't imagine it being a common scenario to complete a stack upgrade with pending agent upgrades still incomplete, e.g. a pile of actions in ES. Is there another case where a user might run into this or am I misunderstanding the edge-case?

kpollich:
> Yes, I am referring to this scenario to have in progress actions before a stack upgrade.

Thanks for clarifying. I think it's okay for Fleet Server to exhibit this edge case and ship this functionality then.

juliaElastic:
@michalpristas @kpollich are you ready to approve this pr?

kpollich (approved):
Functionality LGTM, but I'll leave a more thorough code-focused review to the other folks pinged here.

juliaElastic:
Tried to test this today as the [latest fleet-server snapshot](https://artifacts-api.elastic.co/v1/versions/8.7-snapshot/builds/latest) contains the change, however the instance on cloud staging didn't work as expected. It could be that cloud staging doesn't have the latest fleet-server snapshot yet.

What I experienced:
- if the action was expired, it was not delivered to the agent, while the latest fleet-server code doesn't take into account the expiration time value.

PR #2299, opened 2023-01-25
Openapi docs for File Upload routes

pzl:

This is a followup PR to #1902 to add route documentation to the openapi spec

You may be able to preview the generated output by pasting the contents of `model/openapi.yml` into the editor side of https://editor.swagger.io/


model/openapi.yml
```diff
@@ -831,6 +920,8 @@ paths:
       security:
         - agentApiKey: []
       requestBody:
+        required: true
+        description: Information about the file to be uploaded. Minimum required fields are marked as required. Additional fields may be specified and are allowed. For information about the file itself, ECS.file paths are recommended. For archived files, information about the archive should be placed in `file`. Information about the archive members may be placed in a `contents` array.
```
michel-laterman, line 924: This is in the wrong spot

model/openapi.yml
```diff
@@ -911,4 +1002,99 @@ paths:
           $ref: '#/components/responses/throttle'
         '503':
           $ref: '#/components/responses/unavailable'
-# TODO: Include the file-upload endpoints?
+  /api/fleet/uploads:
```
michel-laterman, line 1005: Can you add the request id header as well:
```yaml
parameters:
  - $ref: '#/components/parameters/requestId'
```
should be added to all paths

model/openapi.yml
```diff
@@ -911,4 +1002,99 @@ paths:
...
+            schema:
+              $ref: '#/components/schemas/uploadBeginRequest'
+      responses:
+        '200':
+          description: Information about the upload procedure
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/uploadBeginResponse'
+        '400':
```
michel-laterman, line 1023: I think (for completeness) 401, 408, and 503 may also be returned. (also on other paths)

model/openapi.yml
```diff
@@ -911,4 +1002,99 @@ paths:
...
+          required: true
+          description: the SHA256 hash of the body contents for this request
+          schema:
+            type: string
+            example: 0c4a81b85a6b7ff00bde6c32e1e8be33b4b793b3b7b5cb03db93f77f7c9374d1
+      requestBody:
+        description: The chunk contents as bytes
+        required: true
+        content:
+            "*/*":
```
michel-laterman, line 1058: do we want to use `application/octet-stream` instead?

model/openapi.yml
```diff
@@ -911,4 +1002,99 @@ paths:
...
+              schema:
+                type: string
+                format: binary
+      responses:
+        '200':
+          description: Successful chunk upload
+        '400':
```
michel-laterman, line 1065: add other responses?

I'm also assuming there's no 428 response anymore

model/openapi.yml
```diff
@@ -475,6 +475,95 @@ components:
...
+              description: Name of the file including the extension, without the directory
+              type: string
+              example: yankees-stats.zip
+            mime_type:
+              description: MIME type of the file
+              type: string
+              example: application/zip
+            size:
+              description: Size of the file contents, in bytes
+              type: integer
```
michel-laterman, line 513: add `format: int64`?

michel-laterman (commented):
mostly have nitpicks.

model/openapi.yml
pzl: to be fair, `*/*` is the most accurate. The content-type is not checked, and the actual content could be plain text, could be binary, could be anything.

A client could send

`Content-Type: text/csv` and send just text in the body and that would work.

model/openapi.yml
```diff
@@ -911,4 +1002,123 @@ paths:
...
+    post:
+      summary: Initiate a file upload process
+      description: ""
+      security:
+        - agentApiKey: []
+      parameters:
+        - $ref: '#/components/parameters/requestId'
+      requestBody:
+        required: true
+        description: Information about the file to be uploaded. Minimum required fields are marked as required. Additional fields may be specified and are allowed. For information about the file itself, ECS.file paths are recommended. For archived files, information about the archive should be placed in `file`. Information about the archive members may be placed in a `contents` array.
```
michel-laterman, line 1015: Is the `contents` array just an array of objects? Is there any definition for it anywhere?

michel-laterman (approved):
lgtm

model/openapi.yml
pzl: yeah I wasn't sure how much to explain.

The required fields are explicitly mapped. Everything else is unmapped, but `dynamic: true` is set on the mapping. So integrations can just index whatever they want, in whatever format.

Considering zip/archive is a common use case, I wanted to suggest a convention (have `file` describe the zip, and `contents: [ {}, {}, {} ...]` describe the files inside). But it is not required, enforced, anything. Just a suggested convention.

The mappings are not shared between integrations, so any client can do what they want. That's why this isn't a very prescriptive definition

PR #2300, opened 2023-01-25
Fix Upload file chunk endpoint for successful uploads

michel-laterman:
chunk endpoint returns 200 on success, fix vagrantfile

### DIFF
--- a/internal/pkg/api/handleUpload.go
+++ b/internal/pkg/api/handleUpload.go
@@ -209,6 +209,7 @@ func (ut *UploadT) handleUploadChunk(zlog *zerolog.Logger, w http.ResponseWriter
 		return uploader.ErrHashMismatch
 	}
 
+	w.WriteHeader(http.StatusOK)
 	return nil
 }
 
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -43,7 +43,7 @@ Vagrant.configure("2") do |config|
           curl \
           delve \
           make \
-          unzip
+          unzip \
           vim \
           wget \
           zip


pzl (approved):

internal/pkg/api/handleUpload.go
```diff
@@ -209,6 +209,7 @@ func (ut *UploadT) handleUploadChunk(zlog *zerolog.Logger, w http.ResponseWriter
 		return uploader.ErrHashMismatch
 	}
 
+	w.WriteHeader(http.StatusOK)
```
jsoriano, line 212: Would this be needed in other handlers here too?

jsoriano (approved):

internal/pkg/api/handleUpload.go
AndersonQ: By default Go takes care of it if no call has been made to `WriteHeader` before `Write`:
> 	// If WriteHeader is not called explicitly, the first call to Write
	// will trigger an implicit WriteHeader(http.StatusOK).

AndersonQ (approved):

internal/pkg/api/handleUpload.go
blakerouse: Is this needed because `w.Write(...)` is not called? I would expect this is required here because there is no call to `w.Write(...)`.

michel-laterman: currently nothing gets written for a successful chunk upload so we need to explicitly write the header

PR #2331, opened 2023-02-09
Allow the the Warn log level to be specified as "warning" and "warn"

### LINKED ISSUE #2328: Fleet server does not translate the Warning log level from agent correctly.
Fleet server sets the log level received from agent here:

https://github.com/elastic/fleet-server/blob/66e36584f16c6378d14a629db9351fc055ca08c2/internal/pkg/server/agent.go#L367-L375

The log level `.String()` implementation is from https://github.com/elastic/elastic-agent-client/blob/4477e3ace394ef1abfec55afa5cc5e1f6f87980c/pkg/client/unit.go#L54-L70

```go
// String returns string representation for the unit log level.
func (l UnitLogLevel) String() string {
	switch l {
	case UnitLogLevelError:
		return "error"
	case UnitLogLevelWarn:
		return "warn"
```

Note that it it converts `UnitLogLevelWarn` to the string "warn". The Fleet server code to validate the log level is at:

https://github.com/elastic/fleet-server/blob/66e36584f16c6378d14a629db9351fc055ca08c2/internal/pkg/config/fleet.go#L70-L76

Fleet server wants the level string to be `warning` and not `warn`. Setting the agent log level to Warning causes Fleet server to fail to start because it fails to parse configurations that set the Fleet server log level to `warn` instead of `warning`.

michel-laterman:

The log level provided by the Agent is "warn", and crashes the fleet-server as it expects "warning".

Config will parse "warning" and "warn" as `zerolog.WarnLevel`

## Checklist

## Related issues

- Closes #2328

### DIFF
--- a/internal/pkg/config/fleet.go
+++ b/internal/pkg/config/fleet.go
@@ -67,7 +67,7 @@ func strToLevel(s string) (zerolog.Level, error) {
 		l = zerolog.DebugLevel
 	case "info":
 		l = zerolog.InfoLevel
-	case "warn":
+	case "warn", "warning":
 		l = zerolog.WarnLevel
 	case "error":
 		l = zerolog.ErrorLevel
--- a/internal/pkg/config/fleet.go
+++ b/internal/pkg/config/fleet.go
@@ -67,12 +67,12 @@ func strToLevel(s string) (zerolog.Level, error) {
 		l = zerolog.DebugLevel
 	case "info":
 		l = zerolog.InfoLevel
-	case "warning":
+	case "warn":
 		l = zerolog.WarnLevel
 	case "error":
 		l = zerolog.ErrorLevel
 	default:
-		return l, fmt.Errorf("invalid log level; must be one of: trace, debug, info, warning, error")
+		return l, fmt.Errorf("invalid log level; must be one of: trace, debug, info, warn, error")
 	}
 
 	return l, nil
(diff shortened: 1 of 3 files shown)


internal/pkg/config/fleet.go
```diff
@@ -67,12 +67,12 @@ func strToLevel(s string) (zerolog.Level, error) {
 		l = zerolog.DebugLevel
 	case "info":
 		l = zerolog.InfoLevel
-	case "warning":
+	case "warn":
```
cmacknz, line 70: Are there backwards compatibility concerns with this?

I suppose there aren't since agent is the only thing that can start Fleet server, and the versions are always in sync. I would feel better about not forgetting something if we just accepted "warning" and "warn", but maybe I am just being paranoid.

cmacknz (approved):

internal/pkg/config/fleet.go
jen-huang, line 70: Sorry, we didn't update the original issue but @cmacknz and I discussed that both `warn` and `warning` should be allowed

jen-huang (approved):

internal/pkg/config/fleet.go
cmacknz: Thanks!

cmacknz:
You need to backport to 8.7 now as well, added the label.

internal/pkg/config/fleet.go
michel-laterman: Already changed to accept both values

PR #2340, opened 2023-02-13
updatecli: fix the bump elastic stack

v1v:

Fixes the bump automation by allowing running in parallel all the bump for all the active branches
Fixing the regex group that somehow remove the major...

### DIFF
--- a/.ci/bump-elastic-stack-snapshot.yml
+++ b/.ci/bump-elastic-stack-snapshot.yml
@@ -3,7 +3,7 @@ name: Bump elastic-stack to latest snapshot version
 
 actions:
   default:
-    title: '[updatecli] update elastic stack version for testing {{ source "latestVersion" }}'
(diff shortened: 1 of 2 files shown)


cmacknz (approved):

.ci/bump-elastic-stack-snapshot.yml
```diff
@@ -40,5 +40,5 @@ targets:
     scmid: default
     spec:
       file: dev-tools/integration/.env
-      matchpattern: '(ELASTICSEARCH_VERSION=)\d+.\d+.\d+-.*-SNAPSHOT'
-      replacepattern: '$1{{ source "latestVersion" }}-SNAPSHOT'
```
v1v, line 44: for some reason it produced https://github.com/elastic/fleet-server/pull/2338/files#diff-22766f0cd0d5cfcb04aebb38f81f912b6e52b1bc417abc1671954b3c4e8364b2R1

PR #2341, opened 2023-02-13
Increase checkin long-poll time to 28m and write timeout to 29m

### LINKED ISSUE #2337: Increase long poll default to ~30 minutes
We've been doing scale testing over the past few months using a ~30 minute long poll duration (rather than current default of 5m) and we are seeing much better results for very large clusters.

We're now ready to make this the default setting for Fleet Server. A corresponding change to Agent will be made, I'll link the issue soon.

### Tasks

The timeout at the proxy will be slightly over 30 minutes at 30m 20s. We should configure Fleet Server to timeout shortly before that, so we'll use a 28m long poll duration with a 29m response write timeout.

- [ ] Update default for `server.timeouts.checkin_long_poll` to 28m here: https://github.com/elastic/fleet-server/blob/c8527d3d9f55c44d4f4764a017c93731f595f342/internal/pkg/config/timeouts.go#L57
- [ ] Update default for `server.timeouts.write` to 29m here: https://github.com/elastic/fleet-server/blob/c8527d3d9f55c44d4f4764a017c93731f595f342/internal/pkg/config/timeouts.go#L47-L51

michel-laterman:

Large deployments have performance degradations due to low timeout values.

Increase default checkin long-poll time to 28m and write timeout to 29m. Adjust comments that reference these timeouts.

## Design Checklist

Adjustment is a result of scale testing

## Checklist

## Related issues

- Closes #2337


juliaElastic (approved):
LGTM

PR #2344, opened 2023-02-14
updatecli: ensure one branch per release branch

v1v:
## What does this PR do?

Use one unique branch per PR targeting the release branches

## Why is it important?

Otherwise, it reused the same branch for different target releases and the automation will produce PRs with reused commits, such as https://github.com/elastic/beats/pull/34570

https://github.com/updatecli/updatecli/issues/1159 contains a bit more of context and the solution to solve this particular corner case, thanks @olblak 🙇


olblak:
@v1v Quick feedback, if you put the Updatecli manifest(s) in the directory `updatecli.d` (or below) then any IDE supporting jsonschema store like VScode or Intellij will automatically have validation and auto-completion enabled without additional configuration.

While you may not use those IDE yourself, some of your colleague may at some point.

amannocci (approved):

PR #2354, opened 2023-02-15
Fix fleet-server discarding extra server.* keys when assembling config

### LINKED ISSUE #2303: [Fleet] User are not able to configure server fields from Fleet
## Description 

It seems that before 8.6 user were able to configure allowed ssl protocol using that config and it does not work anymore
Screen Shot 2023-01-27 at 3 54 42 PM

when running nmap with `nmap -script ssl-enum-ciphers -p 8220 127.0.0.1` you can see that the tls protocol is not restricted
Screen Shot 2023-01-27 at 3 57 05 PM

### Investigation details

The following config is generated in Fleet 
Screen Shot 2023-01-27 at 3 54 59 PM

I tried to manual edit a policy and it seems that policy works well without the two `server` and `server.ssl.supported_protocol` 

```
id: fleet-server-fleet_server-8c1f2fbd-4e5b-450a-8565-2229c4437b81
revision: 3
name: fleet_server-1
type: fleet-server
data_stream:
  namespace: default
use_output: default
package_policy_id: 8c1f2fbd-4e5b-450a-8565-2229c4437b81
server:
  port: 8220
  host: 0.0.0.0
  ssl:
    supported_protocols:
    - TLSv1.2
meta:
  package:
    name: fleet_server
    version: 1.2.0

```

michel-laterman:

The policy that runs on fleet-server may include multiple "server.*" keys as part of the input.
These keys should be combined before being passed

Passing the input soruce as a map directly in one step when creating the
fleet-server config does not merge all keys correctly.
In order to resolve we explicitly create a ucfg.Config from the passed
input and unpack to a map, then use that map when creating the server
config.

## Checklist

## Related issues

- Closes #2303


michel-laterman:
New unit test currently fails, i'm not sure if this can be solved by passing an `Option` to go-ucfg, or if we need to alter `go-ucfg` to account for this type of input

michel-laterman:
We may want to consider using [viper](https://github.com/spf13/viper) to handle config, a quick test shows that it can manage unmarshalling yaml correctly see [my example here](https://gist.github.com/michel-laterman/469a9a49f2bd0026b9f7cb7e733e0958).

kpollich (approved):
LGTM - appreciate the comments + test case

nchaulet (approved):
🚀

Issue #2352, opened 2023-02-15
Add CI pipeline to publish docker images for standalone mode


jsoriano:
Once https://github.com/elastic/fleet-server/issues/2294 is ready, add a pipeline to publish docker images.

These docker images should be published on each commit on `main` by now.

joshdover:
A couple of notes about requirements here:
- We don't need this image to be public, but it's ok if it is. It should be clear though that this is not a supported artifact for customers to deploy.
- We should publish a new image on each commit to main and tag it with `latest`
- We should publish a new image on each git tag push
- We should use the same base image we use for elastic-agent's image

alexsapran:
[ cross-posting from #2359 ]
We discussed this with @joshdover.

We could use the [internal docker registry](https://github.com/elastic/infra/blob/master/docs/container-registry/accessing-the-docker-registry.md?rgh-link-date=2023-02-21T11%3A27%3A07Z) and publish the container image under the employee's path.

After some quick searching in GH I noticed how it's done in [endpoint-dev repo for some dockery-related builds](https://github.com/elastic/endpoint-dev/blob/46819ce5fce831cf6fddd2d8561d6e6a3b4d6133/Containers/Makefile#L7), setting an NS (namespace) to docker.elastic.co/employees/ instead of docker.elastic.co/.

jsoriano:
> We could use the [internal docker registry](https://github.com/elastic/infra/blob/master/docs/container-registry/accessing-the-docker-registry.md?rgh-link-date=2023-02-21T11%3A27%3A07Z) and publish the container image under the employee's path.

These images are going to be used in the production service, we souldn't use an employee's path.

In https://github.com/elastic/fleet-server/pull/2359 I am using `docker.elastic.co/elastic-agent/fleet-server`, but I am fine with changing this name. Maybe another option would be to use `docker.elastic.co/fleet-server/fleet-server`.

jlind23:
@mrodm could you make a high priority on your end as this is needed for the fleet service project. Anything we can do to help you here?

mrodm:
@jlind23 :+1: 
to avoid any possible blockers or issues, I was thinking to do it for the time being with Jenkins CI pipelines.

As mentioned here, I'll check the pipelines and Makefile to use the URL mentioned in https://github.com/elastic/fleet-server/issues/2352#issuecomment-1455872594
`docker.elastic.co/elastic-agent/fleet-server` , ok ? @jsoriano @joshdover

[...]

PR #2359, opened 2023-02-20
Dockerized standalone Fleet Server

### LINKED ISSUE #2294: Add a Dockerfile for running Fleet Server in standalone mode
**Describe the enhancement:** We added standalone mode in https://github.com/elastic/fleet-server/pull/2212 for development. We should start building a Docker image for this so we can use it more easily.

**Describe a specific use case for the enhancement or feature:** Development, Fleet Service

### LINKED ISSUE #2351: Remove self-registration code for standalone mode
Fleet Server still contains self-registration code that prevents it to be run in completely standalone mode, without Elastic Agent.

Remove this code for this use case.

jsoriano:

Allow to run Fleet Server standalone without check in in production builds.

* Standalone mode previously used in development mode can be also used now in release builds.
* Elasticsearch version check is not done in standalone mode.
* Migrations are not done in standalone mode.
* Health is checked in standalone mode by checking that the service can read the policies index.

After this, `fleet-server` only needs an Elasticsearch service account to start.

`fleet-server -c ./fleet-server.yml -E output.elasticsearch.ssl.verification_mode=none -E fleet.agent.checkin=false -E output.elasticsearch.service_token=....`

Or `make release-docker`, and use the built docker image in your docker or kubernetes environment.

## Design Checklist

## Checklist

## Related issues

- Fix #2351
- Fix #2294


internal/pkg/server/fleet.go
```diff
@@ -458,15 +458,18 @@ func (f *Fleet) runSubsystems(ctx context.Context, cfg *config.Config, g *errgro
...
-	sm := policy.NewSelfMonitor(cfg.Fleet, bulker, pim, cfg.Inputs[0].Policy.ID, f.reporter)
+	var sm policy.SelfMonitor
 	var agent *model.Agent
-	if f.standAlone {
-		agent, err = f.standAloneSetup(ctx, bulker, sm, cfg.Inputs[0].Policy.ID, cfg.Fleet.Agent.ID)
-		if err != nil {
-			return err
+	if cfg.Fleet.Agent.Checkin {
+		sm := policy.NewSelfMonitor(cfg.Fleet, bulker, pim, cfg.Inputs[0].Policy.ID, f.reporter)
+		if f.standAlone {
```
jsoriano, line 465: I am assuming by now that "check in" and "standalone" mode are two different things, but I wonder if there is any use case for running standalone with check in (maybe only for development?), or to run in agent without checkin.

jsoriano:
@michel-laterman please take a look to this draft to check if I am going in the good direction :slightly_smiling_face: Thanks!

amitkanfer:
@jsoriano won't it be best to have a new "standalone" config that will toggle the new `checkin` flag you're adding here? i guess that in the future we'll need to control more configs than just this one...
WDYT?

alexsapran:
We discussed this with @joshdover. 

We could use the [internal docker registry](https://github.com/elastic/infra/blob/master/docs/container-registry/accessing-the-docker-registry.md) and publish the container image under the `employees` path. 

After some quick searching in GH I noticed how it's done in [endpoint-dev repo for some dockery-related builds](https://github.com/elastic/endpoint-dev/blob/46819ce5fce831cf6fddd2d8561d6e6a3b4d6133/Containers/Makefile#L7), setting an NS (namespace) to `docker.elastic.co/employees/` instead of `docker.elastic.co/`.

jsoriano:
> @jsoriano won't it be best to have a new "standalone" config that will toggle the new `checkin` flag you're adding here? i guess that in the future we'll need to control more configs than just this one... WDYT?

Yeah, I guess this is related to my other question in https://github.com/elastic/fleet-server/pull/2359#discussion_r1112292612. It will depend on the use cases we support, if for all standalone Fleet Servers we want to disable the checkin, then we can have a single "standalone" config that also disables the checkin.

But, currently we also have an `-agent-mode` flag that disables standalone, and I think that we may have cases where we want to enable or disable the checkin independently of running in agent mode or in standalone mode:

| Use case | `-agent-mode` (no standalone) | Check in |
|--|:--:|:--:|
| Current Fleet Server managed by Agent | true | true |
| Current standalone Fleet Server for development | false | true |
| Standalone Fleet Server on premise | false | true |
| Standalone managed Fleet Server (this PR) | false | false |

jsoriano:
> We discussed this with @joshdover.
> 
> We could use the [internal docker registry](https://github.com/elastic/infra/blob/master/docs/container-registry/accessing-the-docker-registry.md) and publish the container image under the `employees` path.
> 
> After some quick searching in GH I noticed how it's done in [endpoint-dev repo for some dockery-related builds](https://github.com/elastic/endpoint-dev/blob/46819ce5fce831cf6fddd2d8561d6e6a3b4d6133/Containers/Makefile#L7), setting an NS (namespace) to `docker.elastic.co/employees/` instead of `docker.elastic.co/`.

@alexsapran sounds good. There is a separate issue about publishing the image being added here. Let's keep the discussion about this there. It is https://github.com/elastic/fleet-server/issues/2352.

internal/pkg/server/fleet.go
jsoriano: More about this in https://github.com/elastic/fleet-server/pull/2359#issuecomment-1438380909

alexsapran:
> > We discussed this with @joshdover.
> > 
> > We could use the [internal docker registry](https://github.com/elastic/infra/blob/master/docs/container-registry/accessing-the-docker-registry.md) and publish the container image under the `employees` path.
> > 
> > After some quick searching in GH I noticed how it's done in [endpoint-dev repo for some dockery-related builds](https://github.com/elastic/endpoint-dev/blob/46819ce5fce831cf6fddd2d8561d6e6a3b4d6133/Containers/Makefile#L7), setting an NS (namespace) to `docker.elastic.co/employees/` instead of `docker.elastic.co/`.
> 
> @alexsapran sounds good. There is a separate issue about publishing the image being added here. Let's keep the discussion about this there. It is https://github.com/elastic/fleet-server/issues/2352.

Makes sense, thanks for the link, I was not aware of that other issue, will cross post my comment there.

internal/pkg/server/fleet.go
nchaulet: I do not think we need to support checkin at all in standalone, but for these we probably need to make a change in Kibana to remove the restrictions of having an healthy fleet server agent, I can make a PR that introduce that change with a feature flag if you need it for testing.

fleet-server.yml
```diff
@@ -3,12 +3,15 @@ output:
   elasticsearch:
     hosts: '${ELASTICSEARCH_HOSTS:https://localhost:9200}'
     service_token: '${ELASTICSEARCH_SERVICE_TOKEN}'
-    ssl.ca_trusted_fingerprint: '${ELASTICSEARCH_CA_TRUSTED_FINGERPRINT}'
+    # ssl.ca_trusted_fingerprint: '${ELASTICSEARCH_CA_TRUSTED_FINGERPRINT}'
```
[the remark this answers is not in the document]
michel-laterman: By default elasticsearch runs in https

Dockerfile
```diff
@@ -0,0 +1,22 @@
+ARG GO_VERSION
+FROM golang:${GO_VERSION}-buster AS builder
```
michel-laterman, line 2: Why are we using this instead of the same image as we base https://github.com/elastic/fleet-server/blob/main/Dockerfile.build off of?

internal/pkg/server/fleet.go
michel-laterman: I agree with @nchaulet, registration + checkin was added just to get around the kibana restrictions.
If we wanted to release a true "stand-alone" server it should not even need to enroll itself

Dockerfile
jsoriano: Both Dockerfiles have different purpouses, `Dockerfile.build` is used to build the release packages for all platforms, and writes the generated packages to the working copy. This one only builds the binary in the platform it is executed, without writing to the working copy, and from it builds the docker image.

I could make a common Dockerfile for both to import, but I think they wouldn't have so much common code. We would also need to somehow handle the different platforms here.

It would be good though to use here `make release`, to ensure that we are building the same release artifacts in both cases.

I will give another thought to this. I am mainly using this Dockerfile now to have an image for testing on Kubernetes. We could have a more final Dockerfile later.

internal/pkg/server/fleet.go
jsoriano: We can then fully remove `standAloneSetup`?

jsoriano: > I can make a PR that introduce that change with a feature flag if you need it for testing.

@nchaulet that would be great. I guess this would solve part of the _"Add Fleet Server" UI Components_ in https://github.com/elastic/ingest-dev/issues/1530. Instead of using a feature flag we could maybe check if Kibana is runing in standalone mode, if this check is already available.

jsoriano:
Opening for review. @michel-laterman @nchaulet please take another look.

juliaElastic:
This is great! It would be nice to add the example command to start to the README.

nchaulet:
Looks great to me, I tested this locally and it worked well!

internal/pkg/policy/standalone_test.go
```diff
@@ -0,0 +1,71 @@
+// Copyright Elasticsearch B.V. and/or licensed to Elasticsearch B.V. under one
+// or more contributor license agreements. Licensed under the Elastic License;
+// you may not use this file except in compliance with the Elastic License.
+
+//go:build !integration
+// +build !integration
```
michel-laterman, line 6: (nitpick) I don't think we need the `// +` directives

michel-laterman (approved):
lgtm; the only other thing I can think of is to add instructions in the readme's dev section on how to run Kibana so that you can run a stand-alone fleet-server.

Dockerfile
jsoriano: I have replace the build command in the Dockerfile to use `make release-linux/amd64` instead, so we build with the same options as other release builds.

We will have to revisit this if we support other OSs or architectures.

internal/pkg/policy/standalone_test.go
jsoriano: :+1: Removed from all files.

jsoriano:
> It would be nice to add the example command to start to the README.

> the only other thing I can think of is to add instructions in the readme's dev section on how to run Kibana so that you can run a stand-alone fleet-server.

Added docs about building the docker image for standalone and about the experimental flag to be able to use the Fleet UI. @juliaElastic @michel-laterman please take another look.

juliaElastic (approved):
LGTM, thanks for adding the commands to the README.

internal/pkg/server/fleet.go
```diff
@@ -406,22 +405,24 @@ func (f *Fleet) runServer(ctx context.Context, cfg *config.Config) (err error) {
 func (f *Fleet) runSubsystems(ctx context.Context, cfg *config.Config, g *errgroup.Group, bulker bulk.Bulk, tracer *apm.Tracer) (err error) {
 	esCli := bulker.Client()
 
-	// Check version compatibility with Elasticsearch
-	remoteVersion, err := ver.CheckCompatibility(ctx, esCli, f.bi.Version)
-	if err != nil {
-		if len(remoteVersion) != 0 {
-			return fmt.Errorf("failed version compatibility check with elasticsearch (Agent: %s, Elasticsearch: %s): %w",
-				f.bi.Version, remoteVersion, err)
+	if !f.standAlone {
```
joshdover, line 408: Can we add a comment that explains why the version check and migrations are disabled for standalone mode (for now)?

README.md
```diff
@@ -84,6 +112,11 @@ More instructions for setup can be found in the [Elastic Stack Installation Guid
 
 #### fleet-server stand alone
 
+Fleet UI requires a managed Fleet Server, to be able to use stand alone Fleet
+server, you need to enroll a managed Fleet Server or disable this requirement.
+You can disable this requirement since Kibana 8.8.0, starting it with
+`xpack.fleet.enableExperimental: ['fleetServerStandalone']`.
```
joshdover, line 118:
```suggestion
Fleet UI requires a managed Fleet Server, to be able to use stand alone Fleet
server, you need to enroll a managed Fleet Server or disable this requirement.
You can disable this requirement since Kibana 8.8.0, starting it with
`xpack.fleet.enableExperimental: ['fleetServerStandalone']`. This is only
supported internally and is not intended for end-users at this time.
```

joshdover (commented):
We should make it very clear in migration.go that no new migrations should be added going forward. Even better if we make the application crash if a new one is added

internal/pkg/server/fleet.go
jsoriano: Added.

PR #2360, opened 2023-02-21
Upload route rate limits

pzl:

This PR re-adds limits to upload file routes

Conservative limits are placed on each of the routes

- start/end upload are not as expensive to call, not large byte payloads. Are generally 1:1 calls within a "file"
- chunk uploads are 100mb, and has its own rate limit to account for the difference in usage

## Design Checklist

**This PR specifically addresses these checks here**

## Checklist

~~- [ ] I have added an entry in `CHANGELOG.next.asciidoc` or `CHANGELOG-developer.next.asciidoc`.~~
This is a followup of a new feature that has a changelog entry for this release


aleksmaus (approved):
LGTM
Probably would separate into 3 limits parameters and 3 separate stats: uploadBegin uploadChunk, uploadComplete.
Even though it's the same route for two of these, there is precedent with ```/api/fleet/agents```

```
	RouteEnroll         = "/api/fleet/agents/:id"
	RouteCheckin        = "/api/fleet/agents/:id/checkin"
	RouteAcks           = "/api/fleet/agents/:id/acks"
```
, could default 2 of the limits to the same values.

internal/pkg/config/env_defaults.go
```diff
@@ -50,6 +50,21 @@ const (
 	defaultStatusBurst    = 25
 	defaultStatusMax      = 50
 	defaultStatusMaxBody  = 0
+
+	defaultUploadStartInterval = time.Second * 3
+	defaultUploadStartBurst    = 8
+	defaultUploadStartMax      = 3
+	defaultUploadStartMaxBody  = 1024 * 1024 * 5
```
michel-laterman, line 57: Why so large?

michel-laterman (approved):

internal/pkg/config/env_defaults.go
pzl: this was a shot in the dark --

The file start payloads describe any metadata about the file to be stored. The mapping is dynamic, so any integrations/clients can send what fields they are interested in, which could potentially be large. _This_ large, I have no idea. But I also didn't have a case for necessitating it smaller for the unknown payloads

PR #2371, opened 2023-02-22
[Fleet] Tool to build cloud image

nchaulet:
## Descrption 

Related to https://github.com/elastic/ingest-dev/issues/1421

That PR allow to build a custom docker image and push it to be able to test on cloud.

Allow you to build a custom fleet server and run it in cloud.

```
make build-and-push-cloud-image
```

Than you can use the docker image in a cloud deployment or a perf test.


dev-tools/cloud/docker/build.sh
```diff
@@ -0,0 +1,41 @@
...
+# flags (e.g. -t <name>) will be passed to `docker build`.
+
+set -eu
+
+REPO_ROOT=$(cd $(dirname $(readlink -f "$0"))/../../.. && pwd)
+
+USER_NAME=${USER}
+CI_ELASTIC_AGENT_DOCKER_IMAGE=docker.elastic.co/observability-ci/elastic-agent
+
+DEFAULT_IMAGE_TAG=8.8.0-681a23b0-SNAPSHOT
```
kpollich, line 14: Should this be left in place? Seems like we might need to update this over time to prevent accidentally deploying outdated versions by not specifying an argument. Maybe better to just not have a default here?

nchaulet: it make things easier to use, passing and I think it's okay if the agent image is not always up to date as long the fleet server one, but we should probably update it sometimes. WDYT?

kpollich: Yeah let's just try not to let it fall too out of date that it's annoying or confusing to folks using this script.

kpollich (approved):

michel-laterman (approved):
Thanks! this should make our testing much easier

Issue #1596, opened 2022-06-21
[Fleet]: Duplicate tags are displayed under Agents tab when agent is installed with same tags during installation.


amolnater-qasource:
**Kibana version:**  8.3 BC5 Kibana cloud environment

**Host OS and Browser version:** All, All

**Build details:**
```
VERSION: 8.3.0
BUILD: 53457
COMMIT: ba70bdf9fe5d28e832c40b7dbdade1e582133af1
```

**Preconditions:**
1. 8.3 BC5 Kibana cloud environment should be available.

**Steps to reproduce:**
1. Login to Kibana environment.
2. Install Agent with `--tag=hey,hey,hey,hey,hey`.
3. Navigate to Fleet>Agents tab.
4. Observe agent is installed with duplicate tags.
5. Click `Tags` filter and observe duplicacy is removed under this filter.
`hey` tag is only available once.

**Expected Result:**
Duplicate tags should not be displayed under Agents tab when agent is installed with same tags during installation.
As duplicacy is managed under tags dropdown filter, it should also be managed for table under Agents tab.

**Screenshot:**
![8](https://user-images.githubusercontent.com/77374876/174742669-f42d156d-ec77-4e46-bed8-82518e8c948e.JPG)

PR #2386, opened 2023-02-27
Remove agent tag duplicates

jillguyonnet:

When an Elastic Agent is enrolled with the same tag multiple tags, they are treated as separate tags.
Issue: https://github.com/elastic/fleet-server/issues/1596

This PR adds a helper method to remove duplicate tags during Elastic Agent enrollment.

1. Run fleet-server locally with these changes.
2. Enroll a new Elastic Agent with duplicate tags. For instance, if running elastic-agent locally using Docker, pass the flag `-e ELASTIC_AGENT_TAGS=foo,foo,foo,foo`.
3. The duplicated tag should only appear once in the new Agent.

## Design Checklist

## Checklist

## Related issues

- Closes [#123](https://github.com/elastic/fleet-server/issues/1596)

### DIFF
--- a/internal/pkg/api/handleEnroll.go
+++ b/internal/pkg/api/handleEnroll.go
@@ -199,7 +199,7 @@ func (et *EnrollerT) _enroll(
 			LocalMetadata:        agentData.LocalMetadata,
 			PolicyId:             agentData.PolicyID,
 			Status:               "online",
-			Tags:                 removeDuplicateStr(agentData.Tags),
+			Tags:                 agentData.Tags,
 			Type:                 agentData.Type,
 			UserProvidedMetadata: agentData.UserProvidedMetadata,
 		},
--- a/internal/pkg/api/handleEnroll.go
+++ b/internal/pkg/api/handleEnroll.go
@@ -13,6 +13,7 @@ import (
 	"net/http"
 	"time"
 
+	"github.com/elastic/elastic-agent-libs/str"
 	"github.com/elastic/fleet-server/v7/internal/pkg/apikey"
 	"github.com/elastic/fleet-server/v7/internal/pkg/bulk"
 	"github.com/elastic/fleet-server/v7/internal/pkg/cache"
@@ -211,17 +212,10 @@ func (et *EnrollerT) _enroll(
 	return &resp, nil
 }
 
+// Helper function to remove duplicate agent tags.
+// Note that this implementation will also sort the tags alphabetically.
 func removeDuplicateStr(strSlice []string) []string {
-	allKeys := make(map[string]bool)
-	list := []string{}
-	for _, item := range strSlice {
-		if _, value := allKeys[item]; !value {
-			allKeys[item] = true
-			list = append(list, item)
-		}
-	}
-
-	return list
+	return str.MakeSet(strSlice...).ToSlice()
 }
 
 func deleteAgent(ctx context.Context, zlog zerolog.Logger, bulker bulk.Bulk, agentID string) error {
--- a/internal/pkg/api/handleEnroll.go
+++ b/internal/pkg/api/handleEnroll.go
@@ -175,7 +175,7 @@ func (et *EnrollerT) _enroll(
 			ID:      agentID,
 			Version: ver,
 		},
-		Tags: req.Metadata.Tags,
+		Tags: removeDuplicateStr(req.Metadata.Tags),
 	}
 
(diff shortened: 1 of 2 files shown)


internal/pkg/api/handleEnroll.go
```diff
@@ -211,6 +211,19 @@ func (et *EnrollerT) _enroll(
 	return &resp, nil
 }
 
+func removeDuplicateStr(strSlice []string) []string {
```
michel-laterman, line 214: looks good, just for completeness, can you add a unit test for this?

jillguyonnet:
@michel-laterman I've added a unit test for the helper function. Can you review please?

internal/pkg/api/handleEnroll.go
```diff
@@ -211,6 +211,19 @@ func (et *EnrollerT) _enroll(
...
+	allKeys := make(map[string]bool)
```
ycombinator, line 215: Nit: `map[string]struct{}` will use less memory than `map[string]bool`.

https://dave.cheney.net/2014/03/25/the-empty-struct

internal/pkg/api/handleEnroll.go
```diff
@@ -175,7 +175,7 @@ func (et *EnrollerT) _enroll(
 			ID:      agentID,
 			Version: ver,
 		},
-		Tags: req.Metadata.Tags,
+		Tags: removeDuplicateStr(req.Metadata.Tags),
```
ycombinator, line 178: You may also want to consider reusing [`str.Set`](https://github.com/elastic/elastic-agent-libs/blob/f17cb1de6337603f16629d4754bcf8c27e0cc8be/str/stringset.go#L71) from the `elastic-agent-libs` module.  Normally I don't think we'd want to depend on another library _just_ for such simple functionality but since we are [already importing `elastic-agent-libs`](https://github.com/elastic/fleet-server/blob/fe0ec1e969018df23bb8d577aa91490324637722/go.mod#L10), I think it might make sense to reuse functionality from it here.

michel-laterman (approved):

internal/pkg/api/handleEnroll.go
jillguyonnet: Thanks for this suggestion @ycombinator - I see we can reduce the helper function to a one-liner using
```
return str.MakeSet(strSlice...).ToSlice()
```
but this would have the side effect of sorting the tags alphabetically, thus changing the current behaviour. I am not sure this would be desirable. Unless you had another suggestion in mind?

internal/pkg/api/handleEnroll.go
jillguyonnet: Thanks for the link! I'm not sure about the readability vs. memory gain tradeoff. @michel-laterman any preference?

ycombinator: I'm not sure there's a readability tradeoff.  Using `map[string]struct{}` is fairly common usage in Golang IME.  For example, there are plenty of occurrences in the Beats repo: https://github.com/search?q=repo%3Aelastic%2Fbeats%20map%5Bstring%5Dstruct%7B%7D&type=code

internal/pkg/api/handleEnroll.go
ycombinator: Good point re: sorting and how that would change the existing behavior.  Reading through the [documentation on tags](https://www.elastic.co/guide/en/fleet/current/filter-agent-list-by-tags.html) and the [original issue that requested the tags feature](https://github.com/elastic/elastic-agent/issues/149), it seems that tags are agent metadata and order is not significant.  

But it'll be good to get confirmation about this from someone more experienced with the project.  @michel-laterman @blakerouse, WDYT? (both of you reviewed the [original PR](https://github.com/elastic/elastic-agent/pull/336) introducing tags during Agent enrollment)

michel-laterman: I don't think the order matters for tags

jillguyonnet: 👍 If the order doesn't matter, then I can change the implementation to the one above.

internal/pkg/api/handleEnroll.go
jillguyonnet: Thanks, I'm very new to go so these comments are super helpful. I'll make the change with the implementation using Set from your other comment unless there are objections to it.

PR #2412, opened 2023-03-07
mergify: delete branches created by the updatecli, update backports

v1v:

Remove closed branches created by the automation

Remove backports to old 8.x branches but 8.6 and 8.7
Delete updatecli branches when PRs were merged/closed

### DIFF
--- a/.mergify.yml
+++ b/.mergify.yml
@@ -33,34 +33,6 @@ pull_request_rules:
         message: |
           This pull request has been automatically closed by Mergify.
           There are some other up-to-date pull requests.
-  - name: backport patches to 8.2 branch
-    conditions:
-      - merged
-      - base=main
-      - label=backport-v8.2.0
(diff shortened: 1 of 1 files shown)


.mergify.yml
```diff
@@ -127,36 +99,10 @@ pull_request_rules:
         - closed
       - and:
         - label=automation
-        - head~=^update-.*-version
+        - head~=^updatecli.*
```
v1v, line 102: https://github.com/elastic/fleet-server/blob/main/.ci/bump-elastic-stack-snapshot.yml and https://github.com/elastic/fleet-server/blob/main/.ci/bump-golang.yml are the ones creating branches in the upstream with the name `updatecli`, for instance https://github.com/elastic/fleet-server/pull/2408

michel-laterman (approved):

PR #2415, opened 2023-03-07
Log listening address

lucabelluccini:
Make sure we log the Fleet Server listening address as INFO instead of DEBUG.

Unable to verify without tests which is the listening address & port.

Logs the information as INFO

Run the Fleet Server


michel-laterman (approved):
lgtm

michel-laterman:
@lucabelluccini,  should this be backported for the 8.7.1 release?

PR #2437, opened 2023-03-20
Fix missing Mock for MUpdate in benchmark tests

alexsapran:

The benchmark test for the benchmarkBulk was missing the mock for the MUpdate method and thus it was constantly failing with a panic.

```bash
...
panic:
assert: mock: I don't know what to return because the method call was unexpected.
....
```

This commit adds the required mock.on to make the test pass.

From the main branch try running 
```bash
go test -bench=. -count=8 -run=Bench  ./... 
```

It will fail with a panic 

```bash
.....
BenchmarkBulkFlush_1
panic: 
assert: mock: I don't know what to return because the method call was unexpected.
	Either do Mock.On("MUpdate").Return(...) first, or remove the MUpdate() call.
.....
```

Then using the code in this PR try re-running the same benchmark test, it should complete. 

The focus of this PR is not to make the benchmark test better or change the code, but to fix a broken test.


jlind23 (approved):

pchila (approved):
LGTM

PR #2441, opened 2023-03-21
Add benchmark and benchstat make targets

alexsapran:

This commit adds 2 new Makefile targets for running go benchmarks.

The first `make benchmark` will execute all the go benchmarks and output them into a file under `build/benchmark-{git_sha}.out`. This is useful to run before making any change to establish a baseline measurement.

After making the changes you can rerun the `make benchmark` providing the `BENCH_BASE` variable to write into a different file, if you already have committed your changes you don't need to set the variable as it will get a different git-sha.

Once you have the 2 benchmark outputs you can compare them with benchstats as follows.

`make benchstat BENCH_BASE=benchmark-baseline.out BENCH_NEXT=benchmark-next.out` and it will parse the output and show you a comparison between the baseline and your changes.

This PR add 2 (3 with the install-benchstat, which is not really used in a normal workflow, its a dependency) new Makefile targets useful for running the go benchmarks

After checking out this PR you can run 

- `make benchmark BENCH_BASE=test-base.out` and it will execute the go benchmarks 
- Run one more time but different output `make benchmark BENCH_BASE=test-diff.out` 
- Run `make benchstat BENCH_BASE=test-base.out BENCH_NEXT=test-diff.out` and you will get a report of differences between the 2 runs.


Makefile
```diff
@@ -14,6 +14,13 @@ BUILDMODE_darwin_arm64=-buildmode=pie
 
 BUILDER_IMAGE=docker.elastic.co/observability-ci/fleet-server-builder:latest
 
+#Benchmark related targets
+BENCH_BASE ?= benchmark-$(COMMIT).out
+BENCH_NEXT ?=
+BENCHMARK_ARGS := -count=8
+BENCHMARK_PACKAGE ?= ./...
+BENCHMARK_FILTER ?= Bench#
```
michel-laterman, line 22: please add a comment to state that the default for this filter is to only run `Benchmark` test cases in test files and exclude any unit tests

michel-laterman (commented):
Can you add a section to the README about running benchmarks and using the new make targets?

Makefile
alexsapran: Thanks for the review @michel-laterman, 
I pushed a change, let me know if you are ok with it.

michel-laterman (approved):
LGTM

PR #2448, opened 2023-03-22
[Fleet] Fleet server standalone should be healthy before .fleet-policies is created

### LINKED ISSUE #2421: [Fleet] Fleet server standalone not healthy until `.fleet-policies` exists
## Descripton 

Fleet server in standalone mode is stucked with status `STARTING` until the `.fleet-policies` exists. I think we probably should have an healthy fleet server as long an ES connection exists.

It seems it may be problematic with the policy coordinator and the unused leader election mechanism.

nchaulet:
## Description 

Resolve #2421 

Mark Fleet server as healthy in standalone even if the `.fleet-policies` index is not yet created.

## Tests

I updated the unit test.

I manually tested this like this:
1. With a fresh ES and Kibana
2. Create a service token
```
curl -X POST http://localhost:9200/_security/service/elastic/fleet-server/credential/token/test -u elastic:changeme
```
3.  Run fleet server in standalone
```shell
ELASTICSEARCH_CA_TRUSTED_FINGERPRINT=notused ELASTICSEARCH_SERVICE_TOKEN=<TOKEN_CREATED_BEFORE> ELASTICSEARCH_HOSTS=http://localhost:9200  go run main.go
```
4. Get the status and check it's healthy `http GET http://127.0.0.1:8220/api/status`
5. Create a policy in Kibana and  it's still healthy `http GET http://127.0.0.1:8220/api/status`

### DIFF
--- a/internal/pkg/policy/standalone.go
+++ b/internal/pkg/policy/standalone.go
@@ -92,7 +92,7 @@ func (m *standAloneSelfMonitorT) check(ctx context.Context) client.UnitState {
 	if err != nil {
 		if errors.Is(err, es.ErrIndexNotFound) {
 			m.log.Debug().Str("index", m.policiesIndex).Msg(es.ErrIndexNotFound.Error())
-			return m.updateState(client.UnitStateStarting, "Policies not available yet")
+			return m.updateState(client.UnitStateHealthy, "Running: Policies not available yet")
 		}
 		if err != nil {
 			return m.updateState(client.UnitStateFailed, fmt.Sprintf("Failed to request policies: %s", err))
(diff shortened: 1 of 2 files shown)


juliaElastic (approved):
LGTM
How can we detect if `.fleet-policies` creation fails? Is the state going to be updated again in that case?

joshdover (commented):
Is this error even useful in managed mode? Does the upstream code handle retrying until the index is ready?

nchaulet:
> Is this error even useful in managed mode? Does the upstream code handle retrying until the index is ready?

@joshdover in managed mode Fleet server will need to access `.fleet-policies` to retrieve his own policy so I think in this case it make sense to have Fleet not ready until the index exists.

joshdover (approved):

michel-laterman (approved):
lgtm

PR #2449, opened 2023-03-23
Add github workflow for running go benchmarks on pull requests

alexsapran:

This commit adds a new GitHub workflow that will trigger the benchmark workflow introduced by https://github.com/elastic/fleet-server/pull/2441 and report back the diff between the head of the pull_request and the base branch. 

This will be going to be executed for every pull request. An example of how it will look can be seen https://github.com/alexsapran/fleet-server/pull/1

We are going to be running the go microbenchmarks and report the difference for each PR. This allows the reviewers and the author to review quickly if the changes are actually adding any regression to the code (given that there is an existing test for it)
We would be able to start writing benchmark tests and be part of our CI pipeline.

I am not sure there is a way to test this PR locally, as it's a GitHub workflow action.

FYI, given that this is something that is going to be part of every pull request. 

@amitkanfer, @joshdover , @jlind23 , @jen-huang, @kpollich 

If we see that this is causing issues, we can disable it.


jlind23 (approved):

alexsapran:
I removed the conditional local as the benchmark test is not a required check for a PR.

PR #2450, opened 2023-03-23
Add comments and unit test for gathering variables from env var

michel-laterman:
added a really simple unit test and comments to explicitly state how fleet-server handles env vars


nchaulet (approved):
LGTM

PR #2459, opened 2023-03-29
Add support for gathering secret values from files.

michel-laterman:

Allow users to specify secret-files to provide the the values for:
- output.elasticsearch.service_token
- input[0].server.instrumentation.api_key
- input[0].server.instrumentation.secret_token
- input[0].server.ssl.key_passphrase

by using the same attribute with the `_file` suffix, for example `service_token_file`

This will allow fleet-server to run without a reference to a secret in a command line arg or env var.
We assume that the secrets will be provisioned through some outside means.

Add support for gathering secret values from files. The secret values are:
- output.elasticsearch.service_token
- input[0].server.instrumentation.api_key
- input[0].server.instrumentation.secret_token

Removed the output.elasticsearch.api_key option as configuration using it has been rejected by fleet-server for two years.

Update to elastic-agent-libs to version that adds support for specifying key passphrases through files.
The elastic-agent-libs support for certificates already includes gathering certs from files.

Instrumentation configuration has changed to remove as many settings through env vars as it could and instead provide a more complete APMHTTPTransportOptions struct when initializing the tracer.

## Design Checklist

config changes

## Checklist


internal/pkg/config/output.go
```diff
@@ -56,8 +63,15 @@ func (c *Elasticsearch) InitDefaults() {
 
 // Validate ensures that the configuration is valid.
 func (c *Elasticsearch) Validate() error {
-	if c.APIKey != "" {
-		return fmt.Errorf("cannot connect to elasticsearch with api_key; must use service_token")
+	if c.ServiceToken == "" {
```
michel-laterman, line 66: Our validation will now ensure that there is some service_token value; either directly specified, or from a non-empty file.

internal/pkg/config/output.go
```diff
@@ -56,8 +63,15 @@ func (c *Elasticsearch) InitDefaults() {
...
+		if c.ServiceTokenFile == "" {
+			return fmt.Errorf("service_token is undefined")
+		}
+		if p, err := os.ReadFile(c.ServiceTokenFile); err != nil {
+			return fmt.Errorf("unable to read service_token_file: %w", err)
+		} else if len(p) == 0 {
+			return fmt.Errorf("empty service_token_file")
+		}
```
nchaulet, line 74: Does it make sense to populate `c.ServiceToken` here to avoid rereading the file later?

internal/pkg/config/output.go
nchaulet: nitpick :does it make sense to create a function here for the serviceToken validation it start to be a few if imbrication

nchaulet (approved):
some nitpick questions, but otherwise I tested this locally with a `service_token_file` and it work as expected 🚀

internal/pkg/config/output.go
michel-laterman: I don't think this complex enough

internal/pkg/config/output.go
michel-laterman: I'm hesitant to do so as then our validation method may have a state change.
Validate is only called when new config is received (when running under the agent) or on startup. I don't think the potential 2nd file read impacts performance in a significant way

PR #2472, opened 2023-04-05
Add action APM metadata

juliaElastic:

Help debugging agent actions.

Adding APM metadata to actions, see https://github.com/elastic/fleet-server/pull/1794

## Design Checklist

## Checklist

## Related issues

- Reviving Josh's pr by resolving conflicts with main: https://github.com/elastic/fleet-server/pull/1794
- Relates https://github.com/elastic/ingest-dev/issues/1489


internal/pkg/api/handleCheckin.go
```diff
@@ -234,27 +237,54 @@ func (ct *CheckinT) ProcessRequest(zlog zerolog.Logger, w http.ResponseWriter, r
...
 		Actions:  &actions,
 	}
 
-	return ct.writeResponse(zlog, w, r, resp)
+	return ct.writeResponse(zlog, w, r, agent, resp)
 }
 
-func (ct *CheckinT) writeResponse(zlog zerolog.Logger, w http.ResponseWriter, r *http.Request, resp CheckinResponse) error {
+func (ct *CheckinT) writeResponse(zlog zerolog.Logger, w http.ResponseWriter, r *http.Request, agent *model.Agent, resp CheckinResponse) error {
+	// TODO: only process this if we have an apm tracer
```
juliaElastic, line 250: @joshdover What do you mean by this?

joshdover: It would be possible to avoid doing this extra processing to construct if APM is not configured in Fleet Server. It would require detecting if the `tracer` created here is `nil`: https://github.com/elastic/fleet-server/blob/6b18cfd65ddf7bdcd30fb804e6e1d7e642e11a65/internal/pkg/server/fleet.go#L346

juliaElastic: added the condition in the latest commits

juliaElastic:
Hi! Can I get a review on this?

AndersonQ (approved):

PR #2475, opened 2023-04-05
Fix APM dependency

michel-laterman:
Fix the APM instrumentation import; use `apmchiv5` instead of `apmchi`.
fleet-server will panic if instrumentation is used without this change


michel-laterman:
it's a bug fix for an unreleased bug, i don't think this needs a changelog entry

PR #2476, opened 2023-04-06
Rename secret file attributes to use path suffix

michel-laterman:
Rename secret files attributes to use `_path` suffix to be consistent with existing naming conventions in the elastic-agent


nchaulet (approved):
🚀

PR #2482, opened 2023-04-10
Update go to v1.20.3

michel-laterman:
Update go to v1.20.3

Replace use of `math/rand` with `crypto/rand` in test as it is now deprectated


michel-laterman:
As discussed in https://github.com/elastic/fleet-server/issues/2337 we should update to go 1.20 to get access to new stdlib features. 
This should also replace https://github.com/elastic/fleet-server/pull/2478

kpollich (approved):

PR #2491, opened 2023-04-11
Add poll_timeout attribute to checkin endpoint

### LINKED ISSUE #2337: Increase long poll default to ~30 minutes

michel-laterman:

Allow clients to request longer polls for checkin requests.
This results in better performance in larger deployments

The approach is backwards compatible as no defaults are changed.
Changing the default times to larger values may result in older agents requests timing out.

Add a `poll_duration` string attribute to checkin request bodies that inform's fleet-server of the client's poll duration. And a `checkin_max_poll` config option to enforce an upper limit to requested times.

If present the `poll_duration` string is parsed into a `time.Duration`, if the result is a non-zero value fleet-server will use the provided `poll_duration` to set it's poll-time and write deadline for the request

Fleet-server will set it's poll time to `min(max(1m, poll_duration-2m), checkin_max_poll)` before doing the normal setup/jitter offsets.
And the request's write deadline will be set to `max(2m, poll time + 1m)`.
This leads to the min values a poll duration of 1m and the write timeout to 2m.
If `checkin_max_poll` is set to less then 1m it is effectively ignored and a 1m value is used.

The min accepted `poll_time` from the client is 3m, and the max is `checkin_max_poll`-2m

If `poll_duration` is not set (or parses to a zero value) the configured checkin values will be used instead (defaults to a 5m poll and 10m write deadline)

If `poll_duration` cannot be parsed an error is returned.

## Design Checklist

- ~~I have or intend to scale test my changes, ensuring it will work reliably with 100K+ agents connected.~~ will require client changes to test at scale

## Checklist

## Related issues

- Closes #2337


internal/pkg/api/handleCheckin.go
```diff
@@ -108,26 +108,51 @@ func (ct *CheckinT) handleCheckin(zlog zerolog.Logger, w http.ResponseWriter, r
...
+		}
+	}
+
+	pollDuration := ct.cfg.Timeouts.CheckinLongPoll
+	// set the pollDuration if pDur parsed from poll_timeout was a non-zero value
+	// sets timeout to max(1m, pDur-2m)
+	// sets the response write timeout to max(2m, pDur-1m)
+	if pDur != time.Duration(0) {
+		pollDuration = pDur - (2 * time.Minute)
+		if pollDuration < time.Minute {
```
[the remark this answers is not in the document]
blakerouse: I agree, please add a configuration setting with an upper limit.

blakerouse (approved):
Overall this looks really good. I agree with @jsoriano on adding an upper limit.

model/openapi.yml
```diff
@@ -255,6 +255,13 @@ components:
           type: string
           format: application/json
           x-go-type: json.RawMessage
+        poll_timeout:
+          description: |
+            An optional timeout value that informs fleet-server of when a client will time out on it's checkin request.
+            If not specified fleet-server will use the timeout values specified in the config (defaults to 5m polling and a 10m write timeout).
+            If specified fleet-server will set its poll timeout to `max(1m, poll_timeout-2m)` and its write timeout to `max(2m, poll_timout-1m)`.
```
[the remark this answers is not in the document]
michel-laterman: We want to have the write timeout after poll timeout so that the fleet-server still has time to write the response for a long running poll.

internal/pkg/config/timeouts.go
```diff
@@ -58,4 +59,9 @@ func (c *ServerTimeouts) InitDefaults() {
 
 	// Jitter subtracted from c.CheckinLongPoll.  Disabled if zero.
 	c.CheckinJitter = 30 * time.Second
+
+	// MaxPoll is the maximum allowed value for a long poll when the client specified poll_timeout value is used.
+	// The long poll value is poll_timeout-2m, and the request's write timeout is set to poll_timeout-1m
+	// CheckinMaxPoll values of less then 1m are effectively ignored and a 1m limit is used.
+	c.CheckinMaxPoll = 2 * time.Hour
```
michel-laterman, line 66: @blakerouse @jsoriano added `checkin_max_poll` to timeouts. It only applies to client requested `poll_timeout` values (it does not effect `checkin_long_poll`).
I've extended the config unit test and smoke integration test to verify it works as expected.

Open to feedback for the default value, not sure if we want to use 2h or not

go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/elastic/fleet-server/v7
 
-go 1.19
+go 1.20
```
jsoriano, line 3: Should this change go on its own PR?

internal/pkg/config/timeouts.go
jsoriano: Looks good to me, maybe we can start with a lower upper limit. Reducing it later may be breaking.

Should we add this new setting to the documentation?

jsoriano (approved):
Nice :+1:

internal/pkg/config/timeouts.go
blakerouse: 2h seems high, I agree that lowering it would become breaking where raising it would not. So I would start at 1 hour, being its a setting it can be adjusted.

blakerouse (approved):
Looks good!

jlind23:
@blakerouse do you think we can use the integration fwk you are working on in order to test this change? Two tests one where the checkin body contains the new field and one where it does not contain it would be great.

PR #2496, opened 2023-04-12
Fix issue with newlines when reading secret files

michel-laterman:
Fix an issue that didn't occur in unit tests, but occurred in a real end to end test case where os.ReadFile would attach a newline to secret files.

Running under the agent and in stand-alone mode showed an issue where secret file values were being incorrectly returned with a newline character at the end.

For example when the service_token was incorrctly read the error message was:
```
21:45:39.813 ERR fail elasticsearch info error.message="net/http: invalid header field value for \"Authorization\"" cluster.addr=["https://192.168.4.20:9200"] cluster.maxConnsPersHost=128 ecs.version=1.6.0 service.name=fleet-server
```

To recreate:
```
echo MY-SERVICE-TOKEN > /tmp/service-token
```
In fleet-server.yml:
```
output.elasticsearch:
  ...
  service_token_path: /tmp/service-token
```

### DIFF
--- a/internal/pkg/config/instrumentation.go
+++ b/internal/pkg/config/instrumentation.go
@@ -11,6 +11,7 @@ import (
 	"fmt"
 	"net/url"
 	"os"
+	"strings"
 
 	"github.com/elastic/elastic-agent-libs/transport/tlscommon"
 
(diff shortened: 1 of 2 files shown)


michel-laterman:
`echo -n STRING` does what I expected;
however this may be something we want to implement in order to make it easier for a user

PR #2513, opened 2023-04-19
[8.7] Bump go version to 1.19.8

cmacknz:
What the summary says. Main is already at Go 1.20 so this targets 8.7.


cmacknz:
Depends on https://github.com/elastic/golang-crossbuild/pull/282

pierrehilbert (approved):

Issue #2308, opened 2023-01-31
Automatically choose a default load limit present based on available memory


joshdover:
Related to:
- https://github.com/elastic/fleet-server/issues/1860
- https://github.com/elastic/fleet-server/issues/1857

We want to make scaling Fleet Server easy and stable, for both end users and scale testing purposes. Automatically configuring as much as possible will help our scale tests to more accurately reflect what a typical user will see in production when using Fleet Server on Cloud or on-premise.

In addition, this configuration setting is hard to set because it's on a per-instance basis, while customers tend to think about the size of their entire Fleet. We have no way to automatically determine how many instances are running in order to calculate the per-instance agent count.

### Requirements
- When "max agents" config is not set, Fleet Server should be able to use most of available resources in order to maximize the number of agents it can serve and how 
- When "max agents" is set, we should not automatically calculate any load limits. The purpose of this is to either:
	1. Support mixed APM Server / Fleet Server uses cases on Cloud where these processes share a single container. If the user has high APM usage, they may desire to throttle the amount of resources (mostly memory) that Fleet Server will use and this is the only knob we currently have.
	2. If Fleet is putting too much load on Elasticsearch, this knob is an easy way to turn down that traffic without having to set each individual rate limit setting.
- We should not assume that the entire memory space is available just for Fleet Server, because we need to be sure that we can run a minimal APM Server, Metricbeat, and Filebeat in the same container for metric and logs collection. We should ensure that the presets already take this into account.

michel-laterman:
This should not be too complex to do, we can alter our limit loading to work on mem size if agents is not set:
https://github.com/elastic/fleet-server/blob/25be4e26e479b0263d4499ecf8123bf082901a88/internal/pkg/config/env_defaults.go#L187-L204

PR #2514, opened 2023-04-19
added memory limit config

juliaElastic:

Add memory limit config to set a soft limit: https://pkg.go.dev/runtime/debug#SetMemoryLimit

This helps prevent OOM errors for quick agent actions on scale: https://weaviate.io/blog/gomemlimit-a-game-changer-for-high-memory-applications

## Design Checklist

## Checklist

## Related issues

- Relates https://github.com/elastic/ingest-dev/issues/1674


changelog/fragments/1681906741-memory_limit.yaml
```diff
@@ -0,0 +1,32 @@
...
+#description:
+
+# Affected component; a word indicating the component this changeset affects.
+component:
+
+# PR URL; optional; the PR number that added the changeset.
+# If not present is automatically filled by the tooling finding the PR where this changelog fragment has been added.
+# NOTE: the tooling supports backports, so it's able to fill the original PR number instead of the backport PR number.
+# Please provide it if you are adding a fragment for a different PR.
+pr: https://github.com/elastic/fleet-server/pull/2514
```
michel-laterman, line 28:
```suggestion
pr: 2514
```

internal/pkg/config/runtime.go
```diff
@@ -5,7 +5,8 @@
 package config
 
 type Runtime struct {
-	GCPercent int `config:"gc_percent"`
+	GCPercent   int   `config:"gc_percent"`
+	MemoryLimit int64 `config:"memory_limit"`
```
michel-laterman, line 9: Should we add a comment stating that this is set to `math.MaxInt64` by default; and setting this too low will result in a lot more GC behaviour (as outlined in the docs)?

juliaElastic:
> Is the plan for this setting to be set alongside our other limits when we specify the max agents count? Will we be setting the soft limit as a part of #2308?

We discussed to set the limit in the cloud stackpack to be 50% of the container max memory: https://github.com/elastic/ingest-dev/issues/1674#issuecomment-1512811360

I started setting this in manual perf test runs to see what is the effect until we implement addition to the stackpack.

internal/pkg/config/runtime.go
juliaElastic: updated

Issue #2517, opened 2023-04-20
[Buildkite] - Migrate the existing pipelines to Buildkite


alexsapran:
## Context 
We are moving to a new build system and we are going to migrate the existing Jenkins pipelines to Buildkite.

This issue is to track the progress of the migration

PR #2518, opened 2023-04-20
add catalog-info.yaml

SHarbuz:

This PR creates the catalog information for the BackStage internal system

This PR doesn't affect anything. The data for this PR from the Backstage system

## Related issues

Related to https://github.com/elastic/fleet-server/issues/2517

### DIFF
--- a/catalog-info.yaml
+++ b/catalog-info.yaml
@@ -1,0 +1,29 @@
+---
+# yaml-language-server: $schema=https://gist.githubusercontent.com/elasticmachine/988b80dae436cafea07d9a4a460a011d/raw/e57ee3bed7a6f73077a3f55a38e76e40ec87a7cf/rre.schema.json
+apiVersion: backstage.io/v1alpha1
+kind: Resource
+metadata:
+  name: buildkite-pipeline-fleet-server
+  description: Buildkite pipeline for the Elastic Fleet Server project
+  links:
+    - title: Pipeline
+      url: https://buildkite.com/elastic/fleet-server
+
+spec:
+  type: buildkite-pipeline
+  owner: group:ingest-fp
+  system: buildkite
+  implementation:
+    apiVersion: buildkite.elastic.dev/v1
+    kind: Pipeline
+    metadata:
+      name: fleet-server
+      description: Buildkite pipeline for the Elastic Fleet Server project
+    spec:
+      repository: elastic/fleet-server
+      pipeline_file: ".buildkite/pipeline.yml"
+      teams:
+        ingest-fp:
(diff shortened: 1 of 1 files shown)


alexsapran (approved):

catalog-info.yaml
```diff
@@ -0,0 +1,28 @@
+---
+# yaml-language-server: $schema=https://gist.githubusercontent.com/elasticmachine/988b80dae436cafea07d9a4a460a011d/raw/e57ee3bed7a6f73077a3f55a38e76e40ec87a7cf/rre.schema.json
+apiVersion: backstage.io/v1alpha1
+kind: Resource
+metadata:
+  name: buildkite-pipeline-fleet-server
+  description: Buildkite Pipeline for fleet-server
```
mrodm, line 7: Update this description based on the one set in the main Jenkins pipeline
https://github.com/elastic/fleet-server/blob/main/.ci/jobs/fleet-server.yml#L5


```suggestion
  description: Buildkite pipeline for the Elastic Fleet Server project

```

catalog-info.yaml
```diff
@@ -0,0 +1,28 @@
...
+
+spec:
+  type: buildkite-pipeline
+  owner: group:ingest-fp
+  system: buildkite
+  implementation:
+    apiVersion: buildkite.elastic.dev/v1
+    kind: Pipeline
+    metadata:
+      name: fleet-server
```
mrodm, line 20: Add the description too here:

```suggestion
      name: fleet-server
      description: Buildkite pipeline for the Elastic Fleet Server project
```

catalog-info.yaml
```diff
@@ -0,0 +1,28 @@
...
+    spec:
+      repository: elastic/fleet-server
+      pipeline_file: ".buildkite/pipeline.yml"
+      teams:
+        ingest-fp:
+          access_level: MANAGE_BUILD_AND_READ
```
mrodm, line 26: As in this https://github.com/elastic/package-registry/pull/992#discussion_r1173720764

Maybe inges-tfp should have just "BUILD_AND_READ"

> BUILD_AND_READ: can view and create builds but cannot edit pipeline settings.

But some team should have the permissions MANAGE_BUILD_AND_READ just in case @alexsapran WDYT?

alexsapran: I think that the ingest-fp should be able to manage and build, but I don't see a reason why we should lock down the job.

mrodm: Let's keep for the time being MANAGE_BUILD_AND_READ for ingest-fp team. If it is needed, I think it could be updated later

alexsapran (approved):

PR #2529, opened 2023-04-26
Add fleet-server.reference.yml

michel-laterman:
Add fleet-server.reference.yml with all possible config options


jsoriano (approved):

kpollich (approved):
This is great, thank you!

PR #2531, opened 2023-04-26
Make Internal Server Error default error

### LINKED ISSUE #2417: [Fleet Server] Change default error code to 500 instead of 400 to avoid masking bugs
When fleet server does not know what the error is, it will send a badRequest http error code to the user even if the problem is coming from fleet server itself. 
https://github.com/elastic/fleet-server/blob/main/internal/pkg/api/error.go#:~:text=//%20Default,%7D

We should change this and rather return an Http 500 by default

michalpristas:

This PR makes 500 a Default API error.
I also went through errors returned and made 'known errors' from most of them to keep BadRequest behavior for what we expect. This way it will be easier to spot unwanted behavior of API.

Please go through errors translation table and suggest changes if you feel that particular error should not be classified a `Bad Request`
## Checklist

Fixes #2417


michalpristas:
I tried to do this not in a breaking manner
if we specify BadRequest behavior somewhere it could be changed though, will take a look

joshdover:
In general I think we're still using 400 for too many things. Many of these should be things like 401, 403, and 413. I think the challenge is that agent may have some code that expects 400 in some cases to shut down on invalid API keys?

cmacknz:
> I think the challenge is that agent may have some code that expects 400 in some cases to shut down on invalid API keys?

We unenroll if we get too many Unauthorized errors:

Status to error conversion:
https://github.com/elastic/elastic-agent/blob/ceb3ca1d09c1a35a364f13a11e9a8e2643b861e2/internal/pkg/fleetapi/client/round_trippers.go#L54-L58
```go
	if resp.StatusCode == http.StatusUnauthorized {
		defer resp.Body.Close()
		return nil, ErrInvalidAPIKey
	}
```

Is Unauth helper:
https://github.com/elastic/elastic-agent/blob/ceb3ca1d09c1a35a364f13a11e9a8e2643b861e2/internal/pkg/agent/application/gateway/fleet/fleet_gateway.go#L374-L376
```
func isUnauth(err error) bool {
	return errors.Is(err, client.ErrInvalidAPIKey)
}
```

Unenrollment logic:
https://github.com/elastic/elastic-agent/blob/ceb3ca1d09c1a35a364f13a11e9a8e2643b861e2/internal/pkg/agent/application/gateway/fleet/fleet_gateway.go#L338-L347

```go
	resp, took, err := cmd.Execute(ctx, req)
	if isUnauth(err) {
		f.unauthCounter++

		if f.shouldUnenroll() {
			f.log.Warnf("retrieved an invalid api key error '%d' times. Starting to unenroll the elastic agent.", f.unauthCounter)
			return &fleetapi.CheckinResponse{
				Actions: []fleetapi.Action{&fleetapi.ActionUnenroll{ActionID: "", ActionType: "UNENROLL", IsDetected: true}},
			}, took, nil
		}
```

We unenroll after 6 failures:

https://github.com/elastic/elastic-agent/blob/ceb3ca1d09c1a35a364f13a11e9a8e2643b861e2/internal/pkg/agent/application/gateway/fleet/fleet_gateway.go#L27-L28

```go
// Max number of times an invalid API Key is checked
const maxUnauthCounter int = 6
```

This seems a little bit dangerous, but https://github.com/elastic/beats/pull/19722 explains it is to prevent unenrolled agents from continuing to contact Fleet server.

PR #2532, opened 2023-04-26
Disable Buildkite builds

alexsapran:

Set the appropriate settings for the build_pull* parameters so that we dont try and trigger anything.

Right now with onboarding to Backstage, the system provisioned a new Buildkite pipeline for the repository, this pipeline is not yet ready, so we should not trigger it. 
The trigger will happen through the API and not the automatic build trigger. 

No tests mechanism to locally test buildkite/backstage

## Related issues

- Relates #2517

### DIFF
--- a/catalog-info.yaml
+++ b/catalog-info.yaml
@@ -37,6 +37,9 @@ spec:
     spec:
       repository: elastic/fleet-server
       pipeline_file: ".buildkite/pipeline.yml"
+      provider_settings:
+        build_pull_request_forks: false
+        build_pull_requests: false # PR builds are managed by buildkite-pr-bot
       teams:
         ingest-fp:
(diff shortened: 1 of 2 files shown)


catalog-info.yaml
```diff
@@ -37,6 +37,9 @@ spec:
     spec:
       repository: elastic/fleet-server
       pipeline_file: ".buildkite/pipeline.yml"
+      provider_settings:
+        build_pull_request_forks: false
+        build_pull_requests: false # PR builds are managed by buildkite-pr-bot
```
alexsapran, line 42: the buildkite-pr-bot will be added shortly so the `are` for now is ok to be used in this context for now.

pierrehilbert (approved):

PR #2539, opened 2023-04-27
Add go-benchmark pipeline entity definition

alexsapran:

We have already a go-benchmark test running as part of the GH actions, but it does not work as we want it to be. 
For this reason, we are trying out Buildkite.

This commit adds the required definition for adding a new Buildkite
pipeline that would run the go-benchmark steps.

Signed-off-by: Alexandros, Sapranidis <alexandros@elastic.co>

This PR introduce a new Buildkite pipeline that is intended to run the go-benchmark tests. 

There is no local testing for running pipelines with Buildkite

### DIFF
--- a/.buildkite/pipeline.bennchmark.yml
+++ b/.buildkite/pipeline.bennchmark.yml
@@ -1,0 +1,4 @@
+# yaml-language-server: $schema=https://raw.githubusercontent.com/buildkite/pipeline-schema/main/schema.json
+steps:
+  - label: ":wave: Initial step"
+    command: echo "Hello!"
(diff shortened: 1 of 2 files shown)


mrodm (approved):
LGTM, if you plan to create PRs in forks, you will need to configure the buildkite-pr-bot

alexsapran:
> LGTM, if you plan to create PRs in forks, you will need to configure the buildkite-pr-bot

Good catch, I will work on adding the pr-bot and then merge

alexsapran:
not see myself having time to finish this up. Will raise a new one in the future.

PR #2543, opened 2023-04-27
add steps: check-ci, make-local, unit-test, int-test, annotate

sharbuz:

Migrated: check-ci, make-local, unit-test, int-test, and annotate

## Checklist

## Related issues

Related the issue: https://github.com/elastic/fleet-server/issues/2517

**Pipeline steps:**
- [x] Check -> Run check-ci
- [x] Local -> Run local build
- [x] Unit Test -> Run unit tests
- [x] Integration Test -> Run integration tests
- [x] Junit annotate
- [ ] Publish Docker images
- [ ] Release Test
- [ ] Downstream - Package


alexsapran:
We are waiting to sort out some details so, please hold reviewing this PR.

.buildkite/pipeline.yml
```diff
@@ -1,3 +1,65 @@
...
+          image: "golang:1.20.3"
+          cpu: "8"
+          memory: "4G"
+
+      - label: ":junit: Junit annotate"
+        plugins:
+          - junit-annotate#v2.4.1:
+              artifacts: "build/*.xml"
+        agents:
+          provider: "golang:1.20.3"
```
mrodm, line 59: This plugin requires to use `docker`. And for that, it is necessary to use agents with the provider `gcp`

.buildkite/scripts/setenv.sh
```diff
@@ -0,0 +1,28 @@
...
+
+echo "UNMET DEP: Installing docker-compose"
+
+DC_CMD="${HOME}/bin/docker-compose"
+
+mkdir -p "${HOME}/bin"
+
+curl -sSLo "${DC_CMD}" "https://github.com/docker/compose/releases/download/${DOCKER_COMPOSE_VERSION}/docker-compose-$(uname -s)-$(uname -m)"
+chmod +x "${DC_CMD}"
+echo -e "\nPATH="${HOME}/bin:${PATH}"" >> dev-tools/integration/.env
```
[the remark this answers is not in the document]
SHarbuz: the function `add_bin_pat` is working properly only in its own terminal. When the `make` command starts the exported PATH isn't accessible. That's why I use export PATH (https://github.com/elastic/fleet-server/blob/66df4f2e0f17fc499e31d64355f34586a16fa0e6/Makefile#L252-L253) through the `.env` file.

SHarbuz:
**Finished steps:**
Check -> Run check-ci
Local -> Run local build
Unit Test -> Run unit tests
Integration Test -> Run integration tests
Junit annotate

**Still need to finish:**
Publish Docker images
Release Test
Release - Package Registry Distribution
Downstream - Package
cleanup

.buildkite/pipeline.yml
```diff
@@ -1,3 +1,66 @@
+# yaml-language-server: $schema=https://raw.githubusercontent.com/buildkite/pipeline-schema/main/schema.json
+
+env:
+  SETUP_GVM_VERSION: 'v0.5.0'
+  DOCKER_COMPOSE_VERSION: '1.25.5'
+  ELASTIC_PACKAGE_COMPOSE_DISABLE_ANSI: "true"
+  KIND_VERSION: 'v0.17.0'
+  K8S_VERSION: 'v1.26.0'
+
```
mrodm, line 9: At least these 3 variables are not used. I would delete them

.buildkite/pipeline.yml
```diff
@@ -1,3 +1,66 @@
...
+      - label: "Run unit test"
+        key: unit-test
+        commands:
+          - "chmod +x .buildkite/scripts/setenv.sh"
+          - ".buildkite/scripts/setenv.sh with-go"
+          - "make test-unit"
+          - "make junit-report"
+          - "buildkite-agent artifact upload build/*.xml"
+        agents:
+          provider: "gcp"
```
mrodm, line 41: the last command set can be specified using `artifact_paths` key:

```suggestion
      - label: "Run unit test"
        key: unit-test
        commands:
          - "chmod +x .buildkite/scripts/setenv.sh"
          - ".buildkite/scripts/setenv.sh with-go"
          - "make test-unit"
          - "make junit-report"
        agents:
          provider: "gcp"
        artifact_paths:
          - build/*.xml
```

.buildkite/pipeline.yml
```diff
@@ -1,3 +1,66 @@
...
+
+
+      - label: "Run integration tests"
+        key: int-test
+        commands:
+          - "chmod +x .buildkite/scripts/setenv.sh"
+          - ".buildkite/scripts/setenv.sh with-go-docker-compose"
+          - "make test-int"
+          - "make junit-report"
+          - "buildkite-agent artifact upload build/*.xml"
```
mrodm, line 51: Same as in previous step, here it can be used the `artifact_paths` key from Buildkite.

.buildkite/pipeline.yml
```diff
@@ -1,3 +1,66 @@
...
+        agents:
+          provider: "gcp"
+
+      - label: ":junit: Junit annotate"
+        commands:
+          - "buildkite-agent artifact download 'build/*.xml' build/"
```
mrodm, line 57: Is it necessary to add this command ? At least with elastic-package it was not needed. The plugin download the files/artifacts specified with the key `artifacts`.

.buildkite/pipeline.yml
```diff
@@ -1,3 +1,66 @@
...
+        plugins:
+          - junit-annotate#v2.4.1:
+              artifacts: "build/*.xml"
+        agents:
+          provider: "gcp"
+        depends_on:
+          - "unit-test"
+          - "int-test"
+    depends_on: "tests"
```
mrodm, line 66: This depends_on forces to wait for the "Run check-ci" and "Run local build" steps ?

.buildkite/pipeline.yml
```diff
@@ -1,3 +1,66 @@
...
+        commands:
+          - "chmod +x .buildkite/scripts/setenv.sh"
+          - ".buildkite/scripts/setenv.sh with-go-docker-compose"
+          - "make test-int"
+          - "make junit-report"
+          - "buildkite-agent artifact upload build/*.xml"
+        agents:
+          provider: "gcp"
+
+      - label: ":junit: Junit annotate"
```
mrodm, line 55: Is this step executed when unit test or integration test fail ? 

Not sure if it is needed to change the `depends_on` relationship, to a specific step to wait and continue with failure.

For instance, add this step between the unit/integration test and the junit report

```yaml
      - wait: ~
        continue_on_failure: true
```

.buildkite/pipeline.yml
```diff
@@ -1,3 +1,66 @@
+# yaml-language-server: $schema=https://raw.githubusercontent.com/buildkite/pipeline-schema/main/schema.json
+
+env:
+  SETUP_GVM_VERSION: 'v0.5.0'
+  DOCKER_COMPOSE_VERSION: '1.25.5'
```
mrodm, line 5: are these the versions used currently for gvm and docker-compose ?

.buildkite/pipeline.yml
SHarbuz: The command step doesn't support downloading the artifacts, but I need to use it in the next steps:
https://buildkite.com/docs/pipelines/artifacts
That's why I use the Buildkite agent

.buildkite/pipeline.yml
SHarbuz: In my opinion - YES, because if we didn't pass those steps we shouldn't start the next steps.
Of course, we can skip this part, but we can save only up to 1 minute by waiting for these steps (unit-test and int-test).

.buildkite/pipeline.yml
SHarbuz: The value of the variable SETUP_GVM_VERSION I used from https://github.com/elastic/elastic-package/blob/main/.buildkite/pipeline.yml because I didn't find it in the fleet-server repo.
The value of the variable DOCKER_COMPOSE_VERSION I used from https://github.com/elastic/fleet-server/blob/main/.ci/Jenkinsfile
If it needs to change, please let share with me the new values of these variables.

.buildkite/pipeline.yml
SHarbuz: let me check it, I've changed it, but haven't commited yet.

.buildkite/pipeline.yml
SHarbuz: Yes, you are right, mentioned you solution is also working properly, let's use your way because of the same approach.
Thank you!

.buildkite/pipeline.yml
SHarbuz: Yes, you are right, mentioned you solution is also working properly, let's use your way because of the same approach.
Thank you!

.buildkite/pipeline.yml
mrodm: What it is interesting is to run the junit step even if any of the unit or integration test steps fail to be able to show the failed tests. Allowing that, junit step could add the annotation for the tests failed in the UI.

Would that relationship allow that ?

.buildkite/pipeline.yml
mrodm: I see that the other steps have been updated including the `artifact_paths` key :+1: 

I think it is just missing to remove the command key in the junit step, as in the example: https://github.com/buildkite-plugins/junit-annotate-buildkite-plugin#example

```suggestion
```

.buildkite/pipeline.yml
mrodm: Ok, I see that you have added `allow_failure`  :+1: 
```
        depends_on:
          - step: "unit-test"
            allow_failure: true
          - step: "int-test"
            allow_failure: true
```

.buildkite/scripts/setenv.sh
```diff
@@ -0,0 +1,82 @@
...
+
+with_go() {
+    mkdir -p ${WORKSPACE}
+    retry 5 curl -sL -o ${WORKSPACE}/gvm "https://github.com/andrewkroh/gvm/releases/download/${SETUP_GVM_VERSION}/gvm-linux-amd64"
+    chmod +x ${WORKSPACE}/gvm
+    eval "$(gvm $(cat .go-version))"
+    go version
+    which go
+    echo "Installing dependencies..."
+    go get github.com/jstemmer/go-junit-report
```
alexsapran, line 20: Why do we need to get the `go-junit-report`

.buildkite/pipeline.yml
```diff
@@ -1,3 +1,88 @@
...
-  - label: "Example test"
-    command: echo "Hello!"
+  - group: "Check and build"
+    key: "check"
+    steps:
+      - label: ":go: Run check-ci"
+        key: check-ci
+        command: "make check-ci"
+        agents:
+          image: "golang:1.20.3"
```
alexsapran, line 21: This applies to the rest of the references about the golang version. 
Ideally having this be sourced from the `.go-version` would be ideal, alternative it should be defined as env var

.buildkite/pipeline.yml
```diff
@@ -1,3 +1,88 @@
...
+          cpu: "8"
+          memory: "4G"
+
+  - group: "Run tests"
+    key: "tests"
+    steps:
+      - label: "Run unit tests"
+        key: unit-test
+        commands:
+          - "chmod +x .buildkite/scripts/setenv.sh"
```
alexsapran, line 39: What if we commited the file with the executable bit, so we dont need to `chmod` it.

.buildkite/pipeline.yml
```diff
@@ -1,3 +1,88 @@
...
+          - ".buildkite/scripts/setenv.sh with-go"
+          - "make test-unit"
```
alexsapran, line 41: You can define multiple targets in Makefiles at are going to be executed in that order.
```suggestion
          - "make test-unit junit-report"
```

alexsapran (commented):
Left some comments.

.buildkite/scripts/setenv.sh
sharbuz: I suppose the `go-junit-report` tool is installed on the Jenkins server, but in the Buildkite pipeline, I got the error https://buildkite.com/elastic/fleet-server/builds/161#0187e582-99f6-48ce-b791-b0b90b0a7d6f in the pipeline.
Probably we can install all dependencies at one time, I saw the step with downloading several dozen necessary components/tools. I thought about it, but we need to be ready for the prod with the pipeline and after that analyze all dependencies again.

.buildkite/pipeline.yml
sharbuz: fixed, waiting for the result: https://buildkite.com/elastic/fleet-server/builds/244

.buildkite/scripts/setenv.sh
alexsapran: Looking at the Makefile, we actually dont need to install it here. 
Its done as part of the Makefile target https://github.com/elastic/fleet-server/blob/5cf1e3874c85e4e63d8835aa8461e4709ea53735/Makefile#L162

alexsapran: I think you installed it as part of the line 20, the Makefile target installs it (but because it already exists does nothing), but it was not working because the PATH needed adjustment. 
You changed the `PATH` in commit [`9f9e036` (#2543)](https://github.com/elastic/fleet-server/pull/2543/commits/9f9e0361e625f1a1eefc0cff12f7bbdd0e196e09) so I think removing line 20 and having only the Makefile should now work just fine.

jlind23:
@sharbuz I think the best path forward here would be to stop working in this PR as soon as soon as the buildkite pipeline is stable and then continue is other follow up PRs.


@mrodm @alexsapran I believe the clean up step is not needed anymore as we already have the PR bots and slack notifications. Thoughts?

.buildkite/pipeline.yml
mrodm: Is that command in this step needed @sharbuz ?

.buildkite/scripts/imageprep.sh
```diff
@@ -0,0 +1,62 @@
...
+source .buildkite/scripts/setenv.sh
+
+DOCKER_REGISTRY = 'docker.elastic.co'
+DOCKER_REGISTRY_SECRET_PATH = "kv/ci-shared/platform-ingest/docker_registry_prod"
+
+publish_docker_image() {
+    echo "Pushing the docker image "$DOCKER_IMAGE:$DOCKER_IMAGE_TAG" to the ${DOCKER_REGISTRY} registry..."
+    dockerLogin
+    docker push "$DOCKER_IMAGE:$DOCKER_IMAGE_TAG"
+    docker logout "$DOCKER_REGISTRY"
```
mrodm, line 14: This command should be added as a bash trap , to ensure that always (even with error) those credentials are removed from the host.

jlind23:
> Probably, those two could be moved to a follow-up PR ? @alexsapran @jlind23

Yup, sounds reasonable.

.buildkite/scripts/imageprep.sh
sharbuz: Done, please check it.
I didn't find a better way to delete credentials. 
Please let me know if you have other ideas...

.buildkite/pipeline.yml
sharbuz: Let me check it...

.buildkite/scripts/setenv.sh
```diff
@@ -0,0 +1,82 @@
...
+    echo -e "\nPATH="${PATH}"" >> dev-tools/integration/.env
+}
+
+with_docker_compose() {
+    mkdir -p ${WORKSPACE}
+    retry 5 curl -SL -o ${WORKSPACE}/docker-compose "https://github.com/docker/compose/releases/download/${DOCKER_COMPOSE_VERSION}/docker-compose-linux-x86_64"
+    chmod +x ${WORKSPACE}/docker-compose
+    docker-compose version
+    export PATH="${WORKSPACE}:${PATH}"
+    echo -e "\nPATH="${PATH}"" >> dev-tools/integration/.env
```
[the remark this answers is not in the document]
sharbuz: yes, please take a look here: 
https://github.com/elastic/fleet-server/blob/5cf1e3874c85e4e63d8835aa8461e4709ea53735/Makefile#LL253C1-L253C1

.buildkite/scripts/setenv.sh
sharbuz: Yes, you are right, I've commented these lines and everything is OK, the issue was with GOPATH/bin
Will delete them in the next commit

.buildkite/scripts/imageprep.sh
```diff
@@ -2,23 +2,47 @@
...
+    DOCKER_PASSWORD=$(retry 5 vault kv get -field password "${DOCKER_REGISTRY_SECRET_PATH}")
     docker login -u "${DOCKER_USER}" -p "${DOCKER_PASSWORD}" "${DOCKER_REGISTRY}" 2>/dev/null
+    trap cleanup EXIT
+}
+
+cleanup() {
+    echo "Deleting credentials..."
+    DOCKER_USER=""
+    DOCKER_PASSWORD=""
+    echo "Done."
```
mrodm, line 25: I guess environment variables could be kept. Their values are going to be lost when the script finishes. But, it could be kept :+1: 

I mean specially to run in this cleanup function the `docker logout` command. Thereby, the file generated by `docker` at `${HOME}/.docker/config.json` with the credentials/auth is removed or at least the authentication for that docker registry entry from the host/agent.

```suggestion
    echo "Deleting credentials..."
    DOCKER_USER=""
    DOCKER_PASSWORD=""
    docker logout "${DOCKER_REGISTRY}"
    echo "Done."
```

.buildkite/scripts/imageprep.sh
```diff
@@ -2,23 +2,47 @@
...
 
-
-dockerLogin() {
-    DOCKER_USER=$(retry 5 vault kv get -field user ${DOCKER_REGISTRY_SECRET_PATH})
-    DOCKER_PASSWORD=$(retry 5 vault kv get -field password ${DOCKER_REGISTRY_SECRET_PATH})
+docker_login() {
+    DOCKER_USER=$(retry 5 vault kv get -field user "${DOCKER_REGISTRY_SECRET_PATH}")
+    DOCKER_PASSWORD=$(retry 5 vault kv get -field password "${DOCKER_REGISTRY_SECRET_PATH}")
     docker login -u "${DOCKER_USER}" -p "${DOCKER_PASSWORD}" "${DOCKER_REGISTRY}" 2>/dev/null
+    trap cleanup EXIT
```
mrodm, line 18: If more cleanup processes are added later, this call could be moved to the top of this shell script. But, there is not need to move it now.

.buildkite/scripts/imageprep.sh
```diff
@@ -2,23 +2,47 @@
...
+    docker push "${DOCKER_IMAGE}":"${DOCKER_IMAGE_TAG}"
     docker logout "$DOCKER_REGISTRY"
 }
 
-
-dockerLogin() {
-    DOCKER_USER=$(retry 5 vault kv get -field user ${DOCKER_REGISTRY_SECRET_PATH})
-    DOCKER_PASSWORD=$(retry 5 vault kv get -field password ${DOCKER_REGISTRY_SECRET_PATH})
+docker_login() {
+    DOCKER_USER=$(retry 5 vault kv get -field user "${DOCKER_REGISTRY_SECRET_PATH}")
```
mrodm, line 15: Following what it is done in `apm-pipeline-library` ([withDockerEnv](https://github.com/elastic/apm-pipeline-library/blob/bafc85a84dda5436498e1abd5ede9ee9499e6e6c/vars/withDockerEnv.groovy#L43-L45)), please set this environment variable to be redacted too. cc @alexsapran 

So the change here would be could be something like (other suffixes could be used):

```suggestion
    DOCKER_USER_SECRET=$(retry 5 vault kv get -field user "${DOCKER_REGISTRY_SECRET_PATH}")
```
In one of the latest builds ([link](https://buildkite.com/elastic/fleet-server/builds/268#01880f32-a528-4ef4-9bd5-def139ab5273)), I don't see that string (the user). But I would follow the same as done in apm-pipeline-library.

.buildkite/scripts/imageprep.sh
sharbuz: But the `docker logout` command is already in the dpublish_docker_image func and in the build-image step as well. Could you please check it and tell me if I need to do anything else or just delete the cleanup func?

.buildkite/scripts/imageprep.sh
sharbuz: I'm not sure I understood your question, please correct me if I'm wrong. The mentioned variables aren't global variables, I defined them in the docker_login func. It means they are using only inside the func. In addition in the mentioned URL the variables with the same name as I used.
Also, in the doc https://buildkite.com/docs/pipelines/managing-log-output#redacted-environment-variables I found _SECRET and _PASSWORD patterns as well. 
Again, perhaps I didn't get the question :)

.buildkite/scripts/imageprep.sh
mrodm: In any of them, if any previous command fails, those `docker logout` commans won't be executed.

For instance, in `publish_docker_image` function , if docker push command fails, docker logout won't be executed since the falgs `-euo pipefail` have been set at the top of the script.

So, I would add that docker logout command there in the cleanup function to ensure it is executed always.

Maybe those others docker logout can be removed in favor of the cleanup function one ?

.buildkite/scripts/imageprep.sh
mrodm: Even if they are declared inside the docker_login func, I would follow the same conditions to redact them. We need to put these mechanisms to ensure that they are not leaked.

And talking about that, what it has been done in some other repositories to get those secrets, it is to use the `pre-command` hook. For instance:
- kibana:
    - https://github.com/elastic/kibana/blob/main/.buildkite/hooks/pre-command 
    -  https://github.com/elastic/kibana/blob/main/.buildkite/scripts/lifecycle/pre_command.sh
- elastic-pacakge:
    - https://github.com/elastic/elastic-package/blob/main/.buildkite/hooks/pre-command
    
    
I would try to follow that pattern.

mrodm: > Should I add something else to enable the `pre-command` hook?

No, it is automatically enabled when it is detected (ensure that it has execution permission that script)

.buildkite/scripts/imageprep.sh
mrodm: Ok, I think I got your point. As it is now, `docker_login` just is executed when `build-image` is run. So the other steps could fail.

Looking at the definition of the step:
```yaml
      - label: "Build docker image"
        commands:
          - ".buildkite/scripts/imageprep.sh build-image"
          - ".buildkite/scripts/imageprep.sh push-image"
          - ".buildkite/scripts/imageprep.sh retag-and-push-image"
        agents:
          provider: "gcp"
```

I see some options here:
- Every execution of the script runs `docker_login`, and every execution of that scripts executes a cleanup trap to do a `docker logout`:
    - Every execution of imageprep.sh would be in charge of login and logout.
- The script `.buildkite/scripts/imageprep.sh` is used a helper script:
    - `imageprep.sh`  would assume that credentials are in-place.
    - You can create another script that manages the credentials required:
      ```bash
      set -euo pipefail

      source ./buildkite/scripts/imageprep.sh

      cleanup() {
        echo "Deleting credentials..."
        docker logout ${DOCKER_REGISTRY}
        echo "Done."
      }
      trap cleanup EXIT

      docker_login

      .buildkite/scripts/imageprep.sh build-image
      .buildkite/scripts/imageprep.sh push-image
      .buildkite/scripts/imageprep.sh retag-and-push-image

      docker logout ${DOCKER_REGISTRY}  # maybe this is not needed, since it is already executed by the cleanup
      ```
    - Step in the pipeline should be updated too:
      ```yaml
            - label: "Build docker image"
              commands:
                - ".buildkite/scripts/push_image.sh"
              agents:
                provider: "gcp"
      ```

Would any of those options make sense ? @sharbuz 
As a note, `build-image` probably does not require to run `docker login`. The other two (`push-image` and `retag-and-push-image`) should require to be logged in.

alexsapran: I like the idea of the wrapper script.

.buildkite/pipeline.yml
```diff
@@ -1,3 +1,82 @@
...
+        agents:
+          provider: "gcp"
+        depends_on:
+          - step: "unit-test"
+            allow_failure: true
+          - step: "int-test"
+            allow_failure: true
+    depends_on: "check"
+
+  - group: "Publish docker image"
```
alexsapran, line 70: This step should be executed following the existing conditions from the [Jenkinsfile](https://github.com/elastic/fleet-server/blob/e7d04c9d6eb40e04b679dc8f99cc5d7f00ec553a/.ci/Jenkinsfile#LL136C4-L138C8)

sharbuz:
I've deleted the publishing stage from the PR.
@alexsapran, @mrodm please review and I'll update/merge it.
I'll prepare another PR with follow-up stages.
@jlind23

jlind23:
Thanks @sharbuz for the update!

alexsapran (approved):

.buildkite/scripts/common.sh
```diff
@@ -0,0 +1,50 @@
...
+    echo -e "\nPATH="${PATH}"" >> dev-tools/integration/.env
+}
+
+with_docker_compose() {
+    mkdir -p ${WORKSPACE}
+    retry 5 curl -SL -o ${WORKSPACE}/docker-compose "https://github.com/docker/compose/releases/download/${DOCKER_COMPOSE_VERSION}/docker-compose-linux-x86_64"
+    chmod +x ${WORKSPACE}/docker-compose
+    docker-compose version
+    export PATH="${WORKSPACE}:${PATH}"
+    echo -e "\nPATH="${PATH}"" >> dev-tools/integration/.env
```
mrodm, line 29: I realised that this file it is committed in the repository. Usually, I try to keep the files as they are in the CI.

Could it be done this in some other way to keep this file without modifications ? @sharbuz 

Is this file being modified already when it is run in Jenkins ? 

cc @alexsapran

alexsapran, line 29: Why do we need to change the `dev-tools/integration/.env`?
Looking at the existing Jenkinsfile this env is something that is required to be available as env vars inside the docker-compose containers, why we had to add the path to the workspace?

alexsapran: After discussing this a bit, here is what we can do. 

1. Create a pre-command hook with this
```bash
....
WORKSPACE="$(pwd)/bin"
mkdir -p ${WORKSPACE}
export PATH="${PATH}:${WORKSPACE}"
....
```

2. don't do any `.env` manipulation in these scripts, so we don't need to do these (this also applies for the `with_docker`)
```
export PATH="${WORKSPACE}:${PATH}"
echo -e "\nPATH="${PATH}"" >> dev-tools/integration/.env
```

3. Remove the function `add_bin_path` and its usage.

The rationale is that we always require to do `add_bin_path` so it makes sense to be in pre-command and once it's there, the required dependencies (kind, docker-compose, docker, etc) are going to be available for usage.

sharbuz: @alexsapran , I suppose, it won't work on the pre-command step, because each agent has its own PATH with the name of k8s container please take a look: https://buildkite.com/elastic/fleet-server/builds/323#0188292f-5215-402a-a295-60e819a6b823
Also, the variables which are defined in the one command of the stage (let's say one virtual bash terminal) aren't accessible in other ones.
It means we need to define and export PATH for each agent. We can use 2 ways, as we discussed earlier:
1. use file - I agree with @mrodm, we shouldn't use the current CI files, and we can use our own file for that
or
2. use build meta-data - https://buildkite.com/docs/agent/v3/cli-meta-data
In a second way, I would test it, because I didn't find info regarding the accessibility of the exported variables without the "get" command. I mean if we need to get the value every time when it's needed - it's fine, but it shouldn't be accessible everywhere without the "get" command, because in this case, we'll overwrite the variable in the other agents.

mrodm (approved):
:+1:

mrodm (approved):
:+1:

PR #2559, opened 2023-05-03
[Fleet] Allow to deploy a custom fleet server image in cloud

### LINKED ISSUE #2484: [E2E][Cloud] Spin up a cloud stack on cloud testing region and make sure an Hosted Fleet server is available
In fleet server repository, a new test should be added, this test will:
- [ ] Run on every PR as a mandatory step
- [ ] Build Fleet server with the latest Elastic Agent image
- [ ] Use this specific image built to trigger a deployment
- [ ] This deployment should be done on a Cloud First Testing region
- [ ] Fleet Server should be running and healthy after this deployment
- [ ] Tear down and clean up

nchaulet:
Closes #2484 

That PR add a new task run on each PR that create a cloud deployment with a custom fleet server image, and run a basic test to check fleet server is healthy.

I also added some documentation on how to deploy a custom image in cloud.

That task will run on every PR? it something we want? or running every day make more sense?

## Todo

- [x] Update readme with instructions on how to build/deploy/test on cloud


dev-tools/cloud/terraform/.terraform.lock.hcl
```diff
@@ -0,0 +1,23 @@
+# This file is maintained automatically by "terraform init".
```
michel-laterman, line 1: this file should be ignored

dev-tools/cloud/docker/build.sh
```diff
@@ -9,9 +9,9 @@ set -eu
 REPO_ROOT=$(cd $(dirname $(readlink -f "$0"))/../../.. && pwd)
 
 USER_NAME=${USER}
-CI_ELASTIC_AGENT_DOCKER_IMAGE=docker.elastic.co/observability-ci/elastic-agent
+CI_ELASTIC_AGENT_DOCKER_IMAGE=${CI_ELASTIC_AGENT_DOCKER_IMAGE:-"docker.elastic.co/observability-ci/elastic-agent"}
 
-DEFAULT_IMAGE_TAG=8.8.0-94522507-SNAPSHOT
+DEFAULT_IMAGE_TAG=8.9.0-5c23b5b2-SNAPSHOT
```
michel-laterman, line 14: Perhaps you should source `dev-tools/integration/.env` and use the version from there?

dev-tools/cloud/Makefile
```diff
@@ -0,0 +1,42 @@
...
+	@cd ${TERRAFORM_PATH}; terraform init; TF_VAR_elastic_agent_docker_image="${CI_ELASTIC_AGENT_DOCKER_IMAGE}:${CUSTOM_IMAGE_TAG}" terraform apply -auto-approve
+
+.PHONY: clean-cloud
+clean-cloud: ## Clean cloud deployment
+	@cd ${TERRAFORM_PATH}; terraform destroy -auto-approve
+
+.PHONY: test
+test: build-and-push-cloud-image ## Clean cloud deployment
+	echo $(CUSTOM_IMAGE_TAG)
+	GOARCH=amd64 CUSTOM_IMAGE_TAG=${CUSTOM_IMAGE_TAG} ${REPO_ROOT}/dev-tools/cloud/docker/build.sh
```
michel-laterman, line 40:
```suggestion
test: build-and-push-cloud-image ## Clean cloud deployment
```
Aren't the 1st two lines here just a repeat of `build-and-push-cloud-image`? If so they can just be removed here

dev-tools/cloud/docker/build.sh
kpollich: @nchaulet - This still needs to be done, correct? Agree w/ Michel that we should use the `.env` file's configured version if possible here.

kpollich (approved):
Code LGTM but Michel's previous suggestion is still pending I believe. Happy to see us dogfooding Terraform a bit here.

dev-tools/cloud/terraform/main.tf
```diff
@@ -0,0 +1,45 @@
...
+}
+
+locals {
+  match           = regex("const DefaultVersion = \"(.*)\"", file("${path.module}/../../../version/version.go"))[0]
+  stack_version   = format("%s-SNAPSHOT", local.match)
+  docker_image_ea = var.elastic_agent_docker_image
+}
+
+resource "ec_deployment" "deployment" {
+  name                   = "example"
```
mrodm, line 26: Is there any check or safeguard to avoid creating more than one deployment ? Would terraform fail if that tries to be done since it is using the name resource? 

If it could be created several deployments, I would think adding here some suffix with some random string is needed. Mainly, it would be needed to avoid create another deployment with the same name.

One option it would be to add a new variable that should be filled by the script:
```
  name                   = "example-${var.DEPLOY_ID}"
```

Probably, this would cause some issue with the `cloud-clean` target, since the same value should be used.

dev-tools/cloud/terraform/main.tf
```diff
@@ -0,0 +1,45 @@
...
+provider "ec" {}
+
+variable "elastic_agent_docker_image" {
+  type        = string
+  description = "Elastic agent docker image with tag."
+}
+
+locals {
+  match           = regex("const DefaultVersion = \"(.*)\"", file("${path.module}/../../../version/version.go"))[0]
+  stack_version   = format("%s-SNAPSHOT", local.match)
```
mrodm, line 21: Intended to just run SNAPSHOT versions ? Would it be useful to run for stable versions too like `8.7.0` ?

juliaElastic:
I think it's good to run this on every pr to start with, and move to daily if we find it slow/too expensive. I think currently we don't have that frequent prs in this repo.

juliaElastic (approved):
LGTM 🚀

dev-tools/cloud/terraform/main.tf
mrodm: If this is going to be run in every PR , I guess it should be added some kind of suffix here.

mrodm (commented):
Just a comment about the terraform code in case it is needed a suffix to avoid using the same resource  name.

dev-tools/cloud/docker/build.sh
nchaulet: Yes I am going to address that

Makefile
```diff
@@ -295,6 +298,15 @@ test-int-set: ## - Run integration tests without setup
 ##################################################
 # Cloud testing targets
 ##################################################
-.PHONY: build-and-push-cloud-image
```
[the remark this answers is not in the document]
nchaulet: Yes the target is still here, it in a different makefile to make things clearer as it need new variables I updated the READNE accordingly

Issue #2519, opened 2023-04-20
Upgrade action sometimes leaves agent in Updating state


juliaElastic:
While investigating the issue of some horde drones converging slowly during an upgrade, encountered an issue where a few drones were left in Updating state indefinitely (no change in a few hours).

When looking at the logs, I saw that the action was delivered, and during ack it encountered an error when writing out action result (temporary connection issue with ES): https://github.com/elastic/fleet-server/blob/main/internal/pkg/api/handleAck.go#L323

It seems that when hitting this error, the upgrade action didn't continue and didn't retry, the agent was left in `Updating` state.

This could be solved by the redesign of upgrade action https://github.com/elastic/ingest-dev/issues/1621

joshdover:
Let's not couple this to the https://github.com/elastic/ingest-dev/issues/1621 as that is likely to be smaller in scope to begin with.

Do you have any logs of what the error was? I remember @blakerouse fixed an issue where we were silently dropping errors and not retrying. I believe we added retries up to 3 attempts, but maybe it's not enough, or only covered one error (I think was retrying on conflict errors): https://github.com/elastic/fleet-server/pull/1896

joshdover:
We discussed that this old PR may also be a potential workaround/fix for these transient issues by ensuring that the agent's version and upgrading status is always updated on every checkin: https://github.com/elastic/fleet-server/pull/1731

juliaElastic:
The error was a connection issue to ES when trying to create action result, as far as I saw the code didn't retry after this error at all.

cmacknz:
> I believe we added retries up to 3 attempts, but maybe it's not enough, or only covered one error (I think was retrying on conflict errors): https://github.com/elastic/fleet-server/pull/1896

> The error was a connection issue to ES when trying to create action result, as far as I saw the code didn't retry after this error at all.

Let's do a thought experiment here. In the middle of an attempt to upgrade 10K agents, the Elasticsearch deployment for Fleet server has 20 minutes of continuous downtime. What happens to all of our upgrade actions? I don't consider "many agents get stuck in updating until the user manually intervenes" to be the right answer. 

Note that I have used upgrades as an easy example, and some of the changes in https://github.com/elastic/ingest-dev/issues/1621 could be a bandaid for this specifically for upgrades. This would do absolutely nothing for any other action type however. What if instead of upgrades they were security response actions? 

How do we ensure that any type of action is eventually acknowledged?

cmacknz:
[...]

PR #2582, opened 2023-05-12
fix for action_seq_no being overwritten with 0

juliaElastic:
Bugfix

Fix for a bug noticed in upgrades, where the upgrade action is the first action (`_seq_no:0`), and the agent doesn't get the action delivered on the next checkin if it misses the first checkin.

In `resolveSeqNo`, the agent's `action_seq_no` was incorrectly overwritten with 0 if the `ackToken` was not provided, which is the case for the first action.
In this case the agent doesn't get the action with `seq_no:0`, as Fleet Server considers it already delivered.
The agent should keep `action_seq_on:-1` until it gets an action delivered.

The overwrite is also wrong after Fleet Server restart, where the `ackToken` is empty (if I understand this correctly).

- Start a fresh ES, kibana, fleet-server (no actions created before)
- Enroll 1 drone with horde
- restart fleet-server
- trigger an upgrade action, the action is pesristed with `_seq_no:0`
- wait for the drone to checkin, it should receive the action and upgrade successfully.

Agent (drone) after enroll:
```
        "_source": {
          "action_seq_no": [
            -1
          ],
          "local_metadata": {
            "elastic": {
              "agent": {
                "id": "e28a2fa7-346f-42da-a7e9-76d8f0fe17fd",
                "version": "8.2.0"
              }
            },
```

After upgrade:
```
          "action_seq_no": [
            0
          ],
          "local_metadata": {
            "elastic": {
              "agent": {
                "id": "e28a2fa7-346f-42da-a7e9-76d8f0fe17fd",
                "version": "8.2.1"
              }
            },
```

Action with seq_no:
```
 {
        "_index": ".fleet-actions-7",
        "_id": "5347065f-2b26-4c5b-a253-9f34e6fcddc5",
        "_seq_no": 0,
        "_primary_term": 1,
        "_score": 1,
        "_source": {
          "@timestamp": "2023-05-12T12:12:27.247Z",
          "expiration": "2023-06-11T12:12:27.247Z",
          "agents": [
            "e28a2fa7-346f-42da-a7e9-76d8f0fe17fd"
          ],
          "action_id": "3d8ade67-388a-48ba-aa15-842bd8dd64a8",
          "data": {
            "version": "8.2.1"
          },
          "type": "UPGRADE",
          "traceparent": "00-330370b9dfdceb2b46606abefe21224d-7600178b8a212d24-01"
        }
      }
```

Fleet server delivered the action:
```
12:13:21.535 INF Action delivered to agent on checkin ackToken=5347065f-2b26-4c5b-a253-9f34e6fcddc5 createdAt=2023-05-12T12:12:27.247Z ecs.version=1.6.0 fleet.access.apikey.id=Fs7fD4gBC7gel4XV8JLp fleet.agent.id=e28a2fa7-346f-42da-a7e9-76d8f0fe17fd http.request.id=aea58b3c-18f8-4675-9ad5-74077d5a3228 id=3d8ade67-388a-48ba-aa15-842bd8dd64a8 inputType= server.address= service.name=fleet-server timeout=0 type=UPGRADE
```

## Design Checklist

## Checklist

## Related issues

Relates https://github.com/elastic/fleet-server/issues/2519

### DIFF
--- a/internal/pkg/api/handleCheckin.go
+++ b/internal/pkg/api/handleCheckin.go
@@ -387,7 +387,7 @@ func (ct *CheckinT) resolveSeqNo(ctx context.Context, zlog zerolog.Logger, req C
 				zlog.Debug().Str("token", *ackToken).Msg("revision token not found")
 				err = nil
 				// should be left the ActionSeqNo if no ackToken, otherwise would be overwritten with 0 on a Fleet Server restart
-				return seqno, nil
+				return seqno, err
 			} else {
 				return seqno, fmt.Errorf("resolveSeqNo: %w", err)
 			}
--- a/internal/pkg/api/handleCheckin.go
+++ b/internal/pkg/api/handleCheckin.go
@@ -295,12 +295,14 @@ func (ct *CheckinT) writeResponse(zlog zerolog.Logger, w http.ResponseWriter, r 
 		}
 	}
 
-	span, _ := apm.StartSpanOptions(r.Context(), "action delivery", "fleet-server", apm.SpanOptions{
-		Links: links,
-	})
-	span.Context.SetLabel("action_count", len(fromPtr(resp.Actions)))
-	span.Context.SetLabel("agent_id", agent.Id)
-	defer span.End()
+	if len(fromPtr(resp.Actions)) > 0 {
+		span, _ := apm.StartSpanOptions(r.Context(), "action delivery", "fleet-server", apm.SpanOptions{
+			Links: links,
+		})
+		span.Context.SetLabel("action_count", len(fromPtr(resp.Actions)))
+		span.Context.SetLabel("agent_id", agent.Id)
+		defer span.End()
+	}
 
 	for _, action := range fromPtr(resp.Actions) {
 		zlog.Info().
@@ -384,6 +386,8 @@ func (ct *CheckinT) resolveSeqNo(ctx context.Context, zlog zerolog.Logger, req C
 			if errors.Is(err, dl.ErrNotFound) {
 				zlog.Debug().Str("token", *ackToken).Msg("revision token not found")
 				err = nil
+				// should be left the ActionSeqNo if no ackToken, otherwise would be overwritten with 0 on a Fleet Server restart
+				return seqno, nil
 			} else {
 				return seqno, fmt.Errorf("resolveSeqNo: %w", err)
 			}
--- a/internal/pkg/api/handleCheckin_test.go
+++ b/internal/pkg/api/handleCheckin_test.go
@@ -7,10 +7,19 @@
 package api
 
 import (
+	"context"
 	"encoding/json"
 	"testing"
+	"time"
 
+	"github.com/elastic/fleet-server/v7/internal/pkg/cache"
+	"github.com/elastic/fleet-server/v7/internal/pkg/checkin"
+	"github.com/elastic/fleet-server/v7/internal/pkg/config"
 	"github.com/elastic/fleet-server/v7/internal/pkg/model"
+	"github.com/elastic/fleet-server/v7/internal/pkg/monitor/mock"
+	"github.com/elastic/fleet-server/v7/internal/pkg/policy"
+	"github.com/elastic/fleet-server/v7/internal/pkg/sqn"
+	ftesting "github.com/elastic/fleet-server/v7/internal/pkg/testing"
 	testlog "github.com/elastic/fleet-server/v7/internal/pkg/testing/log"
 
 	"github.com/stretchr/testify/assert"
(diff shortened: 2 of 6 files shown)


internal/pkg/api/handleCheckin.go
```diff
@@ -295,12 +295,14 @@ func (ct *CheckinT) writeResponse(zlog zerolog.Logger, w http.ResponseWriter, r
 		}
 	}
 
-	span, _ := apm.StartSpanOptions(r.Context(), "action delivery", "fleet-server", apm.SpanOptions{
-		Links: links,
-	})
-	span.Context.SetLabel("action_count", len(fromPtr(resp.Actions)))
-	span.Context.SetLabel("agent_id", agent.Id)
-	defer span.End()
+	if len(fromPtr(resp.Actions)) > 0 {
```
juliaElastic, line 298: this is just moving the trace into a condition to reduce noise ("action deliver" span was showing up even though there were no actions)

internal/pkg/api/handleCheckin.go
```diff
@@ -384,6 +386,8 @@ func (ct *CheckinT) resolveSeqNo(ctx context.Context, zlog zerolog.Logger, req C
 			if errors.Is(err, dl.ErrNotFound) {
 				zlog.Debug().Str("token", *ackToken).Msg("revision token not found")
 				err = nil
+				// should be left the ActionSeqNo if no ackToken, otherwise would be overwritten with 0 on a Fleet Server restart
+				return seqno, nil
```
juliaElastic, line 390: without the return, the code returned `seqno = []int64{sn}`, and `sn` had the value 0 if `ackToken` not defined.

juliaElastic:
Plan to add an automated test for this.

kpollich (approved):
Code LGTM - thanks for tracking this down

internal/pkg/checkin/bulk.go
```diff
@@ -208,6 +208,11 @@ func (bc *Bulk) flush(ctx context.Context) error {
 			if pendingData.extra.ver != "" {
 				fields[dl.FieldAgent] = map[string]interface{}{
 					dl.FieldAgentVersion: pendingData.extra.ver,
+					// Duplicates the writes done when an upgrade is ack'd by the agent.
+					// This helps eliminate the case where the agent is stuck in upgrading after it's already upgraded if the ack
+					// is never received.
+					dl.FieldUpgradeStartedAt: nil,
+					dl.FieldUpgradedAt:       nowTimestamp,
```
juliaElastic, line 215: Revived [this draft](https://github.com/elastic/fleet-server/pull/1731/) as a safety net, in case the ack is lost and the agent is upgraded to a new version, we can safely reset updating status (`upgrade_started_at`)

internal/pkg/api/handleCheckin_test.go
```diff
@@ -127,3 +136,41 @@ func TestFilterActions(t *testing.T) {
...
+		agent: &model.Agent{
+			ActionSeqNo: []int64{sqn.UndefinedSeqNo},
+		},
+		resp: []int64{sqn.UndefinedSeqNo},
+	}}
+	for _, tc := range tests {
+		t.Run(tc.name, func(t *testing.T) {
+			// setup mock CheckinT
+			logger := testlog.SetLogger(t)
+			ctx, cancel := context.WithCancel(context.Background())
```
juliaElastic, line 160: I took most of the setup code from `server_test.go`, might be simplified for a unit test

internal/pkg/checkin/bulk.go
juliaElastic, line 215: it's not really clear how to assert in the unit test that extra fields were set, can we capture the args that the mock was called with in go? https://github.com/elastic/fleet-server/blob/c8966edac2e560671c9b2caabc9beda25bafacaa/internal/pkg/checkin/bulk_test.go#L184

internal/pkg/checkin/bulk.go
```diff
@@ -209,6 +209,11 @@ func (bc *Bulk) flush(ctx context.Context) error {
 				fields[dl.FieldAgent] = map[string]interface{}{
 					dl.FieldAgentVersion: pendingData.extra.ver,
 				}
+				// Duplicates the writes done when an upgrade is ack'd by the agent.
+				// This helps eliminate the case where the agent is stuck in upgrading after it's already upgraded if the ack
+				// is never received.
+				fields[dl.FieldUpgradeStartedAt] = nil
+				fields[dl.FieldUpgradedAt] = nowTimestamp
```
juliaElastic, line 216: Did some local testing, and I think this doesn't work well, the version is coming in every checkin, and the `upgraded_at` field keeps updating, which can be misleading.
Added logging in line 212:

```
07:19:07.751 INF new version ecs.version=1.6.0 service.name=fleet-server ver=8.2.1
07:19:37.751 INF new version ecs.version=1.6.0 service.name=fleet-server ver=8.2.1
07:20:07.750 INF new version ecs.version=1.6.0 service.name=fleet-server ver=8.2.1
07:20:37.749 INF new version ecs.version=1.6.0 service.name=fleet-server ver=8.2.1
07:21:07.748 INF new version ecs.version=1.6.0 service.name=fleet-server ver=8.2.1
```

I think the right thing to do here would be to read the existing action doc, and only update the upgrade fields if the version is newer, and the `upgrade_started_at` field is set.
I'll revert for now.

joshdover: I also think it might mess up the action status tracking in the UI if we don't create the action result document?

juliaElastic: yes, the action would stay in progress

PR #2604, opened 2023-05-18
add the "Publish Docker images" step to the Buildkite pipeline

sharbuz:

Migrated: docker-push part

## Checklist

## Related issues

**Pipeline steps:**
- [x] Publish Docker images


.buildkite/scripts/imageprep.sh
```diff
@@ -0,0 +1,46 @@
+#!/bin/bash
```
alexsapran, line 1: Do we expect to get more functionality into this script?

It might be better to incorporate these functions directly into `.buildkite/scripts/build_image.sh`.

.buildkite/scripts/build_image.sh
alexsapran, line 1: nit suggestion, rename to `build_push_docker_image.sh`

.buildkite/hooks/pre-command
```diff
@@ -0,0 +1,10 @@
+#!/bin/bash
+
+set -euo pipefail
+
+DOCKER_REGISTRY_SECRET_PATH="kv/ci-shared/platform-ingest/docker_registry_prod"
+
+if [[ "$BUILDKITE_PIPELINE_SLUG" == "temp-fleet-server" && "$BUILDKITE_STEP_KEY" == "publish" ]]; then #should be changed to "fleet-server" after tests
```
alexsapran, line 7: Where is the `temp-fleet-server` defined?

.buildkite/hooks/pre-command
```diff
@@ -0,0 +1,10 @@
...
+  export DOCKER_USER=$(retry 5 vault kv get -field user "${DOCKER_REGISTRY_SECRET_PATH}")
+  export DOCKER_USER_SECRET=$(retry 5 vault kv get -field password "${DOCKER_REGISTRY_SECRET_PATH}")
```
alexsapran, line 9: Thinking out loud, what do we think if this does not export the secret at all, but rather performs directly the docker_login?

.buildkite/scripts/build_image.sh
```diff
@@ -0,0 +1,13 @@
+#!/bin/bash
+
+set -euo pipefail
+
+source .buildkite/scripts/common.sh
+
+trap docker_logout EXIT
```
alexsapran, line 7: Instead of using `trap` for the `docker_logout` we could use a `pre-exit`[[1](https://buildkite.com/docs/agent/v3/hooks#job-lifecycle-hooks)] hook that would take care of that, similar to doing the `docker login`.
@mrodm what do you think?

.buildkite/hooks/pre-command
mrodm, line 9: Just to be in the safe side, I would follow as in Jenkins and set as redacted both environment variables here. Even if they are just used here in the pre-command as suggested in https://github.com/elastic/fleet-server/pull/2604/files#r1197548706

Let's avoid any possible issue with those having them as redacted vars from the beginning.

```suggestion
  export DOCKER_USERNAME_SECRET=$(retry 5 vault kv get -field user "${DOCKER_REGISTRY_SECRET_PATH}")
  export DOCKER_PASSWORD_SECRET=$(retry 5 vault kv get -field password "${DOCKER_REGISTRY_SECRET_PATH}")
```

`_SECRET` is a valid suffix for redacted-vars:
https://buildkite.com/docs/pipelines/managing-log-output#redacted-environment-variables

.buildkite/scripts/imageprep.sh
```diff
@@ -0,0 +1,46 @@
...
+    echo "Retagging images..."
+    if ${BUILDKITE_TAG}; then
+        docker tag "${DOCKER_IMAGE}":"${DOCKER_IMAGE_SHA_TAG}" "${DOCKER_IMAGE}":"${DOCKER_IMAGE_GIT_TAG}"
+        DOCKER_IMAGE_TAG="${DOCKER_IMAGE_GIT_TAG}"
+        publish_docker_image
+    else
+        docker tag "${DOCKER_IMAGE}":"${DOCKER_IMAGE_SHA_TAG}" "${DOCKER_IMAGE}":"${DOCKER_IMAGE_LATEST_TAG}"
+        DOCKER_IMAGE_TAG="${DOCKER_IMAGE_LATEST_TAG}"
+        publish_docker_image
+    fi
```
mrodm, line 40: I would change this condition to avoid errors if BUILDKITE_TAG is set:
```bash
 $ export BUILDKITE_TAG="2"
 $ if ${BUILDKITE_TAG} ; then echo "TAG"; else echo "NO TAG"; fi
2: command not found
NO TAG
```

```suggestion
    if [ -n "${BUILDKITE_TAG}" ]; then
        docker tag "${DOCKER_IMAGE}":"${DOCKER_IMAGE_SHA_TAG}" "${DOCKER_IMAGE}":"${DOCKER_IMAGE_GIT_TAG}"
        DOCKER_IMAGE_TAG="${DOCKER_IMAGE_GIT_TAG}"
        publish_docker_image
    else
        docker tag "${DOCKER_IMAGE}":"${DOCKER_IMAGE_SHA_TAG}" "${DOCKER_IMAGE}":"${DOCKER_IMAGE_LATEST_TAG}"
        DOCKER_IMAGE_TAG="${DOCKER_IMAGE_LATEST_TAG}"
        publish_docker_image
    fi
```

.buildkite/scripts/build_image.sh
mrodm: I like that approach of using Buildkite hooks. :+1:

.buildkite/hooks/pre-command
mrodm: Indeed that would be similar since that hook runs first for each step :+1: 
If another step needs this docker_login (e2e tests ? don't know) we can create a function here and reuse it in all the "ifs" needed.

The only disadvantage of doing the docker login here  I think it is we're coupling on how Buildkite works. But I would say it is ok, we shouldn't move to another CI framework anytime soon :D , and we take advantage of Buikdkite

.buildkite/pipeline.yml
```diff
@@ -62,3 +62,15 @@ steps:
           - step: "int-test"
             allow_failure: true
     depends_on: "check"
+
+  - label: "Publish docker image"
+    key: "publish"
+    steps:
+      - label: "Build docker image"
+        command: ".buildkite/scripts/build_image.sh"
+        if: "!build.pull_request"
```
mrodm, line 71: :+1:

.buildkite/pipeline.yml
```diff
@@ -62,3 +62,15 @@ steps:
...
+        agents:
+          provider: "gcp"
+    depends_on:
+      - step: "tests"
+        allow_failure: true
```
mrodm, line 76: Currently in Jenkins, if unit and integration tests fail, is the docker image pushed anyway ?

.buildkite/pipeline.yml
```diff
@@ -62,3 +62,15 @@ steps:
...
     depends_on: "check"
+
+  - label: "Publish docker image"
+    key: "publish"
+    steps:
+      - label: "Build docker image"
+        command: ".buildkite/scripts/build_image.sh"
+        if: "!build.pull_request"
+        agents:
+          provider: "gcp"
```
mrodm, line 73: Should this step be like this ? 

```suggestion
  - label: "Publish docker image"
    key: "publish"
    command: ".buildkite/scripts/build_image.sh"
    if: "!build.pull_request"
    agents:
      provider: "gcp"
```

.buildkite/hooks/pre-command
alexsapran: > we shouldn't move to another CI framework anytime soon 

I hope not!!!!!

sharbuz: That's a good idea, but unfortunately, it doesn't work. I have the "access denied" error. It looks like pre-command is working somewhere in the dedicated environment but we need to push images from the particular agent and we have to run docker_login in this particular agent to establish a connection with the docker registry.

.buildkite/scripts/build_image.sh
sharbuz: It's also a good idea, but unfortunately, it also doesn't work.  I got the error `docker: command not found`. This error confirms my assumption about the dedicated environment for hooks. If it's true we have to run the `docker logout` command in the environment where we ran the `docker login` command i.e. directly in the agent

.buildkite/hooks/pre-command
sharbuz: fixed

.buildkite/pipeline.yml
sharbuz: I changed the condition to:
if: build.env("BUILDKITE_PULL_REQUEST") == null

.buildkite/hooks/pre-command
sharbuz: will try it once again, I found one mistake...

.buildkite/pipeline.yml
```diff
@@ -62,3 +62,13 @@ steps:
...
+
+  - label: "Publish docker image"
+    key: "publish"
+    command: ".buildkite/scripts/build_image.sh"
+    if: build.env("BUILDKITE_PULL_REQUEST") != null   #the condition is temporarily changed to opposite to check and fix the functionality
+    agents:
+      provider: "gcp"
+    depends_on:
+      - step: "tests"
+        allow_failure: false
```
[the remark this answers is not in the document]
sharbuz: Thank you @ramonbutter for the suggestion, unfortunately, the error 404 for me :(

.buildkite/hooks/pre-command
sharbuz: the result of tests:
`docker login` in the pre-command os OK
`docker logout` in the pre-exit -  `docker: command not found`
@alexsapran @mrodm @leo-ri

sharbuz: I tied to use post-command as well, but unfortunately, the same error:
`.buildkite/hooks/post-command: line 6: docker: command not found`

.buildkite/pipeline.yml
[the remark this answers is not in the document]
ramonbutter: hey sorry @sharbuz, the permissions may have been to restrictive on that repo.
Here is another example that should work for you: https://github.com/elastic/elasticsearch-controller/pull/147

Just wanted to highlight, that nearly around 20 services in the cloud cp needed to implement a similar logic like this, 
and we packed it in a dedicated buildkite agent so that we don't have to replicate the steps like everywhere and keep consistence in our build process.
It would have come further with the support to build ARM images, which is a little bit of a todo for this repository.

.buildkite/scripts/imageprep.sh
sharbuz: done

.buildkite/hooks/pre-command
sharbuz: It's an environment variable. 
I removed this extra option ($BUILDKITE_PIPELINE_SLUG) because there is we have only one fleet-server pipeline and we don't need to verify it. We need to define only step.

.buildkite/pipeline.yml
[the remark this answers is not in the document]
sharbuz: Thank you @ramonbutter, I'll try to get access to the provided you examples because in the second example the same error 404 :(
CC @alexsapran , @mrodm could you help with the access or example?

.buildkite/hooks/pre-command
```diff
@@ -0,0 +1,15 @@
...
+
+set -euo pipefail
+
+source .buildkite/scripts/common.sh
+
+DOCKER_REGISTRY_SECRET_PATH="kv/ci-shared/platform-ingest/docker_registry_prod"
+DOCKER_REGISTRY="docker.elastic.co"
+export DOCKER_REGISTRY="${DOCKER_REGISTRY}"
+
+if [[ "$BUILDKITE_STEP_KEY" == "publish" ]]; then
```
mrodm, line 11: nit. Add also the "$BUILDKITE_PIPELINE_SLUG" == "fleet-server" 

```suggestion
if [[ "$BUILDKITE_PIPELINE_SLUG" == "fleet-server" && "$BUILDKITE_STEP_KEY" == "publish" ]]; then
```

This pre-command is going to be used by all the pipelines defined in the repository. It is better to narrow here where these secrets are going to be retrieved and used.

mrodm (commented):
About the usage of `pre-exit` or `pre-commands` hooks, those hooks are run in all the steps. Therefore, in those steps where docker is not available will fail. For instance that happens in the steps run by k8s agents.

To use those hooks and being able to run `docker logout` command , I would add some if clauses to ensure that the docker command is used where it is available. Following the approach of `pre-command` hook with the secrets.

It would be something like:
```
if [[ "$BUILDKITE_PIPELINE_SLUG" == "fleet-server" && "$BUILDKITE_STEP_KEY" == "docker_step_key" ]]; then
   docker logout...
fi
```

.buildkite/hooks/pre-command
sharbuz: Fixed.
I thought if we have only one pipeline we shouldn't define extra conditions to check the pipeline because it could confuse smb with the questions "Why?" and "Where is the second pipeline?"
And probably it's better to add a comment with an example of using the multi-pipeline solution.
What do you think, @mrodm ?

mrodm: At least, there is one more CI job to migrate to Buildkite from fleet-server at some point:
https://github.com/elastic/fleet-server/blob/main/.ci/packaging.groovy

Not checked, but it is likely that we need to add more code in these hooks for some steps.

We could add some comment about the reasoning of this if that could help other users, sure. The main reason for me is to narrow the scope of the secrets to the steps where they are need. And to identify them it would be needed both pipeline slug and step key.

.buildkite/pipeline.yml
```diff
@@ -3,7 +3,6 @@
 env:
   SETUP_GVM_VERSION: 'v0.5.0'
   DOCKER_COMPOSE_VERSION: '1.25.5'
-  DOCKER_REGISTRY: "docker.elastic.co"
   DOCKER_IMAGE: "${DOCKER_REGISTRY}/observability-ci/temp-fleet-server" # needs to rename after tests
```
mrodm, line 6: Should not be declared `DOCKER_REGISTRY` here? 
`DOCKER_IMAGE` variable is expanded correctly with the value of "DOCKER_REGISTRY" ?

I would add back `DOCKER_REGISTRY` here for readability. To see how `DOCKER_IMAGE` would be expanded without checking other files.

.buildkite/pipeline.yml
```diff
@@ -3,7 +3,6 @@
...
   DOCKER_IMAGE_TAG: ""
```
mrodm, line 7: nit. Maybe this could be removed, since it is created and used inside the script.

mrodm, line 7: I cannot add the message to the corresponding line, adding it here.

About `DOCKER_IMAGE_GIT_TAG` it is used `BUILDKITE_BRANCH`, but I see there is a variable in buildkite for tags `BUILDKITE_TAG`. What is the difference ? Should we use `BUILDKITE_TAG` here? 

```
DOCKER_IMAGE_GIT_TAG: "temp-${BUILDKITE_TAG}" # needs to rename after tests
```

mrodm: It looks like they are both set in case of pushing a tag ([link](https://buildkite.com/elastic/elastic-package/builds/748#01881e9e-15a3-4ee9-809d-8cd01598cf47)):
```
BUILDKITE_TAG="v0.80.0"
BUILDKITE_BRANCH="v0.80.0"
```

And for a push to main branch ([link](https://buildkite.com/elastic/elastic-package/builds/779#01884815-cb47-4b4c-bb95-2b39cbe79978)):
```
BUILDKITE_BRANCH="main"
BUILDKITE_TAG=""
```

Should we use `BUILDKITE_TAG` ?

.buildkite/pipeline.yml
```diff
@@ -62,3 +61,13 @@ steps:
           - step: "int-test"
             allow_failure: true
     depends_on: "check"
+
+  - label: "Publish docker image"
+    key: "publish"
+    command: ".buildkite/scripts/build_push_docker_image.sh"
+    if: build.env("BUILDKITE_PULL_REQUEST") != null
```
mrodm, line 68: I think it is needed to ensure which branches or tags are being executed in Buildkite. since by default `build_branches` is enabled.

And for instance this branch was run in Buildkite (without PR):
- Branch of this PR: https://github.com/elastic/fleet-server/pull/2566
- Buildkite job for the branch event: https://buildkite.com/elastic/fleet-server/builds?branch=updatecli_28b583ca5753aa9b9dea5ffba70409e7bc6ea043c84ea6c3dee72907d8c6e04a
- Buildkite job for the PR event (bot): https://buildkite.com/elastic/fleet-server/builds?branch=elastic:updatecli_28b583ca5753aa9b9dea5ffba70409e7bc6ea043c84ea6c3dee72907d8c6e04a 

cc @alexsapran I think we should try to limit what it is run in Buildkite to the branches to those that are protected (main, 8.* and 7.*) and the corresponding tags.

Catalog-info: 
https://github.com/elastic/fleet-server/blob/main/catalog-info.yaml#L27-L51

It should be added `branch_configuration` as in elastic-package or package-registry to just allow main, 8.* or 7.* branches or v8.* , v7.* tags 

Example: https://github.com/elastic/package-registry/blob/main/catalog-info.yaml#L25



Currently, all

.buildkite/pipeline.yml
sharbuz: I don't understand why you are seeing DOCKER_REGISTRY here...
https://github.com/sharbuz/fleet-server/blob/b35a235fadd19ef254a26e7838fda17a743592c6/.buildkite/pipeline.yml#L3-L11
It's already moved to pre-command hook.
And I moved it to pre-command hook because I faced an issue with the undefined value of the DOCKER_REGISTRY when I started to use it in the pre-command hook.
let me check it one more time.

.buildkite/pipeline.yml
sharbuz: ok

sharbuz: I just used the same logic:
https://github.com/elastic/fleet-server/blob/3229ead9040bc7f9c7bb1e6e7cc2730ab8f1dabf/.ci/Jenkinsfile#L140-L142
I can't find the output of the pipeline with values of these variables, because it looks like the Buildkite stores limited history.

.buildkite/pipeline.yml
sharbuz: Should we change that in this PR, or we can move on next PR?
Because it's changing the logic of the current Jenkins pipeline...

.buildkite/pipeline.yml
mrodm: Jenkins stores the in BRANCH_NAME branch or tag. 

I guess looking at the output of BUILDKITE env. vars it could be used BUILDKITE_BRANCH too. However, as this step is intended to create docker tags according to the git tags pushed (or latest), I would use BUILDKITE_TAG environment variable here. Even that variable is the one used in the if clause.

.buildkite/pipeline.yml
mrodm: It could be done in another Pull Request. This is unrelated to the step defined here in this PR. Just realized about these Buildkite builds while looking at the options of the filtering.

IIRC the required change should be updating the [`catalog-info.yml` file](https://github.com/elastic/fleet-server/blob/main/catalog-info.yaml#L27-L51) to add `branch_configuration` setting with the corresponding values (to be checked the regex to be used).

.buildkite/pipeline.yml
mrodm: Yes, exactly it has been moved to the pre-command hook, and in those changesets it has been removed from the pipeline yaml file.

As DOCKER_IMAGE uses DOCKER_REGISTRY env. var to build the value, I would try (if possible) to set DOCKER_REGISTRY here in the pipeline yaml file. That would help to debug or check DOCKER_IMAGE value.

> And I moved it to pre-command hook because I faced an issue with the undefined value of the DOCKER_REGISTRY when I started to use it in the pre-command hook.

I thought these env. vars defined in the `env` section (pipeline yaml file) should be available in the hooks too. In this case, in the pre-command hook. Those hooks are running in the same agent as the steps/commands, and I would expect to have the same environment. Let's see the results of those tests.

.buildkite/pipeline.yml
mrodm: Created PR for this https://github.com/elastic/fleet-server/pull/2624

.buildkite/pipeline.yml
alexsapran: I guess for now we can leave it be the `branch_name`. If this is causing an issue we can do a follow up.

alexsapran (approved):

mrodm (approved):

PR #2606, opened 2023-05-18
[Fleet] Fake agent in fleet server standalone

### LINKED ISSUE #2605: [Fleet] Fleet server standalone coordination is not working
## Description 

When running fleet server standalone it seems the coordinator is not working

It seems it because we need an agent ID  https://github.com/elastic/fleet-server/blob/main/internal/pkg/coordinator/monitor.go#L103-L108 

## Potential solution 

It probably make sense to generate an agent ID in standalone @michel-laterman @jsoriano what do you think?

nchaulet:
## Description 

Resolve #2605

This create a fake agent id and version for fleet server standalone to allow the coordinator to work.

## How to test?

Run a standalone Fleet server and verify that you can enroll an agent and the agent is fully functional (get the policy ...)


nchaulet:
> How hard would it be to disable the coordinator completely? It does serve any function right now and will actually incur costs for us due to extra refreshes and the associated API calls on stateless.

Yes it's probably a better solution long term to remove the coordinator, @michel-laterman we do not have plan to use it? I remember you talking about maybe using this for remote ES output

michel-laterman:
note that tests are failing due to an issue with the docker v24.0.0 release

michel-laterman:
I don't think we'll need the coordinator for remote outputs, we've decided to revoke an agent's remote output key on checkin.

joshdover (dismissed):
I'm ok with this change to unblock testing, but we should open an issue that is prioritized in the near-future to remove the coordinator.

joshdover (approved):

michel-laterman:
Have we tested to make sure that the it still grabs the ID that an agent generates if it's running under an agent?

joshdover:
> Have we tested to make sure that the it still grabs the ID that an agent generates if it's running under an agent?

Improving the unit test and also going to manually test the Agent case right now, but I want to understand more your concern. How would this change be an issue if it's only called when Fleet Server **is not** in agent mode?

PR #2628, opened 2023-05-25
Fs release test

sharbuz:

Migrating the "Realase Test" part

## Checklist

## Related issues

**Pipeline steps:**
- [ ] Release Test


.buildkite/scripts/common.sh
```diff
@@ -69,3 +71,32 @@ with_Terraform() {
...
+}
+
+upload_packages_to_gcp_backet() {
+    pattern=${1}
+    baseUrl="gs://${JOB_GCS_BUCKET}/${REPO}"
+    bucketUrlCommit="${baseUrl}"/commits/${BUILDKITE_COMMIT}
+    bucketUrlDefault="${baseUrl}"/snapshots
+
+    if [[ ${BUILDKITE_PULL_REQUEST} != "false" ]]; then
+        bucketUrlDefault="${baseUrl}"/pull-requests/pr-${BUILDKITE_COMMIT}
```
mrodm, line 89: This is generating links using the commit instead of the PR id (e.g. 2628 for this  pull request):

```
gs://fleet-ci-artifacts/fleet-server/pull-requests/pr-b10bfc08046aa6a26c05141af4a348a3306613e5
gs://fleet-ci-artifacts/fleet-server/commits/b10bfc08046aa6a26c05141af4a348a3306613e5
```

This could be changed to use the GITHUB_PR_NUMBER environment variable. However as this would match to the URL used by Jenkins, probably it could be added some sufix (e.g. PR-<id>-buildkite?)

```suggestion
        bucketUrlDefault="${baseUrl}"/pull-requests/pr-${GITHUB_PR_NUMBER}-buildkite
```

This should be updated once it is finally migrated to Buildkite from Jenkins.

.buildkite/scripts/common.sh
```diff
@@ -69,3 +71,32 @@ with_Terraform() {
...
 }
+
+google_cloud_auth() {
+    secretFileLocation=$(mktemp -d -p . -t "${TMP_FOLDER_TEMPLATE_BASE}.XXXXXXXXX")/google-cloud-credentials.json
+    echo "${PRIVATE_CI_GCS_CREDENTIALS_SECRET}" > ${secretFileLocation}
+    gcloud auth activate-service-account --key-file ${secretFileLocation} 2> /dev/null
+    export GOOGLE_APPLICATIONS_CREDENTIALS=${secretFileLocation}
+}
+
+upload_packages_to_gcp_backet() {
```
mrodm, line 82: nit.
```suggestion
upload_packages_to_gcp_bucket() {
```

.buildkite/scripts/common.sh
```diff
@@ -69,3 +71,32 @@ with_Terraform() {
...
+    bucketUrlCommit="${baseUrl}"/commits/${BUILDKITE_COMMIT}
+    bucketUrlDefault="${baseUrl}"/snapshots
+
+    if [[ ${BUILDKITE_PULL_REQUEST} != "false" ]]; then
+        bucketUrlDefault="${baseUrl}"/pull-requests/pr-${BUILDKITE_COMMIT}
+    fi
+
+    for backetUrl in "${bucketUrlCommit}" "${bucketUrlDefault}"; do
+        gsutil -m -q cp -a public-read -r ${pattern} "${backetUrl}"
+    done
```
mrodm, line 94: nit
```suggestion
    for bucketUrl in "${bucketUrlCommit}" "${bucketUrlDefault}"; do
        gsutil -m -q cp -a public-read -r ${pattern} "${bucketUrl}"
    done
```

.buildkite/pipeline.yml
```diff
@@ -84,3 +84,30 @@ steps:
...
+        allow_failure: false
+
+  - label: ":jenkins: Downstream - Package"
+    key: "downstream-package"
+    command: echo "Triggering the Jenkins Job 'fleet-server/fleet-server-package-mbp/env.JOB_BASE_NAME'..."
+    agents:
+      provider: "gcp"
+    depends_on:
+      - step: "release-package-registry"
+        allow_failure: false
```
mrodm, line 113: As these two steps are just placeholders for the time being, we could just use k8s agents instead of creating new VMs for those two echo commands. It could be removed the agent key definition:

```suggestion
  - label: ":jenkins: Release - Package Registry Distribution"
    key: "release-package-registry"
    command: echo "Triggering the Jenkins Job 'release-package-registry-distribution'..."
    depends_on:
      - step: "release-test"
        allow_failure: false

  - label: ":jenkins: Downstream - Package"
    key: "downstream-package"
    command: echo "Triggering the Jenkins Job 'fleet-server/fleet-server-package-mbp/env.JOB_BASE_NAME'..."
    depends_on:
      - step: "release-package-registry"
        allow_failure: false
```

.buildkite/scripts/common.sh
```diff
@@ -69,3 +71,32 @@ with_Terraform() {
     chmod +x ${WORKSPACE}/terraform
     terraform version
 }
+
+google_cloud_auth() {
+    secretFileLocation=$(mktemp -d -p . -t "${TMP_FOLDER_TEMPLATE_BASE}.XXXXXXXXX")/google-cloud-credentials.json
```
mrodm, line 76: Where is this stored at $(pwd) or ${WORKSPACE} ?

.buildkite/scripts/common.sh
```diff
@@ -69,3 +71,32 @@ with_Terraform() {
...
+    fi
+
+    for backetUrl in "${bucketUrlCommit}" "${bucketUrlDefault}"; do
+        gsutil -m -q cp -a public-read -r ${pattern} "${backetUrl}"
+    done
+}
+
+cleanup() {
+    echo "Deleting temporal files..."
+    cd ${WORKSPACE}
```
mrodm, line 99: Depending on where the credentials is stored, this `cd` should be changed.

WORKSPACE var points to the bin folder, but probably this is stored at the buildkite workspace folder (repository folder).

.buildkite/scripts/common.sh
sharbuz: fixed

.buildkite/pipeline.yml
sharbuz: fixed

.buildkite/scripts/common.sh
sharbuz: fixed

.buildkite/scripts/common.sh
sharbuz: fixed above, nothing to fix here

.buildkite/scripts/common.sh
```diff
@@ -69,3 +71,32 @@ with_Terraform() {
     chmod +x ${WORKSPACE}/terraform
     terraform version
 }
+
+google_cloud_auth() {
+    secretFileLocation=${WORKSPACE}/$(mktemp -d -p . -t "${TMP_FOLDER_TEMPLATE_BASE}.XXXXXXXXX")/google-cloud-credentials.json
```
mrodm, line 76: To create the temporal folder under ${WORKSPACE}, I think it is needed to change `-p` parameter.

.buildkite/scripts/common.sh
```diff
@@ -69,3 +71,32 @@ with_Terraform() {
...
+    baseUri="gs://${JOB_GCS_BUCKET}/${REPO}"
+    bucketUriCommit="${baseUri}"/commits/${BUILDKITE_COMMIT}
+    bucketUriDefault="${baseUri}"/snapshots
+
+    if [[ ${BUILDKITE_PULL_REQUEST} != "false" ]]; then
+        bucketUriDefault="${baseUri}"/pull-requests/pr-${BUILDKITE_COMMIT}
+    fi
+
+    for backetUri in "${bucketUriCommit}" "${bucketUriDefault}"; do
+        gsutil -m -q cp -a public-read -r ${pattern} "${backetUri}"
```
mrodm, line 93: nit. backet -> bucket 
```suggestion
    for bucketUri in "${bucketUriCommit}" "${bucketUriDefault}"; do
        gsutil -m -q cp -a public-read -r ${pattern} "${bucketUri}"
```

sharbuz: :) fixed.

.buildkite/scripts/common.sh
```diff
@@ -69,3 +71,32 @@ with_Terraform() {
...
+}
+
+upload_packages_to_gcp_backet() {
+    pattern=${1}
+    baseUri="gs://${JOB_GCS_BUCKET}/${REPO}"
+    bucketUriCommit="${baseUri}"/commits/${BUILDKITE_COMMIT}
+    bucketUriDefault="${baseUri}"/snapshots
+
+    if [[ ${BUILDKITE_PULL_REQUEST} != "false" ]]; then
+        bucketUriDefault="${baseUri}"/pull-requests/pr-${BUILDKITE_COMMIT}
```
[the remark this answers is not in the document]
sharbuz: fixed

mrodm (approved):
LGTM

.buildkite/pipeline.yml
```diff
@@ -61,34 +62,20 @@ steps:
...
-      - label: "E2E Test"
-        key: "e2e-test"
-        command: ".buildkite/scripts/e2e_test.sh"
-        agents:
-          provider: "gcp"
-        depends_on:
-          - step: "unit-test"
-            allow_failure: false
-          - step: "int-test"
-            allow_failure: false
```
mrodm, line 74: This step must be recovered.
It was introduced in this PR https://github.com/elastic/fleet-server/pull/2646

.buildkite/pipeline.yml
```diff
@@ -61,34 +62,20 @@ steps:
...
-        command: ".buildkite/scripts/cloud_e2e_test.sh"
-        agents:
-          provider: "gcp"
-        depends_on:
-          - step: "unit-test"
-            allow_failure: false
-          - step: "int-test"
-            allow_failure: false
-          - step: "e2e-test"
-            allow_failure: false
```
mrodm, line 87: This condition was also added as part of the PR https://github.com/elastic/fleet-server/pull/2646
It needs to be recovered too.

mrodm (changes requested):
Latest merge it has removed some steps introduced in a pull request that was merged in main. They need to be restored.

.buildkite/pipeline.yml
sharbuz: restored

.buildkite/scripts/common.sh
```diff
@@ -69,3 +71,32 @@ with_Terraform() {
...
+    echo "${PRIVATE_CI_GCS_CREDENTIALS_SECRET}" > ${secretFileLocation}
+    gcloud auth activate-service-account --key-file ${secretFileLocation} 2> /dev/null
+    export GOOGLE_APPLICATIONS_CREDENTIALS=${secretFileLocation}
+}
+
+upload_packages_to_gcp_bucket() {
+    pattern=${1}
+    baseUri="gs://${JOB_GCS_BUCKET}/${REPO}"
+    bucketUriCommit="${baseUri}"/commits/${BUILDKITE_COMMIT}
+    bucketUriDefault="${baseUri}"/snapshots
```
mrodm, line 86: This probably needs to be prefixed by buildkite too ? @sharbuz 

```suggestion
    bucketUriDefault="${baseUri}"/buildkite-snapshots
```

sharbuz: I'm not sure where this PATH is used in other systems/services, perhaps it will need to change too.
We can easily change this line:
`bucketUriCommit="${baseUri}"/commits/${BUILDKITE_COMMIT}`
to
`bucketUriCommit="${baseUri}"/commits/buildkite-${BUILDKITE_COMMIT}`
WDYT @mrodm ?

mrodm: Probably it's better to prefix both commits and snapshots with `buildkite-` until the migration to Buildkite is completed ?

Once it is done and Jenkins does not run, all these paths should be the same as they were in Jenkins (without the prefix). Same for line https://github.com/elastic/fleet-server/pull/2628/files#diff-68c0021e761ebe6b8a42d32dfdf3dda2200f0f2fee50106c85c9aea530ccf3c3R89

sharbuz: Probably in this case is better to use the next approach:
```
bucketUriCommit="${baseUri}"/**buildkite**/commits/${BUILDKITE_COMMIT}
bucketUriDefault="${baseUri}"/**buildkite**/snapshots
bucketUriDefault="${baseUri}"/**buildkite**/pull-requests/pr-${GITHUB_PR_NUMBER}
```
WDYT @mrodm ?

mrodm: :+1: 

Even you could add this directly to the `baseUri` variable, so it is just in one place for all the other variables.

```
baseUri="gs://${JOB_GCS_BUCKET}/${REPO}/buildkite"
```

Please, add a TODO comment to remove it once it is completed the migration.
It looks like there would be some tasks to do for the final migration.

sharbuz: done, please check it

mrodm: thanks!

PR #2633, opened 2023-05-25
Refactor Upload portions into more generic: File

pzl:

An upcoming PR will handle files being downloaded/delivered as opposed to the previous feature, "upload". Many conventions are shared between this: chunks, metadata format, status field, etc.

To support this, the common definitions and functions are being extracted from `uploader` and `upload` modules to a more generic named `file`.  Only things specific to writing and uploading documents remain in `uploader`. 

There is little to no usage of the `os.file` module in all of fleet server, so there is likely very little conflict in sharing the module name.

This PR contains no functional changes. Only renaming and namespace changes.

Only renames, but everything compiles and all tests in the module pass. No tests were deleted, all references updated.

## Design Checklist

## Checklist


internal/pkg/file/es.go
```diff
@@ -105,73 +66,75 @@ func GetFileDoc(ctx context.Context, bulker bulk.Bulk, uploadID string) ([]es.Hi
...
+	}
+	err = json.Unmarshal(out, &chunk)
+	return chunk, err
 }
 
 // Retrieves a subset of chunk document fields, specifically omitting the Data payload (bytes)
 // but adding the calculated field "size", that is the length, in bytes, of the Data field
 // the chunk's ordered index position (Pos) is also parsed from the document ID
-func GetChunkInfos(ctx context.Context, bulker bulk.Bulk, baseID string) ([]ChunkInfo, error) {
+func GetChunkInfos(ctx context.Context, bulker bulk.Bulk, indexPattern string, baseID string) ([]ChunkInfo, error) {
```
michel-laterman, line 129: Do the read functions here need the index pattern as a param as part of the preparation to allow more generic file reads? The upload functions (in `internal/pkg/file/uploader/es.go`) do not have an index param

pzl: yes, the index patterns will be different for file delivery, so it became a parameter

michel-laterman (approved):
lgtm

PR #2639, opened 2023-05-29
Packaging pipeline initialization

sharbuz:

Pipeline initialization for migration the "packaging.groovy" 

## Checklist

## Related issues

**Pipeline steps:**
- [x] Pipeline init
- [ ] Filter build
- [ ] Package
- [ ] DRA Snapshot
- [ ] DRA Release Staging


catalog-info.yaml
```diff
@@ -45,6 +45,44 @@ spec:
...
+      repository: elastic/fleet-server
+      pipeline_file: ".buildkite/pipeline.package.mbp.yml"
+      branch_configuration: "main 8.* 7.* v8.* v7.*"
+      provider_settings:
+        build_pull_request_forks: false
+        build_pull_requests: true
+        build_tags: false
+        filter_enabled: true
+        filter_condition: |
+          build.pull_request.id == null || (build.creator.name == 'elasticmachine' && build.pull_request.id != null)
```
mrodm, line 85: Looking where this pipeline is executed. It looks like it just runs when it is triggered from the main pipeline.

https://github.com/elastic/fleet-server/commits/main
https://github.com/elastic/fleet-server/commits/8.8

Did you see any other location where it is triggered from some push event @sharbuz ?

If it does not, I would change the configuration to disallow `build_branches`

```suggestion
      provider_settings:
        build_pull_request_forks: false
        build_pull_requests: true
        build_tags: false
        build_branches: false
        filter_enabled: true
        filter_condition: |
          build.pull_request.id == null || (build.creator.name == 'elasticmachine' && build.pull_request.id != null)
```

mrodm: Not sure how to allow triggering this to debug.
One option is to add a new entry in the buildkite-pr-bot configuration `.buildkite/pull-requests.json` file. 

That new entry should follow the same settings as this one

sharbuz: I didn't see any places where it executes with the exception of the main pipeline.

.buildkite/pipeline.yml
```diff
@@ -71,3 +71,10 @@ steps:
     depends_on:
       - step: "tests"
         allow_failure: false
+
+  - label: "Upload the fleet-server-package-mbp pipeline"
+    command: "buildkite-agent pipeline upload .buildkite/pipeline.package.mbp.yml"
+#    if:  "build.env('BUILDKITE_BRANCH') != null && build.env('BUILDKITE_TAG') == null && build.env('BUILDKITE_PULL_REQUEST') != null"
+    depends_on:
+      - step: "publish"
+        allow_failure: false
```
mrodm, line 80: Could it be used a trigger step instead of the uploading the pipeline ?

https://buildkite.com/docs/pipelines/trigger-step

catalog-info.yaml
mrodm: As this step is intended to be run just when the main pipeline triggers it (in some specific branches), I think it could be added the buildkite-pr-bot configuration.

That would help you to debug/develop this Buildkite pipeline.
You can follow the same settings as in https://github.com/elastic/elastic-agent/blob/main/.buildkite/pull-requests.json#L19-L34

mrodm (commented):
I think it would be helpful for developing this pipeline adding the buildkite-pr-bot, allowing just to trigger the pipeline when some specific comment is added.

mrodm (approved):
:+1: 
We'll need to check whether or not it is interesting to add the bot config to help triggering the pipeline in a standalone way.

Issue #2593, opened 2023-05-16
Fleet Server API testing


jlind23:
Fleet server APIs listed below should be tested with integration tests. The purpose is to ensure that the behaviours remain the same as we should ensure backwards compatibility with Elastic Agent versions. All status code should be tested.
Api spec can be found [here](https://github.com/elastic/fleet-server/blob/e7d04c9d6eb40e04b679dc8f99cc5d7f00ec553a/model/openapi.yml#L1)

```[tasklist]
### Tasks
- [x] /api/status:
- [x] /api/fleet/agents/enroll:
- [x] /api/fleet/agents/{id}/checkin:
- [x] /api/fleet/agents/{id}/acks:
- [x] /api/fleet/artifacts/{id}/{sha2}:
- [x] /api/fleet/uploads:
- [x] /api/fleet/uploads/{id}/{chunkNum}:
- [x] /api/fleet/uploads/{id}
```

michel-laterman:
We have a smoke test that makes calls to check the OK status on enroll, checkin, and ack: https://github.com/elastic/fleet-server/blob/main/internal/pkg/server/fleet_integration_test.go#L429

Other integration tests include checks that test various auth failures, and various request errors, and checkin timeout behaviour.

amitkanfer:
this is before we added versioning.. correct?

jlind23:
We do not have any versioning yet, as soon as we introduce it these tests needs to be extended the same way.

michel-laterman:
200 responses for all endpoint are validated in the end to end tests: https://github.com/elastic/fleet-server/pull/2645
the existing integration tests validate some basic auth issues and request issues: https://github.com/elastic/fleet-server/blob/main/internal/pkg/server/fleet_integration_test.go#L544 and https://github.com/elastic/fleet-server/blob/main/internal/pkg/server/fleet_integration_test.go#L664.

If needed we could add additional validation to response bodies/headers as well as we currently don't check a lot.
We can also add coverage for:
* `X-Request-ID` header is populated by the server, or copied if the client specifies it
* `Accept` headers for gzip responses
* enrollment tags
* signed actions

jlind23:
I believe this is good for now, we will continue on it as soon as the api versioning is introduced.

PR #2645, opened 2023-05-29
Expand E2E test suite

michel-laterman:

E2E test of fleet-server's endpoints.
Test elastic-agent/fleet-server bootstrap.

Add tests to the stand alone and fleet-server in an agent container test suite that use autogenerated client code in order to test most API endpoints (only the artifacts endpoint is missing). Add a test to the container suite that enrolls another elastic-agent container instance into fleet and verifies its status.

```sh
make test-e2e
```

## Design Checklist

## Checklist

## Related issues

- Relates #2593


testing/e2e/agent_container_test.go
```diff
@@ -186,5 +208,365 @@ func (suite *AgentContainerSuite) TestWithSecretFiles() {
...
+	time.Sleep(time.Minute * 10)
+
+	suite.FleetIsHealthy(bCtx, endpoint)
+}
+
+// TestDockerAgent will enroll a real agent running in a docker container.
+// It uses the dummy-policy (no monitoring or integrations).
+// The test is successful if it reaches the "online" state in Kibana within 5 minutes of the container starting.
+//
+// NOTE: If we wanted to do more tests against an actual agent (such as sending actions or even running integrations) we should make it a test suite.
```
michel-laterman, line 427: I'm not sure if we want to test elastic-agent and fleet-server action handling (using the binaries) as part of this test suite or the one in the agent repo 
cc @cmacknz @pierrehilbert

cmacknz: My opinion is that the server should test that it implements its own API contract correctly independent of any client implementation. The client implementation (Elastic Agent) is responsible for ensuring it interacts correctly with the server implementation (Fleet Server).

This would lead you to testing action handling and the other APIs in the Fleet Server repository, but using a perfect reference client that you wrote independently and not Elastic Agent itself.

The agent would then have an integration test against a real Fleet Server and Fleet Server doesn't depend on Elastic Agent itself for testing.

testing/e2e/agent_container_test.go
```diff
@@ -186,5 +208,365 @@ func (suite *AgentContainerSuite) TestWithSecretFiles() {
...
+		tester.TestFullFileUpload(agentKey, agentID, actionID, 8192) // 8KiB file
+	})
+
+	// TODO: Not sure how to setup and test (what to use as ID+SHA256 values)
+	suite.Run("test artifact", func() {
+		suite.T().Skip("unimplemented")
+	})
+}
+
+// TestSkeep10m checks if the fleet-server is healthy after 10m
```
juliaElastic, line 353: typo: TestSkeep10m -> TestSleep10m

juliaElastic:
Could we enable e2e tests in buildkite in the same pr to see the output and how long they run?

testing/e2e/e2e_test.go
```diff
@@ -153,3 +187,272 @@ func (suite *BaseE2ETestSuite) FleetServerStatusOK(ctx context.Context, url stri
...
+	suite.Require().NoError(err)
+	defer resp.Body.Close()
+	suite.Require().Equal(http.StatusOK, resp.StatusCode)
+	var obj struct {
+		ID string `json:"id"`
+	}
+	err = json.NewDecoder(resp.Body).Decode(&obj)
+	suite.Require().NoError(err)
+	return obj.ID
+}
```
michel-laterman, line 414: I'm not sure if this is actually needed for the artifacts test.
Or how to reference whatever this creates in the fleet-server endpoint

testing/e2e/stand_alone_test.go
```diff
@@ -120,3 +120,130 @@ func (suite *StandAloneSuite) TestWithSecretFiles() {
...
+			ctx,
+			suite.client,
+			"https://localhost:8220",
+		}
+		_, agentKey := tester.TestEnroll(enrollmentKey)
+		suite.AddSecurityContainer(ctx)
+		suite.AddSecurityContainerItem(ctx)
+
+		hits := suite.FleetHasArtifacts(ctx)
+		tester.TestArtifact(agentKey, hits[0].Source.Identifier, hits[0].Source.DecodedSHA256)
```
michel-laterman, line 244: Before these steps are ran, the test setup calls the setup methods for fleet in kibana, and manually inserts the endpoint package

testing/e2e/client_api_test.go
```diff
@@ -0,0 +1,219 @@
...
+	tester.Require().NoError(err)
+
+	resp, err := client.ArtifactWithResponse(tester.ctx,
+		id,
+		sha2,
+		&api.ArtifactParams{},
+	)
+	tester.Require().NoError(err)
+	tester.Require().Equal(http.StatusOK, resp.StatusCode())
+	// TODO more validation?
```
michel-laterman, line 218: would like some feedback here, or if just ensuring that the call returns a 200 is enough

testing/e2e/e2e_test.go
juliaElastic: you could check with @kevinlog to clarify the use case of artifacts API

testing/e2e/client_api_test.go
nchaulet: Can we test that the body contains the proper artifact? maybe comparing the sha256 as we already have it here?

michel-laterman: comparing with the encoded hash

nchaulet (approved):
LGTM 🚀

testing/e2e/agent_container_test.go
michel-laterman: I'll keep this test as a sanity check.
We'll use the generated API client for endpoint testing

PR #2651, opened 2023-05-31
package step

sharbuz:

Migration of the "Filter build" step of the "packaging.groovy" pipeline.

## Checklist

## Related issues

**Pipeline steps:**
- [x] Pipeline init
- [x] Filter build - Skipped because this filter is required by Jenkins auto-scan and we already have all additional conditions for starting the pipeline
- [x] Package
- [ ] DRA Snapshot
- [ ] DRA Release Staging


michel-laterman:
Jenkins issue should be solved by https://github.com/elastic/fleet-server/pull/2672

sharbuz:
analyzed the **Filter build** part, decided (confirmed by @mrodm ) to skip this part because this filter is required by Jenkins auto-scan and we already have all additional conditions for starting the pipeline. We can start this part manually from the UI https://buildkite.com/elastic/fleet-server-package-mbp, from the upstream pipeline with conditions - https://github.com/elastic/fleet-server/blob/820ae55a8652775a2ab7a6e98325ca4c4c471ddb/.buildkite/pipeline.yml#L120 and by commenting the PR https://github.com/elastic/fleet-server/blob/d5babdea11fbc440f3e059794868a55aca6203ac/.buildkite/pull-requests.json#L14-L26

.buildkite/scripts/package.sh
```diff
@@ -0,0 +1,38 @@
...
+    if [[ ${PLATFORM_TYPE} == "arm" || ${PLATFORM_TYPE} == "aarch64" ]]; then
+        PLATFORMS="linux/arm64"
+        PACKAGES="docker"
+    fi
+
+    add_bin_path
+    with_go
+    with_mage
+
+    if [[ ${MATRIX_TYPE} == "staging" ]]; then
```
alexsapran, line 26: Looking at the Jenkinsfile https://github.com/elastic/fleet-server/blob/820ae55a8652775a2ab7a6e98325ca4c4c471ddb/.ci/packaging.groovy#L169C14-L178

It looks like the `make` target depending on the type are `release-manager-release` or `release-manager-snapshot`

.buildkite/scripts/common.sh
```diff
@@ -94,6 +94,33 @@ upload_packages_to_gcp_bucket() {
...
+    pattern=${1}
+    type=${2}
+    baseUri="gs://${JOB_GCS_BUCKET}/jobs/buildkite"              #TODO: needs to delete the "/buildkite" part after the migration from Jenkins
+
+    if [[ ${type} == "snapshot" ]]; then
+        bucketUri="${baseUri}"/commits/${BUILDKITE_COMMIT}
+    else
+        bucketUri="${baseUri}"/${type}/${BUILDKITE_COMMIT}
+    fi
+    gsutil -m -q cp -a public-read -r ${pattern} "${bucketUri}"
```
mrodm, line 107: nit, I wold use `local` when declaring variables inside a function

```suggestion
    local pattern=${1}
    local type=${2}
    local baseUri="gs://${JOB_GCS_BUCKET}/jobs/buildkite"              #TODO: needs to delete the "/buildkite" part after the migration from Jenkins
    local bucketUri=""

    if [[ ${type} == "snapshot" ]]; then
        bucketUri="${baseUri}"/commits/${BUILDKITE_COMMIT}
    else
        bucketUri="${baseUri}"/${type}/${BUILDKITE_COMMIT}
    fi
    gsutil -m -q cp -a public-read -r ${pattern} "${bucketUri}"
```

.buildkite/scripts/common.sh
```diff
@@ -94,6 +94,33 @@ upload_packages_to_gcp_bucket() {
...
+            "github.com/magefile/mage"
+            "github.com/elastic/go-licenser"
+            "golang.org/x/tools/cmd/goimports"
+            "github.com/jstemmer/go-junit-report"
+            "gotest.tools/gotestsum"
+    )
+
+    for pckg in "${install_packages[@]}"; do
+    go install "${pckg}@latest"
+    done
```
mrodm, line 121: nit
```suggestion
    for pkg in "${install_packages[@]}"; do
        go install "${pkg}@latest"
    done
```

.buildkite/scripts/package.sh
```diff
@@ -0,0 +1,38 @@
...
+    else
+        make SNAPSHOT=true release
+    fi
+fi
+
+if [[ ${BUILDKITE_BRANCH} == "main" && ${MATRIX_TYPE} == "staging" ]]; then
+    echo "INFO: staging artifacts for the main branch are not required."
+else
+    google_cloud_auth
+    upload_mbp_packages_to_gcp_bucket "build/distributions/" "${MATRIX_TYPE}"
```
mrodm, line 37: what is it `mbp` ?

sharbuz: It's a good question, @mrodm :),
I got this name here: https://github.com/elastic/fleet-server/blob/820ae55a8652775a2ab7a6e98325ca4c4c471ddb/.ci/Jenkinsfile#L226
It's part of the current stage name. Also, the pipeline has the same part of the name.
In the fleet-server repo we have 2 pipelines with 2 targets and conditions for the artifacts. That's why I created one more func which has `mbp` in the name.

.buildkite/scripts/package.sh
```diff
@@ -0,0 +1,38 @@
...
+source .buildkite/scripts/common.sh
+
+VERSION=$(awk '/const DefaultVersion/{print $NF}' version/version.go | tr -d '"')
+PLATFORM_TYPE=$(uname -m)
+MATRIX_TYPE="$1"
+INFRA_REPO="https://github.com/repos/elastic/infra/contents"
+
+if [[ ${BUILDKITE_BRANCH} == "main" && ${MATRIX_TYPE} == "staging" ]]; then
+    echo "INFO: staging artifacts for the main branch are not required."
+else
```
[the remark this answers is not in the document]
sharbuz: Yes, @mrodm, thanks for highlighting that, I've checked the logic again and changed the code.

.buildkite/scripts/package.sh
sharbuz: Yes, you are right, I implemented there the same logic. There is only one difference - `SNAPSHOT=true` in case `matrix_type != "staging"`
https://github.com/elastic/fleet-server/blob/820ae55a8652775a2ab7a6e98325ca4c4c471ddb/Makefile#L260-L266

.buildkite/scripts/package.sh
mrodm: Ah, then this `mbp` could stand for `multibranch project`

mrodm (approved):
LGTM

PR #2654, opened 2023-06-01
feat: support static tokens to be used for agent enrollment

olegsu:
This PR adds support for static tokens for an agent enrollment.
The reason we need this capability is to support a flow for serverless security projects when an agent should be installed at the same time as the other project services (kibana, es, fleet). 
To be able to install an agent at that point, we need a predictable token that both the agent and the fleet accept.

By accepting static tokens, well just the `key` part of the token, fleet server will authenticate a request that have this `key` in the enrollment token it uses and will send back the API key to communicate with ES. For example, passing to the agent token like `base64(0123456789:abcdefg)` the fleet server will accept it if the configuration includes the following
```yaml
inputs:
  - type: fleet-server
    policy.id: "${FLEET_SERVER_POLICY_ID:fleet-server-policy}"
    server:
      auth: statis
      static_policy_tokens:
        enabled: true
        policy_tokens:
          # token.key: policy_id
          abcdefg: 901b70f0-fefa-11ed-aa5e-5974b0535e80
```

Today flow looks something like that 
image

And this PR will support the following flow
image

1. Create local stack `elastic-package stack up -s "elasticsearch,kibana,package-registry" --version 8.8.0 -v -d`
2. Build and run fleet server local image ( attached to the network of the stack )
3. Run Elastic-Agent docker image (attached to the same network )
4. See that the Elastic-Agent is healthy and the spawned Cloudbeat sending findings.

Steps 2,3 looks simple but they require more configuration and changes (especially if I want to debug the fleet process)
First building docker image for the server will not work on arm architecture as the `make build-docker` [compiles](https://github.com/elastic/fleet-server/blob/main/Dockerfile#L15) it for amd ( I guess that I was missing something).
Later to command to run the image looks like 
```bash
docker run -it --rm \
    --network elastic-package-stack_default \
    -v $(PWD)/fleet-server.yml:/etc/fleet-server.yml \
    -v $(HOME)/.elastic-package/profiles/default/certs/fleet-server:/etc/ssl/certs \
    -v $(HOME)/.elastic-package/profiles/default/certs/ca-cert.pem:/etc/ssl/certs/elastic-package.pem \
    -e FLEET_TOKEN_POLICY_NAME="Fleet Server (elastic-package)" \
    -e FLEET_URL=https://fleet-server:8220 \
    -e KIBANA_HOST=https://kibana:5601 \
    -e ELASTIC_CONTAINER=true \
    -e FLEET_SERVER_SERVICE_TOKEN=**REDACTED**\
    -e FLEET_SERVER_ENABLE=1 \
    -e ELASTICSEARCH_HOSTS=https://elasticsearch:9200 \
    -e KIBANA_FLEET_SERVICE_TOKEN=**REDACTED** \
    -e KIBANA_FLEET_HOST=https://kibana:5601 \
    -e FLEET_SERVER_ELASTICSEARCH_HOST=https://elasticsearch:9200 \
    -e FLEET_SERVER_HOST=0.0.0.0 \
    -e FLEET_SERVER_CERT_KEY=/etc/ssl/certs/key.pem \
    -e FLEET_SERVER_CERT=/etc/ssl/certs/cert.pem \
    -e KIBANA_FLEET_SETUP=1 \
    -p 8220:8220 \
    --name fleet-server \
    docker.elastic.co/fleet-server/fleet-server:8.9.0
```
Where the change in `fleet-server.yml` was to update `inputs[0].server` with 
```yaml
      static_policy_tokens:
        enabled: true
        policy_tokens:
          # token.key: policy_id
          abcdefg: 901b70f0-fefa-11ed-aa5e-5974b0535e80
```

The command to run Elastic Agent looks like
```bash
docker run -it \
    --network elastic-package-stack_default \
    -v $(HOME)/.elastic-package/profiles/default/certs/ca-cert.pem:/etc/ssl/certs/elastic-package.pem \
    -e FLEET_ENROLLMENT_TOKEN=MDEyMzQ6YWJjZGVmZw== \
    -e FLEET_ENROLL=1 \
    -e FLEET_URL=https://fleet-server:8220 \
    -e KIBANA_HOST=https://kibana:5601 \
    -e ELASTIC_CONTAINER=true \
    docker.elastic.co/elastic-agent/elastic-agent-complete:8.8.0
```

## Design Checklist

## Checklist

## Related issues
- https://github.com/elastic/security-team/issues/6762
- https://github.com/elastic/security-team/issues/6648


olegsu:
Thank you for the quick feedback, @joshdover 


> * What is the behavior on checkin if the specified policy ID does not actually exist yet?
>   
>   * Because the policies are initialized by Kibana asynchronously from Fleet Server startup, I think we absolutely need to handle this case. Probably the agent could receive some default / mostly empty policy that doesn't do much but I'm not sure what this would break on the Agent side, for instance if there are no inputs and no output.

This is a good point, I agree we need to validate the policy is been created already.  I see two additional options this can be achieved:
1. On a check-in flow, check if the policy exists and return an error to the agent if not. This will cause the agent to exit and k8s will restart it. Eventually, Kibana will create one (my suggestion).
2.  On fleet setup flow check that all the policy exists and either wait or exit if at least one is not. Same as before, eventually it will be added.

On the other side, sending an empty policy to the agent will also force us to update it once it has been created, which I am not sure is something we want to handle at this point.

> * I think we will want to limit who can enroll agents using a static token.
>   
>   * I'd suggest we do this by IP address / range so we can lock this down to only agents deployed within our infrastructure internally.
>   * This doesn't necessarily need to be implemented in the first PR but we should think about what the final config schema should be so we can add this config later without making a breaking change to the structure

Right, we need to have some ability to do that, not sure how though.


> * This definitely needs unit and [e2e tests](https://github.com/elastic/fleet-server/tree/main/testing/e2e) before we'll be able to merge the first PR.

Thank you, will try to do that.

joshdover:
> This is a good point, I agree we need to validate the policy is been created already. I see two additional options this can be achieved:
> 
>     1. On a check-in flow, check if the policy exists and return an error to the agent if not. This will cause the agent to exit and k8s will restart it. Eventually, Kibana will create one (my suggestion).
> 
>     2. On fleet setup flow check that all the policy exists and either wait or exit if at least one is not. Same as before, eventually it will be added.
> 
> 
> On the other side, sending an empty policy to the agent will also force us to update it once it has been created, which I am not sure is something we want to handle at this point.

Thinking about this more, I think the enrollment should reject until the policy is created. This way we avoid having agents in the `.fleet-agents` index that point to a policy_id that doesn't exist yet, which could break several things. Having k8s restart the agent until it exists seems acceptable for now.

> Right, we need to have some ability to do that, not sure how though.

Yeah that's fine, let's just make sure the config will allows to add it later. If we needed to have different allowed IPs per policy_token, that would be harder to add without a breaking change to the config schema you proposed. Maybe we should change the `policy_tokens` to an array so we can add options later. I like this more too since it makes the config more self-describing:

```yaml
inputs:
  - type: fleet-server
    policy.id: "${FLEET_SERVER_POLICY_ID:fleet-server-policy}"
    server:
      auth: statis
      static_policy_tokens:
        enabled: true
        allowed_ips: [10.0.0.0/24] # allowlist for all tokens
        policy_tokens:
          # token.key: policy_id
          - policy_id: 901b70f0-fefa-11ed-aa5e-5974b0535e80
            token: abcdefg
            allowed_ips: [10.0.0.0/24] # allowlist for just this token
```

A couple other notes:
- Let's call these `enrollment_tokens` to be more consistent with other naming
- We should enforce some minimum length on these (I'd suggest 32 characters)

eyalkraft:
Great work @olegsu!

@joshdover Thanks for your comments.

> - I think we will want to limit who can enroll agents using a static token.
>    - I'd suggest we do this by IP address / range so we can lock this down to only agents deployed within our infrastructure internally.

Could you elaborate why do you think this limitation is required?
I get why we'd only want "internal" agents using static enrollment tokens. My question is - doesn't the static enrollment token itself ensure that? Where would an external agent/user get a static enrollment token from? Given that static tokens would be generated by the project controller upon project creation.
We can decide on whatever key length to have the required encryption strength.

If we still want to defend ourselves from a scenario where an external user gets a static enrollment token somehow, and only allow "internal" agents to enroll using a static token, I'd argue against basing this limitation on IP address or range.
Basing this limitation on IP would require the project controller to configure this IP to the fleet servers it creates, which adds additional complexity and mix of responsibility domains (for example, I don't know how IP aware the project controller even is). Even if we go for the most permissive path of configuring all the fleet servers with the same static IP range list that covers all MKI clusters - IP ranges changes, and sooner or later this mechanism will break or require updates. Seems very fragile.
The better design here IMO would be using a cryptographic challenge of some sort. Like requiring the token to be cryptographically signed using a certificate only the project controller has, or a cluster-scoped certificate available only to the fleet server and agents running on this cluster. This goes back to the question - what is our fear here exactly? what are we protecting from?

joshdover:
@eyalkraft To be clear - I wasn't proposing we implement IP range restrictions in this PR, it is just an example. My main point was that we need to make the config more flexible to add more options later if we need.

The token itself may be good enough - my thinking here was just adding an additional layer of protection in the case that this token is compromised or we use a really bad default (0000) for some reason. Happy to have a deeper conversation about this - but I don't think we need to have this problem solved now in order to move forward with the current PR and unblock validation of this feature.

olegsu:
> Thinking about this more, I think the enrollment should reject until the policy is created. This way we avoid having agents in the `.fleet-agents` index that point to a policy_id that doesn't exist yet, which could break several things. Having k8s restart the agent until it exists seems acceptable for now.

Sounds good, will add that. 

> Yeah that's fine, let's just make sure the config will allows to add it later. If we needed to have different allowed IPs per policy_token, that would be harder to add without a breaking change to the config schema you proposed. Maybe we should change the `policy_tokens` to an array so we can add options later. I like this more too since it makes the config more self-describing:
> 
> ```yaml
> inputs:
>   - type: fleet-server
>     policy.id: "${FLEET_SERVER_POLICY_ID:fleet-server-policy}"
>     server:
>       auth: statis
>       static_policy_tokens:
>         enabled: true
>         allowed_ips: [10.0.0.0/24] # allowlist for all tokens
>         policy_tokens:
>           - policy_id: 901b70f0-fefa-11ed-aa5e-5974b0535e80
>             token: abcdefg
>             allowed_ips: [10.0.0.0/24] # allowlist for just this token
> ```
> 
> A couple other notes:
> 
> * Let's call these `enrollment_tokens` to be more consistent with other naming
> * We should enforce some minimum length on these (I'd suggest 32 characters)

I will update the names, sure. Having the tokens as an array may have runtime implications later if one fleet server will get a lot of static tokens, for some reason. Do you still think its better to have an array?

internal/pkg/api/handleEnroll.go
```diff
@@ -96,12 +95,27 @@ func (et *EnrollerT) handleEnroll(zlog zerolog.Logger, w http.ResponseWriter, r
 	return writeResponse(zlog, w, resp, ts)
 }
 
-func (et *EnrollerT) processRequest(zlog zerolog.Logger, w http.ResponseWriter, r *http.Request, rb *rollback.Rollback, enrollmentAPIKeyID, ver string) (*EnrollResponse, error) {
+func (et *EnrollerT) processRequest(zlog zerolog.Logger, w http.ResponseWriter, r *http.Request, rb *rollback.Rollback, enrollmentAPIKey *apikey.APIKey, ver string) (*EnrollResponse, error) {
+	var enrollAPI *model.EnrollmentAPIKey
+	if et.cfg.StaticPolicyTokens.Enabled {
+		// Validate that an enrollment record exists for a key with this id.
+		if policy, ok := et.cfg.StaticPolicyTokens.PolicyTokens[enrollmentAPIKey.Key]; ok {
+			enrollAPI = &model.EnrollmentAPIKey{
```
michel-laterman, line 103: We probably want a debug message at this point as well.

internal/pkg/api/handleEnroll.go
```diff
@@ -96,12 +95,27 @@ func (et *EnrollerT) handleEnroll(zlog zerolog.Logger, w http.ResponseWriter, r
 	return writeResponse(zlog, w, resp, ts)
 }
 
-func (et *EnrollerT) processRequest(zlog zerolog.Logger, w http.ResponseWriter, r *http.Request, rb *rollback.Rollback, enrollmentAPIKeyID, ver string) (*EnrollResponse, error) {
+func (et *EnrollerT) processRequest(zlog zerolog.Logger, w http.ResponseWriter, r *http.Request, rb *rollback.Rollback, enrollmentAPIKey *apikey.APIKey, ver string) (*EnrollResponse, error) {
+	var enrollAPI *model.EnrollmentAPIKey
+	if et.cfg.StaticPolicyTokens.Enabled {
```
michel-laterman, line 100: I think we should try to keep our authentication handling in `handleEnroll` (currently line 76), we may need to move some existing logic out of `processRequest` to do so

michel-laterman (commented):
I agree with what @joshdover has stated, we should not accept enroll requests if the policy does not exist.

I know that we're adding this to support our cloud efforts, but it's also likely that this feature will be used in infrastructure as code deployments by customers so we'll need to add user facing documentation and guidelines as well.

internal/pkg/config/input.go
```diff
@@ -56,21 +58,30 @@ func (c *ServerBulk) InitDefaults() {
...
+	Timeouts           ServerTimeouts          `config:"timeouts"`
+	Profiler           ServerProfiler          `config:"profiler"`
+	CompressionLevel   int                     `config:"compression_level"`
+	CompressionThresh  int                     `config:"compression_threshold"`
+	Limits             ServerLimits            `config:"limits"`
+	Runtime            Runtime                 `config:"runtime"`
+	Bulk               ServerBulk              `config:"bulk"`
+	GC                 GC                      `config:"gc"`
+	Instrumentation    Instrumentation         `config:"instrumentation"`
+	StaticPolicyTokens StaticPolicyTokens      `config:"static_policy_tokens"`
```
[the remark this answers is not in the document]
olegsu: >I think that having enrollment tokens as a list as @joshdover suggested is the most straightforward.
And we should add a description in fleet-server.reference.yml.

Thank you for the feedback, added the reference to `fleet-server.reference.yml`

>I don't think we need to be concerned with list/map performance implications on our initial implementation, it can be changed later if we need to.

I dont think that it would be that easy to change in the future but I have updated it to hold a slice.

internal/pkg/api/handleEnroll.go
olegsu: This is done to prevent the call to elastic to fetch the token since it does not really exists there.

olegsu:
@joshdover @michel-laterman please have another look at this, I have added e2e and other tests as well.

@jkakavas, we discussed this feature also with @eyalkraft, @colings86, and @sebgl. It is important to bring this to your knowledge as it added an additional option to enroll in the fleet.

internal/pkg/api/handleEnroll_test.go
```diff
@@ -90,5 +95,182 @@ func TestEnroll(t *testing.T) {
...
+				cache:  cacheBuilder(),
+				bulker: bulkerBuilder(),
+			},
+			args: args{
+				enrollmentAPIKey: &apikey.APIKey{},
+			},
+			want:    nil,
+			wantErr: false,
+		},
+		// flaky test
```
olegsu, line 175: For some reason, this is a flaky one.
We see logs such as
```bash
{"level":"trace","ok":true,"id":"dummy-policy","cost":16,"ttl":3600000,"time":"2023-06-20T12:19:04+03:00","message":"Policy cache SET"}
{"level":"trace","id":"dummy-policy","time":"2023-06-20T12:19:04+03:00","message":"Policy cache MISS"}
```
Where the cache was set, but it was missed later.

internal/pkg/api/handleEnroll_test.go
```diff
@@ -90,5 +95,185 @@ func TestEnroll(t *testing.T) {
...
+				bulker: bulkerBuilder(),
+			},
+			args: args{
+				enrollmentAPIKey: &apikey.APIKey{},
+			},
+			want:    nil,
+			wantErr: false,
+		},
+		{
+			name: "found in cache13",
```
michel-laterman, line 180: cache13?

michel-laterman, line 180:
```suggestion
			name: "policy found in cache",
```
(nit) makes test name a little clearer

internal/pkg/api/handleEnroll_test.go
```diff
@@ -90,5 +95,185 @@ func TestEnroll(t *testing.T) {
...
+			},
+			want: &model.EnrollmentAPIKey{
+				APIKey:   "abcdefg",
+				Active:   true,
+				PolicyID: "dummy-policy",
+			},
+			wantErr: false,
+		},
+		{
+			name: "found in es",
```
michel-laterman, line 209:
```suggestion
			name: "policy found in es",
```
(nit)

internal/pkg/api/handleEnroll_test.go
```diff
@@ -90,5 +95,185 @@ func TestEnroll(t *testing.T) {
...
+			},
+			want: &model.EnrollmentAPIKey{
+				APIKey:   "abcdefg",
+				Active:   true,
+				PolicyID: "dummy-policy",
+			},
+			wantErr: false,
+		},
+		{
+			name: "not found",
```
michel-laterman, line 238:
```suggestion
			name: "policy not found",
```
(nit)

internal/pkg/api/handleEnroll.go
```diff
@@ -122,7 +131,54 @@ func (et *EnrollerT) processRequest(zlog zerolog.Logger, w http.ResponseWriter,
...
-	return et._enroll(r.Context(), rb, zlog, req, erec.PolicyID, ver)
+	return et._enroll(r.Context(), rb, zlog, req, enrollAPI.PolicyID, ver)
+}
+
+func (et *EnrollerT) fetchStaticTokenPolicy(ctx context.Context, zlog zerolog.Logger, enrollmentAPIKey *apikey.APIKey) (*model.EnrollmentAPIKey, error) {
+	if !et.cfg.StaticPolicyTokens.Enabled {
+		return nil, nil
+	}
+
+	zlog.Info().Msgf("Checking static enrollment token %s", enrollmentAPIKey.Key)
```
michel-laterman, line 142: (nit) change this to a debug and put an info log if a static token is found

michel-laterman (approved):
Thanks for adding an e2e test!
We just need a changelog fragment for this as well.

Otherwise I mostly have nitpicks

joshdover (commented):
Changes LGTM - my only concern right now is that adding the caching in the same PR seems unnecessary. I'd prefer to be able to split that change out separately so we can revert either change independent of one another.

olegsu:
> Changes LGTM - my only concern right now is that adding the caching in the same PR seems unnecessary. I'd prefer to be able to split that change out separately so we can revert either change independent of one another, if necessary.

Thanks for the review, I will remove the caching

joshdover (approved):

olegsu:
Merging this after a discussion we have with @jkakavas, @michel-laterman, and @eyalkraft 
There are things that we need to keep in mind for future releases 
1. As @joshdover suggested, to come up with an additional strategy to block the usage of the tokens to CIDR ranges
2. Using a cache to reduce the number of calls to ES for enrollment

PR #2538, opened 2023-04-26
fleet-server blackbox e2e tests

michel-laterman:
## What is the problem this PR solves?

It is difficult to test fleet-server's initialization process, especially when running under the agent as we tend to fake a lot of things.
Our current integration test suite also mocks a lot of calls so it is difficult to get a full picture of everything we test.
We need a way to ensure that fleet-server operates as expected if it runs stand-alone or under the agent.

## How does this PR solve the problem?

We add a new end-to-end test harness for black box tests.
At a high level the end to end test setup will:
* compile the fleet-server locally with the go build `-cover` flag introduced in go 1.20
* pull an elastic-agent docker image and inject a locally built fleet-server into it
* provision a CA and cert for fleet-server with an encrypted private key
* start an elasticsearch docker instance with the integration test harness
* start a kibana instance in the docker integration test network

All e2e tests run the fleet-server binary with the `GOCOVERDIR` env var set so coverage data is generated and written to `build/e2e-cover`. See New [go integration test coverage](https://go.dev/testing/coverage/)  for more information.

We have 3 test suites defined that run the fleet-server in different ways:
* as a stand-alone binary
* under the elastic-agent in a container
* ~from an elastic-agent installed on the host (if no agent is detected on the host)~ (disabled as CI environment is tricky to make this run)

We have two basic tests per suite:
* fleet-server starts an http API listener when given the service_token value
* fleet-server starts an https API listener with an encrypted private key (and passphrase file), and the service_token secret file

Future work can include updating the endpoints tested.

## How to test this PR locally

```sh
make test-e2e
```

## Related issues

- Closes https://github.com/elastic/elastic-agent/issues/2501

PR #2655, opened 2023-06-01
Using enrollment Id in agent enroll to delete duplicate agent on failed enroll

### LINKED ISSUE #2254: Duplicate agents can be created during enrollment at scale
- Version: Likely all versions
- Steps to Reproduce: 
1. Enroll a high number of agents at once against a Fleet Server instance, must be more agents than the Fleet Server can handle, triggering rate limiting
2. Observe that there may be some duplicate agents, with some stuck in "updating" state

image

This can happen because the first attempt to enroll appears to fail from the Agent's perspective (eg. due to a request timeout), but Fleet Server already created a new agent record. The agent then retries the enrollment, creating a new agent record (and ID) for the same hostname, resulting in duplicate agents in the UI. Because the first agent ID was never checked in, the first agent will stay in "updating" state.

Ideally, this is automatically detected or prevented by Fleet Server somehow. In the meantime, the [inactivity timeout feature](https://github.com/elastic/kibana/issues/143455) should phase these duplicates out of the default UI views after some time.

Note: we have not actually verified this happens with real Agents yet, only with Horde-emulated agents.

juliaElastic:

In scale tests we were seeing duplicate agents when the enrollment failed for the first time. This pr fixes that. 

Support `enrollment_id` in enrollment request, which is a unique enrollment identifier.  When fleet server finds the `enrollment_id` in `.fleet-agents`, it deletes the old agents and invalidates the old API key.

This solves the duplicate agent issue in case of a failed first enrollment (for example horde didn't receive the enrollment response for the first time and tries to enroll again).

- call enroll API twice with the same `enrollment_id`
- verify that one agent is enrolled at the end with the latter `enrolled_at` time and healthy

```
   curl -sk -XPOST -H 'content-type:application/json'  \
   -H "Authorization: ApiKey myapikey" \
   http://localhost:8221/api/fleet/agents/enroll  -d '
   {
      "type": "PERMANENT",
      "enrollment_id": "1234",
      "metadata": {
         "user_provided": {},
         "local": {},
         "tags": []
      }
   }' 
```

TODO integration test, changelog

## Design Checklist

## Checklist

## Related issues

Closes #2254


nchaulet:
@juliaElastic should we add another condition than the shared_id, like the policy revision index is not set? to be sure that we do not unenroll agent that successfully enrolled?

juliaElastic:
> should we add another condition than the shared_id, like the policy revision index is not set? to be sure that we do not unenroll agent that successfully enrolled?

We can add a check, though I'm not sure if it still solves the issue where the agent doesn't get the created API key at first. I think it should be fine if the policy revision index is updated on successful checkin.

model/openapi.yml
```diff
@@ -140,9 +140,7 @@ components:
           type: string
           description: |
             The shared ID of the agent.
-            To support pre-existing installs.
-
-            Never implemented.
+            To replace an agent on enroll fail.
```
joshdover, line 143: I think we need to 'undeprecate' this field on line 139

joshdover (changes requested):
Implementation looks good, but this needs to be tested.

I get the reasoning behind using the existing field which had this same intention, but I don't really like the name of it. Seems like it should be something like `enrollment_id` or similar. Thoughts?

juliaElastic:
> Implementation looks good, but this needs to be tested.
> 
> I get the reasoning behind using the existing field which had this same intention, but I don't really like the name of it. Seems like it should be something like `enrollment_id` or similar. Thoughts?

I agree that calling it `enrollment_id` would be more meaningful, I can rename it, unfortunately it needs a change in es repo as well.

joshdover:
Thanks for renaming, I think it will be more clear. LGTM but still needs unit tests 👍

juliaElastic:
It seems that the SonarCloud check is failing the build. I had a check and we require [20% code coverage](https://sonarcloud.io/summary/new_code?id=elastic_fleet-server&pullRequest=2655) on new code.

I can add new unit tests for existing uncovered code, though wondering if we should exclude `*.gen.go` files, it seems exceessive to try to add unit tests on openapi or types.
We also have [e2e tests](https://github.com/elastic/fleet-server/pull/2538) on the client API (`client.gen.go`) that is not included in the coverage.
@jlind23 @michel-laterman @alexsapran 

image

michel-laterman:
I think we should exclude the `*.gen.go` files

michel-laterman (changes requested):
I have a few thoughts on this

First, instead of adding an enrollmentID into the body I think we should consider adding a `PUT /api/fleet/agents/{id}` request as it is more explicit in it's intentions. 

Second, we really need to consider the fact that this new option can be severely misused by an attacker; If someone has access to an enrollment token they can invalidate an any agent's API key, and and remove any references from fleet? Or does the policy revision check prevent this? (we should add a test case to verify)

nchaulet:
> Or does the policy revision check prevent this? 

At first I thought the policy revision check will prevent it, but they may be one case during policy reassign where we set back the policy revision idx to 0. We may want a stronger condition here.

internal/pkg/api/handleEnroll.go
```diff
@@ -146,6 +154,30 @@ func (et *EnrollerT) _enroll(
 	}
 
 	agentID := u.String()
+	// only delete existing agent if it never checked in
```
michel-laterman, line 157: What do we do if the client gives an enrollmentID for an agent that has checked in?
Would it result in two agents with different IDs that share the same enrollmentID?
Can you add the behaviour to the description of the attribute in the openapi spec?

internal/pkg/server/fleet_integration_test.go
```diff
@@ -541,6 +543,80 @@ func Test_SmokeTest_Agent_Calls(t *testing.T) {
...
+	require.True(t, ok, "expected attribute item to be an object")
+	agentID := mm["id"]
+	return agentID.(string)
+}
+
+func Test_Agent_Enrollment_Id(t *testing.T) {
+	enrollBodyWEnrollmentID := `{
+	    "type": "PERMANENT",
+	    "shared_id": "",
+		"enrollment_id": "123456",
```
michel-laterman, line 576:
```suggestion
	    "enrollment_id": "123456",
```
(nit)

michel-laterman (approved):
looks good, the checkin value should make sure it does not replace a running agent.
I just think that we should be a little more clear in our docs with the expected behaviour when using this new attribute

We should hold off on the merge until the es changes are in and the tests pass

internal/pkg/api/handleEnroll.go
juliaElastic: Yes, for an agent that has checked in, there would be 2 agents with the same enrollment_id.

internal/pkg/api/handleEnroll.go
```diff
@@ -146,6 +154,30 @@ func (et *EnrollerT) _enroll(
...
+		err := invalidateAPIKey(ctx, zlog, et.bulker, agent.AccessAPIKeyID)
+		if err != nil {
+			zlog.Debug().
+				Str("EnrollmentId", enrollmentID).
+				Str("AgentId", agent.Id).
+				Str("APIKeyID", agent.AccessAPIKeyID).
+				Msg("Error when trying to invalidate API key of old agent with enrollment id")
+		}
+		// delete existing agent to recreate with new api key
+		err = deleteAgent(ctx, zlog, et.bulker, agent.Id)
```
cmacknz, line 173: If either of these operations fail I think it would be better to fail the entire request with an InternalServerError?

The agent is able to safely retry the enrollment on failure. If you just proceed here you risk leaving the API key valid or not deleting the agent and ending up with a duplicate.

Either of these operations has a low be realistic chance of failure in the real world. I would just fail the enrollment request and let the agent retry to keep the system in a consistent state.

internal/pkg/api/handleEnroll.go
```diff
@@ -132,9 +132,17 @@ func (et *EnrollerT) _enroll(
...
 
-	if req.SharedId != "" {
-		// TODO: Support pre-existing install
-		return nil, errors.New("preexisting install not yet supported")
+	var agent model.Agent
+	var enrollmentID string
+	if req.EnrollmentId != nil {
+		enrollmentID = *req.EnrollmentId
+		var err error
+		agent, err = dl.FindAgent(ctx, et.bulker, dl.QueryAgentByEnrollmentID, dl.FieldEnrollmentID, enrollmentID)
```
cmacknz, line 140: This query can also fail regardless of whether an agent with this enrollment ID actually exists. Elasticsearch could be temporarily unavailable or returning 429s. Proceeding and duplicating the agent seems like the wrong choice.

As I mentioned below, I would just return the error and let the agent retry.

internal/pkg/api/handleEnroll.go
cmacknz: Under what circumstances do we think this could happen? A user executing `elastic-agent enroll` on an already enrolled agent?

cmacknz (changes requested):
Requesting changes to make sure we handle failures during enrollment properly. The whole reason we have this implementation is because parts of the enrollment process can fail at scale and require retries from the agent. We need to make sure we don't introduce a different version of this problem by silently ignoring errors that lead to duplicate agents.

internal/pkg/api/handleEnroll.go
juliaElastic: Agent currently doesn't use the `enrollment_id`. This is a safety measure against someone making a destructive action by deleting a healthy agent with an API call with an existing `enrollment_id`

juliaElastic:
Finally seeing the new mapping in the latest es snapshot locally, and could test locally.
image

juliaElastic:
Added a few log messages and the logic seems to be stuck in the loop that tries to [read the api key](https://github.com/elastic/fleet-server/blob/42af87680750a003a0b54056a1b86044928071f7/internal/pkg/api/handleEnroll.go#L285) before invalidation:

```
{"log.level":"debug","ecs.version":"1.6.0","service.name":"fleet-server","http.request.id":"5f167e84-111d-46e7-82a1-6b0bc8daa3b7","server.address":"","mod":"enroll","fleet.enroll.apikey.id":"CTq_oIgBf48Nt7iUw3or","EnrollmentId":"123456","AgentId":"f24d8919-0d7b-48c0-a462-bed3cb03ed64","APIKeyID":"trsNoIgBfXEwhkEH78Ee","@timestamp":"2023-06-09T15:21:43.078Z","message":"Invalidate old api key and remove existing agent with the same enrollment_id"}
{"log.level":"debug","ecs.version":"1.6.0","service.name":"fleet-server","http.request.id":"5f167e84-111d-46e7-82a1-6b0bc8daa3b7","server.address":"","mod":"enroll","fleet.enroll.apikey.id":"CTq_oIgBf48Nt7iUw3or","fleet.apikey.id":"trsNoIgBfXEwhkEH78Ee","@timestamp":"2023-06-09T15:21:43.078Z","message":"api key read"}
{"log.level":"debug","ecs.version":"1.6.0","service.name":"fleet-server","http.request.id":"5f167e84-111d-46e7-82a1-6b0bc8daa3b7","server.address":"","mod":"enroll","fleet.enroll.apikey.id":"CTq_oIgBf48Nt7iUw3or","fleet.apikey.id":"trsNoIgBfXEwhkEH78Ee","@timestamp":"2023-06-09T15:21:44.081Z","message":"api key read"}
{"log.level":"info","ecs.version":"1.6.0","service.name":"fleet-server","ctx":"policy leader manager","policy_id":"191d062d-fb12-42e3-a587-5403497456db","count":0,"coordinator":"v0","@timestamp":"2023-06-09T15:21:44.4Z","message":"Starting policy coordinator"}
{"log.level":"info","ecs.version":"1.6.0","service.name":"fleet-server","ctx":"policy leader manager","policy_id":"0ab91865-d403-4d6d-9885-21cb01295013","count":0,"coordinator":"v0","@timestamp":"2023-06-09T15:21:44.4Z","message":"Starting policy coordinator"}
{"log.level":"info","ecs.version":"1.6.0","service.name":"fleet-server","ctx":"policy leader manager","policy_id":"d19debb0-06bc-11ee-ae0f-6b1e7c8504e5","count":0,"coordinator":"v0","@timestamp":"2023-06-09T15:21:44.4Z","message":"Starting policy coordinator"}
{"log.level":"info","ecs.version":"1.6.0","service.name":"fleet-server","ctx":"policy leader manager","policy_id":"2b7baf84-4283-4f39-bf9f-93855af34e03","count":0,"coordinator":"v0","@timestamp":"2023-06-09T15:21:44.4Z","message":"Starting policy coordinator"}
{"log.level":"info","ecs.version":"1.6.0","service.name":"fleet-server","ctx":"policy leader manager","policy_id":"5820402c-6c6a-4346-ac6a-21315a2e67da","count":0,"coordinator":"v0","@timestamp":"2023-06-09T15:21:44.4Z","message":"Starting policy coordinator"}
{"log.level":"info","ecs.version":"1.6.0","service.name":"fleet-server","ctx":"policy leader manager","policy_id":"5ea6647b-f91b-478b-8dbc-ce593057ffbc","count":0,"coordinator":"v0","@timestamp":"2023-06-09T15:21:44.4Z","message":"Starting policy coordinator"}
{"log.level":"info","ecs.version":"1.6.0","service.name":"fleet-server","ctx":"policy leader manager","policy_id":"f6e20db5-cf85-4c04-96a6-84c769ddcd16","count":0,"coordinator":"v0","@timestamp":"2023-06-09T15:21:44.4Z","message":"Starting policy coordinator"}
{"log.level":"info","ecs.version":"1.6.0","service.name":"fleet-server","ctx":"policy leader manager","policy_id":"333646d3-c5c5-4e87-900a-593bd2c7c328","count":0,"coordinator":"v0","@timestamp":"2023-06-09T15:21:44.4Z","message":"Starting policy coordinator"}
{"log.level":"info","ecs.version":"1.6.0","service.name":"fleet-server","ctx":"policy leader manager","policy_id":"72ea0c21-7a25-4de2-836c-7227f75a28d2","count":0,"coordinator":"v0","@timestamp":"2023-06-09T15:21:44.4Z","message":"Starting policy coordinator"}
{"log.level":"info","ecs.version":"1.6.0","service.name":"fleet-server","ctx":"policy leader manager","policy_id":"2b81dfa8-0554-4086-9375-a355bafa2709","count":0,"coordinator":"v0","@timestamp":"2023-06-09T15:21:44.4Z","message":"Starting policy coordinator"}
{"log.level":"info","ecs.version":"1.6.0","service.name":"fleet-server","ctx":"policy leader manager","policy_id":"fleet-server-policy","count":0,"coordinator":"v0","@timestamp":"2023-06-09T15:21:44.4Z","message":"Starting policy coordinator"}
{"log.level":"info","ecs.version":"1.6.0","service.name":"fleet-server","ctx":"policy leader manager","policy_id":"79f76396-ba20-47ed-81d6-9a7e674c927f","count":0,"coordinator":"v0","@timestamp":"2023-06-09T15:21:44.4Z","message":"Starting policy coordinator"}
{"log.level":"info","ecs.version":"1.6.0","service.name":"fleet-server","ctx":"policy leader manager","policy_id":"baf068ce-3008-4211-893c-eb8d4cd36f4f","count":0,"coordinator":"v0","@timestamp":"2023-06-09T15:21:44.4Z","message":"Starting policy coordinator"}
{"log.level":"info","ecs.version":"1.6.0","service.name":"fleet-server","ctx":"policy leader manager","policy_id":"818f7727-f1dc-4449-a946-2d5836d2680f","count":0,"coordinator":"v0","@timestamp":"2023-06-09T15:21:44.4Z","message":"Starting policy coordinator"}
{"log.level":"info","ecs.version":"1.6.0","service.name":"fleet-server","ctx":"policy leader manager","policy_id":"79f76396-ba20-47ed-81d6-9a7e674c927f","revision_idx":1,"coordinator_idx":1,"revision_id":1,"@timestamp":"2023-06-09T15:21:44.781Z","message":"Policy coordinator added a new policy revision"}
{"log.level":"debug","ecs.version":"1.6.0","service.name":"fleet-server","index":".fleet-policies","ctx":"index monitor","checkpoints":[30],"@timestamp":"2023-06-09T15:21:44.783Z","message":"updated checkpoint"}
{"log.level":"debug","ecs.version":"1.6.0","service.name":"fleet-server","index":".fleet-policies","ctx":"index monitor","count":1,"@timestamp":"2023-06-09T15:21:44.783Z","message":"hits found after notify"}
{"log.level":"debug","ecs.version":"1.6.0","service.name":"fleet-server","index":".fleet-policies","ctx":"index monitor","checkpoints":[30],"@timestamp":"2023-06-09T15:21:44.783Z","message":"updated checkpoint"}
{"log.level":"info","ecs.version":"1.6.0","service.name":"fleet-server","state":"HEALTHY","@timestamp":"2023-06-09T15:21:44.783Z","message":"Running on default policy with Fleet Server integration"}
{"log.level":"debug","ecs.version":"1.6.0","service.name":"fleet-server","@timestamp":"2023-06-09T15:21:44.783Z","message":"Policy self monitor exited"}
{"log.level":"info","ecs.version":"1.6.0","service.name":"fleet-server","ctx":"policy agent monitor","fleet.policy.id":"79f76396-ba20-47ed-81d6-9a7e674c927f","rev":1,"coord":1,"fleet.policy.id":"79f76396-ba20-47ed-81d6-9a7e674c927f","@timestamp":"2023-06-09T15:21:44.783Z","message":"New policy found on update and added"}
{"log.level":"debug","ecs.version":"1.6.0","service.name":"fleet-server","http.request.id":"5f167e84-111d-46e7-82a1-6b0bc8daa3b7","server.address":"","mod":"enroll","fleet.enroll.apikey.id":"CTq_oIgBf48Nt7iUw3or","fleet.apikey.id":"trsNoIgBfXEwhkEH78Ee","@timestamp":"2023-06-09T15:21:45.083Z","message":"api key read"}
{"log.level":"debug","ecs.version":"1.6.0","service.name":"fleet-server","http.request.id":"5f167e84-111d-46e7-82a1-6b0bc8daa3b7","server.address":"","mod":"enroll","fleet.enroll.apikey.id":"CTq_oIgBf48Nt7iUw3or","fleet.apikey.id":"trsNoIgBfXEwhkEH78Ee","@timestamp":"2023-06-09T15:21:46.084Z","message":"api key read"}
{"log.level":"debug","ecs.version":"1.6.0","service.name":"fleet-server","http.request.id":"5f167e84-111d-46e7-82a1-6b0bc8daa3b7","server.address":"","mod":"enroll","fleet.enroll.apikey.id":"CTq_oIgBf48Nt7iUw3or","fleet.apikey.id":"trsNoIgBfXEwhkEH78Ee","@timestamp":"2023-06-09T15:21:47.086Z","message":"api key read"}
{"log.level":"debug","ecs.version":"1.6.0","service.name":"fleet-server","http.request.id":"5f167e84-111d-46e7-82a1-6b0bc8daa3b7","server.address":"","mod":"enroll","fleet.enroll.apikey.id":"CTq_oIgBf48Nt7iUw3or","fleet.apikey.id":"trsNoIgBfXEwhkEH78Ee","@timestamp":"2023-06-09T15:21:48.088Z","message":"api key read"}
{"log.level":"debug","ecs.version":"1.6.0","service.name":"fleet-server","http.request.id":"5f167e84-111d-46e7-82a1-6b0bc8daa3b7","server.address":"","mod":"enroll","fleet.enroll.apikey.id":"CTq_oIgBf48Nt7iUw3or","fleet.apikey.id":"trsNoIgBfXEwhkEH78Ee","@timestamp":"2023-06-09T15:21:49.09Z","message":"api key read"}
{"log.level":"debug","ecs.version":"1.6.0","service.name":"fleet-server","http.request.id":"5f167e84-111d-46e7-82a1-6b0bc8daa3b7","server.address":"","mod":"enroll","fleet.enroll.apikey.id":"CTq_oIgBf48Nt7iUw3or","fleet.apikey.id":"trsNoIgBfXEwhkEH78Ee","@timestamp":"2023-06-09T15:21:50.091Z","message":"api key read"}
{"log.level":"debug","ecs.version":"1.6.0","service.name":"fleet-server","http.request.id":"5f167e84-111d-46e7-82a1-6b0bc8daa3b7","server.address":"","mod":"enroll","fleet.enroll.apikey.id":"CTq_oIgBf48Nt7iUw3or","fleet.apikey.id":"trsNoIgBfXEwhkEH78Ee","@timestamp":"2023-06-09T15:21:51.093Z","message":"api key read"}
{"log.level":"debug","ecs.version":"1.6.0","service.name":"fleet-server","http.request.id":"5f167e84-111d-46e7-82a1-6b0bc8daa3b7","server.address":"","mod":"enroll","fleet.enroll.apikey.id":"CTq_oIgBf48Nt7iUw3or","fleet.apikey.id":"trsNoIgBfXEwhkEH78Ee","@timestamp":"2023-06-09T15:21:52.095Z","message":"api key read"}
{"log.level":"debug","ecs.version":"1.6.0","service.name":"fleet-server","schedule":"fleet actions cleanup","interval":3600000,"@timestamp":"2023-06-09T15:21:52.269Z","message":"started"}
{"log.level":"debug","ecs.version":"1.6.0","service.name":"fleet-server","ctx":"fleet actions cleanup","interval":"now-30d","@timestamp":"2023-06-09T15:21:52.27Z","message":"delete expired actions"}
{"log.level":"debug","ecs.version":"1.6.0","service.name":"fleet-server","index":".fleet-actions","@timestamp":"2023-06-09T15:21:52.272Z","message":"index not found"}
{"log.level":"debug","ecs.version":"1.6.0","service.name":"fleet-server","ctx":"fleet actions cleanup","interval":"now-30d","count":0,"@timestamp":"2023-06-09T15:21:52.272Z","message":"deleted expired actions"}
{"log.level":"debug","ecs.version":"1.6.0","service.name":"fleet-server","schedule":"fleet actions cleanup","@timestamp":"2023-06-09T15:21:52.272Z","message":"finished"}
{"log.level":"debug","ecs.version":"1.6.0","service.name":"fleet-server","http.request.id":"5f167e84-111d-46e7-82a1-6b0bc8daa3b7","server.address":"","mod":"enroll","fleet.enroll.apikey.id":"CTq_oIgBf48Nt7iUw3or","fleet.apikey.id":"trsNoIgBfXEwhkEH78Ee","@timestamp":"2023-06-09T15:21:53.098Z","message":"api key read"}
{"log.level":"debug","ecs.version":"1.6.0","service.name":"fleet-server","http.request.id":"5f167e84-111d-46e7-82a1-6b0bc8daa3b7","server.address":"","mod":"enroll","fleet.enroll.apikey.id":"CTq_oIgBf48Nt7iUw3or","fleet.apikey.id":"trsNoIgBfXEwhkEH78Ee","@timestamp":"2023-06-09T15:21:54.101Z","message":"api key read"}
{"log.level":"debug","ecs.version":"1.6.0","service.name":"fleet-server","http.request.id":"5f167e84-111d-46e7-82a1-6b0bc8daa3b7","server.address":"","mod":"enroll","fleet.enroll.apikey.id":"CTq_oIgBf48Nt7iUw3or","fleet.apikey.id":"trsNoIgBfXEwhkEH78Ee","@timestamp":"2023-06-09T15:21:55.104Z","message":"api key read"}
{"log.level":"debug","ecs.version":"1.6.0","service.name":"fleet-server","http.request.id":"5f167e84-111d-46e7-82a1-6b0bc8daa3b7","server.address":"","mod":"enroll","fleet.enroll.apikey.id":"CTq_oIgBf48Nt7iUw3or","fleet.apikey.id":"trsNoIgBfXEwhkEH78Ee","@timestamp":"2023-06-09T15:21:56.106Z","message":"api key read"}
{"log.level":"debug","ecs.version":"1.6.0","service.name":"fleet-server","http.request.id":"5f167e84-111d-46e7-82a1-6b0bc8daa3b7","server.address":"","mod":"enroll","fleet.enroll.apikey.id":"CTq_oIgBf48Nt7iUw3or","fleet.apikey.id":"trsNoIgBfXEwhkEH78Ee","@timestamp":"2023-06-09T15:21:57.109Z","message":"api key read"}
{"log.level":"debug","ecs.version":"1.6.0","service.name":"fleet-server","http.request.id":"5f167e84-111d-46e7-82a1-6b0bc8daa3b7","server.address":"","mod":"enroll","fleet.enroll.apikey.id":"CTq_oIgBf48Nt7iUw3or","fleet.apikey.id":"trsNoIgBfXEwhkEH78Ee","@timestamp":"2023-06-09T15:21:58.112Z","message":"api key read"}
{"log.level":"debug","ecs.version":"1.6.0","service.name":"fleet-server","http.request.id":"5f167e84-111d-46e7-82a1-6b0bc8daa3b7","server.address":"","mod":"enroll","fleet.enroll.apikey.id":"CTq_oIgBf48Nt7iUw3or","fleet.apikey.id":"trsNoIgBfXEwhkEH78Ee","@timestamp":"2023-06-09T15:21:59.115Z","message":"api key read"}
{"log.level":"debug","ecs.version":"1.6.0","service.name":"fleet-server","http.request.id":"5f167e84-111d-46e7-82a1-6b0bc8daa3b7","server.address":"","mod":"enroll","fleet.enroll.apikey.id":"CTq_oIgBf48Nt7iUw3or","fleet.apikey.id":"trsNoIgBfXEwhkEH78Ee","@timestamp":"2023-06-09T15:22:00.117Z","message":"api key read"}
{"log.level":"debug","ecs.version":"1.6.0","service.name":"fleet-server","http.request.id":"5f167e84-111d-46e7-82a1-6b0bc8daa3b7","server.address":"","mod":"enroll","fleet.enroll.apikey.id":"CTq_oIgBf48Nt7iUw3or","fleet.apikey.id":"trsNoIgBfXEwhkEH78Ee","@timestamp":"2023-06-09T15:22:01.119Z","message":"api key read"}
{"log.level":"debug","ecs.version":"1.6.0","service.name":"fleet-server","http.request.id":"5f167e84-111d-46e7-82a1-6b0bc8daa3b7","server.address":"","mod":"enroll","fleet.enroll.apikey.id":"CTq_oIgBf48Nt7iUw3or","fleet.apikey.id":"trsNoIgBfXEwhkEH78Ee","@timestamp":"2023-06-09T15:22:02.121Z","message":"api key read"}
{"log.level":"debug","ecs.version":"1.6.0","service.name":"fleet-server","http.request.id":"5f167e84-111d-46e7-82a1-6b0bc8daa3b7","server.address":"","mod":"enroll","fleet.enroll.apikey.id":"CTq_oIgBf48Nt7iUw3or","fleet.apikey.id":"trsNoIgBfXEwhkEH78Ee","@timestamp":"2023-06-09T15:22:03.123Z","message":"api key read"}
{"log.level":"debug","ecs.version":"1.6.0","service.name":"fleet-server","http.request.id":"5f167e84-111d-46e7-82a1-6b0bc8daa3b7","server.address":"","mod":"enroll","fleet.enroll.apikey.id":"CTq_oIgBf48Nt7iUw3or","fleet.apikey.id":"trsNoIgBfXEwhkEH78Ee","@timestamp":"2023-06-09T15:22:04.125Z","message":"api key read"}
{"log.level":"debug","ecs.version":"1.6.0","service.name":"fleet-server","ctx":"policy leader manager","@timestamp":"2023-06-09T15:22:04.402Z","message":"ensuring leadership of policies"}
{"log.level":"debug","ecs.version":"1.6.0","service.name":"fleet-server","http.request.id":"5f167e84-111d-46e7-82a1-6b0bc8daa3b7","server.address":"","mod":"enroll","fleet.enroll.apikey.id":"CTq_oIgBf48Nt7iUw3or","fleet.apikey.id":"trsNoIgBfXEwhkEH78Ee","@timestamp":"2023-06-09T15:22:05.128Z","message":"api key read"}
{"log.level":"debug","ecs.version":"1.6.0","service.name":"fleet-server","http.request.id":"5f167e84-111d-46e7-82a1-6b0bc8daa3b7","server.address":"","mod":"enroll","fleet.enroll.apikey.id":"CTq_oIgBf48Nt7iUw3or","fleet.apikey.id":"trsNoIgBfXEwhkEH78Ee","@timestamp":"2023-06-09T15:22:06.13Z","message":"api key read"}
{"log.level":"debug","ecs.version":"1.6.0","service.name":"fleet-server","http.request.id":"5f167e84-111d-46e7-82a1-6b0bc8daa3b7","server.address":"","mod":"enroll","fleet.enroll.apikey.id":"CTq_oIgBf48Nt7iUw3or","fleet.apikey.id":"trsNoIgBfXEwhkEH78Ee","@timestamp":"2023-06-09T15:22:07.132Z","message":"api key read"}
{"log.level":"debug","ecs.version":"1.6.0","service.name":"fleet-server","http.request.id":"5f167e84-111d-46e7-82a1-6b0bc8daa3b7","server.address":"","mod":"enroll","fleet.enroll.apikey.id":"CTq_oIgBf48Nt7iUw3or","fleet.apikey.id":"trsNoIgBfXEwhkEH78Ee","@timestamp":"2023-06-09T15:22:08.135Z","message":"api key read"}
{"log.level":"debug","ecs.version":"1.6.0","service.name":"fleet-server","http.request.id":"5f167e84-111d-46e7-82a1-6b0bc8daa3b7","server.address":"","mod":"enroll","fleet.enroll.apikey.id":"CTq_oIgBf48Nt7iUw3or","fleet.apikey.id":"trsNoIgBfXEwhkEH78Ee","@timestamp":"2023-06-09T15:22:09.137Z","message":"api key read"}
{"log.level":"debug","ecs.version":"1.6.0","service.name":"fleet-server","http.request.id":"5f167e84-111d-46e7-82a1-6b0bc8daa3b7","server.address":"","mod":"enroll","fleet.enroll.apikey.id":"CTq_oIgBf48Nt7iUw3or","fleet.apikey.id":"trsNoIgBfXEwhkEH78Ee","@timestamp":"2023-06-09T15:22:10.14Z","message":"api key read"}
{"log.level":"debug","ecs.version":"1.6.0","service.name":"fleet-server","http.request.id":"5f167e84-111d-46e7-82a1-6b0bc8daa3b7","server.address":"","mod":"enroll","fleet.enroll.apikey.id":"CTq_oIgBf48Nt7iUw3or","fleet.apikey.id":"trsNoIgBfXEwhkEH78Ee","@timestamp":"2023-06-09T15:22:11.141Z","message":"api key read"}
{"log.level":"debug","ecs.version":"1.6.0","service.name":"fleet-server","http.request.id":"5f167e84-111d-46e7-82a1-6b0bc8daa3b7","server.address":"","mod":"enroll","fleet.enroll.apikey.id":"CTq_oIgBf48Nt7iUw3or","fleet.apikey.id":"trsNoIgBfXEwhkEH78Ee","@timestamp":"2023-06-09T15:22:12.144Z","message":"api key read"}
{"log.level":"debug","ecs.version":"1.6.0","service.name":"fleet-server","http.request.id":"5f167e84-111d-46e7-82a1-6b0bc8daa3b7","server.address":"","mod":"enroll","fleet.enroll.apikey.id":"CTq_oIgBf48Nt7iUw3or","fleet.apikey.id":"trsNoIgBfXEwhkEH78Ee","@timestamp":"2023-06-09T15:22:13.146Z","message":"api key read"}
{"log.level":"debug","ecs.version":"1.6.0","service.name":"fleet-server","http.request.id":"5f167e84-111d-46e7-82a1-6b0bc8daa3b7","server.address":"","mod":"enroll","fleet.enroll.apikey.id":"CTq_oIgBf48Nt7iUw3or","fleet.apikey.id":"trsNoIgBfXEwhkEH78Ee","@timestamp":"2023-06-09T15:22:14.148Z","message":"api key read"}
{"log.level":"debug","ecs.version":"1.6.0","service.name":"fleet-server","http.request.id":"5f167e84-111d-46e7-82a1-6b0bc8daa3b7","server.address":"","mod":"enroll","fleet.enroll.apikey.id":"CTq_oIgBf48Nt7iUw3or","fleet.apikey.id":"trsNoIgBfXEwhkEH78Ee","@timestamp":"2023-06-09T15:22:15.15Z","message":"api key read"}
{"log.level":"debug","ecs.version":"1.6.0","service.name":"fleet-server","http.request.id":"5f167e84-111d-46e7-82a1-6b0bc8daa3b7","server.address":"","mod":"enroll","fleet.enroll.apikey.id":"CTq_oIgBf48Nt7iUw3or","fleet.apikey.id":"trsNoIgBfXEwhkEH78Ee","@timestamp":"2023-06-09T15:22:16.151Z","message":"api key read"}
{"log.level":"debug","ecs.version":"1.6.0","service.name":"fleet-server","http.request.id":"5f167e84-111d-46e7-82a1-6b0bc8daa3b7","server.address":"","mod":"enroll","fleet.enroll.apikey.id":"CTq_oIgBf48Nt7iUw3or","fleet.apikey.id":"trsNoIgBfXEwhkEH78Ee","@timestamp":"2023-06-09T15:22:17.153Z","message":"api key read"}
{"log.level":"debug","ecs.version":"1.6.0","service.name":"fleet-server","http.request.id":"5f167e84-111d-46e7-82a1-6b0bc8daa3b7","server.address":"","mod":"enroll","fleet.enroll.apikey.id":"CTq_oIgBf48Nt7iUw3or","fleet.apikey.id":"trsNoIgBfXEwhkEH78Ee","@timestamp":"2023-06-09T15:22:18.155Z","message":"api key read"}
{"log.level":"debug","ecs.version":"1.6.0","service.name":"fleet-server","http.request.id":"5f167e84-111d-46e7-82a1-6b0bc8daa3b7","server.address":"","mod":"enroll","fleet.enroll.apikey.id":"CTq_oIgBf48Nt7iUw3or","fleet.apikey.id":"trsNoIgBfXEwhkEH78Ee","@timestamp":"2023-06-09T15:22:19.158Z","message":"api key read"}
{"log.level":"debug","ecs.version":"1.6.0","service.name":"fleet-server","http.request.id":"5f167e84-111d-46e7-82a1-6b0bc8daa3b7","server.address":"","mod":"enroll","fleet.enroll.apikey.id":"CTq_oIgBf48Nt7iUw3or","fleet.apikey.id":"trsNoIgBfXEwhkEH78Ee","@timestamp":"2023-06-09T15:22:20.16Z","message":"api key read"}
{"log.level":"debug","ecs.version":"1.6.0","service.name":"fleet-server","http.request.id":"5f167e84-111d-46e7-82a1-6b0bc8daa3b7","server.address":"","mod":"enroll","fleet.enroll.apikey.id":"CTq_oIgBf48Nt7iUw3or","fleet.apikey.id":"trsNoIgBfXEwhkEH78Ee","@timestamp":"2023-06-09T15:22:21.162Z","message":"api key read"}
{"log.level":"debug","ecs.version":"1.6.0","service.name":"fleet-server","http.request.id":"5f167e84-111d-46e7-82a1-6b0bc8daa3b7","server.address":"","mod":"enroll","fleet.enroll.apikey.id":"CTq_oIgBf48Nt7iUw3or","fleet.apikey.id":"trsNoIgBfXEwhkEH78Ee","@timestamp":"2023-06-09T15:22:22.164Z","message":"api key read"}
{"log.level":"debug","ecs.version":"1.6.0","service.name":"fleet-server","http.request.id":"5f167e84-111d-46e7-82a1-6b0bc8daa3b7","server.address":"","mod":"enroll","fleet.enroll.apikey.id":"CTq_oIgBf48Nt7iUw3or","fleet.apikey.id":"trsNoIgBfXEwhkEH78Ee","@timestamp":"2023-06-09T15:22:23.166Z","message":"api key read"}
{"log.level":"debug","ecs.version":"1.6.0","service.name":"fleet-server","http.request.id":"5f167e84-111d-46e7-82a1-6b0bc8daa3b7","server.address":"","mod":"enroll","fleet.enroll.apikey.id":"CTq_oIgBf48Nt7iUw3or","fleet.apikey.id":"trsNoIgBfXEwhkEH78Ee","@timestamp":"2023-06-09T15:22:24.168Z","message":"api key read"}
{"log.level":"debug","ecs.version":"1.6.0","service.name":"fleet-server","http.request.id":"5f167e84-111d-46e7-82a1-6b0bc8daa3b7","server.address":"","mod":"enroll","fleet.enroll.apikey.id":"CTq_oIgBf48Nt7iUw3or","fleet.apikey.id":"trsNoIgBfXEwhkEH78Ee","@timestamp":"2023-06-09T15:22:25.17Z","message":"api key read"}
{"log.level":"debug","ecs.version":"1.6.0","service.name":"fleet-server","ctx":"policy leader manager","@timestamp":"2023-06-09T15:22:26.042Z","message":"ensuring leadership of policies"}
{"log.level":"debug","ecs.version":"1.6.0","service.name":"fleet-server","http.request.id":"5f167e84-111d-46e7-82a1-6b0bc8daa3b7","server.address":"","mod":"enroll","fleet.enroll.apikey.id":"CTq_oIgBf48Nt7iUw3or","fleet.apikey.id":"trsNoIgBfXEwhkEH78Ee","@timestamp":"2023-06-09T15:22:26.171Z","message":"api key read"}
{"log.level":"debug","ecs.version":"1.6.0","service.name":"fleet-server","http.request.id":"5f167e84-111d-46e7-82a1-6b0bc8daa3b7","server.address":"","mod":"enroll","fleet.enroll.apikey.id":"CTq_oIgBf48Nt7iUw3or","fleet.apikey.id":"trsNoIgBfXEwhkEH78Ee","@timestamp":"2023-06-09T15:22:27.173Z","message":"api key read"}
{"log.level":"debug","ecs.version":"1.6.0","service.name":"fleet-server","http.request.id":"5f167e84-111d-46e7-82a1-6b0bc8daa3b7","server.address":"","mod":"enroll","fleet.enroll.apikey.id":"CTq_oIgBf48Nt7iUw3or","fleet.apikey.id":"trsNoIgBfXEwhkEH78Ee","@timestamp":"2023-06-09T15:22:28.175Z","message":"api key read"}
{"log.level":"debug","ecs.version":"1.6.0","service.name":"fleet-server","http.request.id":"5f167e84-111d-46e7-82a1-6b0bc8daa3b7","server.address":"","mod":"enroll","fleet.enroll.apikey.id":"CTq_oIgBf48Nt7iUw3or","fleet.apikey.id":"trsNoIgBfXEwhkEH78Ee","@timestamp":"2023-06-09T15:22:29.177Z","message":"api key read"}
{"log.level":"debug","ecs.version":"1.6.0","service.name":"fleet-server","http.request.id":"5f167e84-111d-46e7-82a1-6b0bc8daa3b7","server.address":"","mod":"enroll","fleet.enroll.apikey.id":"CTq_oIgBf48Nt7iUw3or","fleet.apikey.id":"trsNoIgBfXEwhkEH78Ee","@timestamp":"2023-06-09T15:22:30.178Z","message":"api key read"}
{"log.level":"debug","ecs.version":"1.6.0","service.name":"fleet-server","http.request.id":"5f167e84-111d-46e7-82a1-6b0bc8daa3b7","server.address":"","mod":"enroll","fleet.enroll.apikey.id":"CTq_oIgBf48Nt7iUw3or","fleet.apikey.id":"trsNoIgBfXEwhkEH78Ee","@timestamp":"2023-06-09T15:22:31.18Z","message":"api key read"}
{"log.level":"debug","ecs.version":"1.6.0","service.name":"fleet-server","http.request.id":"5f167e84-111d-46e7-82a1-6b0bc8daa3b7","server.address":"","mod":"enroll","fleet.enroll.apikey.id":"CTq_oIgBf48Nt7iUw3or","fleet.apikey.id":"trsNoIgBfXEwhkEH78Ee","@timestamp":"2023-06-09T15:22:32.183Z","message":"api key read"}
{"log.level":"debug","ecs.version":"1.6.0","service.name":"fleet-server","http.request.id":"5f167e84-111d-46e7-82a1-6b0bc8daa3b7","server.address":"","mod":"enroll","fleet.enroll.apikey.id":"CTq_oIgBf48Nt7iUw3or","fleet.apikey.id":"trsNoIgBfXEwhkEH78Ee","@timestamp":"2023-06-09T15:22:33.185Z","message":"api key read"}
{"log.level":"debug","ecs.version":"1.6.0","service.name":"fleet-server","http.request.id":"5f167e84-111d-46e7-82a1-6b0bc8daa3b7","server.address":"","mod":"enroll","fleet.enroll.apikey.id":"CTq_oIgBf48Nt7iUw3or","fleet.apikey.id":"trsNoIgBfXEwhkEH78Ee","@timestamp":"2023-06-09T15:22:34.188Z","message":"api key read"}
{"log.level":"debug","ecs.version":"1.6.0","service.name":"fleet-server","http.request.id":"5f167e84-111d-46e7-82a1-6b0bc8daa3b7","server.address":"","mod":"enroll","fleet.enroll.apikey.id":"CTq_oIgBf48Nt7iUw3or","fleet.apikey.id":"trsNoIgBfXEwhkEH78Ee","@timestamp":"2023-06-09T15:22:35.19Z","message":"api key read"}
{"log.level":"debug","ecs.version":"1.6.0","service.name":"fleet-server","http.request.id":"5f167e84-111d-46e7-82a1-6b0bc8daa3b7","server.address":"","mod":"enroll","fleet.enroll.apikey.id":"CTq_oIgBf48Nt7iUw3or","fleet.apikey.id":"trsNoIgBfXEwhkEH78Ee","@timestamp":"2023-06-09T15:22:36.192Z","message":"api key read"}
{"log.level":"debug","ecs.version":"1.6.0","service.name":"fleet-server","http.request.id":"5f167e84-111d-46e7-82a1-6b0bc8daa3b7","server.address":"","mod":"enroll","fleet.enroll.apikey.id":"CTq_oIgBf48Nt7iUw3or","fleet.apikey.id":"trsNoIgBfXEwhkEH78Ee","@timestamp":"2023-06-09T15:22:37.193Z","message":"api key read"}
{"log.level":"debug","ecs.version":"1.6.0","service.name":"fleet-server","http.request.id":"5f167e84-111d-46e7-82a1-6b0bc8daa3b7","server.address":"","mod":"enroll","fleet.enroll.apikey.id":"CTq_oIgBf48Nt7iUw3or","fleet.apikey.id":"trsNoIgBfXEwhkEH78Ee","@timestamp":"2023-06-09T15:22:38.196Z","message":"api key read"}
{"log.level":"debug","ecs.version":"1.6.0","service.name":"fleet-server","http.request.id":"5f167e84-111d-46e7-82a1-6b0bc8daa3b7","server.address":"","mod":"enroll","fleet.enroll.apikey.id":"CTq_oIgBf48Nt7iUw3or","fleet.apikey.id":"trsNoIgBfXEwhkEH78Ee","@timestamp":"2023-06-09T15:22:39.199Z","message":"api key read"}
{"log.level":"debug","ecs.version":"1.6.0","service.name":"fleet-server","http.request.id":"5f167e84-111d-46e7-82a1-6b0bc8daa3b7","server.address":"","mod":"enroll","fleet.enroll.apikey.id":"CTq_oIgBf48Nt7iUw3or","fleet.apikey.id":"trsNoIgBfXEwhkEH78Ee","@timestamp":"2023-06-09T15:22:40.201Z","message":"api key read"}
{"log.level":"debug","ecs.version":"1.6.0","service.name":"fleet-server","http.request.id":"5f167e84-111d-46e7-82a1-6b0bc8daa3b7","server.address":"","mod":"enroll","fleet.enroll.apikey.id":"CTq_oIgBf48Nt7iUw3or","fleet.apikey.id":"trsNoIgBfXEwhkEH78Ee","@timestamp":"2023-06-09T15:22:41.202Z","message":"api key read"}
{"log.level":"debug","ecs.version":"1.6.0","service.name":"fleet-server","http.request.id":"5f167e84-111d-46e7-82a1-6b0bc8daa3b7","server.address":"","mod":"enroll","fleet.enroll.apikey.id":"CTq_oIgBf48Nt7iUw3or","fleet.apikey.id":"trsNoIgBfXEwhkEH78Ee","@timestamp":"2023-06-09T15:22:42.205Z","message":"api key read"}
{"log.level":"debug","ecs.version":"1.6.0","service.name":"fleet-server","http.request.id":"5f167e84-111d-46e7-82a1-6b0bc8daa3b7","server.address":"","mod":"enroll","fleet.enroll.apikey.id":"CTq_oIgBf48Nt7iUw3or","fleet.apikey.id":"trsNoIgBfXEwhkEH78Ee","@timestamp":"2023-06-09T15:22:43.207Z","message":"api key read"}
{"log.level":"error","ecs.version":"1.6.0","service.name":"fleet-server","http.request.id":"5f167e84-111d-46e7-82a1-6b0bc8daa3b7","server.address":"","mod":"enroll","fleet.enroll.apikey.id":"CTq_oIgBf48Nt7iUw3or","fleet.apikey.id":"trsNoIgBfXEwhkEH78Ee","error.message":"apikey index failed to refresh","@timestamp":"2023-06-09T15:22:43.209Z","message":"Abort query attempt on apikey"}
{"log.level":"error","ecs.version":"1.6.0","service.name":"fleet-server","http.request.id":"5f167e84-111d-46e7-82a1-6b0bc8daa3b7","server.address":"","mod":"enroll","fleet.enroll.apikey.id":"CTq_oIgBf48Nt7iUw3or","error.message":"apikey index failed to refresh","EnrollmentId":"123456","AgentId":"f24d8919-0d7b-48c0-a462-bed3cb03ed64","APIKeyID":"trsNoIgBfXEwhkEH78Ee","@timestamp":"2023-06-09T15:22:43.209Z","message":"Error when trying to invalidate API key of old agent with enrollment id"}
{"log.level":"info","ecs.version":"1.6.0","service.name":"fleet-server","http.request.id":"5f167e84-111d-46e7-82a1-6b0bc8daa3b7","server.address":"","error.message":"apikey index failed to refresh","http.response.status_code":500,"event.duration":60543928000,"@timestamp":"2023-06-09T15:22:43.209Z","message":"HTTP request error"}
{"log.level":"error","ecs.version":"1.6.0","service.name":"fleet-server","http.request.id":"5f167e84-111d-46e7-82a1-6b0bc8daa3b7","server.address":"","mod":"enroll","error.message":"apikey index failed to refresh","@timestamp":"2023-06-09T15:22:43.209Z","message":"perform rollback on enrollment failure"}
{"log.level":"info","ecs.version":"1.6.0","service.name":"fleet-server","http.request.id":"5f167e84-111d-46e7-82a1-6b0bc8daa3b7","server.address":"","url.full":"/api/fleet/agents/enroll","http.version":"1.1","http.request.method":"POST","fleet.apikey.id":"CTq_oIgBf48Nt7iUw3or","client.address":"127.0.0.1:61097","tls.established":false,"http.request.body.bytes":161,"http.response.body.bytes":82,"http.response.status_code":500,"event.duration":60544005000,"@timestamp":"2023-06-09T15:22:43.209Z","message":"500 HTTP Request"}
    fleet_integration_test.go:563: map[error:BadRequest message:apikey index failed to refresh statusCode:500]
    fleet_integration_test.go:565: 
                Error Trace:    /Users/juliabardi/fleet-server/internal/pkg/server/fleet_integration_test.go:565
                                                        /Users/juliabardi/fleet-server/internal/pkg/server/fleet_integration_test.go:593
                Error:          Not equal: 
                                expected: 200
                                actual  : 500
                Test:           Test_Agent_Enrollment_Id
--- FAIL: Test_Agent_Enrollment_Id (61.59s)
FAIL
```

internal/pkg/api/handleEnroll.go
cmacknz: It looks like the error handling was changed here and for the other two blocks below without updating any tests, which suggests that we aren't testing the failure handling this function at all.

```go
		err := invalidateAPIKey(ctx, zlog, et.bulker, agent.AccessAPIKeyID)
		if err != nil {
			zlog.Error().Err(err).
				Str("EnrollmentId", enrollmentID).
				Str("AgentId", agent.Id).
				Str("APIKeyID", agent.AccessAPIKeyID).
				Msg("Error when trying to invalidate API key of old agent with enrollment id")
			return nil, err
		}
		// delete existing agent to recreate with new api key
		err = deleteAgent(ctx, zlog, et.bulker, agent.Id)
		if err != nil {
			zlog.Error().Err(err).
				Str("EnrollmentId", enrollmentID).
				Str("AgentId", agent.Id).
				Msg("Error when trying to delete old agent with enrollment id")
			return nil, err
		}
```

cmacknz:
This comment might be a hint for why it is trying to read the API key:

https://github.com/elastic/fleet-server/blob/42af87680750a003a0b54056a1b86044928071f7/internal/pkg/api/handleEnroll.go#L274-L276

Does `invalidateAPIKey` return a successful response if the API key it should invalidate wasn't found to exist? If you remove the read check is there a race condition where you are attempting to invalidate an API key that isn't visible in the index yet, and if you skip this part you end up with a valid API key that should have been invalidated sometime later?

juliaElastic:
> This comment might be a hint for why it is trying to read the API key:
> 
> https://github.com/elastic/fleet-server/blob/42af87680750a003a0b54056a1b86044928071f7/internal/pkg/api/handleEnroll.go#L274-L276

Thanks, I understand now.

> Does `invalidateAPIKey` return a successful response if the API key it should invalidate wasn't found to exist? If you remove the read check is there a race condition where you are attempting to invalidate an API key that isn't visible in the index yet, and if you skip this part you end up with a valid API key that should have been invalidated sometime later?

I tested invalidating a non existent api key and the response was not an error, but an empty array of invalidated api keys.
You're right that if we don't read we could end up in a situation where the invalidate runs earlier than the api key being visible. 
We could change the implementation to try to invalidate first, and if the response didn't invalidate, then try to read in a loop.
What I don't understand though, why in this test the read never seems to find the api key.

```
{
  "invalidated_api_keys": [],
  "previously_invalidated_api_keys": [],
  "error_count": 0
}
```

internal/pkg/api/handleEnroll.go
```diff
@@ -242,7 +282,7 @@ func invalidateAPIKey(ctx context.Context, zlog zerolog.Logger, bulker bulk.Bulk
 LOOP:
 	for {
 
-		_, err := bulker.APIKeyRead(ctx, apikeyID, false)
+		_, err := bulker.APIKeyRead(ctx, apikeyID, true)
```
juliaElastic, line 285: I found that the test works if I use the `withOwner` flag in `APIKeyRead`
`// WithOwner - flag to query api keys owned by the currently authenticated user.`

```
{"log.level":"debug","ecs.version":"1.6.0","service.name":"fleet-server","http.request.id":"5492a01d-efc9-436f-b77f-3c45245236eb","server.address":"","mod":"enroll","fleet.enroll.apikey.id":"tEvjs4gBdXQ3jhGbSbnv","EnrollmentId":"123456","AgentId":"a3a9c74d-3a16-4963-94fe-858fd7c81c3c","APIKeyID":"t0vjs4gBdXQ3jhGbU7kM","@timestamp":"2023-06-13T08:33:20.386Z","message":"Invalidate old api key and remove existing agent with the same enrollment_id"}
{"log.level":"debug","ecs.version":"1.6.0","service.name":"fleet-server","http.request.id":"5492a01d-efc9-436f-b77f-3c45245236eb","server.address":"","mod":"enroll","fleet.enroll.apikey.id":"tEvjs4gBdXQ3jhGbSbnv","fleet.apikey.id":"t0vjs4gBdXQ3jhGbU7kM","error.message":"[404 Not Found] {\"api_keys\":[]}: api key not found","@timestamp":"2023-06-13T08:33:20.388Z","message":"api key read"}

{"log.level":"debug","ecs.version":"1.6.0","service.name":"fleet-server","http.request.id":"5492a01d-efc9-436f-b77f-3c45245236eb","server.address":"","mod":"enroll","fleet.enroll.apikey.id":"tEvjs4gBdXQ3jhGbSbnv","fleet.apikey.id":"t0vjs4gBdXQ3jhGbU7kM","@timestamp":"2023-06-13T08:33:21.409Z","message":"api key read"}
{"log.level":"debug","ecs.version":"1.6.0","service.name":"fleet-server","http.request.id":"5492a01d-efc9-436f-b77f-3c45245236eb","server.address":"","mod":"enroll","fleet.enroll.apikey.id":"tEvjs4gBdXQ3jhGbSbnv","fleet.apikey.id":"t0vjs4gBdXQ3jhGbU7kM","@timestamp":"2023-06-13T08:33:21.409Z","message":"api key invalidate"}
{"error_count":0,"invalidated_api_keys":["t0vjs4gBdXQ3jhGbU7kM"],"previously_invalidated_api_keys":[]}
{"log.level":"info","ecs.version":"1.6.0","service.name":"fleet-server","http.request.id":"5492a01d-efc9-436f-b77f-3c45245236eb","server.address":"","mod":"enroll","fleet.enroll.apikey.id":"tEvjs4gBdXQ3jhGbSbnv","fleet.apikey.id":"t0vjs4gBdXQ3jhGbU7kM","dur":1304.635082,"@timestamp":"2023-06-13T08:33:21.691Z","message":"invalidated apiKey"}
{"log.level":"info","ecs.version":"1.6.0","service.name":"fleet-server","http.request.id":"5492a01d-efc9-436f-b77f-3c45245236eb","server.address":"","mod":"enroll","fleet.enroll.apikey.id":"tEvjs4gBdXQ3jhGbSbnv","fleet.agent.id":"a3a9c74d-3a16-4963-94fe-858fd7c81c3c","@timestamp":"2023-06-13T08:33:21.987Z","message":"agent record deleted"}
{"log.level":"info","ecs.version":"1.6.0","service.name":"fleet-server","http.request.id":"5492a01d-efc9-436f-b77f-3c45245236eb","server.address":"","mod":"enroll","fleet.enroll.apikey.id":"tEvjs4gBdXQ3jhGbSbnv","fleet.agent.id":"66d97083-db90-4ab9-b16a-6b6e12676d46","fleet.policy.id":"eaaa69f5-e3bc-4e4f-bbb0-6cd10717d19e","fleet.access.apikey.id":"ukvjs4gBdXQ3jhGbXLlE","http.response.body.bytes":410,"event.duration":2294719000,"@timestamp":"2023-06-13T08:33:22.35Z","message":"Elastic Agent successfully enrolled"}
    fleet_integration_test.go:614: old agent not found as expected
    fleet_integration_test.go:606: could not clean up first agent
--- PASS: Test_Agent_Enrollment_Id (6.32s)
```

juliaElastic:
@cmacknz Ready for final review, thanks!

internal/pkg/api/handleEnroll.go
cmacknz: Sanity check: What is the "user" in this context? Is it the agent? Or is it the Fleet Server? If it's Fleet Server itself does this have any implications when there is more than one Fleet Server?

internal/pkg/api/handleEnroll.go
```diff
@@ -146,6 +158,33 @@ func (et *EnrollerT) _enroll(
...
+	// only delete existing agent if it never checked in
+	if agent.Id != "" && agent.LastCheckin == "" {
+		zlog.Debug().
+			Str("EnrollmentId", enrollmentID).
+			Str("AgentId", agent.Id).
+			Str("APIKeyID", agent.AccessAPIKeyID).
+			Msg("Invalidate old api key and remove existing agent with the same enrollment_id")
+		// invalidate previous api key
+		err := invalidateAPIKey(ctx, zlog, et.bulker, agent.AccessAPIKeyID)
+		if err != nil {
```
cmacknz, line 170: Do this and the `deleteAgent` call below also need to be guarded in the same way the `FindAgent` call is above with `if !errors.Is(err, dl.ErrNotFound) && !strings.Contains(err.Error(), "no such index")` before returning error?

Are there any expected and non-fatal errors that can happen here? I assume that the agents here are not yet enrolled so there's no way a concurrent operation from the UI could interfere?

internal/pkg/server/fleet_integration_test.go
```diff
@@ -541,6 +543,82 @@ func Test_SmokeTest_Agent_Calls(t *testing.T) {
...
+
+	// Start test server
+	srv, err := startTestServer(t, ctx)
+	require.NoError(t, err)
+
+	t.Log("Enroll the first agent with enrollment_id")
+	firstAgentID := EnrollAgent(enrollBodyWEnrollmentID, t, ctx, srv)
+
+	t.Log("Enroll the second agent with the same enrollment_id")
+	secondAgentID := EnrollAgent(enrollBodyWEnrollmentID, t, ctx, srv)
```
cmacknz, line 596: Could you write a version of this test where the API key of the agent being re-enrolled was invalidated separately? What if someone uses an independent query to revoke all API keys as a security measure? Could that break this implementation?

internal/pkg/api/handleEnroll.go
cmacknz: Following up from https://github.com/elastic/fleet-server/pull/2655/files#r1228834438, does `invalidateAPIKey` need to handle the case where the API key is already invalid?

internal/pkg/api/handleEnroll.go
```diff
@@ -146,6 +158,33 @@ func (et *EnrollerT) _enroll(
...
+			zlog.Error().Err(err).
+				Str("EnrollmentId", enrollmentID).
+				Str("AgentId", agent.Id).
+				Str("APIKeyID", agent.AccessAPIKeyID).
+				Msg("Error when trying to invalidate API key of old agent with enrollment id")
+			return nil, err
+		}
+		// delete existing agent to recreate with new api key
+		err = deleteAgent(ctx, zlog, et.bulker, agent.Id)
```
cmacknz, line 179: Is this safe if two calls to this API are made concurrently with the same enrollment ID? I could see this happening if somebody clones a VM with a pre-generated enrollment ID.

Since enrollment is done from the command line and the API call is not per agent, there is nothing guaranteeing this is impossible.

internal/pkg/api/handleEnroll.go
juliaElastic: I tried the invalidate API with an already invalidated and non existing API key, and the response was successful, so I think we can repeat it safely on a retry. I can add another test for it.

internal/pkg/api/handleEnroll.go
juliaElastic: Do you mean two agents enrolling at the same time with the same enrollment ID and one being deleted and it shouldn't? 
I think the same enrollment ID shouldn't be used by 2 different agents. If this happens we should have a way on the agent to generate a new enrollment ID. 

The goal of this feature is really to solve the agent duplication on enroll failure, this will unblock the flakyness of scale tests.

internal/pkg/api/handleEnroll.go
juliaElastic: It is Fleet Server, according to the error message: `action [cluster:admin/xpack/security/api_key/get] is unauthorized for service account [elastic/fleet-server],`
I think it shouldn't be an issue with more than one FS as they are using the same service account.

internal/pkg/server/fleet_integration_test.go
juliaElastic: Added one more test where the first api key was invalidated before the second enroll

internal/pkg/api/handleEnroll.go
cmacknz: > I think the same enrollment ID shouldn't be used by 2 different agents.  If this happens we should have a way on the agent to generate a new enrollment ID.

It shouldn't happen but people do clone virtual machine filesystems, and there is nothing preventing this from happening it will probably will eventually.

What would `deleteAgent` return if the agent didn't exist or was already deleted? We know the API key can be invalided already without issue already.

cmacknz: Actually I'm not sure it matters, because if there are two calls with the same enrollment ID the deleteAgent call here will remove one of them and we'd be left with only one agent. The user can hopefully detect this.

I'd only expect this to be a problem if someone clones a virtual machine image where the enrollment ID has already been created.

cmacknz (approved):

PR #2656, opened 2023-06-01
fix condition for the docker push step

sharbuz:

fix the condition for the docker push step

changed condition to correct, according to the docs
https://buildkite.com/docs/pipelines/conditionals
https://buildkite.com/docs/pipelines/environment-variables

It was tested here:
https://github.com/elastic/fleet-server/blob/820ae55a8652775a2ab7a6e98325ca4c4c471ddb/.buildkite/pipeline.yml#L120

## Checklist

## Related issues


juliaElastic (approved):

.buildkite/pipeline.yml
```diff
@@ -91,7 +91,7 @@ steps:
   - label: ":docker: Publish docker image"
     key: "publish"
     command: ".buildkite/scripts/build_push_docker_image.sh"
-    if: build.env("BUILDKITE_PULL_REQUEST") == null
+    if: build.env("BUILDKITE_PULL_REQUEST") == ''
```
mrodm, line 94: Here, it would be better to just run this step if it is branch main. Could that be used BUILDKITE_BRANCH for that ?

mrodm (approved):

PR #2664, opened 2023-06-02
Add step to trigger gitops to update image

michel-laterman:
Trigger update fleet-server image update through gitops.


.buildkite/pipeline.yml
```diff
@@ -98,6 +98,21 @@ steps:
       - step: "tests"
         allow_failure: false
 
+  - label: ":argo: Update image tag for Fleet Server"
+    trigger: serverless-gitops-update-stack-image-tag
+    async: true
+    branches: main
+    build:
+      env:
+        IMAGE_TAG: "$DOCKER_IMAGE_TAG"
```
michel-laterman, line 107: I'm not sure if this is actually the right tag

.buildkite/pipeline.yml
```diff
@@ -98,6 +98,21 @@ steps:
...
+        SERVICE: fleet-controller
+        NAMESPACE: fleet-ci
+        IMAGE_NAME: fleet-server
+        COMMIT_MESSAGE: "gitops: update fleet-server tag to elastic/fleet-server@$GIT_ABBREV_COMMIT"
```
mrodm, line 111: I think `GIT_ABBREV_COMMIT` env. variable does not exist in the Buildkite context.

Probably, it should be used:
- GITHUB_PR_HEAD_SHA
- BUILDKITE_COMMIT

They contain the full sha of the commit. Example:
```
GITHUB_PR_HEAD_SHA="74556f61420168ef34592455a2bd42e4c4e01dc9"
BUILDKITE_COMMIT="74556f61420168ef34592455a2bd42e4c4e01dc9"
```

.buildkite/pipeline.yml
mrodm: This environment variable does not exist in Buildkite or when the execution reaches this step.

What value should it be here ?

.buildkite/pipeline.yml
mrodm: If those values needs to be calculated, it looks like it should be created this step dynamically (following the example in Kibana repo).

Here there is another example of creating steps dynamically:
- https://github.com/elastic/elastic-package/blob/main/.buildkite/pipeline.yml#L38-L39
- https://github.com/elastic/elastic-package/blob/main/.buildkite/pipeline.trigger.integration.tests.sh

.buildkite/pipeline.yml
```diff
@@ -98,6 +98,21 @@ steps:
...
 
+  - label: ":argo: Update image tag for Fleet Server"
+    trigger: serverless-gitops-update-stack-image-tag
+    async: true
+    branches: main
+    build:
+      env:
+        IMAGE_TAG: "$DOCKER_IMAGE_TAG"
+        SERVICE: fleet-controller
+        NAMESPACE: fleet-ci
```
mrodm, line 109: What does this namespace refer to ? Is it to kubernetes namespace or the docker namespace (URL) ? 

Should it be `observability-ci` ? 
```suggestion
        NAMESPACE: observability-ci
```

This is where docker images are being pushed right now by Jenkins: https://github.com/elastic/fleet-server/blob/main/.ci/Jenkinsfile#L19

mrodm (approved):
LGTM

jlind23 (approved):

PR #2666, opened 2023-06-02
File delivery

pzl:

Delivering user-uploaded files (via kibana) to integrations (specifically, endpoint use-case)

an API route was added, `/api/fleet/file/[ID]`.  Authenticated calls to this will locate file metadata documents in elasticsearch for the presence of a file, and that the authenticated agent is present in the file's `file.Meta.target_agents` field (i.e. only allowed client IDs can download). Fleet server iterates through the file contents to send on the response pipe

- Create destination indices in Elasticsearch matching these templates (until the PR is merged): https://github.com/elastic/elasticsearch/pull/96504
- Use a recent Dev Endpoint that includes file-requesting feature
- In kibana, create a send-file action to send a file to an Endpoint

Alternatively, a simpler setup would be to check out this PR, and modify the authentication section in `internal/pkg/api/handleFileDelivery.go` line 48. Comment out the `authAgent` call (or otherwise mock it), and put a different, perhaps static ID as the last parameter to the `FindFileForAgent` call a few lines down  (e.g. `... fileID, "mockID")`). As long as a file is uploaded to the right indices (`.fleet-filedelivery-meta-endpoint` and `.fleet-filedelivery-data-endpoint`) with that matching agent_id in the `target_agents` field, then a command-line curl, wget or similar to `https://localhost:8221/api/fleet/file/[ID]` should retrieve the file like any other file-based HTTP url.

## Design Checklist

## Checklist


model/openapi.yml
```diff
@@ -1251,3 +1251,44 @@ paths:
...
+        - name: id
+          in: path
+          description: The file_id as provided to the integration
+          required: true
+          schema:
+            type: string
+            example: ecb30383-6dd1-4b1d-bed0-2386b4e5df51
+      responses:
+        '200':
+          description: file contents
```
michel-laterman, line 1271:
```suggestion
          description: File contents.
          content:
            "*/*":
              schema:
                type: string
                format: binary
```
(nitpick) I don't think this will change any of the generated code

Can you also add the `X-File-Sha2` header to the 200 response description?

internal/pkg/api/api.go
```diff
@@ -128,6 +129,14 @@ func (a *apiServer) UploadChunk(w http.ResponseWriter, r *http.Request, id strin
 	}
 }
 
+func (a *apiServer) GetFile(w http.ResponseWriter, r *http.Request, id string) {
+	zlog := hlog.FromRequest(r).With().Logger()
+	if err := a.ft.handleSendFile(zlog, w, r, id); err != nil {
+		cntFileDeliv.IncError(err)
```
michel-laterman, line 135:
```suggestion
		cntFileDeliv.IncError(err)
		w.Header().Set("Content-Type", "application/json")
```
We should ensure that the content type header of error responses is set.

internal/pkg/file/delivery/delivery.go
```diff
@@ -0,0 +1,100 @@
...
+	// find chunk indices behind alias, doc IDs
+	infos, err := file.GetChunkInfos(ctx, d.bulker, FileDataIndexPattern, fileID, file.GetChunkInfoOpt{})
+	if err != nil {
+		zlog.Error().Err(err).Msg("problem getting infos")
+		return err
+	}
+
+	if len(infos) == 0 {
+		zlog.Warn().Str("fileID", fileID).Msg("chunk documents not found for file")
+		w.WriteHeader(http.StatusNotFound)
```
michel-laterman, line 68: At this point the content headers from `internal/pkg/api/handleFileDelivery.go` have already been written, which means that the response code is (implicitly) set to 200. 
Is there a way to do this check earlier in `handleFileDelivery`?

internal/pkg/file/delivery/delivery.go
```diff
@@ -0,0 +1,100 @@
...
+
+	var fi file.MetaDoc
+	if err := json.Unmarshal(result.Hits[0].Source, &fi); err != nil {
+		return file.MetaDoc{}, fmt.Errorf("file meta doc parsing error: %w", err)
+	}
+
+	return fi, nil
+}
+
+func (d *Deliverer) SendFile(ctx context.Context, zlog zerolog.Logger, w http.ResponseWriter, f file.MetaDoc, fileID string) error {
```
michel-laterman, line 57:
```suggestion
func (d *Deliverer) SendFile(ctx context.Context, zlog zerolog.Logger, w io.Writer, f file.MetaDoc, fileID string) error {
```
I would rather see an `io.Writer` instead of an `http.ResponseWriter` here as at this point the status codes and header have already been set.

internal/pkg/file/es.go
```diff
@@ -123,11 +126,21 @@ func GetChunk(ctx context.Context, bulker bulk.Bulk, indexPattern string, source
...
+}
+
 // Retrieves a subset of chunk document fields, specifically omitting the Data payload (bytes)
-// but adding the calculated field "size", that is the length, in bytes, of the Data field
-// the chunk's ordered index position (Pos) is also parsed from the document ID
-func GetChunkInfos(ctx context.Context, bulker bulk.Bulk, indexPattern string, baseID string) ([]ChunkInfo, error) {
-	query, err := QueryChunkInfo.Render(map[string]interface{}{
+// the chunk's ordered index position (Pos) is also parsed from the document ID.
+// Optionally adding the calculated field "size", that is the length, in bytes, of the Data field.
+// and optionally requiring
```
michel-laterman, line 137: requiring what?

michel-laterman (commented):
looks good so far.

Can you also add a test case for the e2e tests? (or a follow up item to do so)

kevinlog:
I built this and ran Fleet Server locally with a Kibana stack to test everything e2e and it worked!

I can upload multi-chunk files and Endpoint will be able to download them from ES through this API.

In use in Response Console:
image

The file ends up on the host in the expected folder from Endpoint and is executable by root:
image

from a functionality POV, LGTM

kevinlog (approved):

michel-laterman (approved):
LGTM.

Please follow it up with e2e tests. you can ping me if you need help writing them

pzl:
Thanks, adding plenty of tests in a followup. Going to merge this to get builds going over the weekend, for other teams to start integrating

PR #2677, opened 2023-06-07
[Fleet] Add elastic-api-version support

nchaulet:
## Description 

Introduce a new middleware that support `elastic-apiversion` header to allow versioning of the Fleet server APIs.

If the header is not provided we fallback to a default version.

For PR readability I will add some integration tests that duplicate the client and the api tests suite in a following PR.

## Todo

- [x] Unit tests the version middleware


internal/pkg/api/apiVersion.go
```diff
@@ -0,0 +1,81 @@
...
+	"strings"
+)
+
+var (
+	ErrInvalidAPIVersionFormat = errors.New("invalid version format")
+	ErrUnsupportedAPIVersion   = errors.New("version is not supported")
+)
+
+const (
+	ElasticAPIVersionHeader = "elastic-api-version"
```
michel-laterman, line 21:
```suggestion
	ElasticAPIVersionHeader = "Elastic-Api-Version"
```
(nit)

internal/pkg/api/apiVersion.go
```diff
@@ -0,0 +1,81 @@
...
+}
+
+func NewAPIVersion() *apiVersion {
+	return &apiVersion{
+		supportedVersions: SupportedVersions,
+		defaultVersion:    DefaultVersion,
+	}
+}
+
+func (a *apiVersion) validateVersionFormat(version string) (string, error) {
```
michel-laterman, line 41: Why is this returning a string and an error?

internal/pkg/api/apiVersion.go
```diff
@@ -0,0 +1,81 @@
...
+		if headerValue != "" {
+			_, err := a.validateVersionFormat(headerValue)
+			if err != nil {
+				ErrorResp(w, r, err)
+				return
+			}
+
+			if !a.isVersionSupported(headerValue) {
+				w.Header().Add(ElasticAPIVersionHeader, a.defaultVersion)
+				ErrorResp(w, r, fmt.Errorf("received \"%s\", is not supported. supported versions are: %s %w", headerValue, strings.Join(a.supportedVersions, ", "), ErrInvalidAPIVersionFormat))
```
michel-laterman, line 71:
```suggestion
				ErrorResp(w, r, fmt.Errorf("received %q, is not supported. supported versions are: [%s] %w", headerValue, strings.Join(a.supportedVersions, ", "), ErrInvalidAPIVersionFormat))
```
(nit)

model/openapi.yml
```diff
@@ -808,7 +814,8 @@ paths:
     get:
       operationId: status
       parameters:
-        - $ref: '#/components/parameters/requestId'
+        - $ref: "#/components/parameters/requestId"
+        - $ref: "#/components/parameters/apiVersion"
```
michel-laterman, line 818: As we're adding the api version middleware, all endpoint including status can now return a 400, we should add that to the spec

michel-laterman (commented):
think it looks good, mostly have nitpicks.

We may want to consider altering the openapi spec to show responses will contain the version header (we don't do this for request ids, but we should)

The only real concern I have is now we're changing the behaviour of the status endpoint as well

nchaulet:
> The only real concern I have is now we're changing the behaviour of the status endpoint as well

in my opinion is probably okay as if you do not provide the `Elastic-Api-Version` it will still work

internal/pkg/api/apiVersion.go
```diff
@@ -0,0 +1,81 @@
...
+		if headerValue != "" {
+			_, err := a.validateVersionFormat(headerValue)
+			if err != nil {
+				ErrorResp(w, r, err)
+				return
+			}
+
+			if !a.isVersionSupported(headerValue) {
+				w.Header().Add(ElasticAPIVersionHeader, a.defaultVersion)
+				ErrorResp(w, r, fmt.Errorf("received %q, is not supported. supported versions are: [%s] %w", headerValue, strings.Join(a.supportedVersions, ", "), ErrInvalidAPIVersionFormat))
```
jsoriano, line 71: And `ErrUnsupportedAPIVersion` here?
```suggestion
				ErrorResp(w, r, fmt.Errorf("received %q, is not supported. supported versions are: [%s] %w", headerValue, strings.Join(a.supportedVersions, ", "), ErrUnsupportedAPIVersion))
```

jsoriano (approved):

internal/pkg/api/apiVersion.go
```diff
@@ -0,0 +1,81 @@
...
+		headerValue := r.Header.Get(ElasticAPIVersionHeader)
+		if headerValue != "" {
+			_, err := a.validateVersionFormat(headerValue)
+			if err != nil {
+				ErrorResp(w, r, err)
+				return
+			}
+
+			if !a.isVersionSupported(headerValue) {
+				w.Header().Add(ElasticAPIVersionHeader, a.defaultVersion)
```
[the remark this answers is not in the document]
michel-laterman: Perhaps the client should supply an ordered list of protocol versions it understands

This would allow a client running version N+1 to request [N+1, N, N-1].
The fleet-server would respond with the first match it supports, or a default value if none match.

This way clients that don't have version negotiation will use the (oldest) default value but clients that have negotiation can use a more recent version

PR #2678, opened 2023-06-08
Update dependencies in Component

mrodm:

This PR adds the current dependencies related to CI/CD.

This PR adds the dependency with:
- [package-registry repository](https://github.com/elastic/package-registry/)
- pipeline to release a new distribution image when a tag is pushed - [link](https://github.com/elastic/package-registry/blob/main/.buildkite/release-package-registry-distribution.yml)

Documentation about these Backstage relationships: https://backstage.io/docs/features/software-catalog/well-known-relations/#dependson-and-dependencyof

## Design Checklist

## Checklist

### DIFF
--- a/catalog-info.yaml
+++ b/catalog-info.yaml
@@ -12,6 +12,10 @@ spec:
   owner: group:ingest-fp
   system: platform-ingest
   lifecycle: production
+  dependsOn:
+    - "component:package-registry"
+    - "resource:buildkite-pipeline-release-package-registry-distribution"
+
 
 ---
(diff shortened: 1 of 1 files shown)


catalog-info.yaml
```diff
@@ -12,6 +12,10 @@ spec:
   owner: group:ingest-fp
   system: platform-ingest
   lifecycle: production
+  dependsOn:
+    - "component:package-registry"
+    - "resource:buildkite-pipeline-release-package-registry-distribution"
+
```
mrodm, line 18: This dependencies have been added to the Component resource. 
It looks like it is where it is commonly defined.

More info: https://backstage.io/docs/features/software-catalog/well-known-relations/#dependson-and-dependencyof

jlind23 (approved):

PR #2688, opened 2023-06-12
[Fleet] Add e2e tests for api versions

### LINKED ISSUE #2594: Fleet Server testing for versioned API
Fleet server APIs have to be versioned in order to ensure backwards compatibility with all already existing components calling it.
This issue is a placeholder that will be used as soon as the first versioned API is introduced.

```[tasklist]
### Tasks
- [ ] /api/status:
- [ ] /api/fleet/agents/enroll:
- [ ] /api/fleet/agents/{id}/checkin:
- [ ] /api/fleet/agents/{id}/acks:
- [ ] /api/fleet/artifacts/{id}/{sha2}:
- [ ] /api/fleet/uploads:
- [ ] /api/fleet/uploads/{id}/{chunkNum}:
- [ ] /api/fleet/uploads/{id}
```

nchaulet:
## Description

Resolve #2594

That PR add e2e test for versioned APIs.

To do so I duplicated the client for each API version, and created a duplicated test suite for each version.

We probably want to relax our sonar rules for duplication in that case.


testing/e2e/stand_alone_test.go
```diff
@@ -155,94 +156,91 @@ func (suite *StandAloneSuite) TestClientAPI() {
...
-	// Run subtests here
-	suite.Run("test status unauthenicated", func() {
-		ctx, cancel := context.WithCancel(bCtx)
-		defer cancel()
-		tester := &ClientAPITester{
-			suite.Suite,
-			ctx,
-			suite.client,
-			"https://localhost:8220",
+	versions := []string{"current", "2022-06-01"}
```
nchaulet, line 159: Run the test for each the version here, if we add a new version we will have to create a new `api_version/client_api*.go` (probably worth adding this to the readme)

jlind23:
Nice job, I can also see all tests passing on the buildkite build:
Screenshot 2023-06-13 at 07 29 04

jlind23 (approved):

jlind23:
NIT: Can we run the API tests in parallel like one "runner" per API version? Otherwise it might increase the build time.

testing/e2e/api_version/client_api_2023_06_01.go
```diff
@@ -0,0 +1,250 @@
...
+		if withRequestEditorFn != nil {
+			return withRequestEditorFn(ctx, req)
+		}
+
+		return nil
+	}))
+}
+
+// TestStatus tests the status endpoint, if an apiKey is given the authenticated response if verfied.
+func (tester *ClientAPITester20230601) TestStatus(apiKey string) {
```
jsoriano, line 57: Are we going to duplicate all tests for all versions of the API?

testing/e2e/api_version/client_api_2023_06_01.go
```diff
@@ -0,0 +1,250 @@
...
+	} else {
+		resp, err = client.StatusWithResponse(tester.ctx, &api.StatusParams{})
+	}
+	tester.Require().NoError(err)
+	tester.Require().Equal(http.StatusOK, resp.StatusCode())
+	tester.Require().NotEmpty(resp.JSON200.Name)
+	tester.Require().NotEmpty(resp.JSON200.Status)
+	if apiKey != "" {
+		tester.Require().NotNil(resp.JSON200.Version)
+	}
```
jsoriano, line 76: Nit. Use Assert for non-critical assertions so they are all executed.
```suggestion
	tester.Assert().Equal(http.StatusOK, resp.StatusCode())
	tester.Assert().NotEmpty(resp.JSON200.Name)
	tester.Assert().NotEmpty(resp.JSON200.Status)
	if apiKey != "" {
		tester.Assert().NotNil(resp.JSON200.Version)
	}
```

jsoriano (approved):

testing/e2e/api_version/api_version.go
```diff
@@ -0,0 +1,19 @@
...
+// you may not use this file except in compliance with the Elastic License.
+
+package api_version
+
+// enrollMetadataTpl is the template for enrollement metadata
+// It defines elastic.agent.version as that attribute is required by Kibana to successfully request diagnostics from an agent.
+// A diagnostics request is used when testing the file-upload endpoints.
+const enrollMetadataTpl = `{"elastic":{"agent":{"version":"%s"}}}`
+
+type ClientAPITesterInterface interface {
```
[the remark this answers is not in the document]
nchaulet: Looks like if I put in a `_test` file I cannot access it in the parent module

testing/e2e/api_version/client_api_2023_06_01.go
nchaulet: Yes I think it's the easiest way to maintain a test suite for each version as normally we will have difference between version, but I am open to other suggestions

jlind23:
@michel-laterman then we should probably keep an eye on this as it might become pretty slow when number of api version will increase.

testing/e2e/api_version/api_version.go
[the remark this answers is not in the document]
jsoriano: Ah right, this is used in the parent. Well, everything is under `testing`, so it should be fine :slightly_smiling_face:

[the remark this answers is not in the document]
jsoriano: But well, given that this is only used in the parent, I think we should define it there too.

jlind23:
Then @nchaulet should be good to merge this one then.

Issue #2683, opened 2023-06-09
Fleet Server should not crash on startup if connection to Elasticsearch timeouts


jsoriano:
Fleet Server should not crash on startup if connection to Elasticsearch timeouts, at least when running in standalone mode.
In Kubernetes deployments this leads to `CrashLoopBackOff`s, what is unexpected.

When this happens Fleet Server should report to be in unhealthy state, and continue retrying, probably with some kind of exponential backoff in case Elasticsearch is overloaded.

Other issues like unknown hosts and so on should still make Fleet Server to fail because they use to indicate misconfigurations.

This is logged when this happens, it shouldn't print the usage information because is not an error related to usage.
```
{"log.level":"error","ecs.version":"1.6.0","service.name":"fleet-server","cluster.addr":["https://xxxxxxxx:9200"],"cluster.maxConnsPersHost":128,"error.message":"dial tcp X.X.X.X:9200: i/o timeout","@timestamp":"2023-06-09T09:47:30.286Z","message":"fail elasticsearch info"}
{"ecs.version":"1.6.0","service.name":"fleet-server","log.level":"info","log.logger":"fleet-metrics.api","message":"Stats endpoint (127.0.0.1:5066) finished: accept tcp 127.0.0.1:5066: use of closed network connection","@timestamp":"2023-06-09T09:47:30.286Z"}
{"log.level":"info","ecs.version":"1.6.0","service.name":"fleet-server","state":"FAILED","@timestamp":"2023-06-09T09:47:30.286Z","message":"Error - dial tcp X.X.X.X:9200: i/o timeout"}
{"log.level":"error","ecs.version":"1.6.0","service.name":"fleet-server","error.message":"dial tcp X.X.X.X:9200: i/o timeout","@timestamp":"2023-06-09T09:47:30.286Z","message":"Fleet Server failed"}
{"log.level":"error","ecs.version":"1.6.0","service.name":"fleet-server","error.message":"dial tcp X.X.X.X:9200: i/o timeout","@timestamp":"2023-06-09T09:47:30.286Z","message":"Exiting"}
Error: dial tcp X.X.X.X:9200: i/o timeout
Usage:
  fleet-server [flags]

Flags:
  -E, --E setting=value   Overwrite configuration value
      --agent-mode        Running under execution of the Elastic Agent
  -c, --config string     Configuration for Fleet Server (default "fleet-server.yml")
  -h, --help              help for fleet-server

dial tcp 10.253.233.41:9200: i/o timeout
```

PR #2693, opened 2023-06-13
Avoid crashing on startup if Elasticsearch is not available

jsoriano:

Keep Fleet Server process running if Elasticsearch is not available on startup.

* Initial `Info` request is removed, so the client can be eventually used even if during startup Elasticsearch is not available.
* Subsystems that need initialization keep retrying if they fail.
* Standalone monitor keeps running once the service is healthy. If it loses connection with Elasticsearch, it goes to degraded state.

* An e2e is added that uses toxiproxy to simulate connectivity loss.
* In general, this can be tested by:
  * Starting Fleet Server without connectivity with Elasticsearch (or with Elasticsearch down), it should still start, but without reaching the healthy state.
  * Start Fleet Server, and once it is healthy, interrupt connectivity with Elasticsearch.

Locally I have simulated connectivity interruptions with `iptables`.  

## Design Checklist

## Checklist

## Related issues

- Fixes https://github.com/elastic/fleet-server/issues/2683


internal/pkg/es/client.go
```diff
@@ -106,27 +92,3 @@ type InfoResponse struct {
...
-		return nil, err
-	}
-
-	defer res.Body.Close()
-
-	if res.IsError() {
-		return nil, fmt.Errorf("info fail %v", res)
-	}
-
-	var resp InfoResponse
```
michel-laterman, line 124: Is `InfoResponse` used anywhere else? can we get rid of it?

internal/pkg/policy/standalone_test.go
```diff
@@ -25,6 +25,7 @@ func TestStandAloneSelfMonitor(t *testing.T) {
 		title         string
 		searchResult  *es.ResultT
 		searchErr     error
+		initialState  client.UnitState
```
michel-laterman, line 28: 👍

internal/pkg/coordinator/monitor.go
```diff
@@ -107,10 +107,24 @@ func (m *monitorT) Run(ctx context.Context) (err error) {
...
 
+	// Start timer loop to ensure leadership
+	lT := time.NewTimer(m.checkInterval)
+	defer lT.Stop()
+
 	// Ensure leadership on startup
-	err = m.ensureLeadership(ctx)
-	if err != nil {
-		return err
+	for {
```
michel-laterman, line 115: The coordinator and monitors get started  separate goroutines ( in `server/fleet.go`) so I don't think anything should go wrong with these changes.
The only difference should be that the API is "available", but should return a 503 in a non-status endpoint (like enroll) is called. Can you add that as an e2e test?

michel-laterman (commented):
Can you add a changelog fragment?

Also I think with these changes we should have a discussion about health vs readiness endpoints for the fleet-server

internal/pkg/es/client.go
jsoriano: Removed.

internal/pkg/api/router.go
```diff
@@ -128,3 +134,32 @@ func (l *limiter) middleware(next http.Handler) http.Handler {
...
+	message := "unavailable service"
+	statusCode := http.StatusServiceUnavailable
+	errorBody, err := json.Marshal(Error{
+		Message:    &message,
+		StatusCode: statusCode,
+	})
+	if err != nil {
+		// This should never fail.
+		panic(fmt.Sprintf("marshalling error failed, this is probably a bug: %v", err))
+	}
```
michel-laterman, line 150:
```suggestion
	statusCode := http. StatusServiceUnavailable
	errorBody := `{"message":" unavailable service","status":503}`
```

internal/pkg/api/router.go
```diff
@@ -128,3 +134,32 @@ func (l *limiter) middleware(next http.Handler) http.Handler {
...
+
+	return func(next http.Handler) http.Handler {
+		fn := func(w http.ResponseWriter, r *http.Request) {
+			if pathToOperation(r.URL.Path) != "status" {
+				if sm.State() != client.UnitStateHealthy {
+					http.Error(w, string(errorBody), statusCode)
```
michel-laterman, line 156:
```suggestion
					w.Header().Set("Content-Type", "application/json")
					http.Error(w, string(errorBody), statusCode)
```

testing/e2e/api_version/client_api_current.go
```diff
@@ -88,6 +88,31 @@ func (tester *ClientAPITester) TestEnroll(apiKey string) (string, string) {
...
+			},
+			Type: api.PERMANENT,
+		},
+	)
+	tester.Require().NoError(err)
+	tester.Assert().Condition(func() bool {
+		status := enrollResp.StatusCode()
+		tester.T().Logf("status code found: %v", status)
+		return status >= http.StatusInternalServerError
+	})
```
michel-laterman, line 113:
```suggestion
	tester.Assert().GreaterOrEqual(enrollResp.StatusCode(), http.StatusInternalServerError)
```
(nit) I find the provided assertions easier to read, it's also what we're doing for all other checks

internal/pkg/api/router.go
```diff
@@ -5,24 +5,37 @@
...
 	r.Use(logger.Middleware) // Attach middlewares to router directly so the occur before any request parsing/validation
 	r.Use(middleware.Recoverer)
 	r.Use(Limiter(cfg).middleware)
+
+	// In managed mode, fleet-server is expected to be able to operate normally during startup,
+	// so the agent managing it can enroll. So attach the statusChecker middleware only on
+	// stand-alone mode.
+	if _, isStandAlone := sm.(*policy.StandAloneSelfMonitor); isStandAlone {
+		r.Use(statusChecker(sm))
+	}
```
jsoriano, line 37: Not sure if I like to have this difference in behaviour, but I think we may need it, we don't have the same healthcheck needings when running as standalone and when running inside agent.

michel-laterman: I think our alternative is to create a health endpoint and have the platform direct traffic if to fleet-server once ES is healthy. What do you think?

jsoriano: Yeah, once we have healthchecks in place, the platform won't send traffic to us when we cannot reach ES. Should I remove this?
I started adding this after [this comment](https://github.com/elastic/fleet-server/pull/2693#discussion_r1230244522) about returning 503s when the healthcheck fails. But I see now that maybe you were referring to the platform and not directly here?

joshdover:
> Also I think with these changes we should have a discussion about health vs readiness endpoints for the fleet-server

Did we have this discussion? I think we need readiness not to be successful until Fleet Server can accept traffic (so it's connection to ES is working), while liveness should still pass to avoid unnecessary container restarts.

jsoriano:
> > Also I think with these changes we should have a discussion about health vs readiness endpoints for the fleet-server
> 
> Did we have this discussion? I think we need readiness not to be successful until Fleet Server can accept traffic (so it's connection to ES is working), while liveness should still pass to avoid unnecessary container restarts.

Not really, thanks for the heads-up on this.

TLDR; I would add a healthcheck script to the docker image to check readiness based on current status endpoint, and I wouldn't add a liveness probe at least at the moment.

For readiness, we can check now if the status is healthy. We would need a command probe for this.
If we want to use a simpler HTTP probe, and rely only on the status code, we would need to modify the current handler, because it returns 200 now when it is in degraded state, or to  define a new `/healthz` endpoint that returns 200 only when healthy.

So three options:
* Use a command probe that uses the status endpoint to check if the service is healthy. We could add a `healthcheck.sh` script to the image for this.
* Modify current status endpoint to return 200 only when the service is healthy. This is potentially a breaking change.
* Add a new `/healthz` endpoint that returns 200 only when the service is healthy.

I will go for the first option by now so we don't need to modify or extend our APIs.

Regarding liveness, I don't think we need to add a probe for this at the moment. I would add it if we know that the Fleet Server can reach some unrecoverable unhealthy state, that afaik is not the case. If we add a liveness probe now, it will be always succeeding unless the pod crashes, what is no different to not having any probe.

@michel-laterman @joshdover wdyt?

joshdover:
> Modify current status endpoint to return 200 only when the service is healthy. This is potentially a breaking change.

I lean towards this option as I see it more of a bug fix than a breaking change, but also open to start with a command probe for now as we experiment and only make the change on the existing endpoint once we have it working the way we want.

> Regarding liveness, I don't think we need to add a probe for this at the moment. I would add it if we know that the Fleet Server can reach some unrecoverable unhealthy state, that afaik is not the case. If we add a liveness probe now, it will be always succeeding unless the pod crashes, what is no different to not having any probe.

Makes sense to me 👍

jsoriano:
> > Modify current status endpoint to return 200 only when the service is healthy. This is potentially a breaking change.
> 
> I lean towards this option as I see it more of a bug fix than a breaking change, but also open to start with a command probe for now as we experiment and only make the change on the existing endpoint once we have it working the way we want.

Ok, I will give this a try before adding the script.

internal/pkg/api/router.go
jsoriano: I am removing the middleware I added, and we will rely on readiness probes.

internal/pkg/coordinator/monitor.go
jsoriano: And removed after https://github.com/elastic/fleet-server/pull/2693#discussion_r1240095377 and the discussion about health checks.

Not sure then if there is an E2E test that we can add here, maybe in the controllers that use it?

internal/pkg/policy/standalone.go
```diff
@@ -87,17 +81,32 @@ func (m *standAloneSelfMonitorT) State() client.UnitState {
...
-	if err != nil {
-		if errors.Is(err, es.ErrIndexNotFound) {
-			m.log.Debug().Str("index", m.policiesIndex).Msg(es.ErrIndexNotFound.Error())
-			return m.updateState(client.UnitStateHealthy, "Running: Policies not available yet")
-		}
-		if err != nil {
-			return m.updateState(client.UnitStateFailed, fmt.Sprintf("Failed to request policies: %s", err))
+	if errors.Is(err, es.ErrIndexNotFound) {
+		m.log.Debug().Str("index", m.policiesIndex).Msg(es.ErrIndexNotFound.Error())
+		message = "Running: Policies not available yet"
```
jsoriano, line 95: @michel-laterman should it be actually healthy if the policies index hasn't been created yet? Would requests succeed if Fleet Server cannot use this index? :thinking:

michel-laterman: IMO it should be unhealthy if the index does not exist, that way the fleet-controller/k8s can use it for health checks/traffic routing. What do you think @nchaulet?
This PR also changes the status endpoint to return 200 only on a healthy status

nchaulet: > IMO it should be unhealthy if the index does not exist, that way the fleet-controller/k8s can use it for health checks/traffic routing. What do you think @nchaulet?

Not sure the index while not exists if the user do not create any agent policy, so it probably should be healthy even if the index do not exists

michel-laterman: Sounds good, let's add the behaviour to the description in the openapi doc

michel-laterman (approved):

internal/pkg/policy/standalone.go
jsoriano: Description updated in https://github.com/elastic/fleet-server/pull/2693/files#diff-5ad174233c27653aa08c5ce213ddfc13cbf7bcfc10cdf724a547da655e5743c0R830, let me know if this is fine.

jlind23:
Looks like the buildkite failure comes from 
```
go: cloud.google.com/go/pubsublite@v1.7.0: verifying go.mod: cloud.google.com/go/pubsublite@v1.7.0/go.mod: reading https://sum.golang.org/tile/8/1/252: 404 Not Found
```

joshdover:
Ran a rebuild and it seemed to get past that step 👍

jlind23:
🚢

PR #2705, opened 2023-06-16
Dra snapshot

sharbuz:

Migration of the "DRA Snapshot" step of the "packaging.groovy" pipeline.

## Checklist

## Related issues

**Pipeline steps:**
- [x] Pipeline init
- [x] Filter build - Skipped because this filter is required by Jenkins auto-scan and we already have all additional conditions for starting the pipeline
- [x] Package
- [x] DRA Snapshot
- [ ] DRA Release Staging


sharbuz:
tested mktemp command, the same result...

PR #2710, opened 2023-06-20
Fix e2e certificates generation with OpenSSL 3

jsoriano:

Fix private key parsing issue with certificates created for e2e tests.
```
tls: failed to parse private key
```

OpensSSL 3 uses PKCS#8 by default instead of PKCS#1. This changes the format of encrypted private keys, fleet server is not able to parse encrypted PKCS#8 keys (see https://github.com/elastic/elastic-agent-libs/issues/134).

This PR forces OpenSSL 3 to use the old format when creating keys, to ensure that fleet-server is able to read them.

Longer term we should fix https://github.com/elastic/elastic-agent-libs/issues/134 to ensure that private keys generated with more recent libraries can be used.

## Design Checklist

## Related issues

- Root cause: https://github.com/elastic/elastic-agent-libs/issues/134


dev-tools/e2e/validatecerts.go
```diff
@@ -0,0 +1,74 @@
+// Copyright Elasticsearch B.V. and/or licensed to Elasticsearch B.V. under one
+// or more contributor license agreements. Licensed under the Elastic License;
+// you may not use this file except in compliance with the Elastic License.
+
+package main
```
michel-laterman, line 5:
```suggestion

// validatecerts is a script to ensure that certs used in e2e tests can be parsed by tlscommon
package main
```
(nit)

michel-laterman (approved):
lgtm

mrodm:
buildkite test this

PR #2724, opened 2023-06-22
Trigger package registry

sharbuz:
## What is the problem this PR solves?

The partial migration of the Jenkins pipeline to the Buildkite pipeline. 

## How does this PR solve the problem?

Migrating the "Release - Package Registry Distribution" part

## How to test this PR locally

It doesn't be checked locally, that's why it's a partial "step by step" migration

## Checklist

The Buildkite pipeline will be triggered by this PR and we will see the result

## Related issues

Related to the issue: https://github.com/elastic/fleet-server/issues/2517

**Pipeline steps:**
- [x] Emulation - Release - Package Registry Distribution

PR #2729, opened 2023-06-23
add the trigger package-registry pipeline

sharbuz:

Migrating the "Release - Package Registry Distribution" part
Instead the https://github.com/elastic/fleet-server/pull/2724, only changes according to the "Release - Package Registry Distribution" stage

## Checklist

## Related issues

**Pipeline steps:**
- [x] Release - Package Registry Distribution

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -124,7 +124,12 @@ steps:
 
   - label: ":jenkins: Release - Package Registry Distribution"
     key: "release-package-registry"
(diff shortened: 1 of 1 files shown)


mrodm (approved):
LGTM 
As the target pipeline has DRY_RUN environment variable set
https://github.com/elastic/package-registry/blob/main/.buildkite/release-package-registry-distribution.yml#L40

it looks good to have here the actual trigger step.

In the next fleet-server release, this trigger step should be checked  and also the output in the target pipeline whether or not it works successfully in dry-run mode.

PR #2735, opened 2023-06-26
get the go version from the .go-version file

sharbuz:

Fix of the getting go-version

## Checklist

## Related issues


.buildkite/pull-requests.json
```diff
@@ -1,27 +1,27 @@
...
-        "set_commit_status": false,
-        "build_on_commit": false,
-        "build_on_comment": true,
-        "trigger_comment_regex": "^(?:(?:buildkite\\W+)?(?:package)\\W+(?:this|it))|^/package$",
-        "always_trigger_comment_regex": "^(?:(?:buildkite\\W+)?(?:package)\\W+(?:this|it))|^/package$",
-
-        "fail_on_not_mergeable": true
-      }]
-}
+{
```
mrodm, line 1: what is the change in this file ?

.buildkite/hooks/post-checkout
```diff
@@ -50,5 +50,8 @@ echo "Commit information"
 
 git --no-pager log --format=%B -n 1
 
+export GO_VERSION="$(< .go-version)"
+echo "$GO_VERSION"
+
```
mrodm, line 55: echoing the go version is just for debugging ? could it be removed?

mrodm (commented):
If defining environment varaibles can be used in the step definitoin, maybe it could be then updated also in other repositories. And just trust `.go-version`.

There was a related PR to add the needed configuration in the CLI config to update the Golang version from the pipelines

https://github.com/elastic/fleet-server/pull/2714

If this works, that config entry could be removed too

.buildkite/pull-requests.json
sharbuz: In this file was changed just EOL from CRLF to LF

PR #2748, opened 2023-06-29
TLS-446 consistent image tags

kroustou:

## Design Checklist

## Checklist

### DIFF
--- a/.ci/Jenkinsfile
+++ b/.ci/Jenkinsfile
@@ -151,7 +151,7 @@ pipeline {
         not { changeRequest() }
       }
       environment {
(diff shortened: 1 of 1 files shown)


jlind23 (approved):

jlind23 (changes requested):
You just broke the Jenkins build.
`[2023-06-29T10:39:20.980Z] WorkflowScript: 154: expecting '}', found ':' @ line 154, column 58.

[2023-06-29T10:39:20.980Z]    G = "git-${env.GIT_BASE_COMMIT:0:12}"`

PR #2760, opened 2023-07-03
fix go version

sharbuz:

PR solves the defining of the variable GO_VERSION

Manual defining in the pipeline.yml

## Checklist

## Related issues

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -3,6 +3,7 @@
 env:
   SETUP_GVM_VERSION: 'v0.5.0'
   DOCKER_COMPOSE_VERSION: '1.25.5'
+  GO_VERSION: "1.20.5"
   DOCKER_REGISTRY: "docker.elastic.co"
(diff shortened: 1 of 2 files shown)


jlind23 (approved):
LGTM waiting for the build results though

.buildkite/pipeline.yml
```diff
@@ -3,6 +3,7 @@
 env:
   SETUP_GVM_VERSION: 'v0.5.0'
   DOCKER_COMPOSE_VERSION: '1.25.5'
+  GO_VERSION: "1.20.5"
```
mrodm, line 6: In a previous PR, the configuration to bump golang version was updated to include this file too
https://github.com/elastic/fleet-server/blob/540abaecd980355eab8f3628c67ebd36f60b2ade/.ci/bump-golang.yml#L117-L119

If this change is introduced, then the configuration should be updated too.

Another option is to keep just one variable so that configuration would still work (`GO_VERSION` env. var should be removed in that case):

```
  GO_AGVENT_IMAGE: "golang:1.20.5"
```

PR #2781, opened 2023-07-12
Increase e2e tests timeout

jsoriano:

Try to reduce flakiness in E2E tests in Buildkite.

Increase `go test` timeouts.

Go tests default timeout is 10 minutes, and current E2E tests are taking around this time in buildkite, even when they pass:
imagen

When they fail, they are failing on different tests, but always with:
```
panic: test timed out after 10m0s
```

```
make test-e2e
```


nchaulet (approved):
🚀

PR #2799, opened 2023-07-18
Add refresh parameter config option

### LINKED ISSUE #2581: Replace refresh=true with refresh=wait_for in Fleet Server
While calling Elasticsearch APIs, fleet server uses a refresh flag value that default to `true` this should be change to `wait_for`.

- [ ] This change should be testing against a stateful Elasticsearch first to see if whether or not we want to apply it on all deployment types.

michel-laterman:

Fleet-server should not force ES to refresh with all of it's requests.

Add a config option that allows us to define the refresh parameter value instead of using `refresh=true` for all requests. 
Parameter is used for bulk operations and creating chunks while uploading a file.
Parameter is not used (and `true` is used) for [mget search](https://www.elastic.co/guide/en/elasticsearch/reference/current/docs-multi-get.html#docs-multi-get-api-query-params) requests and the [update by query](https://www.elastic.co/guide/en/elasticsearch/reference/current/docs-update-by-query.html#docs-update-by-query-api-query-params) requests as refresh is still defined as a bool param.

## Design Checklist

## Checklist

- [x] I have made corresponding change to the default configuration files~~

## Related issues

- Closes #2581

### DIFF
--- a/changelog/fragments/1689704810-add-refresh-param-config.yaml
+++ b/changelog/fragments/1689704810-add-refresh-param-config.yaml
@@ -11,12 +11,12 @@
 kind: feature
 
 # Change summary; a 80ish characters long description of the change.
-summary: Change refresh=true to refresh=wait_for
+summary: Add refresh param config
 
 # Long description; in case the summary is not enough to describe the change
 # this field accommodate a description without length limits.
 # NOTE: This field will be rendered only for breaking-change and known-issue kinds at the moment.
-description: Change the refresh param from "true" to "wait_for" to better support upcoming serverless offering. Remove refresh param from mget requests.
+description: Add a parameter that allows for the change of the refresh param from "true" to "wait_for" that effects file upload and bulk creation calls.
 
 # Affected component; a word indicating the component this changeset affects.
 component:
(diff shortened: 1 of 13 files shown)


joshdover:
Seems like we'll need to understand which refresh changes cause problems. Probably worth running the integration tests with fewer changes and see which one starts breaking things.

I would avoid us spending too much time on this and instead try utilizing the `fast_refresh` setting if needed.

changelog/fragments/1689704810-add-refresh-param-config.yaml
```diff
@@ -0,0 +1,32 @@
...
+description: Add a parameter that allows for the change of the refresh param from "true" to "wait_for" that effects file upload and bulk creation calls.
+
+# Affected component; a word indicating the component this changeset affects.
+component:
+
+# PR URL; optional; the PR number that added the changeset.
+# If not present is automatically filled by the tooling finding the PR where this changelog fragment has been added.
+# NOTE: the tooling supports backports, so it's able to fill the original PR number instead of the backport PR number.
+# Please provide it if you are adding a fragment for a different PR.
+#pr: https://github.com/owner/repo/1234
```
juliaElastic, line 28:
```suggestion
pr: 2799
```

joshdover:
@michel-laterman should we close this in favor of https://github.com/elastic/elasticsearch-serverless/pull/600?

michel-laterman:
Yeah, I'll close the extra PRs

PR #2815, opened 2023-07-24
[Fleet] Fix docker publish step in buildkite

nchaulet:
## Description 

Fix docker publish step in buildkite

The if condition to build the docker image was not met 
Screenshot 2023-07-24 at 1 00 56 PM

It seems according to buildkite doc that `BUILDKITE_PULL_REQUEST ` should be false if it's not a pull request and not an empty string

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -1,15 +1,15 @@
 # yaml-language-server: $schema=https://raw.githubusercontent.com/buildkite/pipeline-schema/main/schema.json
 
 env:
-  SETUP_GVM_VERSION: 'v0.5.0'
-  DOCKER_COMPOSE_VERSION: '1.25.5'
+  SETUP_GVM_VERSION: "v0.5.0"
+  DOCKER_COMPOSE_VERSION: "1.25.5"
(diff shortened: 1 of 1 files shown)


.buildkite/pipeline.yml
```diff
@@ -105,6 +105,8 @@ steps:
       env:
         SERVICE_COMMIT_HASH: ${BUILDKITE_COMMIT:0:12}
         REMOTE_SERVICE_CONFIG: https://raw.githubusercontent.com/elastic/serverless-gitops/main/gen/gpctl/fleet/config.yaml
+    depends_on:
+      - step: "publish"
```
nchaulet, line 109: I added a dependency here so we do not update tags with non valid image

michel-laterman (approved):
lgtm

nchaulet:
@michel-laterman should we replace the `fleet-server/pr-merge` as required by `buildkite/fleet-server ` ? How and who can do that?

jlind23:
@nchaulet I just fixed this by adding buildkite as a required step and removed the jenkins one.

jlind23:
Merging this with admin permissions as fleet-server jenkins build is blocking this but should not be required anymore.

PR #2833, opened 2023-07-26
add export FOLDER and export OUTPUT_FILE

sharbuz:

fix the issue: https://buildkite.com/elastic/fleet-server-package-mbp/builds/230#018992bc-5d99-45b7-8296-732f9bc7fc7b

define the FOLDER variable

## Related issues

### DIFF
--- a/.buildkite/scripts/dra_release.sh
+++ b/.buildkite/scripts/dra_release.sh
@@ -7,11 +7,10 @@ source .buildkite/scripts/common.sh
 FOLDER_PATH="build/distributions"
 BASE_DIR="${WORKSPACE}/${FOLDER_PATH}"
 DRA_OUTPUT="release-manager.out"
-PROJECT="fleet-server"
+export PROJECT="fleet-server"
 TYPE=${1}
-BRANCH="${BUILDKITE_BRANCH}"
-
-VERSION=$(make get-version)
+export BRANCH="${BUILDKITE_BRANCH}"
+export VERSION=$(make get-version)
 
 if [[ "${VERSION}" == *"-SNAPSHOT"* || "${VERSION}" == "" ]]; then
     echo "The 'version' parameter is required and it cannot contain the suffix '-SNAPSHOT'."
@@ -34,7 +33,7 @@ with_go
 with_mage
 
 if [[ "${TYPE}" == "snapshot" ]]; then
-    SNAPSHOT=true
+    export SNAPSHOT=true
 fi
 
 mkdir -p ${BASE_DIR}/reports
--- a/.buildkite/scripts/dra_release.sh
+++ b/.buildkite/scripts/dra_release.sh
@@ -35,11 +35,9 @@ with_mage
 case "${TYPE}" in
     "snapshot")
         make release-manager-dependencies-snapshot
-        make release-manager-snapshot
         ;;
     "staging")
         make release-manager-dependencies-release
-        make release-manager-release
         ;;
     *)
     echo "The option is unsupported yet"
--- a/.buildkite/scripts/dra_release.sh
+++ b/.buildkite/scripts/dra_release.sh
@@ -8,7 +8,7 @@ FOLDER_PATH="build/distributions"
 BASE_DIR="${WORKSPACE}/${FOLDER_PATH}"
 DRA_OUTPUT="release-manager.out"
 export PROJECT="fleet-server"
-TYPE=${1}
+export TYPE=${1}
 export BRANCH="${BUILDKITE_BRANCH}"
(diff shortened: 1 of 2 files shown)


.buildkite/scripts/dra_release.sh
```diff
@@ -42,6 +42,6 @@ mkdir -p ${BASE_DIR}/reports
 cd ${BASE_DIR}/reports && shasum -a 512 dependencies-${VERSION}.csv > dependencies-${VERSION}.csv.sha512
 
 cd $(dirname ${WORKSPACE})
-FOLDER=${FOLDER_PATH}
-OUTPUT_FILE=${DRA_OUTPUT}
+export FOLDER="${FOLDER_PATH}"
+export OUTPUT_FILE="${DRA_OUTPUT}"
```
mrodm, line 46: BRANCH, PROJECT, TYPE and VERSION are also used by the script 

Add to those vars also "export"

sharbuz: I think the TYPE is the local var, isn't it?

.buildkite/scripts/dra_release.sh
```diff
@@ -34,14 +33,14 @@ with_go
 with_mage
 
 if [[ "${TYPE}" == "snapshot" ]]; then
-    SNAPSHOT=true
+    export SNAPSHOT=true
 fi
 
 mkdir -p ${BASE_DIR}/reports
 ./dev-tools/dependencies-report --csv ${BASE_DIR}/reports/dependencies-${VERSION}.csv
 cd ${BASE_DIR}/reports && shasum -a 512 dependencies-${VERSION}.csv > dependencies-${VERSION}.csv.sha512
```
mrodm, line 41: I would try to use targets instead, to use everywhere the same.

Instead of 
```shell
if [[ "${TYPE}" == "snapshot" ]]; then
    export SNAPSHOT=true
fi

mkdir -p ${BASE_DIR}/reports
./dev-tools/dependencies-report --csv ${BASE_DIR}/reports/dependencies-${VERSION}.csv
cd ${BASE_DIR}/reports && shasum -a 512 dependencies-${VERSION}.csv > dependencies-${VERSION}.csv.sha512
```
I would use make as in Jenkins (https://github.com/elastic/fleet-server/blob/883cb46647979e234f70c88fd4aca1dc9f65800b/.ci/packaging.groovy#L217-L220)
```shell
if [[ "${TYPE}" == "staging" ]]; then
    make release-manager-dependencies-release
else
    make release-manager-dependencies-snapshot
fi
```

mrodm: And I think package.sh should also be updated to ensure that the same as in Jenkins is run:
https://github.com/elastic/fleet-server/blob/883cb46647979e234f70c88fd4aca1dc9f65800b/.ci/packaging.groovy#L170-L176

Instead of:
```shell
if [[ ${MATRIX_TYPE} == "staging" ]]; then
    make release
else
    make SNAPSHOT=true release
fi
```

I would use the makefile targets:
```shell
if [[ ${MATRIX_TYPE} == "staging" ]]; then
    make release-manager-release
else
    make release-manager-snapshot
fi
```

.buildkite/scripts/dra_release.sh
mrodm: That variable in release-manager.sh gets the value from the TYPE variable if it exists
https://github.com/elastic/fleet-server/blob/aeb00c548cd5d3042c9052ef8657e8b677bac101/.buildkite/scripts/release-manager.sh#L20

mrodm (commented):
LGTM but I would consider to use the same targets as it was run in Jenkinsfile now or in a following Pull Request.

Those targets have already all the needed logic to create the dependencies.

Would that be ok to be done in this PR ?

sharbuz:
@mrodm , please check it once again

.buildkite/scripts/dra_release.sh
```diff
@@ -33,15 +32,21 @@ with_go
...
-cd ${BASE_DIR}/reports && shasum -a 512 dependencies-${VERSION}.csv > dependencies-${VERSION}.csv.sha512
+case "${TYPE}" in
+    "snapshot")
+        make release-manager-dependencies-snapshot
+        make release-manager-snapshot
+        ;;
+    "staging")
+        make release-manager-dependencies-release
+        make release-manager-release
+        ;;
```
mrodm, line 43: `make release-manager-snaphot` and `make release-manager-release` should be part of the `.buildkite/scripts/package.sh` in the step "Package Ubuntu-20 x86_64"

mrodm (approved):

PR #2834, opened 2023-07-26
fix package type var

sharbuz:

fix the issue: https://buildkite.com/elastic/fleet-server-package-mbp/builds/232#018993cb-aa77-4e4a-acfe-61749eb103fa

change the variable MATRIX_TYPE to TYPE

## Related issues

### DIFF
--- a/.buildkite/scripts/package.sh
+++ b/.buildkite/scripts/package.sh
@@ -6,10 +6,10 @@ source .buildkite/scripts/common.sh
 
 VERSION=$(awk '/const DefaultVersion/{print $NF}' version/version.go | tr -d '"')
 PLATFORM_TYPE=$(uname -m)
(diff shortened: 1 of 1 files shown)


jlind23 (approved):

sharbuz:
@mrodm I remembered why I didn't use `make release-manager-snapshot` and `make release-manager-release` commands in the package script. Because these commands use the` ./dev-tools/run_with_go_ver` script which was created for Jenkins instance with installed GVM and needed permissions. When I tried to use this script again I got the same error:
https://buildkite.com/elastic/fleet-server-package-mbp/builds/233
And I remembered that I already faced this issue earlier, and I spent a lot of time troubleshooting the issue, and, as a result, I decided to use our common approach - `with_go` func. 
CC @jlind23

PR #2837, opened 2023-07-27
add path to release-manager script

sharbuz:

fix the issue: https://buildkite.com/elastic/fleet-server-package-mbp/builds/240#0189966d-f773-4479-be35-7913b0e5f9db

add `cd ${WORKSPACE}` to the release-manager script

## Related issues

https://github.com/elastic/fleet-server/issues/2517

### DIFF
--- a/.buildkite/scripts/dra_release.sh
+++ b/.buildkite/scripts/dra_release.sh
@@ -41,9 +41,7 @@ mkdir -p ${BASE_DIR}/reports
 ./dev-tools/dependencies-report --csv ${BASE_DIR}/reports/dependencies-${VERSION}.csv
 cd ${BASE_DIR}/reports && shasum -a 512 dependencies-${VERSION}.csv > dependencies-${VERSION}.csv.sha512
 
-cd ${WORKSPACE}
+cd $(dirname ${WORKSPACE})
 export FOLDER="${FOLDER_PATH}"
 export OUTPUT_FILE="${DRA_OUTPUT}"
-echo "$(pwd)"
-ls -la
 ./.buildkite/scripts/release-manager.sh          #TODO use "echo" for rollback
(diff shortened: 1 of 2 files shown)


.buildkite/scripts/release-manager.sh
```diff
@@ -21,8 +21,8 @@ readonly TYPE=${TYPE:-snapshot}
 readonly OUTPUT_FILE=${OUTPUT_FILE:-release-manager-report.out}
 
 # set required permissions on artifacts and directory
-chmod -R a+r "$FOLDER"/*
-chmod -R a+w "$FOLDER"
+chmod -R a+r "${WORKSPACE}"/"$FOLDER"/*
+chmod -R a+w "${WORKSPACE}"/"$FOLDER"
```
mrodm: Would it be the same if 
FOLDER would get that value from dra_release.sh ?
Suggestion added in the other comment

To keep the script as it was (if possible)

.buildkite/scripts/dra_release.sh
```diff
@@ -41,9 +41,7 @@ mkdir -p ${BASE_DIR}/reports
 ./dev-tools/dependencies-report --csv ${BASE_DIR}/reports/dependencies-${VERSION}.csv
 cd ${BASE_DIR}/reports && shasum -a 512 dependencies-${VERSION}.csv > dependencies-${VERSION}.csv.sha512
 
-cd ${WORKSPACE}
+cd $(dirname ${WORKSPACE})
 export FOLDER="${FOLDER_PATH}"
```
mrodm, line 45: Could it be change FOLDER here , so it is used as before inside release-manager.sh ? 

```suggestion
export FOLDER="${WORKSPACE}"/"${FOLDER_PATH}"
```

jlind23: Then it becomes BASE_DIR
`BASE_DIR="${WORKSPACE}/${FOLDER_PATH}"`

mrodm: @sharbuz I see also that inside `release_manager` it is used `$PWD`. what is the expected path where that docker command should run ?

sharbuz:
I added the `cd ${WORKSPACE}` command to avoid possible issues in the next lines.

mrodm (approved):

PR #2838, opened 2023-07-27
add export WORKSPACE

sharbuz:

fix the issue: https://buildkite.com/elastic/fleet-server-package-mbp/builds/242#018996d2-44d7-40e7-b368-59349dc25866

add `export ${WORKSPACE}` 

## Related issues

https://github.com/elastic/fleet-server/issues/2517

### DIFF
--- a/.buildkite/scripts/dra_release.sh
+++ b/.buildkite/scripts/dra_release.sh
@@ -11,7 +11,7 @@ export PROJECT="fleet-server"
 export TYPE=${1}
 export BRANCH="${BUILDKITE_BRANCH}"
 export VERSION="$(make get-version)"
-export ${WORKSPACE}
+export WORKSPACE
 
 if [[ "${VERSION}" == *"-SNAPSHOT"* || "${VERSION}" == "" ]]; then
     echo "The 'version' parameter is required and it cannot contain the suffix '-SNAPSHOT'."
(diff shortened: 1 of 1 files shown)


jlind23 (approved):

.buildkite/scripts/dra_release.sh
```diff
@@ -11,6 +11,7 @@ export PROJECT="fleet-server"
 export TYPE=${1}
 export BRANCH="${BUILDKITE_BRANCH}"
 export VERSION="$(make get-version)"
+export ${WORKSPACE}
```
mrodm, line 14: Should not be just `export WORKSPACE`

sharbuz: I checked it with $, and NP, changed it, please check it

mrodm (approved):

.buildkite/scripts/dra_release.sh
sharbuz: just tried to use one approach ))

mrodm: just tested with this:
```
 $ WORKSPACE="testtest"
 $ export |grep WORKSPACE
 $ 
 $ export ${WORKSPACE}
 $ export |grep WORKS
 $ export |grep testtest
declare -x testtest
 $ export WORKSPACE
 $ export |grep WORKS
declare -x WORKSPACE="a"
```

if it is used `${}` , it's being exported the contents of the env. var

Issue #2485, opened 2023-04-11
[Agent policy secrets] Retrieve agent policy secrets from `.fleet-secrets` index and replace secret references with secret values


hop-dev:
Part of https://github.com/elastic/kibana/issues/154715

🚧 
- parse all references out of agent policy `secret_refs` field
- get secrets by ID from the `.fleet-secrets` index
- replace all secret references with the secret themselves in the policy. Secret references are in the format `$co.elastic.secret{12345}`
- never log a secret value

PR #2863, opened 2023-08-07
replace secret refs

juliaElastic:

Agent policy secrets are sent to agents in agent policy.

On checkin, read secret values of secret references in the agent policy, and replace occurrences in input and stream variables.

- Start local es, kibana, fleet server
- Enable secrets feature flag in `kibana.dev.yml`: `xpack.fleet.enableExperimental: ['secretsStorage']`
- Start local registry with secret package here: https://github.com/elastic/kibana/pull/157176
- Open kibana, add integration: secret package, add values for package, input and stream secret vars
- When the agent policy with the secret integration is created, the agent policy should have 3 secret references
- Enroll an agent locally
- Download diagnostics and check in `computed-config.yml` that the inputs contain the secret values, not references
- Note: the agent is unhealthy with the error message `input not supported` is not related to secrets, looks like a validation [changed](https://github.com/elastic/elastic-agent/commit/64fc9d91dbea21ab51a9a72dcc12e4b42b723854#diff-db2140f630be8cde1ea80e76eff1c58c5a9d95022bff853724e205ee9fe58d9f) on agent side, so that the test package doesn't have a valid input type. Can be ignored for now.

Example:
```
inputs:
- data_stream:
    namespace: default
  id: test_input-secrets-7c6b08d8-dbe5-4278-8929-dfb26ea8e579
  input_var_secret: input
  meta:
    package:
      name: secrets
      version: 1.0.0
  name: secrets-1
  package_policy_id: 7c6b08d8-dbe5-4278-8929-dfb26ea8e579
  package_var_secret: pkg
  policy:
    revision: 5
  revision: 1
  streams:
  - config.version: "2"
    data_stream:
      dataset: secrets.log
      type: logs
    id: test_input-secrets.log-7c6b08d8-dbe5-4278-8929-dfb26ea8e579
    input_var_secret: input
    package_var_secret: pkg
    stream_var_secret: stream
  type: test_input
```

## Design Checklist

## Checklist

## Related issues

- Closes https://github.com/elastic/fleet-server/issues/2485


internal/pkg/dl/secret.go
juliaElastic, line 1: Following the example here: https://github.com/elastic/go-elasticsearch/blob/main/_examples/extension/main.go

internal/pkg/dl/secret.go
```diff
@@ -0,0 +1,54 @@
...
+	return &esapi.Response{StatusCode: res.StatusCode, Body: res.Body, Header: res.Header}, nil
+}
+
+type SecretResponse struct {
+	Id    string
+	Value string
+}
+
+func ReadSecret(client *elasticsearch.Client, secretId string) (string, error) {
+	es := ExtendedClient{Client: client, Custom: &ExtendedAPI{client}}
```
juliaElastic, line 42: the client could be created once and reused for multiple `ReadSecret` calls

internal/pkg/policy/parsed_policy.go
```diff
@@ -86,11 +97,97 @@ func NewParsedPolicy(p model.Policy) (*ParsedPolicy, error) {
 		Default: ParsedPolicyDefaults{
 			Name: defaultName,
 		},
+		Inputs: policyInputs,
 	}
 
+	// TODO remove before merge
```
juliaElastic, line 103: verified by writing out the result, will check to verify on the agent side

internal/pkg/policy/parsed_policy.go
```diff
@@ -38,15 +42,20 @@ type ParsedPolicyDefaults struct {
...
 	Policy  model.Policy
 	Fields  map[string]json.RawMessage
 	Roles   RoleMapT
 	Outputs map[string]Output
 	Default ParsedPolicyDefaults
+	Inputs  []map[string]interface{}
 }
 
-func NewParsedPolicy(p model.Policy) (*ParsedPolicy, error) {
+func NewParsedPolicy(p model.Policy, client *elasticsearch.Client) (*ParsedPolicy, error) {
```
juliaElastic, line 58: I'm not sure if this is the best place to use es client.

internal/pkg/dl/secret.go
joshdover: We should probably use the bulker for this and fetch multiple secrets in a single go with an `mget`. I think the bulker will do this automatically.

By the way - do we cache the policy after the secrets have been replaced? That way we don't have to do this replacement for every agent that needs this same policy update.

juliaElastic: This is a custom API in elastic, I'm not sure if `mget` works here unless we add a bulk get API to ES.
Not sure about the caching, will check.

juliaElastic: It doesn't seem that we cache the policy. We could cache the secret values or the whole policy.

joshdover: Forgot about the custom API endpoints. We'll have to test this at scale, but I think if we can cache the whole policy instead of re-fetching and re-building it every time, we can probably defer on adding an mget API.

I believe this caching should be safe since we don't modify policy revisions, only add new ones. One thing we have to be careful about is to not cache the version of the policy that has the agent-specific output credentials inserted. That should happen after the caching layer.

juliaElastic: Isn't it a concern though to cache the secret values? So that by taking a memory dump, secret values could be exposed.

juliaElastic:
I'm planning to add an integration or e2e test for this, though not sure how to create a secret as it requires `kibana_system` privilege. I would try to create a kibana user, unless someone has a better idea.

internal/pkg/policy/secret.go
```diff
@@ -0,0 +1,111 @@
...
+			if k == "streams" {
+				if streams, ok := input[k].([]any); ok {
+					newStreams := make([]any, 0)
+					for _, stream := range streams {
+						if streamMap, ok := stream.(map[string]interface{}); ok {
+							newStream := make(map[string]interface{})
+							for streamKey, streamVal := range streamMap {
+								if streamRef, ok := streamMap[streamKey].(string); ok {
+									replacedVal := replaceSecretRef(streamRef, secretReferences)
+									newStream[streamKey] = replacedVal
```
joshdover, line 74: Let's work on reducing the nesting here and adding any helpful comments on some of the conditions.

For example, maybe we can invert the condition on line 66 and use a `continue` to avoid on layer of nesting.

May also consider taking all of lines 66-74 and putting it into a separate named function

internal/pkg/policy/monitor.go
```diff
@@ -261,14 +261,14 @@ func (m *monitorT) loadPolicies(ctx context.Context) error {
...
-func (m *monitorT) processPolicies(_ context.Context, policies []model.Policy) error {
+func (m *monitorT) processPolicies(ctx context.Context, policies []model.Policy) error {
 	if len(policies) == 0 {
 		return nil
 	}
 
 	latest := m.groupByLatest(policies)
 	for _, policy := range latest {
-		pp, err := NewParsedPolicy(policy)
+		pp, err := NewParsedPolicy(ctx, m.bulker, policy)
```
joshdover, line 271: Ah we get policies through the monitor - so I think this will be shared with all agents current subscribed to checkin. That should minimize the amount of re-fetching we have to do here.

internal/pkg/policy/secret.go
```diff
@@ -0,0 +1,111 @@
...
+	}
+	ids := make([]string, 0)
+	for _, ref := range secretReferences {
+		ids = append(ids, ref.ID)
+	}
+	results, err := bulker.ReadSecrets(ctx, ids)
+	if err != nil {
+		return nil, err
+	}
+	return results, nil
```
joshdover, line 37: nit: readability cleanup

```suggestion
	if secretRefsRaw == nil {
		return nil, nil
	}

	var secretReferences []SecretReference
	err := json.Unmarshal([]byte(secretRefsRaw), &secretReferences)
	if err != nil {
		return nil, err
	}

	ids := make([]string, 0)
	for _, ref := range secretReferences {
		ids = append(ids, ref.ID)
	}

	results, err := bulker.ReadSecrets(ctx, ids)
	if err != nil {
		return nil, err
	}

	return results, nil
```

internal/pkg/policy/secret.go
```diff
@@ -0,0 +1,111 @@
...
+}
+
+// read inputs and secret_references from agent policy
+// replace values of secret refs in inputs and input streams properties
+func getPolicyInputsWithSecrets(ctx context.Context, fields map[string]json.RawMessage, bulker bulk.Bulk) ([]map[string]interface{}, error) {
+	if fields["inputs"] == nil {
+		return nil, nil
+	}
+	var inputs []map[string]interface{}
```
joshdover, line 46: nit

```suggestion
	if fields["inputs"] == nil {
		return nil, nil
	}

	var inputs []map[string]interface{}
```

internal/pkg/policy/secret.go
```diff
@@ -0,0 +1,111 @@
...
+	err := json.Unmarshal([]byte(fields["inputs"]), &inputs)
+	if err != nil {
+		return nil, err
+	}
+
+	if fields["secret_references"] == nil {
+		return inputs, nil
+	}
+
+	secretReferences, err := getSecretReferences(ctx, fields["secret_references"], bulker)
```
joshdover, line 56: naming nit: I think this function and it's return value should be called `getSecretValues` rather than references.

internal/pkg/policy/secret.go
```diff
@@ -0,0 +1,111 @@
...
+			}
+		}
+		result = append(result, newInput)
+	}
+	return result, nil
+}
+
+// replace values mathing a secret ref regex, e.g. $co.elastic.secret{<secret ref>} -> <secret value>
+func replaceSecretRef(ref string, secretReferences map[string]string) string {
+	regexp := regexp.MustCompile(`^\$co\.elastic\.secret{(.*)}$`)
```
joshdover, line 102: Let's put this in a constant at the top of the file to avoid re-compiling on each invocation.

Doesn't the `^` at the beginning of this regex mean the secret has to be at the beginning of the line? I wouldn't expect that as I think we have some of this form where only part of the value is a secret: `key: partial $co.elastic.secret{1234}`

internal/pkg/server/fleet_secrets_integration_test.go
```diff
@@ -0,0 +1,228 @@
...
+
+func createSecret(t *testing.T, ctx context.Context, bulker bulk.Bulk) string {
+	t.Log("Setup secret for fleet integration test")
+	esClient := bulker.Client()
+
+	req, err := http.NewRequestWithContext(ctx, "POST", "/_fleet/secret/", bytes.NewBuffer([]byte("{\"value\":\"secret_value\"}")))
+	req.Header.Set("Content-Type", "application/json")
+	req.Header.Set("Accept", "application/json")
+	// kibana_system:changeme base64 encoded
+	req.Header.Set("Authorization", "Basic a2liYW5hX3N5c3RlbTpjaGFuZ2VtZQ==")
```
juliaElastic, line 45: found that `kibana_system` already exists in local testing

internal/pkg/policy/secret.go
juliaElastic: yes, I thought we only have secrets for a full field value, I can change to a partial match if that is supported

internal/pkg/policy/secret.go
juliaElastic: Thanks for the suggestions, refactored.

joshdover (approved):
LGTM thanks for the e2e test!

juliaElastic:
I'm seeing some flakyness with integration tests using fake checkin request:
```
FAIL: TestServerInstrumentation (6.16s) | 2s
-- | --
  | === RUN   Test_SmokeTest_Agent_Calls
  | fleet_integration_test.go:598: Enroll an agent
  | fleet_integration_test.go:608: Agent enrollment successful, verify body
  | fleet_integration_test.go:633: Fake a checkin for agent d56ff08b-bad3-452d-8bb5-f012c5e90127
  | fleet_integration_test.go:640:
  | Error Trace:	/opt/buildkite-agent/builds/bk-agent-prod-gcp-1691594516224871631/elastic/fleet-server/internal/pkg/server/fleet_integration_test.go:640
  | Error:      	Received unexpected error:
  | Post "http://localhost:46833/api/fleet/agents/d56ff08b-bad3-452d-8bb5-f012c5e90127/checkin": EOF


```

juliaElastic:
The integration tests all pass now, but there is still an issue that I don't know how to fix:
```
--- PASS: TestCheckCompatibilityInternal/supported_elasticsearch_715-800a1 (0.00s)
--
  | PASS
  | ok  	github.com/elastic/fleet-server/v7/internal/pkg/ver	0.062s
  | ?   	github.com/elastic/fleet-server/v7/pkg/api	[no test files]
  | ?   	github.com/elastic/fleet-server/v7/pkg/api/versions/2022_06_01/api	[no test files]
  | ?   	github.com/elastic/fleet-server/v7/version	[no test files]
  | FAIL
  | make[1]: *** [Makefile:328: test-int-set] Error 1
  | make[1]: Leaving directory '/opt/buildkite-agent/builds/bk-agent-prod-gcp-1691664493893470492/elastic/fleet-server'
  | make: *** [Makefile:317: test-int] Error 2
  | 🚨 Error: The command exited with status 2
  | user command error: exit status 2

```

Line 328 is this:
https://github.com/elastic/fleet-server/blob/4834215261f429bb74d41bdf9bab32f2c013da98/Makefile#L328

joshdover:
Did the last commit fix? Seems like it should. Otherwise I can dig in to help

juliaElastic:
> Did the last commit fix? Seems like it should. Otherwise I can dig in to help

I had a passing build before [this](https://github.com/elastic/fleet-server/pull/2863/commits/56cfcd279be8ec75c2e0d2076a2e318bf8518f5d) commit, which excludes some unit tests from the integration tests run, not sure if this is flakyness, or something breaks when I exclude those.

internal/pkg/policy/secret_test.go
```diff
@@ -0,0 +1,127 @@
+// Copyright Elasticsearch B.V. and/or licensed to Elasticsearch B.V. under one
+// or more contributor license agreements. Licensed under the Elastic License;
+// you may not use this file except in compliance with the Elastic License.
+
+//go:build !integration
```
juliaElastic, line 5: I don't understand though we we have to exclude unit tests like this explicitly, as integration tests are tagged also.
Are we missing something?
the Makefile has this command to tigger integration tests:
`go test -v -tags=integration -count=1 -race -p 1 ./...`

jsoriano: I guess that the idea is to don't run unit tests when running integration tests, so the same tests are not executed at two different stages of testing.

juliaElastic: Yes, I mean shouldn't it be enough to mark integration tests with `//go:build integration`? I think it's easy to forget when adding new unit tests to add the negated condition, and I'm trying to understand why it is needed.

jsoriano: Files without any build tag are always included, so `-tags=integration` wouldn't discard unit test files if they don't have the negated tag. So we need `!integration` for test files that we don't want to execute during integration tests.

I agree that this is easy to forget. Tough I wouldn't see as a great problem to execute some fast unit tests during integration test execution.

If we strongly want to have their execution separated, a possible strategy to don't forget it is to have only unit tests in packages, and move the integration tests to their own package/s, in a similar fashion to what we do for [e2e tests](https://github.com/elastic/fleet-server/tree/main/testing). This also help to validate interfaces. 
Or we could have some check in CI to ensure that all test files have tags.

PR #2880, opened 2023-08-15
[Fleet] Fix static enrollment tokens blocking db enrollemnt tokens

### LINKED ISSUE #2877: Enabling static enrollment tokens disable regular ones
## Description 

Enabling static enrollment tokens seems to disable regular ones, we should support both working together

## How to reproduce

1. Configure static enrollment tokens
2. Create an agent policy in Fleet and try to enroll an agent 

you should get the following error message
```
Error: fail to enroll: fail to execute request to fleet-server: status code: 500, fleet-server returned an error: BadRequest, message: policy not found
```

nchaulet:
## Description 

Resolve #2877 

Fix an issue when enabling static enrollement tokens it was disabling support for regular db enrollment tokens.

That PR fix that and add a unit test to cover that scenario.

## Manual tests

* Add some static tokens to your fleet server config
* Create a policy in the fleet UI and check you can enroll agents into that policy.

### DIFF
--- a/internal/pkg/api/handleEnroll.go
+++ b/internal/pkg/api/handleEnroll.go
@@ -100,7 +100,7 @@ func (et *EnrollerT) handleEnroll(zlog zerolog.Logger, w http.ResponseWriter, r 
 func (et *EnrollerT) processRequest(zlog zerolog.Logger, w http.ResponseWriter, r *http.Request, rb *rollback.Rollback, enrollmentAPIKey *apikey.APIKey, ver string) (*EnrollResponse, error) {
 	// Validate that an enrollment record exists for a key with this id.
 	var enrollAPI *model.EnrollmentAPIKey
-	enrollAPI, err := et.fetchStaticTokenPolicy(r.Context(), zlog, enrollmentAPIKey)
+	enrollAPI, err := et.retrieveStaticTokenEnrollmentToken(r.Context(), zlog, enrollmentAPIKey)
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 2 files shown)


internal/pkg/api/handleEnroll.go
```diff
@@ -161,7 +161,7 @@ func (et *EnrollerT) fetchStaticTokenPolicy(ctx context.Context, zlog zerolog.Lo
 
 	}
 	// no error, just not found
-	return nil, ErrPolicyNotFound
+	return nil, nil
```
jsoriano, line 164: Nit. Instead of returning `nil`, `nil`, wdyt about handing `ErrPolicyNotFound` in the caller? It is usually assumed that when a method returns _something_ and an error, this _something_ can be used if the error is nil.

If we decide to keep returning a `nil` error here, maybe we should do the same in [fetchPolicy](https://github.com/elastic/fleet-server/pull/2880/files#diff-b3d2d3a49ef214fba08e3aa10772162cbb65c378e24ef6c4a0c5df65b051f71aR180) to keep the same semantics in both places.

nchaulet: The method already return `nil,nil` in some case, it seems a little different than `fetchPolicy` as that method conditionally fetch or not a static token, maybe the method could be renamed to something more clear

nchaulet: I renamed the function in https://github.com/elastic/fleet-server/pull/2880/commits/485720e76f2e9f2637f7ecce2ae4c8a4d16b1df2 let me know if it make better sense to you

jsoriano (approved):

michel-laterman (approved):

PR #2898, opened 2023-08-21
Update pipeline.yml to remove a non existing step

jlind23:
Remove the depending step as it does not exist in 7.17

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -69,3 +69,6 @@ steps:
     label: ":esbuild: Downstream - Package"
     key: "downstream-package"
     if: "build.env('BUILDKITE_PULL_REQUEST') == 'false' && build.env('BUILDKITE_TAG') == '' && build.env('BUILDKITE_BRANCH') != ''"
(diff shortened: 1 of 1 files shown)


juliaElastic (approved):

.buildkite/pipeline.yml
```diff
@@ -69,6 +69,3 @@ steps:
     label: ":esbuild: Downstream - Package"
     key: "downstream-package"
     if: "build.env('BUILDKITE_PULL_REQUEST') == 'false' && build.env('BUILDKITE_TAG') == '' && build.env('BUILDKITE_BRANCH') != ''"
-    depends_on:
```
sharbuz, line 72: depends_on:
      - step: "tests"
        allow_failure: false

juliaElastic (approved):

.buildkite/pipeline.yml
sharbuz: "release-test" is correct dependency

sharbuz (approved):

PR #2901, opened 2023-08-21
Add new upgrade_details field used in checkin

michel-laterman:

Support more detailed upgrade status that's defined as part of the checkin request.

Add upgrade_details to checkin request body and to agent index to support the new way of communicating upgrade status. Keep support for previous ack based status.

`checkin_details` is defined by the checkin request body in our openapi definition. `metadata` objects are defined using openapi's `oneOf` option.
The fleet-server's model of the attribute defines it as `json.RawMessage` with the [actual elasticsearch issue](https://github.com/elastic/elasticsearch/issues/97912) handling its implementation.

## Checklist


internal/pkg/api/handleCheckin.go
```diff
@@ -274,6 +282,21 @@ func (ct *CheckinT) ProcessRequest(zlog zerolog.Logger, w http.ResponseWriter, r
 	return ct.writeResponse(zlog, w, r, agent, resp)
 }
 
+func (ct *CheckinT) processUpgradeDetails(ctx context.Context, agent *model.Agent, details *UpgradeDetails) error {
```
michel-laterman, line 285: How much additional validation do we need to do here? Do we want to verify that the actionID is valid for the agent? Should we verify that the `metadata` attribute is valid for the `state`?

model/openapi.yml
```diff
@@ -1,4 +1,4 @@
-openapi: 3.0.0
+openapi: 3.1.0
```
michel-laterman, line 1: Note that the spec version has been upgraded and some minor changes [following the migration guide](https://www.openapis.org/blog/2021/02/16/migrating-from-openapi-3-0-to-3-1-0) have been made.

michel-laterman (commented):
As far as implementation goes;
1. Should an `ack` for an upgrade action clear the `upgrade_details` field?
2. Should a checkin without an `upgrade_details` field clear it?
3. Should a checkin with `upgrade_details` clear the [existing upgrade attributes](https://github.com/elastic/fleet-server/blob/main/internal/pkg/model/schema.go#L205-L215)

jlind23:
Maybe @juliaElastic or @ycombinator know the answer here.

ycombinator:
> As far as implementation goes;
> 
> 1. Should an `ack` for an upgrade action clear the `upgrade_details` field?

No. See answer to 2. below.

> 2. Should a checkin without an `upgrade_details` field clear it?

Yes.  An Agent that's capable of sending an `upgrade_details` field is expected to keep sending it until an upgrade has completed (successfully or unsuccessfully). So if Fleet Server sees a checkin request without the `upgrade_details` field in it, it should clear it.

> 3. Should a checkin with `upgrade_details` clear the [existing upgrade attributes](https://github.com/elastic/fleet-server/blob/main/internal/pkg/model/schema.go#L205-L215)

I need to think about this one and have some follow up questions regarding the upgrade attributes:

1. What clears these today? An ack for the upgrade action?
2. What is the difference between `UpdatedAt` and `UpgradedAt`?
3. What are some examples of values that are stored in `UpgradeStatus`?
4. When does `UpgradeStatus` get set or updated?

jlind23:
@michel-laterman @nchaulet @sharbuz @mrodm does anyone know what can be the E2E tests problems here? The log message looks like this:
```

------
--
  | > COPY --chmod=0755 --chown=elastic-agent cover/fleet-server-8.10.0-SNAPSHOT-linux-x86_64/fleet-server              ./data/elastic-agent-a3fb47/components/fleet-server:
  | ------
  | Dockerfile:1
  | --------------------
  | 1 \| >>> ARG ELASTIC_AGENT_IMAGE # e.g. docker.elastic.co/cloud-release/elastic-agent-cloud:8.8.0-4671daa2-SNAPSHOT
  | 2 \|
  | 3 \|     FROM --platform=linux/amd64 ${ELASTIC_AGENT_IMAGE} as elastic_agent_amd64
  | --------------------
  | ERROR: failed to solve: failed to compute cache key: failed to calculate checksum of ref 30f39f47-e53a-452e-8f08-bcf05c8c193f::0zlbb1tnh0ynawt5syqveugl6: failed to walk /var/lib/docker/tmp/buildkit-mount2000922373/cover/fleet-server-8.10.0-SNAPSHOT-linux-x86_64: lstat /var/lib/docker/tmp/buildkit-mount2000922373/cover/fleet-server-8.10.0-SNAPSHOT-linux-x86_64: no such file or directory

```

juliaElastic:
I think when we clear the `upgrade_details` field, it should go together with resetting the `upgrade_started_at` and setting the `upgraded_at` fields, so the high level status is in consistent state.

internal/pkg/api/handleCheckin.go
michel-laterman: @ycombinator, should we to additional validation for the `upgrade_details` attribute?

ycombinator: In general, I would perform the same validation as we do today for the ack API path. So I would imagine that means verifying that the actionID is valid for the agent.

As for whether the `metadata` attribute is valid for the `state`, I think you are referring to this part of the `upgrade_details` object:

```
    "state": "UPG_*",
    "metadata": {
      "scheduled_at": "2023-08-09T10:11:12Z", // when state == "UPG_SCHEDULED"
      "download_percent": 16.4, // when state == "UPG_DOWNLOADING"
      "failed_state": "UPG_*" // when state == "UPG_FAILED"
      "error_msg": "" // when state == "UPG_FAILED"
    }
```

It would be nice to have validation at this finer level but I'd start with just validating `metadata.scheduled_at` and `metadata.error_msg`.  I think for the remaining two metadata properties, we'll need to see whether Agent is capable of setting those reliably.

changelog/fragments/1692641595-Add-support-for-upgrade_details-on-checkin.yaml
```diff
@@ -0,0 +1,34 @@
...
+  Keep old ack based upgrade logic in place.
+
+# Affected component; a word indicating the component this changeset affects.
+component:
+
+# PR URL; optional; the PR number that added the changeset.
+# If not present is automatically filled by the tooling finding the PR where this changelog fragment has been added.
+# NOTE: the tooling supports backports, so it's able to fill the original PR number instead of the backport PR number.
+# Please provide it if you are adding a fragment for a different PR.
+#pr: https://github.com/owner/repo/1234
```
juliaElastic, line 30:
```suggestion
pr:2901
```

juliaElastic (approved):
LGTM

jlind23:
@michel-laterman can we merge this?

juliaElastic:
@ycombinator Could you review it too?

internal/pkg/api/handleCheckin.go
```diff
@@ -165,6 +166,12 @@ func (ct *CheckinT) ProcessRequest(zlog zerolog.Logger, w http.ResponseWriter, r
 		return err
 	}
 
+	// Handle upgrade details for agents using the new 8.11 upgrade details field of the checkin. // FIXME verify version
```
ycombinator, line 169: Do we still need the FIXME?

ycombinator (commented):
This looks good to me me but I had some questions that I believe are still pending — see end of https://github.com/elastic/fleet-server/pull/2901#issuecomment-1686961935.

internal/pkg/api/handleCheckin.go
michel-laterman: nope, i'll remove and merge

michel-laterman:
@ycombinator sorry about the delay, I updated the relevant attributes after @juliaElastic's comments

> I need to think about this one and have some follow up questions regarding the upgrade attributes:
> 
> 1. What clears these today? An ack for the upgrade action?

yes, currently the ack an agent sends after the new version starts (or an upgrade fails) sets these

> 2. What is the difference between `UpdatedAt` and `UpgradedAt`?

iirc `UpdatedAt` is the agent heartbeat that fleet-server updates while the agent is in the long-poll phase of a checkin, `UpgradedAt` is the time when the agent was last upgraded

> 3. What are some examples of values that are stored in `UpgradeStatus`?

Currently I think it was only used to indicate failed/retring

> 4. When does `UpgradeStatus` get set or updated?

I think just in the code that [handles upgrade acks](https://github.com/elastic/fleet-server/blob/main/internal/pkg/api/handleAck.go#L554)

ycombinator:
Thanks for clarifying, @michel-laterman.

I agree with @juliaElastic's comment (https://github.com/elastic/fleet-server/pull/2901#issuecomment-1689577109) as a general guiding principle. We should do whatever makes sense so "the high level status is in consistent state [between the old and new methods of communicating upgrade state]".

Given that:

> Should a checkin with upgrade_details clear the [existing upgrade attributes](https://github.com/elastic/fleet-server/blob/main/internal/pkg/model/schema.go#L205-L215)

I'm not sure about "clearing" as in deleting but I agree that a checkin with `upgrade_details` should _update_ any relevant upgrade attributes, which is what I believe you've already implemented in this PR. So I think we're good here! Thanks again.

PR #2909, opened 2023-08-23
changes to serverless pipelines

juliaElastic:

Changes based on https://github.com/elastic/hello-world-service/blob/main/.buildkite/pipeline.hello-world-service-tests.yaml

tests-qa: before qa promotion, this will trigger observability-perf tests in a new dev env
tests-staging: before staging promotion, this will trigger observability-perf tests in qa env

## Design Checklist

## Checklist

## Related issues

Relates https://github.com/elastic/ingest-dev/issues/2201


jlind23:
Can't we merge `tests-qa` and `tests-staging` into a single pipeline and use just pass down the environment variable?

juliaElastic:
> Can't we merge `tests-qa` and `tests-staging` into a single pipeline and use just pass down the environment variable?

There will be differences between qa and staging later (like different quality gates, creating dev env before qa deploy), so we should keep them separate.

jlind23 (approved):

.buildkite/pipeline.fleet-server-tests.yaml
```diff
@@ -0,0 +1,25 @@
...
+# The updated environment is provided via ENVIRONMENT variable. The seedling
+# step will branch and execute pipeline snippets at the following location:
+# .buildkite/pipeline.tests-qa.yaml
+# .buildkite/pipeline.tests-staging.yaml
+# .buildkite/pipeline.tests-production.yaml
+#
+# Docs: https://docs.elastic.dev/serverless/qualitygates
+
+env:
+  ENVIRONMENT: ${ENVIRONMENT}
```
ramonbutter, line 14:
```suggestion
  ENVIRONMENT: ${ENVIRONMENT?}
```

.buildkite/pipeline.tests-staging.yaml
```diff
@@ -0,0 +1,11 @@
+# These pipeline steps constitute the quality gate for your service within the QA environment.
```
ramonbutter, line 1: Cool that you got the latest template for this
```suggestion
# These pipeline steps constitute the quality gate for your service within the Staging environment.
```

.buildkite/pipeline.fleet-server-tests.yaml
juliaElastic: This was causing [issues](https://github.com/elastic/fleet-server/pull/2851) before on main when triggered without ENVIRONMENT. How can we prevent this pipeline from being run without ENVIRONMENT?

ramonbutter: This pipeline has to have a Target Environment for its tests, 
otherwise it has to fail. 
And the earlier the better. 
It is the task of any invoker to guarantee that the ENVIRONMENT is set.
When you run manaul builds, it should be handed over via run settings.

But I am not sure If I am following. 
This should NOT run on PRs or main merges or any github actions at all. I think we need to add [after this](https://github.com/elastic/fleet-server/blob/22832f2b17cafd1891952d3c19ef80432b1f24c9/catalog-info.yaml#L134).
```
      provider_settings:
        trigger_mode: none
```

catalog-info.yaml
```diff
@@ -128,10 +128,11 @@ spec:
...
     spec:
       repository: elastic/fleet-server
-      pipeline_file: .buildkite/pipeline.service-tests.yml
+      pipeline_file: .buildkite/pipeline.service-tests.yaml
       branch_configuration: "main"
       provider_settings:
         build_pull_requests: false
+        trigger_mode: none
       teams:
         cloud-k8s-region:
```
ramonbutter, line 137: I think you want to replace this with your team here.
Otherwise you would not be able to manually invoke and test the pipline afaict

jlind23:
buildkite test it

PR #2918, opened 2023-08-25
fix catalog-info

juliaElastic:
Fix https://github.com/elastic/fleet-server/pull/2909
Renamed the pipeline file and should be updated in catalog-info.yaml


jlind23 (approved):
🚀

ramonbutter (approved):

PR #2939, opened 2023-09-01
TLS-626 make sure e2e tests are being triggered before updating gitops

kroustou:
https://elasticco.atlassian.net/browse/TLS-626

We have bundled the e2e tests with the gpctl pipeline.

## Design Checklist

## Checklist

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -100,11 +100,14 @@ steps:
   - label: ":argo: Update fleet image tag for fleet-controller using gpctl"
     async: true
     branches: main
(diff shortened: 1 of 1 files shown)


.buildkite/pipeline.yml
```diff
@@ -100,11 +100,14 @@ steps:
   - label: ":argo: Update fleet image tag for fleet-controller using gpctl"
```
juliaElastic, line 100: looks like a typo here
```suggestion
  - label: ":argo: Update fleet image tag for fleet-server using gpctl"
```

jlind23 (approved):

PR #2962, opened 2023-09-11
Add e2e coverage for sonar

### LINKED ISSUE #2857: export e2e coverage results to sonarcloud
Current sonarcloud builds on main are failing due to less then 40% code coverage.
e2e tests for non-docker deployments generate coverage artifacts.
These artifacts should be exported to sonarcloud in a format it understands so our current code coverage can more accurately reflect our testing efforts

michel-laterman:

e2e test that run in pipeline just make sure the container is healthy and don't test endpoints.
Coverage information is not collected.

Add go integration test coverage export for e2e tests and collect with sonar. Run all stand alone e2e tests instead of just the basic container tests.

`make test-e2e`

## Related issues

- Closes #2857


michel-laterman:
The PR re-enables endpoint tests, but I'm not sure if coverage will actually be collected.
The sonarqube action [runs unit tests](https://github.com/elastic/fleet-server/blob/main/.github/workflows/sonarcloud.yml#L112), then collects coverage information.
@jlind23 do we need to add `test-e2e` as a target step in the workflow, or should we try to adjust buildkite to export the coverage report?

jlind23:
@michel-laterman we should modify the buildkite pipeline in order to collect coverage information from unit test and not e2e tests.

jlind23 (approved):

michel-laterman:
For some reason the sonarqube branch job also fails: https://github.com/elastic/fleet-server/actions/runs/6175649746/job/16762960444

PR #2964, opened 2023-09-12
Update version to 8.10.1

pierrehilbert:

## Design Checklist

## Checklist


juliaElastic (approved):

pierrehilbert:
@juliaElastic / @jlind23 : what should I do with those failing checks?

jlind23:
The E2E test is failing as usual because there are no ES binaries in 8.10.1 yet.

PR #2977, opened 2023-09-19
Add PGP key retrieval endpoint

### LINKED ISSUE #2887: Allow Fleet Server to serve the PGP key when stored on disk
If Elastic Agent are unable to connect with Elastic public URL to retrieve the PGP key, they will fallback to a Fleet Server URL where the public key can be hosted too.

In Fleet Server, we must define a fixed route for hosting the GPG key `downloads/signing/key.pub`  to mirror the public URL and have it read the GPG key in from its configuration using the same secure mechanisms we use to read the TLS private key. 
Mandate HTTPS on this endpoint and let Elastic Agent call it whenever needed.

Related Elastic Agent issue - https://github.com/elastic/elastic-agent/issues/3264

michel-laterman:

Fleet-server must be able to have a request path where a valid PGP key will be served to requesting agents.
This endpoint must only be valid when the server runs HTTPS.
An operator must be able to change the key that is served.

Add a new endpoint `GET /api/agents/upgrades/{major}.{minor}.{patch}/pgp-public-key` that will return a PGP key.
The endpoint needs a valid API key and is rate limited.
Currently the `major`, `minor`, and `patch` values are ignored.

The endpoint will server a PGP key from it's cache if one exists, otherwise it will serve the contents of a PGP key found on disk at `FLEETSERVER_BINARY_DIR/elastic-agent-upgrade-keys/default.pgp`. Note that the directory can be changed through configuration.
The key value is added to cache if it exists.

If this file does not exist the fleet-server will retrieve the key from `https://artifacts.elastic.co/GPG-KEY-elastic-agent` and serve it's contents (writing them to cache and disk). The upstream URL can also be changed through configuration.

## Design Checklist

## Checklist

## Related issues

- Closes #2887


README.md
```diff
@@ -260,6 +260,17 @@ the test BenchmarkXXXX and not unit tests)
 - `BENCHMARK_COUNT`: you can define the number of iterations go test will run. Having larger number helps
 remove run-to-run variations (Default: 8)
 
+#### E2E Tests
```
nchaulet, line 263: 🎉 nice to add some doc on how to run tests

nchaulet:
I am wondering if we should read the key on start as we do for the the other filer we read from the config like (`SecretTokenPath` , ...) it will be less confusing for the user to change the key it will have to restart fleet-server and not rely on a cache expiring no?

internal/pkg/api/handlePGPRequest.go
```diff
@@ -0,0 +1,177 @@
...
+		return nil, err
+	}
+
+	// file does not exist
+	p, err = pt.getPGPFromUpstream(ctx)
+	if err != nil {
+		return nil, err
+	}
+	pt.cache.SetPGPKey(key, p)
+	pt.writeKeyToDir(ctx, zlog, key, p)
```
nchaulet, line 107: If we write the upstream key to disk, this mean if that key is updated user will have to manually update it on their fleet-server?

michel-laterman: Yes, if the upstream key is updated the user will need to manually update the key on disk (or remove it)

internal/pkg/api/handlePGPRequest.go
```diff
@@ -0,0 +1,177 @@
...
+	if !errors.Is(err, fs.ErrNotExist) {
+		return nil, err
+	}
+
+	// file does not exist
+	p, err = pt.getPGPFromUpstream(ctx)
+	if err != nil {
+		return nil, err
+	}
+	pt.cache.SetPGPKey(key, p)
```
michel-laterman, line 106: @nchaulet If we wanted to make changing the key source easier we could use path+key when interacting with the cache.

If an operator needs to do an emergency release in this case just changing `pgp.dir` to a new path would immediately take effect (and either retrieve the key from disk or from a new upstream url if that's changed).
What do you think?

nchaulet: Yes I think it could be a better solution

nchaulet (approved):
Thanks for addressing the cache changes 🚀

PR #2991, opened 2023-09-26
turn down error logs in serverless

juliaElastic:

Serverless quality gates started failing since disabling debug logs: https://github.com/elastic/fleet-controller/pull/219

Modify most common error logs, and turning them to warn if matching the error message around ES not being available/DNS not ready.
Increased retry delay in monitor.

## Design Checklist

## Checklist

### DIFF
--- a/internal/pkg/api/server.go
+++ b/internal/pkg/api/server.go
@@ -11,7 +11,6 @@ import (
 	slog "log"
 	"net"
 	"net/http"
-	"strings"
 
 	"github.com/elastic/elastic-agent-libs/transport/tlscommon"
 	"github.com/elastic/fleet-server/v7/internal/pkg/build"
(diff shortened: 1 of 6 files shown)


internal/pkg/api/server.go
```diff
@@ -102,7 +103,11 @@ func (s *server) Run(ctx context.Context) error {
 	defer func() {
 		err := ln.Close()
 		if err != nil {
-			log.Error().Err(err).Msg("error while closing listener.")
+			if strings.Contains(err.Error(), "use of closed network connection") {
+				log.Warn().Err(err).Msg("server.Run: error while closing listener, closed network connection.")
+			} else {
+				log.Error().Err(err).Msg("server.Run: error while closing listener.")
```
juliaElastic, line 109: sonar check failed, any concerns if we turn these errors down to `Warn` without condition?

PR #2994, opened 2023-09-26
Add ActionLimit and a gzip writer pool to handle checkin responses

michel-laterman:

Small containers OOM when issuing updates to many clients, this could be mitigated by adding a rate limiter and using a pool for the gzip writter.

Add a sync.Pool to share gzip writers among all resquests in order to avoid new gzip.Writer allocations. Each writer allocation is around 1.2MB and is a limiting factor of hitting larger scales. Add action dispatch limiter to control how quickly the checkin endpoint writes responses in order to make better use of the writer pool.

## Design Checklist

- ~~I have or intend to scale test my changes, ensuring it will work reliably with 100K+ agents connected.~~ 10k test for serverless complete

## Checklist

### DIFF
--- a/fleet-server.reference.yml
+++ b/fleet-server.reference.yml
@@ -141,15 +141,18 @@ fleet:
 #       max_agents: 0
 #       # policy_throttle is the duration that the fleet-server will wait in between attempts to dispatch policy updates to polling agents # TODO verify this
 #       policy_throttle: 5ms # 1ms min is forced
-#       # action_throttle is added as a delay before the in checkin requests to help space response writes
-#       # This is done in order to be able to reuse gzip writers if gzip is requested as allocating new writers is expensive
-#       # should never me more then 2m to prevent edge cases with assumptions around poll-times and server writes from becoming invalid
-#       # a 0 value disables the throttle (default behavior)
-#       action_throttle: 0
 #       # max_header_byte_size is the request header size limit
 #       max_header_byte_size: 8192 # 8Kib
 #       # max_connections is the maximum number of connnections per API endpoint
 #       max_connections: 0
+#
+#       # action_limit is a limiter for the action dispatcher, it is added to control how fast the checkin endpoint reponds when an action effecting multiple agents is detected.
+#       # This is done in order to be able to reuse gzip writers if gzip is requested as allocating new writers is expensive
+#       # If the value is too high it may effect the server/request write times (generally 2m higher then checkin poll times).
+#       # a 0 value disables the limiter by using an infinite value instead (default behavior)
+#       action_limit:
+#         interval: 0
+#         burst: 10
 #
 #       # endpoint specific limits below
 #       checkin_limit:
--- a/fleet-server.reference.yml
+++ b/fleet-server.reference.yml
@@ -120,10 +120,6 @@ fleet:
 #       # checkin_jitter time may be subtracted from the long_poll time.
 #       # a 0 value disables jitter
 #       checkin_jitter: 30s
-#       # checkin_response_jitter is added as a random delay before the checkin method starts writing a response to a checkin
-#       # should never me more then 2m to prevent edge cases with assumptions around poll-times and server writes from becoming invalid
-#       # a 0 value disables jitter
-#       checkin_response_jitter: 100ms
 #       # checkin_max_poll is the maximum long_poll value a client can request.
 #       checkin_max_poll: 1h
 #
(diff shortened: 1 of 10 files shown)


fleet-server.reference.yml
```diff
@@ -141,6 +141,11 @@ fleet:
 #       max_agents: 0
 #       # policy_throttle is the duration that the fleet-server will wait in between attempts to dispatch policy updates to polling agents # TODO verify this
 #       policy_throttle: 5ms # 1ms min is forced
+#       # action_throttle is added as a delay before the in checkin requests to help space response writes
+#       # This is done in order to be able to reuse gzip writers if gzip is requested as allocating new writers is expensive
+#       # should never me more then 2m to prevent edge cases with assumptions around poll-times and server writes from becoming invalid
+#       # a 0 value disables the throttle (default behavior)
+#       action_throttle: 0
```
cmacknz, line 148:
```suggestion
#       # action_throttle is added as a delay before writing checkin responses to help space out response writes
#       # This is done in order to be able to reuse gzip writers if gzip is requested as allocating new writers is expensive
#       # should never be more then 2m to prevent edge cases with assumptions around poll-times and server writes from becoming invalid
#       # a 0 value disables the throttle (default behavior)
#       action_throttle: 0
```

A few typos.

internal/pkg/action/dispatcher.go
```diff
@@ -122,6 +124,14 @@ func (d *Dispatcher) process(ctx context.Context, hits []es.HitT) {
 	}
 
 	for agentID, actions := range agentActions {
+		if d.throttle != 0 {
+			select {
+			case <-ctx.Done():
+				return
+			case <-time.After(d.throttle):
```
cmacknz, line 131: The reference documentation above doesn't make it clear that the throttle delay is cumulative for each agent dispatching an action.

internal/pkg/action/dispatcher_test.go
```diff
@@ -224,9 +262,13 @@ func Test_Dispatcher_Run(t *testing.T) {
 			}()
 
 			ticker := time.NewTicker(time.Second * 5)
+
 			select {
 			case actions := <-d.subs["agent1"].Ch():
 				compareActions(t, tt.expect["agent1"], actions)
+				if tt.throttle != 0 {
+					assert.Greater(t, time.Now(), now.Add(1*tt.throttle))
```
cmacknz, line 270: You might want these conditions to be `GreaterOrEqual` to avoid flakiness.

internal/pkg/api/handleCheckin.go
```diff
@@ -498,13 +510,12 @@ func (ct *CheckinT) writeResponse(zlog zerolog.Logger, w http.ResponseWriter, r
 	if len(payload) > compressThreshold && compressionLevel != flate.NoCompression && acceptsEncoding(r, kEncodingGzip) {
 		wrCounter := datacounter.NewWriterCounter(w)
 
-		zipper, err := gzip.NewWriterLevel(wrCounter, compressionLevel)
-		if err != nil {
-			return fmt.Errorf("writeResponse new gzip: %w", err)
-		}
+		zipper, _ := ct.gwPool.Get().(*gzip.Writer)
```
cmacknz, line 513: Add a comment that the effectiveness of the pool depends on spreading these responses out over time.

cmacknz: This comment should probably explicitly reference the importance of the ActionThrottle configuration parameter.

cmacknz (commented):
Generally looks good, a couple of places comments can be improved.

internal/pkg/config/env_defaults.go
```diff
@@ -25,6 +25,7 @@ const (
 
 	defaultMaxConnections = 0 // no limit
 	defaultPolicyThrottle = time.Millisecond * 5
+	defaultActionThrottle = 0 // no throttle
```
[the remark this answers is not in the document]
nchaulet: I put this as a default so this change do not impact current on-prem/cloud deployment, but it probably make sense to match it

juliaElastic (commented):
LGTM, comments should be fixed.

internal/pkg/api/handleCheckin.go
michel-laterman: added in struct defintion

internal/pkg/config/env_defaults.go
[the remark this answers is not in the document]
michel-laterman: The point of the 0 values was to not change existing behaviour; with the new limiter a 0 results in an infinite rate limit (unchanged behaviour)

nchaulet (approved):
LGTM 🚀 tested locally work as expected

PR #3014, opened 2023-10-05
Buildkite migration is now complete, removing fleet-server jobs

jlind23:

## Design Checklist

## Checklist


michel-laterman:
Looks like `.ci/scripts/test-release.sh` is used by buildkite, it should probably be moved into .buildkite/scripts

Makefile
```diff
@@ -154,7 +154,7 @@ test: prepare-test-context  ## - Run all tests
 
 .PHONY: test-release
 test-release:  ## - Check that all release binaries are created
-	./.ci/scripts/test-release.sh $(DEFAULT_VERSION)
+	./.buildkite/scripts/test-release.sh $(DEFAULT_VERSION)
```
juliaElastic, line 157: Should the `test-release.sh` file be moved to `.buildkite`?

jlind23: Yes 🤦🏼

juliaElastic (approved):

jlind23:
the buildkite build will fail until the unified release for 8.12 is done.

jlind23:
buildkite test this

jlind23:
buildkite test this

jlind23:
Blocked by https://github.com/elastic/fleet-server/pull/2999

PR #3023, opened 2023-10-10
[Fleet] Fix kibana config in e2e tests

nchaulet:
## Description 

Fix kibana config file after that change  https://github.com/elastic/kibana/pull/167085

Also improved the script that is running kibana to fail on docker error and show the logs


kpollich (approved):
LGTM 🚀

PR #3024, opened 2023-10-10
Update oapi-codegen to v1.15.0, regenerate old api client

michel-laterman:
Update oapi-codegen to v1.15.0 and regenerate the 2022_06_01 client from commit 1d28f4ca4ec62f940771e5eb72b77699d256525f with the new version in order to get rid of unneeded dependencies.


kpollich (approved):
Bad dependency is gone so LGTM 🚀

michel-laterman:
the snyk check is failing because `github.com/oapi-codegen/runtime` has an indirect import of `github.com/gomarkdown/markdown/parser` that has an issue, the import in runtime is:
```
# github.com/gomarkdown/markdown/parser
github.com/oapi-codegen/runtime/strictmiddleware/iris
github.com/kataras/iris/v12
github.com/kataras/iris/v12/context
github.com/gomarkdown/markdown
github.com/gomarkdown/markdown/parser
```
fleet-server does not do any markdown parsing so I think we are safe to ignore this until it is fixed upstream

PR #3032, opened 2023-10-12
[Fleet] Quality gates for production canary

nchaulet:
## Description 

Add quality gates for production canary.

Also add baking period for staging and production:
* 30 minute for staging
* 1h for production canary

### DIFF
--- a/.buildkite/pipeline.tests-production-canary.yaml
+++ b/.buildkite/pipeline.tests-production-canary.yaml
@@ -2,22 +2,12 @@
 # A failure in this pipeline build will prevent further progression to the subsequent stage.
 
 steps:
+  - label: ":pipeline::rocket::seedling: Trigger control-plane e2e tests"
+    trigger: "ess-k8s-production-e2e-tests" # https://buildkite.com/elastic/ess-k8s-production-e2e-tests
+    build:
+      env:
+        REGION_ID: aws-us-east-1
+        NAME_PREFIX: ci_test_kibana-promotion_
+      message: "${BUILDKITE_MESSAGE} (triggered by pipeline.tests-production-canary.yaml)"
   - label: ":cookie: 1h bake period before continuing promotion"
     command: "sleep 3600"
-  - wait
-  - label: ":rocket: Run observability gates"
-    trigger: "serverless-quality-gates"
-    build:
-      branch: main
-      commit: HEAD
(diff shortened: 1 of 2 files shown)


.buildkite/pipeline.tests-production-canary.yaml
```diff
@@ -0,0 +1,23 @@
+# These pipeline steps constitute the quality gate for fleet server within the production canary environment.
+# A failure in this pipeline build will prevent further progression to the subsequent stage.
+
+steps:
+  - label: ":cookie: 1h bake period before continuing promotion"
+    command: "sleep 3600"
+  - wait
+  - label: ":rocket: Run observability gates"
+    trigger: "serverless-quality-gates"
```
juliaElastic, line 9: unfortunately these quality gates don't work in production, I tried to enable before: https://github.com/elastic/serverless-quality-gates/blob/main/observability/execute-query.sh#L27

I think we have to come up with another check as quality gate, maybe we can trigger the e2e tests like kibana: https://github.com/elastic/kibana/blob/main/.buildkite/pipelines/quality-gates/pipeline.tests-production-canary.yaml#L7

juliaElastic: Yes, we get an alert if the SLO breaks during the baking period in canary.
AFAIK we can manually stop the buildkite job that runs the wait, so the promotion doesn't continue. What I'm not sure about if production-noncanary gets automatically locked or we have to manually lock it, I can ask around to find out.

.buildkite/pipeline.tests-production-canary.yaml
```diff
@@ -0,0 +1,13 @@
+# These pipeline steps constitute the quality gate for fleet server within the production canary environment.
+# A failure in this pipeline build will prevent further progression to the subsequent stage.
+
+steps:
+  - label: ":pipeline::rocket::seedling: Trigger control-plane e2e tests"
+    trigger: "ess-k8s-production-e2e-tests" # https://buildkite.com/elastic/ess-k8s-production-e2e-tests
+    build:
+      env:
+        REGION_ID: aws-us-east-1
+        NAME_PREFIX: ci_test_kibana-promotion_
```
juliaElastic, line 10:
```suggestion
        NAME_PREFIX: ci_test_fleet-promotion_
```

PR #3039, opened 2023-10-13
Relax version checks for snapshot builds

### LINKED ISSUE #2960: Allow the snapshot version of the next minor release of Elastic Agent to connect.
Fleet Server currently performs a version check when an agent connects, and prevents any agent whose version is greater than the version of Fleet Server from connecting. This helps enforce the API compatibility guarantees we currently have in place around breaking changes.

https://github.com/elastic/fleet-server/blob/4c279b0509a5b777f9add26214d1a51613f63f72/internal/pkg/api/userAgent.go#L40

However this problem causes issues during development of the next minor version of agent, in particular when the feature freeze for the next minor release occurs the version of Elastic Agent must be increased to produce the first snapshot build of the next minor. When this happens, the version of agent has been increased but there is no stack deployment with an equal version until the first snapshot build is produced.

This creates a period where all of the agent automated integration test will fail until the first snapshot build is produced, which has at points taken days. To make this process easier and automatic, let's allow the snapshot version of the next release of agent to connect to Fleet server. Snapshot versions are not officially supported for production use, so this should have no consequences besides making the feature freeze and development period easier to manage.

michel-laterman:

Automated testing stalls for a few days when we release a new minor version.

Relax version checking when a SNAPSHOT binary of the fleet-server is built.
Version checking is relaxed to allow the minor version to be ahead by one.
If the fleet-server is running `X.Y.Z`, it allows clients of `X.Y+1.Z` to communicate.

`go test -v -tags=snapshot ./...`

Example output:
```
--- PASS: TestVersionConstraint (0.00s)
    --- PASS: TestVersionConstraint/8.0.0 (0.00s)
    --- PASS: TestVersionConstraint/8.1.0 (0.00s)
    --- PASS: TestVersionConstraint/8.2.0 (0.00s)
PASS
```

## Design Checklist

## Checklist

## Related issues

- Closes #2960

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -74,7 +74,7 @@ list-platforms: ## - Show the possible PLATFORMS
 .PHONY: local
 local: ## - Build local binary for local environment (bin/fleet-server)
 	@printf "${CMD_COLOR_ON} Build binaries using local go installation\n${CMD_COLOR_OFF}"
-	go build $(if $(DEV),-tags="dev",) -gcflags="${GCFLAGS}" -ldflags="${LDFLAGS}" -o ./bin/fleet-server .
+	go build $(if $(SNAPSHOT),-tags="snapshot",) -gcflags="${GCFLAGS}" -ldflags="${LDFLAGS}" -o ./bin/fleet-server .
 	@printf "${CMD_COLOR_ON} Binaries in ./bin/\n${CMD_COLOR_OFF}"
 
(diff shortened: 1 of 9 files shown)


juliaElastic (approved):
LGTM

cmacknz (approved):
Thanks for this

PR #3043, opened 2023-10-17
[Buildkite] Allow failures to triggering package-registry job

mrodm:

Package-Registry pipeline is not triggered if other steps are failing

Example of build where this step was not triggered: https://buildkite.com/elastic/fleet-server/builds/2186

Allow to be executed the trigger of package-registry pipeline without any dependencies on other steps.

As part of this PR, checking how it was configured Jenkins, it has been updated trigger steps to be asynchronous:
- https://github.com/elastic/fleet-server/blob/396d745bc1081190ff5912dcf0b05a266fe56f6a/.ci/Jenkinsfile#L215
-https://github.com/elastic/fleet-server/blob/396d745bc1081190ff5912dcf0b05a266fe56f6a/.ci/Jenkinsfile#L228

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -144,9 +144,6 @@ steps:
       meta_data:
         DOCKER_TAG: "${BUILDKITE_TAG}"
     if: "build.env('BUILDKITE_TAG') != ''"
-    depends_on:
-      - step: "release-test"
-        allow_failure: true
 
   - trigger: "fleet-server-package-mbp"
     label: ":esbuild: Downstream - Package"
@@ -156,5 +153,5 @@ steps:
     build:
       branch: "${BUILDKITE_BRANCH}"
     depends_on:
-      - step: "release-package-registry"
+      - step: "release-test"
         allow_failure: false
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -138,6 +138,7 @@ steps:
   - label: ":jenkins: Release - Package Registry Distribution"
     key: "release-package-registry"
     trigger: "package-registry-release-package-registry-distribution"
+    async: true
     build:
       branch: "main"
       meta_data:
@@ -145,11 +146,12 @@ steps:
     if: "build.env('BUILDKITE_TAG') != ''"
     depends_on:
       - step: "release-test"
-        allow_failure: false
+        allow_failure: true
 
   - trigger: "fleet-server-package-mbp"
     label: ":esbuild: Downstream - Package"
     key: "downstream-package"
+    async: true
     if: "build.env('BUILDKITE_PULL_REQUEST') == 'false' && build.env('BUILDKITE_TAG') == '' && build.env('BUILDKITE_BRANCH') != ''"
     build:
       branch: "${BUILDKITE_BRANCH}"


.buildkite/pipeline.yml
```diff
@@ -138,18 +138,20 @@ steps:
...
     if: "build.env('BUILDKITE_TAG') != ''"
     depends_on:
       - step: "release-test"
-        allow_failure: false
+        allow_failure: true
 
   - trigger: "fleet-server-package-mbp"
     label: ":esbuild: Downstream - Package"
     key: "downstream-package"
+    async: true
```
mrodm, line 154: In Jenkins, this trigger step was asynchronous
https://github.com/elastic/fleet-server/blob/396d745bc1081190ff5912dcf0b05a266fe56f6a/.ci/Jenkinsfile#L228

.buildkite/pipeline.yml
```diff
@@ -138,18 +138,20 @@ steps:
...
     build:
       branch: "main"
       meta_data:
         DOCKER_TAG: "${BUILDKITE_TAG}"
     if: "build.env('BUILDKITE_TAG') != ''"
     depends_on:
       - step: "release-test"
-        allow_failure: false
+        allow_failure: true
```
mrodm, line 149: Allow that `release-step` could fail to trigger this step.

jsoriano (approved):

.buildkite/pipeline.yml
```diff
@@ -138,21 +138,20 @@ steps:
...
     trigger: "package-registry-release-package-registry-distribution"
+    async: true
     build:
       branch: "main"
       meta_data:
         DOCKER_TAG: "${BUILDKITE_TAG}"
     if: "build.env('BUILDKITE_TAG') != ''"
-    depends_on:
-      - step: "release-test"
-        allow_failure: false
```
mrodm, line 148: Allow to start this trigger step just when the pipeline starts. No dependencies.

.buildkite/pipeline.yml
```diff
@@ -138,21 +138,20 @@ steps:
...
   - trigger: "fleet-server-package-mbp"
     label: ":esbuild: Downstream - Package"
     key: "downstream-package"
+    async: true
     if: "build.env('BUILDKITE_PULL_REQUEST') == 'false' && build.env('BUILDKITE_TAG') == '' && build.env('BUILDKITE_BRANCH') != ''"
     build:
       branch: "${BUILDKITE_BRANCH}"
     depends_on:
-      - step: "release-package-registry"
+      - step: "release-test"
```
mrodm, line 156: Change this dependency, to be the one set in `release-package-registry` to continue with the same chain/flow of steps.

jsoriano (approved):
:+1:

mrodm:
Added backport label for v8.11.0 , is that enough @kpollich ?

.buildkite/pipeline.yml
sharbuz: If I'm not wrong in Jenkins all steps were running one by one, why did you decide to change the dependencies?
Are you sure if this sequence doesn't affect the final result? From fleet-server it shouldn't, I'm not sure about the package-registry-distribution. 
@mrodm WDYT?

mrodm: It shouldn't change the final result.
Once the tag is pushed, it can be triggered the package-registry-distribution job to create the new docker images.
@jsoriano do you see any issue here ?

sharbuz (approved):

.buildkite/pipeline.yml
mrodm: At the end triggering this pipeline in package-registry to create the docker images should be just tied to the event of pushing the new tag.
So it should be fine removing the `depends_on` section here @sharbuz

PR #3044, opened 2023-10-17
Add Elastic-Api-Version and X-Request-Id headers in openapi spec

michel-laterman:

`Elastic-Api-Version` and `X-Request-Id` headers are not defined as part of responses for fleet-server

Add doc.go files for minimal information on generated client apis (what they are generated from).
Add missing headers to all req/resp in the openapi spec for the current implementation in `model/openapi.yml`.
Add an `openapi.yml` spec file for the `2023-06-01` based off commit 1d28f4ca4ec62f940771e5eb72b77699d256525f with the missing headers (and version defined in the spec). Correct the directory name for the versioned client.


jlind23:
@michel-laterman  Looks good to me but seems like the test failure might be legit.

kpollich (approved):
LGTM 🚀

PR #3048, opened 2023-10-19
Add top level keys for policy definition into specs

michel-laterman:

Policy is only defined as a generic `interface{}` or `json.RawMessage`.

Define agent policy structure as part of the openapi and json schema specs.
Use the generated json schema struct when fleet-server interacts with Elasticsearch.

Future work will use the struct generated from openapi as part of responses


model/schema.json
```diff
@@ -72,29 +72,36 @@
         },
         "data": {
           "description": "The opaque payload.",
-          "type": "object",
```
michel-laterman, line 75: Our custom [go json schema generator](https://github.com/elastic/go-json-schema-generate/) has some strange behaviour

If `type: object` is found it will define a go struct with that name then use it as the inline type for the attribute, If `format: raw` is found it will define the inline type as `json.RawMessage` (what we want for these cases).
If both are used, it leads to an unused (empty) struct definition.

Also names/attributes can collide, so if we have a top level object named `agent`, and an attribute named `agent` (defined with `type: object` and `format:raw`), things go _**wrong**_.

I removed `type: object` from attributes we want as a `json.RawMessage` in order to avoid this behaviour in the future

model/schema.json
```diff
@@ -647,99 +645,152 @@
...
-        "description": "The current overall status of the Elastic Agent",
-        "type": "string",
-        "enum": ["enrolling", "healthy", "error", "degraded", "offline", "unenrolling", "upgrading"]
-      },
-      "message": {
-        "description": "The current overall status message of the Elastic Agent",
-        "type": "string"
+    },
+
+    "checkin": {
```
michel-laterman, line 650: This was not scoped correctly (as defined in the draft-07 definition) so the go structs for checkin were not created. 

The openapi spec for checkin requests has [less strict definition](https://github.com/elastic/fleet-server/blob/main/model/openapi.yml#L340), @cmacknz should we update the openapi spec to check for the checkin as it would be defined here?

model/schema.json
```diff
@@ -647,99 +645,152 @@
...
+        "_id",
+        "agent",
+        "status",
+        "message",
+        "enrolled_at",
+        "updated_at"
+      ]
+    },
+
+    "policy-data": {
```
michel-laterman, line 744: So far I've only added the top-level keys for the policy here, should we get more detailed?

model/openapi.yml
```diff
@@ -490,7 +490,42 @@ components:
             yaml: "timeout"
         signed:
           $ref: "#/components/schemas/actionSignature"
-
+    policyData:
```
michel-laterman, line 493: No go code has been generated for this because it is not linked as part of an action payload at the moment.

We can likely use `oneOf` to generate fleet-server's structs in order to be backwards compatible if that's required

model/schema.json
```diff
@@ -647,99 +645,152 @@
...
-                "status",
-                "message"
-              ]
+            "id": {
+                "description": "The policy's ID",
+                "type": "string"
+            },
+            "outputs": {
+                "description": "A map of all outputs that the agent running the policy can use to send data to.",
+                "format": "raw"
```
michel-laterman, line 755: @juliaElastic, If we define additional keys as part of the policy definition it should remove some of the boilerplate around casting for the remote output definition.
But we wound need to decide on how much detail we want to get into (should fleet-server care or be aware of attributes it does not use? should we use the `additionalProperties` to hide some of these?)

michel-laterman: I did a little expirementing and was unable to get the json schema generator tool we use to output generic objects (type `map[string]interface`), but it's easily able to do lists of type `[]interface{}`, or just assign it the generic `interface{}` type.

That means if we want to have a stricter definition in fleet-server then a `json.RawMessage` or `interface{}` for outputs we would need to define the properties that exist in the policy in Elasticsearch (the schema we return from our openapi endpoint can differ by removing the secret_token for instance)

internal/pkg/policy/parsed_policy_test.go
```diff
@@ -11,107 +11,65 @@ import (
...
-			}
-		}
-
-		// Now validate output perms hash
-		if len(pp.Roles) != 1 {
-			t.Error("Only expected one role")
-		}
-
-		// Validate that default was found
-		if pp.Default.Name != "other" {
```
michel-laterman, line 65: I'm not sure how this test was passing as the minified policy had a single output named "default"

internal/pkg/policy/policy_output.go
```diff
@@ -74,13 +74,17 @@ func (p *Output) prepareElasticsearch(
...
 	agent *model.Agent,
-	outputMap smap.Map) error {
+	outputMap map[string]map[string]interface{}) error {
 	// The role is required to do api key management
 	if p.Role == nil {
 		zlog.Error().
 			Msg("policy does not contain required output permission section")
 		return ErrNoOutputPerms
 	}
+	if _, ok := outputMap[p.Name]; !ok {
```
michel-laterman, line 84: fail fast here instead of a possible panic on line 245

model/schema.json
michel-laterman: Outputs is tracked as a `map[string]map[string]interface{}` to make it easy to iterate in the fleet-server.
I didn't define attributes in outputs as it can get complex (ssl, proxies, keys, etc may be involved)

internal/pkg/policy/parsed_policy.go
```diff
@@ -166,34 +148,33 @@ func parsePerms(permsRaw json.RawMessage) (RoleMapT, error) {
...
-func findDefaultOutputName(outputsRaw json.RawMessage) (string, error) {
-	outputsMap, err := smap.Parse(outputsRaw)
-	if err != nil {
-		return "", err
-	}
-
+// findDefaultName returns the name of the 1st output with the "elasticsearch" type or falls back to behaviour that relies on deprecated fields.
+//
+// Previous fleet-server and elastic-agent released had a default output which was removed Sept 2021.
+func findDefaultOutputName(outputs map[string]map[string]interface{}) (string, error) {
```
michel-laterman, line 154: From what I can tell `ParsedPolicy.Default` is only set in the fleet-server, nothing fleet-server runs uses it (and it's not transmitted as part of a policy_change action. Do we still want to keep this?
@blakerouse I believe you wrote this, do you remember any additional context?

juliaElastic (approved):
LGTM

kpollich (approved):
Code LGTM - thanks for taking this on.

internal/pkg/policy/policy_output.go
```diff
@@ -238,10 +242,7 @@ func (p *Output) prepareElasticsearch(
 	// in place to reduce number of agent policy allocation when sending the updated
 	// agent policy to multiple agents.
 	// See: https://github.com/elastic/fleet-server/issues/1301
-	if err := setMapObj(outputMap, output.APIKey, p.Name, "api_key"); err != nil {
-		return err
-	}
-
+	outputMap[p.Name]["api_key"] = output.APIKey
```
nchaulet, line 245: @michel-laterman it seems to me it is going to update the parsedPolicy output here and potentially impact other checkin no?

PR #3051, opened 2023-10-24
Support for remote ES output

juliaElastic:
Changes in this pr:
- Fleet Server handles remote_elasticsearch outputs: creates a new child bulker for each, uses this bulker to read, create, update API keys for remote ES
- if remote config changed, stop and start a new child bulker
- API keys invalidated if remote ES output is removed from agent policy
- added unit and integration tests to cover the changes

- will be added in another pr: report error state if remote ES is not accessible

Remote ES output support in Fleet Server

Generate API keys for agents using remote ES output host and service_token created for fleet-server-remote

1. Start kibana locally with the changes in [this pr](https://github.com/elastic/kibana/pull/169252)
2. Create a cloud deployment to use as remote monitoring cluster, tested with [pr deployment](https://kibana-pr-169252.kb.us-west2.gcp.elastic-cloud.com:9243/app), but can be any 8.12-SNAPSHOT
3. Start fleet-server locally with the changes in this pr (used standalone mode)
4. Create a remote ES output with a valid service token generated in the remote cluster
```
// create service token in remote kibana console
POST kbn:/api/fleet/service_tokens
{
  "remote": true
}
```
5. Create an agent policy that uses this output as monitoring output
6. Enroll an agent locally/from VM to this policy
7. Verify that the agent API key is created on the remote cluster, and the monitoring data arrives from the enrolled agent

image

image

## Design Checklist

## Checklist

## Related issues

Closes https://github.com/elastic/ingest-dev/issues/1017


internal/pkg/policy/policy_output.go
```diff
@@ -229,6 +233,35 @@ func (p *Output) prepareElasticsearch(
 		output.PermissionsHash = p.Role.Sha2 // for the sake of consistency
 	}
 
+	if outputMap.GetMap(p.Name).GetString("type") == OutputTypeRemoteElasticsearch {
+		om, _ := outputMap.GetMap(p.Name).Marshal()
+		var outputObj map[string]any
+		json.Unmarshal(om, &outputObj)
+		hosts := outputObj["hosts"].([]interface{})
+		outputAPIKey, err :=
+			generateRemoteOutputAPIKey(ctx, agent.Id, p, []string{hosts[0].(string)})
```
juliaElastic, line 242: It was pretty cumbersome to get `hosts` out of the `outputMap`, I appreciate any suggestions to do it simpler.

internal/pkg/policy/policy_output.go
```diff
@@ -229,6 +233,35 @@ func (p *Output) prepareElasticsearch(
...
+		if err != nil {
+			return fmt.Errorf("failed generate output API key: %w", err)
+		}
+
+		// TODO update agent doc like above
+		// TODO what about update flow?
+
+		output.APIKey = outputAPIKey.Agent()
```
juliaElastic, line 250: @AndersonQ I saw your comment above, I think we have to update the agent doc with the remote api key. Is there any problem with doing that in the current data model? Or do we have to introduce new fields to store these?

https://github.com/elastic/fleet-server/blame/44b6985dbc49b09bc993bf4c98d593a74bd12bdd/internal/pkg/policy/policy_output.go#L191-L193

internal/pkg/policy/policy_output.go
```diff
@@ -411,6 +444,28 @@ func generateOutputAPIKey(
 	)
 }
 
+func generateRemoteOutputAPIKey(ctx context.Context,
```
juliaElastic, line 447: Any suggestion where to put this to reuse the es client? Maybe a new service? The existing api key generation is in bulker, but I can't use that as it uses the local ES client.

michel-laterman: We may want to introduce a bulker manager (or map of bulkers) that can have "named" clients.
Right now it will only be used to support remote outputs in policies, but it may be useful in the future (if we ever want to try to tackle multi tenancy).

I think this item may be better to leave as tech-debt at it would be a pretty big change if we wanted to replace all our bulkers

michel-laterman (commented):
i need to reacquaint  myself with output policy generation in order to better review the actual work being done

internal/pkg/policy/policy_output.go
```diff
@@ -35,9 +38,10 @@ var (
 )
 
 type Output struct {
-	Name string
-	Type string
-	Role *RoleT
+	Name         string
+	Type         string
+	ServiceToken string `json:"service_token,omitempty"`
```
AndersonQ, line 43: Any reason for having json tag on'y for this one?

internal/pkg/policy/policy_output.go
```diff
@@ -229,6 +233,35 @@ func (p *Output) prepareElasticsearch(
 		output.PermissionsHash = p.Role.Sha2 // for the sake of consistency
 	}
 
+	if outputMap.GetMap(p.Name).GetString("type") == OutputTypeRemoteElasticsearch {
+		om, _ := outputMap.GetMap(p.Name).Marshal()
+		var outputObj map[string]any
+		json.Unmarshal(om, &outputObj)
```
AndersonQ, line 239: [Blocker]
was it a test or debug code? It seems `om` isn't used.

internal/pkg/policy/policy_output.go
```diff
@@ -17,16 +17,19 @@ import (
...
 	"github.com/elastic/fleet-server/v7/internal/pkg/model"
 	"github.com/elastic/fleet-server/v7/internal/pkg/smap"
 )
 
 const (
-	OutputTypeElasticsearch = "elasticsearch"
-	OutputTypeLogstash      = "logstash"
-	OutputTypeKafka         = "kafka"
+	OutputTypeElasticsearch       = "elasticsearch"
+	OutputTypeRemoteElasticsearch = "remote-elasticsearch"
```
AndersonQ, line 30: I'm wondering if creating a new type is the best option. Ins't a ES output always an ES output. I mean, the `elasticsearch` and `remote-elasticsearch` have exactly the same data?

Without much thought I'd say the distinction should be made either on the output name, or extending the model for the ES output. I prefer the distinction being based on the output name.

internal/pkg/policy/policy_output.go
```diff
@@ -229,6 +233,35 @@ func (p *Output) prepareElasticsearch(
 		output.PermissionsHash = p.Role.Sha2 // for the sake of consistency
 	}
 
+	if outputMap.GetMap(p.Name).GetString("type") == OutputTypeRemoteElasticsearch {
```
AndersonQ, line 236: I might be missing something, but wouldn't [this switch block](https://github.com/elastic/fleet-server/pull/3051/files#diff-85caf77b270865c04855a90fc60468c90130cd83d2218aa66f362b32c178f4e0R57-R73) prevent `prepareElasticsearch` ever be called for `output.type == OutputTypeRemoteElasticsearch`?

internal/pkg/policy/policy_output.go
AndersonQ: I think you're in the right path when you defined the json tag. You might need to define different models for each output type.

Also there is another thing to consider here, ideally I believe it'd be good to try all the hosts if they fail to generate the API key

internal/pkg/policy/policy_output.go
AndersonQ: I kinda take it back. If `prepareElasticsearch` is being called for the fleet-ES and the remote-ES it should just work without any change.

juliaElastic: The difference in remote ES output is that the service_token comes from the output config, not configured in fleet server itself.

internal/pkg/policy/policy_output.go
juliaElastic: the Name and Type field has the same name in the json, so I though only need to map ServiceToken which is different. I can add json tag for the others too for consistency

internal/pkg/policy/policy_output.go
juliaElastic: `om` is being unmarshalled to `outputObj`
It might be possible to simplify this if I change the type of `outputMap` to contain `hosts` field as `[]string`

internal/pkg/policy/policy_output.go
juliaElastic: The config is different slightly as the remote-elasticsearch has a service_token, and the different type is used on the UI to render different components. I was thinking about it though if it would simplify to use the same type

internal/pkg/policy/policy_output.go
juliaElastic: it is a bit hacky now as I've set `p.Type` to `elasticsearch` so the switch block goes to `prepareElasticsearch`. However the type in `outputMap` is still remote-elasticsearch.

internal/pkg/policy/policy_output.go
juliaElastic: Yeah as far as I understood, there is one bulker now that is initialized with the local ES client, so I can't reuse it for remote ES. 
I could create another bulker instance for the remote ES client for now.

internal/pkg/policy/policy_output.go
AndersonQ: > The config is different slightly as the remote-elasticsearch has a service_token
> [...]
> so fleet-server doesn't have to introduce a new type, it can rely on the presense of the `service_token` field.

I believe relying on the presence of `service_token` isn't ideal. Besides if they're indeed different, then having 2 types makes sense. I thought at first they'd be the same thing, the new type was only to differentiate between the fleet-ES and the remote-ES

internal/pkg/policy/policy_output.go
AndersonQ: I see, ok, but there is still the error being ignored. It needs to be checked, otherwise `outputObj` might not be suitable for use

internal/pkg/policy/policy_output.go
AndersonQ: I see, I think the best is to get it all consistent, without this hack. Right now the way you're dealing with the remote-ES is different. So either it's be better to create another function to do all the handling of the remote-ES OR, my preferred solution, if they indeed share code, make the part that differs generic or configurable. For example the `generateOutputAPIKey` could be changed to take in a interface or ES client, so it can create the API key either on the fleet-ES or the remote-ES.

I see it's not the final version and you're still experimenting, but right now a lot of operation that do not serve the remote ES are performed before the remote-ES specific part which will indeed _prepare~Remote~Elasticsearch_.

internal/pkg/policy/policy_output.go
juliaElastic: Okay, it makes sense. I still have to change the type back to elasticsearch before sending it to the agent, as the agent doesn't know anything about "remote-elasticsearch" output type, and doesn't need to.

internal/pkg/policy/policy_output.go
juliaElastic: > For example the generateOutputAPIKey could be changed to take in a interface or ES client, so it can create the API key either on the fleet-ES or the remote-ES.

This is a good idea, hopefully then we can use the same `prepareElasticsearch` with a different es client.

I was looking at creating another bulker for remote ES but I think it's an overkill, as bulker does a lot of other things, and has to be started as a service.
I think it would be simpler for the existing bulker to store a list of remote ES clients (one per output) and use the ES client per output to generate api key.

juliaElastic: Made some changes in the latest commit, storing remote es clients in bulker to reuse between Prepare calls. It should probably be watching for hosts or service_token changes, to update the client.

Using remote-elasticsearch type, only changing it back to elasticsearch before sending to agent.

This way we can use the same prepareElasticsearch function, and only differ when an api key is created (and updated).
Tested that the functionality still works, and now the correct remote `api_key_id` is stored in the agent doc.

What's missing is using the remote es client when update api key is needed (it seems to be not that straighforward in bulker as create), and when api key is validated or invalidated.

internal/pkg/policy/policy_output.go
AndersonQ: that's true! how could I forget that...
in that case I'd strongly suggest sticking with only `elasticsearch` as type and differentiate them some other way

internal/pkg/bulk/engine.go
```diff
@@ -115,6 +120,22 @@ func (b *Bulker) Client() *elasticsearch.Client {
 	return client
 }
 
+func (b *Bulker) GetRemoteClient(name string) *elasticsearch.Client {
+	remoteEs := b.remoteEsList[name]
+	if remoteEs == nil {
+		return nil
+	}
+	client, ok := remoteEs.(*elasticsearch.Client)
```
AndersonQ, line 128: [Suggestion]
If you need an `*elasticsearch.Client`, use it as the type for the values on `remoteEs`. `SetRemoteClient` doesn't take in an `esapi.Transport`, `Get...` doesn't return one and as `*elasticsearch.Client` implements `esapi.Transport`, it'll be easier if you need to use any `remoteEs` as `esapi.Transport` later.

internal/pkg/bulk/engine.go
```diff
@@ -65,18 +65,22 @@ type Bulk interface {
...
 
 type Bulker struct {
-	es          esapi.Transport
-	ch          chan *bulkT
-	opts        bulkOptT
-	blkPool     sync.Pool
-	apikeyLimit *semaphore.Weighted
-	tracer      *apm.Tracer
+	es           esapi.Transport
+	remoteEsList map[string]esapi.Transport
```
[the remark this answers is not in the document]
juliaElastic: renamed, and change the type to map[string]*elasticsearch.Client

internal/pkg/policy/policy_output.go
```diff
@@ -245,6 +269,46 @@ func (p *Output) prepareElasticsearch(
 	return nil
 }
 
+func (p *Output) createRemoteEsClientIfNotExists(ctx context.Context, bulker bulk.Bulk, outputMap smap.Map) error {
+	remoteEs := bulker.GetRemoteClient(p.Name)
+	if remoteEs != nil {
+		// TODO replace if hosts or service token changed?
```
juliaElastic, line 275: I am wondering how can I decide if the `hosts` or `service_token` is different in the output config than the config used to create the esClient originally. I could store these values in the bulker too.

Does anyone know how does it work for the local ES, does the bulker pick up the hosts or other config changes and recreate the es client?

internal/pkg/policy/policy_output.go
juliaElastic: I'm looking at the code on how to create and run a new bulker, though it seems like a core logic to fleet server. 
Do I have to copy all of this code to create a new bulker? Probably `bulker.Run` and `runSubsystems` is needed too.
Maybe this code should be moved out to a separate function to decouple from the rest of the fleet server run.

https://github.com/elastic/fleet-server/blob/475e3ee0acdc0b9e33eee60cea1e5cabe91ced64/internal/pkg/server/fleet.go#L363-L417

internal/pkg/policy/policy_output.go
```diff
@@ -73,6 +89,7 @@ func (p *Output) prepareElasticsearch(
 	ctx context.Context,
 	zlog zerolog.Logger,
 	bulker bulk.Bulk,
+	outputBulker bulk.Bulk,
```
juliaElastic, line 92: the `outputBulker` is the same for local ES, and different for remote ES. It is only used to create/update Api keys in the right ES

internal/pkg/policy/policy_output.go
```diff
@@ -246,6 +271,71 @@ func (p *Output) prepareElasticsearch(
...
+		return bulker, nil
+	}
+	bulkCtx, bulkCancel := context.WithCancel(context.Background())
+	defer bulkCancel()
+	es, err := p.createRemoteEsClient(bulkCtx, outputMap)
+	if err != nil {
+		return nil, err
+	}
+	// starting a new bulker to create/update API keys for remote ES output
+	newBulker := bulk.NewBulker(es, nil, mainBulker.Opts())
```
juliaElastic, line 287: this code is mostly copying how a bulker is started in https://github.com/elastic/fleet-server/blob/475e3ee0acdc0b9e33eee60cea1e5cabe91ced64/internal/pkg/server/fleet.go#L363-L417, without the extra steps like `runSubsystems`.

internal/pkg/bulk/engine.go
```diff
@@ -115,6 +118,11 @@ func (b *Bulker) Client() *elasticsearch.Client {
 	return client
 }
 
+func (b *Bulker) Opts() BulkOpt {
+	return func(o *bulkOptT) {
```
juliaElastic, line 122: this is kind of a dummy function, I needed pass a BulkOpt to `NewBulker` [here](https://github.com/elastic/fleet-server/blob/e901a80e021aad672e716b14a59f1f9b39099bb0/internal/pkg/policy/policy_output.go#L287)

michel-laterman: Opts are optional, you should be able to call `NewBulker` with the transport and tracer

internal/pkg/policy/policy_output.go
```diff
@@ -17,27 +17,34 @@ import (
...
+	OutputTypeElasticsearch       = "elasticsearch"
+	OutputTypeRemoteElasticsearch = "remote_elasticsearch"
+	OutputTypeLogstash            = "logstash"
+	OutputTypeKafka               = "kafka"
 )
 
 var (
 	ErrNoOutputPerms    = errors.New("output permission sections not found")
 	ErrFailInjectAPIKey = errors.New("fail inject api key")
+	// TODO clean up bulkers when a remote output is removed
```
michel-laterman, line 39: We should not remove a bulker when the output is removed from the policy, that will lead to all the keys being orphaned. Our workflow now should be:
1. output is removed
2. fleet-server detects change and sends a new policy without (remote) output to agent
3. agent acks policy change
4. fleet-server revokes remote output api key.

It's not going to be safe to remove fleet-server's knowledge of a remote output before the api key is revoked and we'll likely need another way to track outputs in the future

internal/pkg/policy/policy_output.go
michel-laterman: Currently if fleet-server [receives a new config](https://github.com/elastic/fleet-server/blob/main/internal/pkg/server/fleet.go#L196) it will try to detect a [change it the subsystems it runs](https://github.com/elastic/fleet-server/blob/main/internal/pkg/server/fleet.go#L242). If a change is detected, the entire server is [stopped and restarted](https://github.com/elastic/fleet-server/blob/main/internal/pkg/server/fleet.go#L177-L190).

We may want to store the output config in each bulker (as a `map[string]interface{}`) and use `reflect.DeepEqual` to see if it has changed.
I don't think that storing the `hosts` and `service_token` is enough as the output can include additional information (proxy config, certs, etc)

internal/pkg/policy/policy_output.go
michel-laterman: As I commented before, we should be able to use `bulk.NewBulker(es, tracer)`.

Also we should set the tracer to the one that fleet-server uses.

internal/pkg/policy/policy_output.go
```diff
@@ -246,6 +271,71 @@ func (p *Output) prepareElasticsearch(
...
+
+	cfg := config.Config{
+		Output: config.Output{
+			Elasticsearch: config.Elasticsearch{
+				Hosts:        hostsStrings,
+				ServiceToken: p.ServiceToken,
+			},
+		},
+	}
+	es, err := es.NewClient(ctx, &cfg, false)
```
michel-laterman, line 332: We should get APM traces

internal/pkg/policy/policy_output.go
juliaElastic: updated

internal/pkg/policy/policy_output.go
juliaElastic: Do you mean like here? I'm trying to add it, not sure how to pass the bi (build info) from Fleet

https://github.com/elastic/fleet-server/blob/a2f232939f961887a511a5434d84d7c0778e59c5/internal/pkg/server/fleet.go#L458-L459

juliaElastic:
I can't seem to reproduce the integration test failure locally

```
PASS
ok      github.com/elastic/fleet-server/v7/internal/pkg/server  191.494s
```

In buildkite:
```
...
PASS: TestServerUnauthorized (1.16s)
    --- PASS: TestServerUnauthorized/no_auth_header (0.00s)
    --- PASS: TestServerUnauthorized/unauthorized (0.02s)
=== RUN   TestServerInstrumentation
    fleet_integration_test.go:529: Checkin request successful
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x157637a]
goroutine 585 [running]:
github.com/elastic/fleet-server/v7/internal/pkg/policy.NewParsedPolicy({0x1e87258, 0xc00022b360}, {0x1e904b0, 0xc000352640}, {{{0xc0001d0510, 0x14}, 0x0, 0x5a}, 0x1, 0x0, ...})
	/opt/buildkite-agent/builds/bk-agent-prod-gcp-1698847400491575936/elastic/fleet-server/internal/pkg/policy/parsed_policy.go:59 +0x5a
github.com/elastic/fleet-server/v7/internal/pkg/policy.(*monitorT).processPolicies(0xc00017cb60, {0x1e87258, 0xc00022b360}, {0xc0001c0fc0, 0x5, 0x48b489?})
	/opt/buildkite-agent/builds/bk-agent-prod-gcp-1698847400491575936/elastic/fleet-server/internal/pkg/policy/monitor.go:276 +0x5ae
github.com/elastic/fleet-server/v7/internal/pkg/policy.(*monitorT).processHits(0xc00017cb60, {0x1e87258, 0xc00022b360}, {0xc000611600, 0x5, 0x8})
	/opt/buildkite-agent/builds/bk-agent-prod-gcp-1698847400491575936/elastic/fleet-server/internal/pkg/policy/monitor.go:194 +0xdc
github.com/elastic/fleet-server/v7/internal/pkg/policy.(*monitorT).Run(0xc00017cb60, {0x1e87258, 0xc00022b360})
	/opt/buildkite-agent/builds/bk-agent-prod-gcp-1698847400491575936/elastic/fleet-server/internal/pkg/policy/monitor.go:159 +0x56b
github.com/elastic/fleet-server/v7/internal/pkg/server.(*Fleet).runSubsystems.loggedRunFunc.func10()
	/opt/buildkite-agent/builds/bk-agent-prod-gcp-1698847400491575936/elastic/fleet-server/internal/pkg/server/fleet.go:290 +0x99
golang.org/x/sync/errgroup.(*Group).Go.func1()
	/opt/buildkite-agent/go/pkg/mod/golang.org/x/sync@v0.3.0/errgroup/errgroup.go:75 +0x77
created by golang.org/x/sync/errgroup.(*Group).Go in goroutine 570
	/opt/buildkite-agent/go/pkg/mod/golang.org/x/sync@v0.3.0/errgroup/errgroup.go:72 +0x125
FAIL	github.com/elastic/fleet-server/v7/internal/pkg/server	9.375s
```

internal/pkg/policy/policy_output.go
michel-laterman: I think that bi is used to attach the fleet-server user agent header: https://github.com/elastic/fleet-server/blob/main/internal/pkg/es/client.go#L53

michel-laterman:
We need to check if the instrumentation integration test uses a full policy in elasticsearch, or if we should just adjust the `NewParsedPolicy` func to better handle nil `policy.data` or nil property cases

internal/pkg/policy/policy_output.go
juliaElastic: So do you think it makes sense to add all policy output configs to the [Config](https://github.com/elastic/fleet-server/blob/main/internal/pkg/config/config.go#L40) object and use it in [configChangedServer](https://github.com/elastic/fleet-server/blob/main/internal/pkg/server/fleet.go#L242) logic?

I'm not sure though what's the best way to do this. Currently the remote bulkers are in `policy_output.go` which is probably not the best place, I don't think I can call anything from `fleet.go` from here as it would be a circular dependency. Maybe a new shared channel between these files to send config changes, if that's possible?

Another solution could be to store the remove bulkers map in the main bulker, and create a channel in the main bulker to send config changes, to which we could listen to in `fleet.go`.

juliaElastic: Implemented the second approach in this [commit](https://github.com/elastic/fleet-server/pull/3051/commits/a52b32be3d15c8231bf9f92f917feaeb23344d6e)

internal/pkg/policy/policy_output.go
```diff
@@ -246,6 +272,81 @@ func (p *Output) prepareElasticsearch(
...
+		if !ok {
+			return nil, fmt.Errorf("failed to get hosts from output: %v", host)
+		}
+	}
+
+	cfg := config.Config{
+		Output: config.Output{
+			Elasticsearch: config.Elasticsearch{
+				Hosts:        hostsStrings,
+				ServiceToken: p.ServiceToken,
```
juliaElastic, line 329: I'm wondering where the Advanced YAML config of an output should be used. How does it work for local ES output?

This is the `config_yaml` field in output, and appears in agent policy under output too.
image

image

cmacknz: The Advanced YAML box for an output is meant to allow adding configuration for that output that aren't exposed in the UI, worker and bulk_max_size are the two most common. All of the keys in the Advanced YAML box should get merged with the rest of the output configuration and sent to the agent. 

I don't see a reason why this would be different for a remote ES output.

internal/pkg/policy/policy_output.go
```diff
@@ -246,6 +272,80 @@ func (p *Output) prepareElasticsearch(
...
+			},
+		},
+	}
+	es, err := es.NewClient(ctx, &cfg, false, elasticsearchOptions(
```
cmacknz, line 332: If this fails does Fleet server fail to start? If so this is another point of system failure.

Most users will not want their remote monitoring ES output being unavailable to cause all of their Fleet Servers and Agents to go offline. The system should degrade gracefully in this case and treat the remote ES output as a nice to have.

internal/pkg/server/fleet.go
```diff
@@ -414,6 +421,12 @@ func (f *Fleet) runServer(ctx context.Context, cfg *config.Config) (err error) {
 		return err
 	}
 
+	go func() {
+		outputChanged := <-bulker.RemoteOutputCh()
+		f.outputCh <- outputChanged
```
cmacknz, line 426: There is no for loop here so this can only ever receive one update before exiting, is that the intent?

I don't see this channel being closed anywhere and this goroutine doesn't select on a context so if there were a for loop this goroutine will never exit.

juliaElastic: If a remote output changed, fleet server will be restarted, do we need a for loop then?

internal/pkg/policy/policy_output.go
juliaElastic: We want to make fleet-server unhealthy if it can't connect to remote ES, but we don't want it to go offline. Perhaps the client should be created conditionally, or catch the error if it can't be created.

internal/pkg/policy/policy_output.go
juliaElastic: Okay, so I think it doesn't need any new logic, as the advanced config is already sent to the agent under outputs section in agent policy.

internal/pkg/policy/policy_output.go
juliaElastic: I am wondering if we should invalidate api key when the remote output is removed from a policy (and the agent acks), or only when the output is completely removed from Fleet?
Probably the latter, because the users can change the agent policy back to use the remote output as long as it exists.

juliaElastic: What I also don't understand, if I remove the remote ES output from an agent policy, the output is not removed from the agent output list in ES. So if I retire the remote API key and then change back the agent policy to the remote output, the agent will again try to use the old API key, which is invalidated.

internal/pkg/server/fleet.go
cmacknz: Restarting Fleet server takes all agents offline. I think it will almost always be the case that when multiple Fleet Severs are deployed for high availability that they will share a remote ES output.

This means that if the remote ES output goes offline, every Fleet server will restart will restart at once. I think it would be much better if changing the remote ES output did not require a restart. If we cannot do this, then rolling out the change gradually so as not to take every connected agent offline at once would be better.

cmacknz: That said, I wouldn't expect the remote ES output to change frequently, so perhaps this is something to address as a follow up.

In general we should avoid anything that causes Fleet server to go offline as much as possible, even if it is only offline briefly.

At minimum if we keep this behavior, it needs to be very clear to users that if they change the remote ES configuration their Fleet servers are all going to restart. This is not something they will assume by default.

internal/pkg/bulk/engine.go
```diff
@@ -65,18 +68,28 @@ type Bulk interface {
...
-	tracer      *apm.Tracer
+	es                    esapi.Transport
+	ch                    chan *bulkT
+	opts                  bulkOptT
+	blkPool               sync.Pool
+	apikeyLimit           *semaphore.Weighted
+	tracer                *apm.Tracer
+	remoteOutputConfigMap map[string]map[string]interface{}
+	remoteOutputCh        chan bool
+	bulkerMap             map[string]Bulk
```
juliaElastic, line 92: moved the remove bulker map to the bulker itself (out of policy_output) so it can be called from handleAck to invalidate api keys

internal/pkg/bulk/engine.go
```diff
@@ -65,18 +68,28 @@ type Bulk interface {
 	// Accessor used to talk to elastic search direcly bypassing bulk engine
 	Client() *elasticsearch.Client
 
+	CheckRemoteOutputChanged(name string, newCfg map[string]interface{})
```
juliaElastic, line 71: this function is currently only used within bulker and tests, could be made private

internal/pkg/server/fleet.go
joshdover: I don't think we can ship with this behavior. It's going to blow away all the caches and create extra load on the main Elasticsearch cluster when a remote ES config is changed which is quite unexpected. Especially if you consider the use case where a very small management cluster is being used for Fleet + CCS and all data is sent to remote data clusters. Impact on these management clusters may be very high when the remote config is changed since the ES cluster may be sized just large enough to handle Fleet.

Is there a technical limitation to why we can't stop the old bulk and start a new one?

juliaElastic: Not really, I was trying to replicate the same behaviour as for local ES. If this is a concern, I can try to replace the bulker only if the remote es config changes.

juliaElastic: I'm a little concerned though about merging this pr eventually as it would go immediately to prod. Even if the remote ES output is not used yet, there might be changes impacting the existing functionality, which might not be caught by the smoke tests. 
I think we should run the elastic-agent e2e tests too as quality gates.

internal/pkg/policy/standalone.go
```diff
@@ -112,6 +112,20 @@ func (m *standAloneSelfMonitorT) check(ctx context.Context) {
...
+	hasError := false
+	remoteESPayload := make(map[string]interface{})
+	for key, value := range remoteOutputErrorMap {
+		if value != "" {
+			hasError = true
+			remoteESPayload[key] = value
+		}
+	}
+	if hasError {
+		state = client.UnitStateDegraded
```
juliaElastic, line 125: added the same check in standalone monitor too, mostly for local testing

internal/pkg/policy/policy_output.go
cmacknz: Usually if you just do a GET to the ES hostname it should give a response with the version and some other things. I think this works for Serverless too although that may have changed since I last tried it.

juliaElastic: that's good for testing availability, though doesn't verify if the provided service_token is valid to create api keys for agents

juliaElastic: I added logic locally to ping every remote es output host every 5s in self monitor, but I'm not sure if it's the best thing to do, fleet-server would go to degraded even though the inaccessible remote ES is not used by any agents.
Shouldn't we only ping those hosts that are being used? E.g. we could ping on agent checkin.

juliaElastic:
I'm opening this for review, please have a look. I'm trying to come up with meaningful integration tests to make sure I'm not breaking anything.

internal/pkg/policy/standalone.go
```diff
@@ -112,7 +113,43 @@ func (m *standAloneSelfMonitorT) check(ctx context.Context) {
...
+		message = fmt.Sprintf("Could not connect to remote ES output: %+v", remoteESPayload)
+		m.log.Debug().Msg(message)
+	} else {
+		bulkerMap := m.bulker.GetBulkerMap()
+		for outputName, outputBulker := range bulkerMap {
+			res, err := outputBulker.Client().Ping(outputBulker.Client().Ping.WithContext(ctx))
```
juliaElastic, line 133: This logic is not yet added to `self.go` (the monitor when fleet-server is run by agent)

There is a limitation if Fleet Server is restarted, the pinging will stop, because the `bulkerMap` is stored in memory, and it will be wiped out. 
Not sure what's the best way to solve this, should we store internal state on disk between restarts?

juliaElastic:
Ran a smoke test with a custom image built from the pr, which passed: https://buildkite.com/elastic/observability-perf/builds/1947#018bce58-a041-4ad9-8072-e3350b3cb5c6

internal/pkg/server/remote_es_output_integration_test.go
```diff
@@ -0,0 +1,171 @@
...
+	var policyRemoteID = "policyRemoteID"
+	var policyDataRemoteES = model.PolicyData{
+		Outputs: map[string]map[string]interface{}{
+			"default": {
+				"type": "elasticsearch",
+			},
+			"remoteES": {
+				"type": "remote_elasticsearch",
+				// TODO start another fleet-server
+				"hosts": []string{"localhost:9200"},
```
juliaElastic, line 105: To have a real remote ES test, we should start another ES and refer that in hosts, and also generate a service token.
I tried to generate a remote service token with fleet-server privileges, but I got an error that it doesn't have `manage_service_account` privilege.

internal/pkg/testing/esutil/client.go
```diff
@@ -0,0 +1,57 @@
...
+type MockTransport struct {
+	Response    *http.Response
+	RoundTripFn func(req *http.Request) (*http.Response, error)
+}
+
+func (t *MockTransport) RoundTrip(req *http.Request) (*http.Response, error) {
+	return t.RoundTripFn(req)
+}
+
+func MockESClient(t *testing.T) (*elasticsearch.Client, *MockTransport) {
```
juliaElastic, line 32: copied this mock from [here](https://github.com/elastic/fleet-server/blob/a465d093a882d8b4a9ec1ae97d8494b0d2f9ae39/internal/pkg/api/handleUpload_test.go#L1112) to be accessible from another package

model/schema.json
```diff
@@ -545,10 +549,6 @@
           "description": "Last checkin message",
           "type": "string"
         },
-        "last_checkin_message": {
```
juliaElastic, line 548: removed duplicated `last_checkin_message`

internal/pkg/policy/standalone.go
juliaElastic: added the remote es ping to the agent self monitor too in this commit: https://github.com/elastic/fleet-server/pull/3051/commits/8797823a149f773f73dae4ce4727168005a94a98

internal/pkg/bulk/engine.go
```diff
@@ -98,13 +114,119 @@ func NewBulker(es esapi.Transport, tracer *apm.Tracer, opts ...BulkOpt) *Bulker
...
+	cfg := config.Config{
+		Output: config.Output{
+			Elasticsearch: config.Elasticsearch{
+				Hosts:        hostsStrings,
+				ServiceToken: serviceToken,
+			},
+		},
 	}
+	es, err := es.NewClient(ctx, &cfg, false, elasticsearchOptions(
+		true, build.Info{},
```
michel-laterman, line 216: I think the `build.Info` here would be used to set user-agent headers with the fleet-server version number, is there an alternate way for us to set this here?

internal/pkg/bulk/engine.go
```diff
@@ -115,6 +237,30 @@ func (b *Bulker) Client() *elasticsearch.Client {
 	return client
 }
 
+func (b *Bulker) Tracer() *apm.Tracer {
```
michel-laterman, line 240: Is this used anywhere?

internal/pkg/bulk/engine.go
```diff
@@ -115,6 +237,30 @@ func (b *Bulker) Client() *elasticsearch.Client {
...
+}
+
+// check if remote output cfg changed
+func (b *Bulker) CheckRemoteOutputChanged(zlog zerolog.Logger, name string, newCfg map[string]interface{}) bool {
+	curCfg := b.remoteOutputConfigMap[name]
+
+	hasChanged := false
+
+	// TODO remoteOutputConfigMap empty when FS restarts - won't detect changes
+	// when output config first added, not reporting change
```
michel-laterman, line 251: In what cases is this causing issues?
From what I can see, `CheckRemoteOutputChanged` is only called by `CreateAndGetBulker`; if fleet-server restarts/starts up and the config map is empty, the bulker will be nil and a new one is created.

But I think that this function should not be public, and should also be changed to show it may mutate the config map as well

internal/pkg/api/handleAck_test.go
```diff
@@ -571,6 +571,46 @@ func TestInvalidateAPIKeys(t *testing.T) {
...
+						assert.Subset(t, want, ids)
+				})).
+				Return(nil)
+		}
+
+		logger := testlog.SetLogger(t)
+		ack := &AckT{bulk: bulker}
+		ack.invalidateAPIKeys(context.Background(), logger, out.ToRetireAPIKeyIds, "")
+
+		bulker.AssertExpectations(t)
```
michel-laterman, line 610: we should also call `remoteBulker.AssertExpectations(t)`

internal/pkg/bulk/bulk_remote_output_test.go
```diff
@@ -0,0 +1,146 @@
...
+			changed: true,
+		}}
+
+	for _, tc := range testcases {
+		t.Run(tc.name, func(t *testing.T) {
+			log := testlog.SetLogger(t)
+			bulker := NewBulker(nil, nil)
+			bulker.remoteOutputConfigMap["remote1"] = tc.cfg
+			hasChanged := bulker.CheckRemoteOutputChanged(log, "remote1", tc.newCfg)
+			assert.Equal(t, tc.changed, hasChanged)
```
michel-laterman, line 84: We also should assert the config map as `CheckRemoteOutputChanged` may change it

internal/pkg/bulk/engine.go
```diff
@@ -98,13 +114,119 @@ func NewBulker(es esapi.Transport, tracer *apm.Tracer, opts ...BulkOpt) *Bulker
...
+
+func (b *Bulker) GetBulkerMap() map[string]Bulk {
+	return b.bulkerMap
+}
+
+func (b *Bulker) CancelFn() context.CancelFunc {
+	return b.cancelFn
+}
+
+func (b *Bulker) CreateAndGetBulker(zlog zerolog.Logger, outputName string, serviceToken string, outputMap map[string]map[string]interface{}) (Bulk, bool, error) {
```
michel-laterman, line 149: please add a comment here as well

internal/pkg/policy/policy_output.go
```diff
@@ -97,6 +111,64 @@ func (p *Output) prepareElasticsearch(
...
 
+	// retire api key of removed remote output
+	var toRetireAPIKeys *model.ToRetireAPIKeyIdsItems
+	var removedOutputKey string
+	// find the first output that is removed - supposing one output can be removed at a time
+	for agentOutputKey, agentOutput := range agent.Outputs {
+		found := false
+		for outputMapKey := range outputMap {
+			if agentOutputKey == outputMapKey {
+				found = true
```
michel-laterman, line 122: we can probably break out of the inner loop here

internal/pkg/policy/policy_output.go
```diff
@@ -97,6 +111,64 @@ func (p *Output) prepareElasticsearch(
...
+			}
+		}
+		if !found {
+			zlog.Info().Str("APIKeyID", agentOutput.APIKeyID).Str("output", agentOutputKey).Msg("Output removed, will retire API key")
```
michel-laterman, line 126: use the [ecs compliant](https://github.com/elastic/fleet-server/blob/main/internal/pkg/logger/ecs.go#L63) label?

internal/pkg/policy/policy_output.go
```diff
@@ -183,7 +258,24 @@ func (p *Output) prepareElasticsearch(
 
 		ctx := zlog.WithContext(ctx)
 		outputAPIKey, err :=
-			generateOutputAPIKey(ctx, bulker, agent.Id, p.Name, p.Role.Raw)
+			generateOutputAPIKey(ctx, outputBulker, agent.Id, p.Name, p.Role.Raw)
+			// reporting output error status to self monitor and not returning the error to keep fleet-server running
+		if outputAPIKey == nil && p.Type == OutputTypeRemoteElasticsearch {
+			if err != nil {
+				zerolog.Ctx(ctx).Warn().Msg("Could not create API key in remote ES")
```
michel-laterman, line 265:
```suggestion
				zerolog.Ctx(ctx).Warn().Err(err).Msg("Could not create API key in remote ES")
```

internal/pkg/policy/policy_output_integration_test.go
```diff
@@ -202,3 +203,108 @@ func createAgent(ctx context.Context, t *testing.T, index string, bulker bulk.Bu
...
+
+	// need to wait a bit before querying the agent again
+	// TODO: find a better way to query the updated agent
+	time.Sleep(time.Second)
+
+	got, err := dl.FindAgent(
+		ctx, bulker, dl.QueryAgentByID, dl.FieldID, agentID, dl.WithIndexName(index))
+	if err != nil {
+		require.NoError(t, err, "failed to find agent ID %q", agentID)
+	}
```
michel-laterman, line 245: Perhaps put this in a loop with some max retries before erroring the test out?

internal/pkg/policy/self.go
```diff
@@ -218,6 +218,45 @@ func (m *selfMonitorT) updateState(ctx context.Context) (client.UnitState, error
...
+				m.log.Debug().Msg(message)
+				m.reporter.UpdateState(m.state, message, nil) //nolint:errcheck // not clear what to do in failure cases
+				hasError = true
+				break
+			}
+		}
+		if hasError {
+			return m.state, nil
+		}
+	}
```
michel-laterman, line 258: We should describe this behaviour (degraded if a remote output fails) as part of the openapi spec

internal/pkg/policy/standalone_test.go
```diff
@@ -94,3 +102,96 @@ func TestStandAloneSelfMonitor(t *testing.T) {
...
+	// back to healthy
+	mocktrans.Response = &http.Response{
+		StatusCode: http.StatusOK,
+		Body:       ioutil.NopCloser(strings.NewReader(`{}`)),
+	}
+
+	sm.check(context.Background())
+	state = sm.State()
+
+	// TODO does not work, gives error "the client noticed that the server is not Elasticsearch and we do not support this unknown product"
```
michel-laterman, line 193: Is this still an error?

internal/pkg/server/remote_es_output_integration_test.go
michel-laterman: We likely need to add a new ES "cluster" as part of the https://github.com/elastic/fleet-server/blob/main/dev-tools/integration/docker-compose.yml file

internal/pkg/bulk/engine.go
juliaElastic: actually it's not anymore, I'll remove

internal/pkg/policy/policy_output.go
joshdover: ⚠️  Does `agentOutputKey` contain the credential? If so we cannot log this out. If not, can we rename this to be less scary looking on a log line?

internal/pkg/policy/policy_output.go
```diff
@@ -97,6 +111,64 @@ func (p *Output) prepareElasticsearch(
...
+				"lang":   "painless",
+				"source": fmt.Sprintf("ctx._source['outputs'].remove(\"%s\")", removedOutputKey),
+			},
+		})
+		if err != nil {
+			return fmt.Errorf("could not create request body to update agent: %w", err)
+		}
+
+		if err = bulker.Update(ctx, dl.FleetAgents, agent.Id, body, bulk.WithRefresh(), bulk.WithRetryOnConflict(3)); err != nil {
+			zlog.Error().Err(err).Msg("fail update agent record")
```
joshdover, line 167: Could these `err` strings that we log contain the API key itself?

internal/pkg/policy/policy_output.go
juliaElastic: No, it is only the output name. I'll rename.

internal/pkg/policy/policy_output.go
juliaElastic: good question, let me check

internal/pkg/bulk/engine.go
juliaElastic: I think the issue is that after a fleet-server restart, if the remote config changed (e.g. hosts), the code doesn't detect that it changed, and not going to recreate the api key. 
However when I tested this, I think eventually the changes were detected, so might not be an actual issue.

I'm renaming it and make it private.

internal/pkg/policy/standalone.go
juliaElastic: On thinking about this more it's not really an issue that `bulkerMap` is cleared on FS restart, because the Prepare output logic is called on every agent checkin [here](https://github.com/elastic/fleet-server/blob/main/internal/pkg/api/handleCheckin.go#L753), so the pinging is going to restart as soon as an agent checks in that uses the remote output.

One question, where is the `message` field reported in standalone monitor? I'm seeing the status in `/api/status` call, but not the message.

internal/pkg/bulk/engine.go
```diff
@@ -98,13 +114,119 @@ func NewBulker(es esapi.Transport, tracer *apm.Tracer, opts ...BulkOpt) *Bulker
...
+	if bulker != nil && !hasConfigChanged {
+		return bulker, false, nil
+	}
+	if bulker != nil && hasConfigChanged {
+		cancelFn := bulker.CancelFn()
+		if cancelFn != nil {
+			cancelFn()
+		}
+	}
+	bulkCtx, bulkCancel := context.WithCancel(context.Background())
```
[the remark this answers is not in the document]
juliaElastic: added the logic to do this, how can I test that the remote bulkers are cancelled?

internal/pkg/policy/policy_output.go
juliaElastic: done

internal/pkg/policy/policy_output.go
juliaElastic: tested (with an invalid script) and the API key is not logged:
```
10:47:23.146 ERR fail update agent record error.message="elastic fail 400: illegal_argument_exception: failed to execute script: script_exception: compile error" ecs.version=1.6.0 fleet.access.apikey.id=Pe2914sBdoeXGYc8n11i fleet.agent.id=a047dc88-9409-400a-b66b-3d83df8de26b fleet.ctx=processPolicy fleet.policy.id=9602dfd2-fc1d-423d-8607-e2b46b819560 fleet.policy.output.name=default fleet.policyCoordinator=1 fleet.policyRevision=40 http.request.id=01HFBVVVAHASQPJQ3XA92FG728 server.address= service.name=fleet-server service.type=fleet-server
```

internal/pkg/server/fleet.go
juliaElastic: It's not trivial to run elastic-agent e2e tests with a custom image, I might need agent team's help with it.

internal/pkg/api/handleAck.go
```diff
@@ -556,16 +568,26 @@ func cleanRoles(roles json.RawMessage) (json.RawMessage, int, error) {
...
 	ids := make([]string, 0, len(toRetireAPIKeyIDs))
+	output := ""
 	for _, k := range toRetireAPIKeyIDs {
 		if k.ID == skip || k.ID == "" {
 			continue
 		}
 		ids = append(ids, k.ID)
+		output = k.Output
+	}
+	// using remote es bulker to invalidate api key - supposing all retire api key ids belong to the same remote es
```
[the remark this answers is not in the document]
juliaElastic: done

internal/pkg/policy/policy_output.go
```diff
@@ -475,8 +475,9 @@ if (ctx._source['outputs']['%s']==null)
 			source.WriteString(fmt.Sprintf(`
 if (ctx._source['outputs']['%s'].%s==null)
   {ctx._source['outputs']['%s'].%s=new ArrayList();}
-ctx._source['outputs']['%s'].%s.add(params.%s);
-`, outputName, field, outputName, field, outputName, field, field))
+if (!ctx._source['outputs']['%s'].%s.contains(params.%s))
```
juliaElastic, line 478: only adding `toRetireAPIKey` if does not exist, because I noticed the same API key was added multiple times.
the comment in L473 says `It's an array that gets deleted when the keys are invalidated.` though I haven't found where the invalidated api keys are deleted from the retire list, does anyone know?

internal/pkg/policy/policy_output_integration_test.go
juliaElastic: updated

internal/pkg/policy/standalone_test.go
juliaElastic: Yes, still happening. I'm not sure how the Ping api decides that this is not a valid ES response, I tried to mock a similar return value that an actual ES ping returns, but didn't help.

internal/pkg/policy/standalone.go
juliaElastic: I'm also trying to test the agent based fleet-server self monitor, but I don't see fleet-server going to degraded state or the error logs from the self monitor, when I use an invalid remote es host.
I do see the error log that couldn't create API key, not sure what I'm doing wrong.
```
Could not create API key in remote ES
dial tcp 192.168.64.1:9202: connect: connection refused
```

juliaElastic: I noticed that the self monitor breaks the loop when it reaches healthy state: https://github.com/elastic/fleet-server/blob/772a5bf09e122b577696afe11256ef2f1ef8f2a3/internal/pkg/policy/self.go#L128-L130

If I change that to the loop running, is there any impact to consider? @blakerouse @cmacknz @michel-laterman

juliaElastic: I tested fleet-server with agent by removing the `break LOOP`, and I see fleet-server going to degraded state on the `/api/status` API. 
The fleet-server agent though doesn't change, it stays on healthy, and the component healthy doesn't change either. Is this implemented on Agent side to pick up degraded/unhealthy state from fleet-server?

internal/pkg/policy/self.go
juliaElastic: updated description here: https://github.com/elastic/fleet-server/pull/3051/commits/9c7d6fe3701e42d2b1188855b9f18b4cbe3beb94

internal/pkg/bulk/engine.go
juliaElastic: Added the same `build.Info` as used in Fleet, do you know how to verify it? https://github.com/elastic/fleet-server/pull/3051/commits/10021a89ccdffcdcb63bc8107b60815d7042cc7b

internal/pkg/policy/standalone.go
michel-laterman: I would avoid saving remote output information locally, i would rather see it as part of a fleet index if required.
Right now we would have orphaned API keys if the output is removed from a policy and fleet-server restarts before the checkin requests (where the API keys would be invalidated) are processed. I think this may occur if the policy that the agent managing fleet-server is running has a remote output that is removed.

I think not checking a remote cluster until an agent checks in is OK.

I think breaking the self-monitor after it reaches the healthy state is to help/simplify the elastic-agent/fleet-server bootstrap process.
iirc the agent needs fleet to be degraded so it will send config, then healthy to checkin. (@cmacknz @blakerouse can one of you confirm?)
If we report degraded when a remote ES cluster is down/misconfigured the agent managing the fleet server may never go online. Or it may try to restart the fleet-server then go offline?

I think we need to consider this behaviour, and what it means for the fleet-server to be healthy from the agent's pov as well as a platform pov. This may mean adding a new endpoint so that `/api/status` can be the ready endpoint and another will be for full output health.

michel-laterman (commented):
The only real concern I have is what the platform/elastic-agent will do in response to the fleet-server reporting a degraded state when a remote output goes offline.
We may want to separate the reporting/endpoint

joshdover:
> Good point. I think making Fleet Server degraded (potentially never healthy) on a remote output issue would have an impact on cloud/serverless where the platform is waiting for the `/api/status` to be healthy.
> I agree that introducing a new status endpoint for remote outputs would make reporting clearer, and we could use that from the UI to show an error banner.
> @kpollich @joshdover WDYT? We might want to do the health reporting separately, if it's going to be a new endpoint, this pr is already large enough.

I think the `/api/status` endpoint should be used to signal to orchestrators whether or not this Fleet Server can serve traffic to Agents. Since a Fleet may include agents that aren't using a remote output, we shouldn't block traffic flowing to a Fleet Server simply because one remote output is unreachable or misconfigured. This would lead to an entire Fleet going offline when one remote output is having a problem.

We do still need a way to signal this condition to the user in the UI though, and I think the existing way to do this in Fleet Server under Agent would be through the control protocol's [`UnitObserved`](https://github.com/elastic/elastic-agent-client/blob/21e4fd899bbcbe622248407e90744217da77b2b4/elastic-agent-client.proto#L293-L307) message. The issue is that this won't help us with Serverless where Fleet Servers don't show up as agents.

I think it makes sense to have a dedicated index/data stream where Fleet Server can post periodic health updates and other metrics, and we could read these from the Outputs UI. Ideally we'd do something that users could also alert on if necessary, so maybe something like `logs-fleet_server.output_health-default` data stream would make sense (or maybe `metrics-*`?)

juliaElastic:
Okay, to take this forward, I'll remove the logic of making the FS health status degraded for now, and raise a new issue to publish output health to a new datastream. We already have [metrics-fleet_server*](https://github.com/elastic/elasticsearch/pull/100574) privileges in ES, we could add `logs-fleet_server*` too if necessary.
It's possible that the status reporting won't make it in 8.12, do we consider it a must have for this feature? I think we probably should.

joshdover:
> It's possible that the status reporting won't make it in 8.12, do we consider it a must have for this feature? I think we probably should.

How complex would it be to report this via the control protocol? Since we aren't supporting this on Serverless for now, we may want to just do that option first so we don't have to block this feature or only release it as beta. It could still be a follow up to this PR though.

juliaElastic:
> How complex would it be to report this via the control protocol? Since we aren't supporting this on Serverless for now, we may want to just do that option first so we don't have to block this feature or only release it as beta. It could still be a follow up to this PR though.

I'm not familiar with the control protocol, how can Fleet Server send a UnitObserved message?
Is it the same as the component unit state in the agent doc?

internal/pkg/server/remote_es_output_integration_test.go
```diff
@@ -0,0 +1,170 @@
...
+	require.Equal(t, nil, serviceToken)
+	remoteAPIKey := remoteES["api_key"]
+	defaultOutput, ok := outputs["default"].(map[string]interface{})
+	require.True(t, ok, "expected default to be map")
+	defaultAPIKey := defaultOutput["api_key"]
+	require.False(t, remoteAPIKey == defaultAPIKey, "expected remote api key to be different than default")
+
+}
+
+func Test_Agent_Remote_ES_Output(t *testing.T) {
```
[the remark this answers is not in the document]
juliaElastic: updated the test to verify that the api key exists in remote es

internal/pkg/policy/policy_output.go
AndersonQ: I think it used to happen when it was the `DefaultAPIKeyHistory` list. Now I think we keep the history. I was trying to find when it'd get deleted and I also did not find. The comment was probably left there after the behaviour changed.

juliaElastic: this might be a bug if we don't ever clear the list of retired api keys

AndersonQ:
> I also noticed while testing that when the remote output is not accessible, the Agent doesn't go to unhealthy state (I'm testing with a docker container). Could this be a bug? @AndersonQ

The remote output is the output the agent is using? If so it might be an issue. Do you have the full logs of your test or is it easy to reproduce so I can try it myself?

AndersonQ:
> I'm not familiar with the control protocol, how can Fleet Server send a UnitObserved message?
> Is it the same as the component unit state in the agent doc?

It's almost the same, mainly for fleet as it's got only "one" unit that it's fleet-server itself. I'm not sure how things are right now, I can have a deeper look at the code, but fleet-server could report each output as a unit, so it'd have one input, itself, and a unit for each of the outputs. 
IIRC the agent supports that as everything is an unit, only the unit type changes (either input or output)

AndersonQ:
> I also noticed while testing that when the remote output is not accessible, the Agent doesn't go to unhealthy state (I'm testing with a docker container). Could this be a bug? @AndersonQ I'm seeing the error logs in agent as expected, but the agent status stays healthy.
> 
> ```
> 2023-11-21 11:07:00 {"log.level":"error","@timestamp":"2023-11-21T10:07:00.367Z","message":"Error dialing dial tcp 192.168.64.1:9202: connect: connection refused","component":{"binary":"metricbeat","dataset":"elastic_agent.metricbeat","id":"http/metrics-monitoring","type":"http/metrics"},"log":{"source":"http/metrics-monitoring"},"ecs.version":"1.6.0","address":"192.168.64.1:9202","log.logger":"esclientleg","log.origin":{"file.line":38,"file.name":"transport/logging.go"},"service.name":"metricbeat","network":"tcp","ecs.version":"1.6.0"}
> ```

How was your setup? I believe all our clients retry for a while before failing and the failure start reflecting on the beats status. So if it's quick enough the beat might not have yet transitioned to degraded, then the connection is established. what would let us with the error logs but a health beat.

internal/pkg/policy/policy_output.go
AndersonQ: I don't remember right now, but I think the idea was to keep the history. Besides with the possibility to update API keys permissions the number of retired API keys would be rather small

juliaElastic:
> How was your setup? I believe all our clients retry for a while before failing and the failure start reflecting on the beats status. So if it's quick enough the beat might not have yet transitioned to degraded, then the connection is established. what would let us with the error logs but a health beat.

I waited quite long (30m or so) and the agent was still healthy.
This is how I tested:
- create a remote elasticsearch output, can be a host that does not exist
- set remote elasticsearch output as monitoring output in an agent policy
- enroll an agent to this agent policy
- the agent enrolls as healthy, in agent logs seeing the errors that it can't connect to the remote es host.

internal/pkg/policy/policy_output.go
```diff
@@ -97,6 +111,65 @@ func (p *Output) prepareElasticsearch(
...
+				found = true
+				break
+			}
+		}
+		if !found {
+			zlog.Info().Str(logger.APIKeyID, agentOutput.APIKeyID).Str("outputName", agentOutputName).Msg("Output removed, will retire API key")
+			toRetireAPIKeys = &model.ToRetireAPIKeyIdsItems{
+				ID:        agentOutput.APIKeyID,
+				RetiredAt: time.Now().UTC().Format(time.RFC3339),
+				Output:    agentOutputName,
```
juliaElastic, line 131: As a follow up after https://github.com/elastic/fleet-server/pull/3051#issuecomment-1820608162, I was planning to store the policy id and revision here to query it later when invalidating api keys.
I realized this doesn't work in all cases, because it can happen that the agent is reassigned to another policy, so the policy might be different than the remote output's policy.
I also can't directly query the output from kibana SO as fleet-server doesn't have access and it might be already deleted.

juliaElastic: implemented in this commit: https://github.com/elastic/fleet-server/pull/3051/commits/727daa4e0fe1c4d62bb32286f8b7080efac2aeda

the solution is not perfect as it queries the last 100 policy revisions to find the output config, we should do a follow up to add mapping on `data.outputs` field to be able to filter in ES.

AndersonQ:
> > How was your setup? I believe all our clients retry for a while before failing and the failure start reflecting on the beats status. So if it's quick enough the beat might not have yet transitioned to degraded, then the connection is established. what would let us with the error logs but a health beat.
> 
> I waited quite long (30m or so) and the agent was still healthy. This is how I tested:
> 
> * create a remote elasticsearch output, can be a host that does not exist
> * set remote elasticsearch output as monitoring output in an agent policy
> * enroll an agent to this agent policy
> * the agent enrolls as healthy, in agent logs seeing the errors that it can't connect to the remote es host.
> 
> Here is the agent diagnostics from my local: [elastic-agent-diagnostics-2023-11-22T10-03-07Z-00.zip](https://github.com/elastic/fleet-server/files/13437735/elastic-agent-diagnostics-2023-11-22T10-03-07Z-00.zip)

thanks, I had a look at it and talked to them team and there seems to be a known issue the beats not properly reporting or updating their status when an output unit is failing :/

juliaElastic:
> thanks, I had a look at it and talked to them team and there seems to be a known issue the beats not properly reporting or updating their status when an output unit is failing :/

Is there an open issue that we can track?

internal/pkg/policy/policy_output.go
juliaElastic: @michel-laterman please review again

internal/pkg/bulk/engine.go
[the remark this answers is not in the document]
juliaElastic: I've added a test to verify that the old child bulker ctx is canceled or config change here, but it doesn't seem to happen, not sure what I'm missing, it logs bulk context done, but still tries to make the http request.
[commit](https://github.com/elastic/fleet-server/pull/3051/commits/9382975df36401c4b859516a363de746a52e344f)

```
=== RUN   TestCancelCtxChildBulkerReplaced
    engine.go:261: {"level":"info","name":"remote","message":"remote output configuration has changed"}
    engine.go:194: {"level":"debug","outputName":"remote","message":"Bulk context done"}
    engine.go:183: {"level":"debug","outputName":"remote","message":"Bulker started"}
    engine.go:183: {"level":"debug","outputName":"remote","message":"Bulker started"}
    bulk_test.go:350: apikey auth request : dial tcp: lookup remote-es: no such host
--- PASS: TestCancelCtxChildBulkerReplaced (0.10s)
PASS
ok      github.com/elastic/fleet-server/v7/internal/pkg/bulk    0.502s
```

internal/pkg/api/handleAck.go
```diff
@@ -556,19 +568,48 @@ func cleanRoles(roles json.RawMessage) (json.RawMessage, int, error) {
...
 	}
+	// using remote es bulker to invalidate api key
+	for outputName, outputIds := range remoteIds {
+		outputBulk := ack.bulk.GetBulker(outputName)
+
+		if outputBulk == nil {
+			// read output config from .fleet-policies, not filtering by policy id as agent could be reassigned
+			policy, err := dl.QueryOutputFromPolicy(ctx, ack.bulk, outputName)
+			if err != nil || policy == nil {
+				zlog.Debug().Str("outputName", outputName).Msg("Output policy not found")
```
michel-laterman, line 599: If we can't find the policy associated with an output and need to invalidate the API key, that means it's an orphaned key, right?
Should we emit a WARN log about the key for the agent being orphaned?

internal/pkg/api/handleAck.go
```diff
@@ -556,19 +568,48 @@ func cleanRoles(roles json.RawMessage) (json.RawMessage, int, error) {
...
+			} else {
+				outputBulk, _, err = ack.bulk.CreateAndGetBulker(ctx, zlog, outputName, policy.Data.Outputs)
+				if err != nil {
+					zlog.Debug().Str("outputName", outputName).Msg("Failed to recreate output bulker")
```
michel-laterman, line 603: same here, no output bulk with the output found  would also lead to an orphan

internal/pkg/api/handleAck_test.go
```diff
@@ -571,6 +571,74 @@ func TestInvalidateAPIKeys(t *testing.T) {
...
+	logger := testlog.SetLogger(t)
+	ack := &AckT{bulk: bulker}
+	ack.invalidateAPIKeys(context.Background(), logger, toRetire, "")
+
+	bulker.AssertExpectations(t)
+	remoteBulker.AssertExpectations(t)
+	remoteBulker2.AssertExpectations(t)
+}
+
+func TestInvalidateAPIKeysRemoteOutputReadFromPolicies(t *testing.T) {
```
michel-laterman, line 608: :+1:
great test cases

internal/pkg/bulk/bulk_test.go
```diff
@@ -272,6 +273,90 @@ func TestCancelCtx(t *testing.T) {
...
+	cancelF()
+
+	wg.Wait()
+}
+
+// verify that old bulker ctx is canceled if cfg changed and a new one is started
+func TestCancelCtxChildBulkerReplaced(t *testing.T) {
+	bulker := NewBulker(nil, nil)
+
+	ctx, _ := context.WithCancel(context.Background())
```
michel-laterman, line 317: we can defer the cancel to make the linter happy, or just disable it for this line

internal/pkg/bulk/bulk_test.go
```diff
@@ -272,6 +273,90 @@ func TestCancelCtx(t *testing.T) {
...
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	var wg sync.WaitGroup
+	wg.Add(1)
+	go func() {
+		defer wg.Done()
+
+		_, err := childBulker.APIKeyAuth(ctx, apikey.APIKey{})
```
michel-laterman, line 348: I think that `APIKeyAuth` calls are synchronous; they don't actually get bulked or use the `bulkCtx` that is created for the child context.

internal/pkg/bulk/engine.go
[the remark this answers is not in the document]
michel-laterman: Commented on the test case, i think it's because API key operations don't get processed through bulk calls/the engine so they are not associated with the context created here

internal/pkg/bulk/engine.go
```diff
@@ -98,13 +110,136 @@ func NewBulker(es esapi.Transport, tracer *apm.Tracer, opts ...BulkOpt) *Bulker
...
+		opts:                  bopts,
+		es:                    es,
+		ch:                    make(chan *bulkT, bopts.blockQueueSz),
+		blkPool:               sync.Pool{New: poolFunc},
+		apikeyLimit:           semaphore.NewWeighted(int64(bopts.apikeyMaxParallel)),
+		tracer:                tracer,
+		remoteOutputConfigMap: make(map[string]map[string]interface{}),
+		// remote ES bulkers
+		bulkerMap:         make(map[string]Bulk),
+		remoteOutputLimit: semaphore.NewWeighted(1),
```
michel-laterman, line 122: Is this semaphore only to control updates? Or would it be better to replace this with a `RWLock`?

internal/pkg/policy/policy_output.go
michel-laterman: This seems like a good solution to get this out, thanks for all the effort!

michel-laterman (commented):
Looks good, great test coverage!

I just think we need to warn if we are going to orphan keys, and we should decide on how much concurrency control we need for the bulker map

internal/pkg/bulk/engine.go
juliaElastic: It is only used when updating `bulkerMap`, as I've seen concurrency issues in the logs. 
I can replace it with a `sync.RWMutex`

joshdover:
> > How complex would it be to report this via the control protocol? Since we aren't supporting this on Serverless for now, we may want to just do that option first so we don't have to block this feature or only release it as beta. It could still be a follow up to this PR though.
> 
> I'm not familiar with the control protocol, how can Fleet Server send a UnitObserved message? Is it the same as the component unit state in the agent doc?

This is done here in Fleet Server's "agent" module for interacting with the control protocol: https://github.com/elastic/fleet-server/blob/0dbce8ce079a72459652cb51dfde99a14eaea216/internal/pkg/server/agent.go#L170

I believe we could modify the output unit's state to degraded with an informative message whenever one of the remote outputs are not accessible.

internal/pkg/api/handleAck.go
juliaElastic: changed to warning log

internal/pkg/api/handleAck.go
juliaElastic: done

joshdover:
Yeah we might have to hack it somehow and just use the single output Fleet Server is assigned to report on all remote outputs it connects to. Agree on a follow up 👍

internal/pkg/bulk/bulk_test.go
juliaElastic: @michel-laterman removed this test for now as I couldn't get it working

juliaElastic:
> Yeah we might have to hack it somehow and just use the single output Fleet Server is assigned to report on all remote outputs it connects to. Agree on a follow up 👍

@joshdover I'm coming back to this, and if we start reporting degraded state on at least one output unit, Fleet Server will go to overall unhealthy state and show degraded state on `/api/status` API. We want to avoid this as discussed before, so I'm not sure if there is any other option rather than reporting output state separately from fleet-server component state.

michel-laterman (approved):

PR #3060, opened 2023-10-26
Define action.data and ack event schemas

michel-laterman:

The action data attribute and eck-event struct are not well defined leaving a lot of ambiguity as to what is being communicated.

Define the possible action types along with the accepted values for action.Data. Define the possible permutations for each action ack that the agent is able to send.

## Checklist


model/openapi.yml
```diff
@@ -674,7 +756,11 @@ components:
         events:
           type: array
           items:
-            $ref: "#/components/schemas/event"
+            anyOf:
+              - $ref: "#/components/schemas/genericEvent"
+              - $ref: "#/components/schemas/upgradeEvent"
+              - $ref: "#/components/schemas/diagnosticsEvent"
+              - $ref: "#/components/schemas/inputEvent"
```
michel-laterman, line 763: Event acks do not have a specific type per action as most actions use `genericEvent` with an optional error for most actions.
I don't think we need to have a 1-1 action-type to ack-event mapping as the agent only produces different acks for upgrades (may include an error and payload, but the upgrade communication is being replaced by upgrade_details in the checkin), diagnostics (may include an error and data), and the input action (lots of additional fields); but i'm open to feedback here.

We also can't use a discriminator here as action/ack types are undefined in the object/parent (type info is read from the separate action record)

model/openapi.yml
```diff
@@ -542,129 +607,146 @@ components:
...
-            An embedded JSON object that has the data about the ack.
-
-            Used by REQUEST_DIAGNOSTICS actions.
-            Contains a `upload_id` attribute used to communicate the successfullly uploaded diagnostics ID.
+        timestamp:
+          description: The timestamp of the acknowledgement event. Has the format of "2006-01-02T15:04:05.99999-07:00"
           type: string
-          format: application/json
-          x-go-type: json.RawMessage
+          format: date-time
```
michel-laterman, line 673: Note that we are now forcing serialization for the `timestamp`, `started_at`, and `completed_at` properties. additionally the `timestamp` attribute will now be set in the results as well

internal/pkg/api/handleAck.go
```diff
@@ -215,20 +237,19 @@ func (ack *AckT) handleAckEvents(ctx context.Context, zlog zerolog.Logger, agent
...
 
 	for n, ev := range events {
+		event, _ := ev.AsGenericEvent()
 		span, ctx := apm.StartSpan(ctx, "ackEvent", "process")
 		span.Context.SetLabel("agent_id", agent.Agent.ID)
-		span.Context.SetLabel("action_id", ev.ActionId)
+		span.Context.SetLabel("action_id", event.ActionId)
 		log := zlog.With().
-			Str("actionType", string(ev.Type)).
-			Str("actionSubType", string(ev.Subtype)).
```
michel-laterman, line 223: I'm taking the opportunity to remove deprecated/unused attributes from our codebase (`stream_id` and `policy_id` have been fully removed)

michel-laterman (commented):
I think we should mark a new API version, this pr defines specific formats for actions sent by fleet-server as part of a checkin request  (these are backwards compatible with the released api version) and ack events received in the ack endpoint (should also be backwards compatible but is a bit more picky).
We also have a few unversioned changes in the openapi spec since the 2023-06-01 release:
- enrollement requests have new `enrollement_id` param
- enrollment request `shared_id` removed
- checkin requests have `upgrade_details` param
- pgp endpoint added

I think our options are:
1. Ensure old version functions as described by using the api version param (`enrollement_id` and `upgrade_details` will be ignored, pgp endpoint cannot be acccessed, use more generic ack event format)
2. the new version is the default behaviour, the api version "2023-06-01" is treated the same as "2023-11-01", api compatibility to be added in the future
3. a limited version of proposal 1, pick what we actually want to differ between version numbers

Out of all of these I like option 2 the most as it describes the current behaviour of our stack. action.data and ack events are not being _changed_ with this pr, the openapi spec is being updated to better reflect current behaviour; same with the use of `enrollment_id` and `upgrade_details`. It also gives us more time to figure out how to make fleet-server have better api-version defined behaviour.

model/openapi.yml
```diff
@@ -526,6 +533,64 @@ components:
...
+      type: object
+      required:
+        - target_id
+      properties:
+        target_id:
+          decription: The action to attempt to cancel.
+          type: string
+    actionInputAction:
+      description: The INPUT_ACTION action data.
+      type: object # FIXME: needs security team to define fields as the action is passed from the agent to their componenets.
```
michel-laterman, line 593: Right now the data attribute for an `INPUT_ACTION` is a generic `map[string]interface{}`. Do we want a more in depth definition?

cc: @aleksmaus @scunningham

juliaElastic (approved):
LGTM

Regarding the versioning, AFAIK new fields can be added without a version update, only breaking changes need a new API version. See this [doc](https://docs.google.com/document/d/1YpF6hXIHZaHvwNaQAxWFzexUF1nbqACTtH2IfDu0ldA/edit#bookmark=id.kgm0pthdpzrt)

kpollich:
> Regarding the versioning, AFAIK new fields can be added without a version update, only breaking changes need a new API version. See this [doc](https://docs.google.com/document/d/1YpF6hXIHZaHvwNaQAxWFzexUF1nbqACTtH2IfDu0ldA/edit#bookmark=id.kgm0pthdpzrt)

This is my understanding as well. Purely additive changes don't require a new version as they'll be non breaking to any API consumers.

PR #3067, opened 2023-10-30
Fix mutation issue when using parsed policies

michel-laterman:

Fix parsed policy mutation issue inadvertently added as part of #3048 that @nchaulet found.

Clone policy data before doing any operations that alter parsed policy references, serialize and return cloned data.


nchaulet (approved):
Thanks for addressing this LGTM 🚀

PR #3075, opened 2023-11-02
Add unit tests for supporting build metadata

joshdover:

This updates our user agent validation and parsing tests to ensure that we accept Agent builds that include the `\+build[0-9]$` metadata suffix. These tests also verify that we will retain the build metadata suffix in the version that we report to the `.fleet-agents` index.

## Design Checklist

## Checklist


jlind23:
[Clean up] Closing for now until it becomes a priority again.

PR #3088, opened 2023-11-07
Use system memory to determine limits

### LINKED ISSUE #2308: Automatically choose a default load limit present based on available memory

michel-laterman:

fleet-server should pick out defaults based on system memory if an agent count is not provided.

Adjust limit defaults and agent count settings based off of new scale tests.
Allow the fleet-server to get matching limits settings based off system memory if no agent max is provided.
Allow default settings to be used if a negative value is specified as the agents max setting.

## Design Checklist

- [ ] I have or intend to scale test my changes, ensuring it will work reliably with 100K+ agents connected. [40k](https://buildkite.com/elastic/observability-perf/builds/1993) [100k](https://buildkite.com/elastic/observability-perf/builds/1991)

## Checklist

## Related issues

- Closes #2308


internal/pkg/config/env_defaults.go
```diff
@@ -40,43 +40,44 @@ const (
 	defaultArtifactMaxBody  = 0
 
 	defaultEnrollInterval = time.Millisecond * 10
-	defaultEnrollBurst    = 100
-	defaultEnrollMax      = 50
```
michel-laterman, line 44: I added it as a comment to the `limit` struct below, having max < burst functionally will result in burst == max and the burst setting is for the `rate.Limiter` and the max setting is for our request controller

internal/pkg/config/env_defaults.go
```diff
@@ -40,43 +40,44 @@ const (
...
 	defaultStatusBurst    = 25
 	defaultStatusMax      = 50
 	defaultStatusMaxBody  = 0
 
-	defaultUploadStartInterval = time.Second * 3
-	defaultUploadStartBurst    = 8
-	defaultUploadStartMax      = 3
+	defaultUploadStartInterval = time.Second * 2
+	defaultUploadStartBurst    = 5
+	defaultUploadStartMax      = 10
```
michel-laterman, line 59: upped the limit and made it the same as UploadEnd

internal/pkg/config/env_defaults.go
```diff
@@ -40,43 +40,44 @@ const (
...
+	defaultFileDelivBurst    = 5
+	defaultFileDelivMax      = 10
+	defaultFileDelivMaxBody  = 0
 
-	defaultPGPRetrievalInterval = time.Millisecond * 10
-	defaultPGPRetrievalBurst    = 100
+	defaultPGPRetrievalInterval = time.Millisecond * 5
+	defaultPGPRetrievalBurst    = 25
 	defaultPGPRetrievalMax      = 50
+	defaultPGPRetrievalMaxBody  = 0
```
michel-laterman, line 80: now the same as artifacts

internal/pkg/config/defaults/gt10000_limits.yml
```diff
@@ -1,26 +0,0 @@
-num_agents:
-  min: 10000
-  max: 12499
```
michel-laterman, line 3: Turns out this setting was never changed and actually leads to a max agents val of 12500 using default values, incorrectly causing one of the previous unit tests to pass.
My current changes have caused the test to fail.

internal/pkg/config/cache.go
```diff
@@ -28,9 +28,7 @@ type Cache struct {
 	APIKeyJitter time.Duration `config:"jitter_api_key"`
 }
 
-func (c *Cache) InitDefaults() {
-	c.LoadLimits(loadLimits(0))
-}
+func (c *Cache) InitDefaults() {}
```
michel-laterman, line 31: tldr: loading the defaults here causes bugs

The `Cache` and `ServerLimits`'s `InitDefaults` methods have been purposefully changed to `nop`s as they may have been ignoring a bug that the TestConfig/input-config test was not catching because the agent limits in default/gt10000_limits.yml was incorrectly specified.

The test was incorrectly generating the default config output because the agent count was not included in the bounds of the limits.yml files. (note that the test to use agent count derived limits instead of memory limits was moved to the "config specifies agent count")

[`Config.Unpack`](https://pkg.go.dev/github.com/elastic/go-ucfg#Config.Unpack) calls `InitDefaults` on the struct (our top level config) and all children that have an `InitDefaults` function (this leads to extra initialization calls which is [another issue](https://github.com/elastic/fleet-server/issues/3090)).

If the `Cache` or `ServerLimits` loads defaults with the `InitDefaults` function then the `LoadLimits` calls with an agent count will not set the limits correctly as the config does not have zero values.

Without the change:
1. Input yml is read and `Unpack` is called. the config struct is loaded with data from the yml.
2. `Unpack` calls `InitDefaults`, which calls `c.LoadLimits(loadLimits(0))` resulting cache/limit values being set to a number that reflects the instance memory.
3. `cfg.LoadServerLimits` is called to use the `limits.max_agents` value that's part of the input
4. envLimits that correspond to the agent limit are found
5. `LoadLimits` compares the config (with memory limit values) and envLimits and does not replace anything.

Using a `nop` means that after `Unpack`, `Server.Limits` and `Server.Cache` are just copies of whatever is explicitly specified in the input, so when `LoadServerLimits` is called it can replace the 0 values correctly

nchaulet (approved):
Not tested locally but code LGTM! in cloud we use max agent so it should not affect it and in serverless it will use the config values and not the memory default correct?

michel-laterman:
Yes, setting an agent limit will override the memory inferred defaults. And passing any config values (as we do for serverless) overrides any memory inferred or max-agent defaults

juliaElastic (approved):
LGTM

PR #3092, opened 2023-11-10
Remove internal/pkg/testing from sonarqube coverage

michel-laterman:
Remove internal/pkg/testing from sonarqube coverage so it no longer appears in coverage reports


nchaulet (approved):
🚀

PR #3100, opened 2023-11-14
QX-337: QG pipeline runs should not skip intermediate builds

ramonbutter:
same as here:
https://github.com/elastic/hello-world-service/pull/58


juliaElastic (approved):
LGTM

PR #3124, opened 2023-11-27
Add additional transaction labels with error details to requests.

### LINKED ISSUE #3098: Enrich traces and request logs with Elasticsearch error details
We want to avoid being paged for Elasticsearch problems that the Fleet team cannot take any action on. In order to filter these alerts from our SLOs and/or route these pages to the Elasticsearch team, we need to enrich our observability data with more details about the Elasticsearch error that caused a request to fail. This will allow us to distinguish between real Fleet related problems that we can solve (eg. misconfigured ES URL or service token) from Elasticsearch or other infrastructure problems that we cannot take action on.

We should add a field that helps us do this to all traces and request logs.

TODO: fill in list of ES errors that we can safely ignore

michel-laterman:

It's difficult to determine if a fleet-server error is caused by a downstream elasticsearch error, or it indicates an actual issue (fleet-server misconfiguration etc.)

Add transaction labels with more error details when a request would send an APM error.

## Checklist

## Related issues

- Closes #3098


internal/pkg/api/error.go
```diff
@@ -518,13 +520,26 @@ func (er HTTPErrResp) Write(w http.ResponseWriter) error {
...
 	if resp.StatusCode >= 500 {
+		if trans := apm.TransactionFromContext(r.Context()); trans != nil {
+			esErr := &es.ErrElastic{}
+			if errors.As(err, &esErr) {
+				trans.Context.SetLabel("error.type", "ErrElastic")
+				trans.Context.SetLabel("error.details.status", esErr.Status)
+				trans.Context.SetLabel("error.details.type", esErr.Type)
+				trans.Context.SetLabel("error.details.reason", esErr.Reason)
+				trans.Context.SetLabel("error.details.cause.type", esErr.Cause.Type)
+				trans.Context.SetLabel("error.details.cause.reason", esErr.Cause.Reason)
```
michel-laterman, line 538: @joshdover is this closer to what you had in mind (in the issue description)?

joshdover: Yes this makes sense to me. One thing I can't tell from just reading this is if any of these fields contain anything instance-specific content that will create noise and make it harder to group related errors, such as specific disk usage stats, node names, etc. This would probably be tricky to normalize though at this point and I think we can do this at query time later if it's an issue.

joshdover (approved):

internal/pkg/api/error.go
michel-laterman: ok, we'll merge as is for now and we can revist/clean up the labels in the future

Issue #3116, opened 2023-11-21
Report remote output health


juliaElastic:
Related to https://github.com/elastic/kibana/issues/104986
Follow up after https://github.com/elastic/fleet-server/pull/3051#issuecomment-1820648919

In the original plans of remote ES output, we planned to make Fleet Server status degraded if a remote ES output is not accessible, however this has too large impact, so we decided to do output health reporting separately.

Implementation tasks:
- Report periodic health updates per remote output to `logs-fleet_server.output_health-default` data stream or `metrics-*`
  - Check if privileges are needed in ES for fleet-server service account to write to these data streams, and kibana_system to read it ([metrics privilege](https://github.com/elastic/elasticsearch/pull/100574) already exists for kibana_system)
  - Add ILM/DLM
- From Fleet UI, read this data stream and show error status on Output UI

juliaElastic:
I picked this up as I think this is require to roll out remote ES: https://github.com/elastic/fleet-server/pull/3051#issuecomment-1829894290

juliaElastic:
Raised a few prs, I have the base functionality working, with some notes/suggestions here:, as the design doc was not very detailed regarding the UX: https://github.com/elastic/kibana/pull/172181
I'll continue to add tests, feel free to comment.

juliaElastic:
@amolnater-qasource This enhancement is related to Remote ES output, adding a health reporting for remote outputs.
See screenshots and verification steps here: https://github.com/elastic/kibana/pull/172181

amolnater-qasource:
Hi @juliaElastic 

Thank you for the update.

We have revalidated this issue on latest 8.12.0 Snapshot kibana cloud environment and found it **working fine**:

**Observations:**
- Unhealthy badge is visible with invalid Remote Elasticsearch output.
- Appropriate error message is displayed while editing invalid Remote Elasticsearch output.
- Healthy badge is visible with valid Remote Elasticsearch output.
- Healthy message is also displayed while editing valid Remote Elasticsearch output.

**Screenshot:**
![image](https://github.com/elastic/fleet-server/assets/77374876/723d2f40-76f1-40eb-ab72-5448574adcae)

![image](https://github.com/elastic/fleet-server/assets/77374876/7b0098cb-d895-4b08-833d-c528578aa4ee)

![image](https://github.com/elastic/fleet-server/assets/77374876/14d13bb0-24e0-48ef-a53a-699d0ffc53fd)

[...]

PR #3127, opened 2023-11-29
report output health

juliaElastic:

Report state of remote es outputs

Report `HEALTHY/DEGRADED` state of remote es outputs to `logs-fleet_server.output_health-default`.
 - report state when api key is being generated
 - report state regularly in self monitor by pinging the remote es hosts

## Design Checklist

## Checklist

## Related issues

Resolves https://github.com/elastic/fleet-server/issues/3116


internal/pkg/policy/self.go
```diff
@@ -253,6 +255,32 @@ func (m *selfMonitorT) updateState(ctx context.Context) (client.UnitState, error
 	return state, nil
 }
 
+func reportOutputHealth(ctx context.Context, bulker bulk.Bulk, logger zerolog.Logger) {
+	//pinging logic
+	bulkerMap := bulker.GetBulkerMap()
```
juliaElastic, line 260: as mentioned on the previous pr, the regular health reporting will stop if fleet-server is restarted, and doesn't restart until an agent tries to create an API key again (e.g. due to change in output config), because the `bulkerMap` is stored in memory and output bulkers are created when there is a config change or a new output used for the first time by an agent.

internal/pkg/policy/self.go
```diff
@@ -218,6 +218,8 @@ func (m *selfMonitorT) updateState(ctx context.Context) (client.UnitState, error
 		return client.UnitStateStarting, nil
 	}
 
+	reportOutputHealth(ctx, m.bulker, m.log)
```
juliaElastic, line 221: Currently pinging remote outputs every 5s (default monitor interval) and writing out a doc to the output health data stream
We could change this to only write out a doc if the state changed.

internal/pkg/dl/output_health.go
```diff
@@ -0,0 +1,54 @@
...
+	Namespace string `json:"namespace,omitempty"`
+}
+
+type OutputHealth struct {
+	Output     string     `json:"output,omitempty"`
+	State      string     `json:"state,omitempty"`
+	Message    string     `json:"message,omitempty"`
+	Timestamp  string     `json:"@timestamp,omitempty"`
+	DataStream DataStream `json:"data_stream,omitempty"`
+}
```
michel-laterman, line 28: If this is something that is being written to ES, does it make more sense to define in in model/schema.json instead?

juliaElastic: moved to schema.json

PR #3135, opened 2023-11-30
Unauthorized keys return 401

### LINKED ISSUE #2861: Check enrollment status before authenticating agents
In cases where Agents are unenrolled from the backend without uninstalled, they may continue to check in with Fleet Server or fetch artifacts from the API. In this case, we incur the costs of attempting to hash the agent's API key on each request. Though we have caching in place, on a cache miss this is an expensive operation.

We can avoid this overhead by first checking the Agent's enrollment status and avoid trying to auth the key if it's already unenrolled in https://github.com/elastic/fleet-server/blob/main/internal/pkg/api/auth.go

michel-laterman:

Any issues with the API key result in a 400 status code

Issues with API keys will result in a 401, or a 403 if the key does not match the agent.

## Checklist

## Related issues

- Closes #2861


internal/pkg/api/error.go
```diff
@@ -167,7 +167,7 @@ func NewHTTPErrResp(err error) HTTPErrResp {
 		{
 			ErrAgentIdentity,
 			HTTPErrResp{
-				http.StatusBadRequest,
+				http.StatusForbidden,
```
michel-laterman, line 170: I'm pretty sure I changed all the relevant auth errors to return a 401 here.

However I think it makes sense to use a 403 instead of a 401 in this case where the API key does not match the requested agent. I've altered the openapi spec to include it as a return for the relevant endpoints

cc @joshdover

jlind23: Makes sense to me.
So IIUC if the key doesn't match it will be a `403` everything else will be `401`.

If the key is not found, shouldn't we also return a `403` instead of a `401`?

jsoriano: > If the key is not found, shouldn't we also return a `403` instead of a `401`?

This would be an unauthorized case, right? So 401 would be fine.

model/openapi.yml
```diff
@@ -996,6 +996,25 @@ components:
                 statusCode: 401
                 error: Unauthorized
                 message: ApiKey not enabled
+    forbidden:
+      description: 403 response when the API key does not match the agent
```
jsoriano, line 1000: For cases where the API key is invalid we return 401, right?
```suggestion
      description: 403 response when the API key is valid but does not match the agent
```

internal/pkg/api/error.go
```diff
@@ -185,7 +185,7 @@ func NewHTTPErrResp(err error) HTTPErrResp {
 		{
 			ErrAgentInactive,
 			HTTPErrResp{
-				http.StatusBadRequest,
+				http.StatusUnauthorized,
```
jsoriano, line 188: Why is this case unauthorized? Should it be forbidden if the key was valid?

jsoriano: :+1:

jsoriano (approved):
Change makes sense to me :+1: 

Do we need to change anything in how the Agent handles authorization errors to completely close https://github.com/elastic/fleet-server/issues/2861?

michel-laterman:
I don't think so, the [`isUnauth` check](https://github.com/elastic/elastic-agent/blob/main/internal/pkg/agent/application/gateway/fleet/fleet_gateway.go#L344) for the checkin would trigger based on the [response code the agent gets](https://github.com/elastic/elastic-agent/blob/main/internal/pkg/fleetapi/client/round_trippers.go#L56), so the agent would unenroll iteself after too many failures.

PR #3138, opened 2023-12-01
Remove logger and timings from tests

michel-laterman:
## What is the problem this PR solves?

Tests are flaky

## How does this PR solve the problem?

Remove test logger that is being used after test terminates

## How to test this PR locally

`make test-unit`

## Related issues

- Closes #3134

Issue #3134, opened 2023-11-30
TestCancelCtxChildBulker may be flaky


michel-laterman:
## Flaky Test

* **Test Name:** TestCancelCtxChildBulker.
* **Link:** [Link to file/line number in github](https://github.com/elastic/fleet-server/blob/main/internal/pkg/bulk/engine.go#L176).
* **Branch:** [buildkite build](https://buildkite.com/elastic/fleet-server/builds/2585#018c2148-cff9-4a20-8caf-3e01795801fb/946)
* **Notes:** It looks like the engine tries to log the `Bulker started` message after the test has exited, meaning that the output that the logger uses has been invalidated.

### Stack Trace

```
=== RUN   TestCancelCtxChildBulker
--
  | PASS: TestCancelCtxChildBulker (0.00s) | 1s
  | PASS
  | panic: Log in goroutine after TestCancelCtxChildBulker hengine.go:176: {"level":"debug","outputName":"remote","message":"Bulker started"}
  |  
  | goroutine 107 [running]:
  | testing.(*common).logDepth(0xc00040a000, {0xc0000c6600, 0x5f}, 0x3)
  | /opt/buildkite-agent/.gvm/versions/go1.21.4.linux.amd64/src/testing/testing.go:1022 +0x674
  | testing.(*common).log(...)
  | /opt/buildkite-agent/.gvm/versions/go1.21.4.linux.amd64/src/testing/testing.go:1004
  | testing.(*common).Logf(0xc00040a000, {0x147c7fd, 0xb}, {0xc000091100, 0x4, 0x4})
  | /opt/buildkite-agent/.gvm/versions/go1.21.4.linux.amd64/src/testing/testing.go:1055 +0x9a
  | github.com/rs/zerolog.TestWriter.Write({{0x7f239056b1b0?, 0xc00040a000?}, 0x7f239056a530?}, {0xc0000ba400, 0x43, 0x1f4})
  | /opt/buildkite-agent/go/pkg/mod/github.com/rs/zerolog@v1.31.0/writer.go:142 +0x448
  | github.com/rs/zerolog.LevelWriterAdapter.WriteLevel(...)
  | /opt/buildkite-agent/go/pkg/mod/github.com/rs/zerolog@v1.31.0/writer.go:27
  | github.com/rs/zerolog.(*Event).write(0xc0000cad90)
[...]

PR #3139, opened 2023-12-01
Use nop loggers to prevent test failures

michel-laterman:
Use nop loggers to fix some test failures.

Some tests that not fixed by #3138 are failing, see [the build](https://buildkite.com/elastic/fleet-server/builds/2598#018c26f6-64a8-48c5-aeea-402a774a78c7)

## Related issues

- Relates #3134


juliaElastic (approved):
LGTM, thanks for fixing

PR #3141, opened 2023-12-05
Add go benchmark as part of the buildkite pipeline

jlind23:

## Design Checklist

## Checklist

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -89,6 +89,12 @@ steps:
           - step: "e2e-test"
             allow_failure: false
 
+      - label: "Run go benchmark"
+        key: "go-benchmark"
+        command: ".buildkite/scripts/run_benchmark.sh
+        agents:
+          provider: "gcp"
+
       - label: ":gcloud: Cloud e2e Test"
         key: "cloud-e2e-test"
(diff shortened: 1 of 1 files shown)


.buildkite/pipeline.yml
```diff
@@ -30,6 +30,23 @@ steps:
...
+      - label: "Run go benchmark for PR branch"
+        key: "go-benchmark-pr"
+        command: ".buildkite/scripts/run_benchmark.sh pr"
+        agents:
+          provider: "gcp"
+
+      - label: "Run go benchmark for base branch"
+        key: "go-benchmark-base"
+        command: ".buildkite/scripts/run_benchmark.sh base"
+        branches: ${BUILDKITE_BRANCH}
```
alexsapran, line 46: Maybe 
```suggestion
        branches: ${BUILDKITE_PULL_REQUEST_BASE_BRANCH}
```

alexsapran (approved):

jlind23:
@nchaulet For now we decided not to fail the build as it seems that benchmarks have some flakiness.
Example, this PR didn't change anything apart from the buildkite pipeline and even though the benchmarks are failing.

nchaulet (approved):
LGTM 🚀 it will be great if it can report something in the PR when it failed, like a failed check that do not block the merge

PR #3157, opened 2023-12-08
ci(action): auto-bump daily and change slack notification channel

v1v:
## What does this PR do?

Change the slack channel 
Run daily

## Why is it important?

There will be only 24 hours daily between the golang-crossbuild/golang auto-bumps and this one.

## Design Checklist

## Checklist


.github/workflows/bump-golang.yml
```diff
@@ -22,3 +25,5 @@ jobs:
           vaultRoleId: ${{ secrets.VAULT_ROLE_ID }}
           vaultSecretId: ${{ secrets.VAULT_SECRET_ID }}
           pipeline: ./.ci/bump-golang.yml
+          notifySlackChannel: "#ingest-notifications"
+          messageIfFailure: ":traffic_cone: updatecli failed for `${{ github.repository }}@${{ github.ref_name }}`, `@update-me-with-the-slack-team-to-be-poked` please look what's going on <${{ env.JOB_URL }}|here>"
```
v1v, line 29: For instance, currently,  we are the ones receiving those notifications in Slack:

image

In this PR the ping will happen to `@update-me-with-the-slack-team-to-be-poked` - an unknown user/team in Slack. @jlind23, would you like to enable a specific ping in Slack? If so, what's the team name in Slack?

juliaElastic (approved):
LGTM

PR #3154, opened 2023-12-07
invalidate remote api keys on unenroll

juliaElastic:
## What is the problem this PR solves?

If an agent using remote es output is unenrolled, the remote API key was not invalidated.

## How does this PR solve the problem?

Enhanced the logic that invalidates api keys on unenroll to take into account the output, so that remote api keys are invalidated with the remote es client (bulker).

### Invalidate remote API key on agent unenroll
1. start es,kibana,fleet-server locally with these changes, start another cluster for monitoring with 8.12-snapshot version
2. add a remote es output in local cluster using the cluster2 es
3. create an agent policy with remote output as monitoring output
4. enroll an agent to the agent policy, wait until the API key is created on the remote cluster and data is showing up
5. unenroll the agent, wait for agent to ack the action
6. verify that remote API key is invalidated

these logs show up in Fleet Server:
```
15:51:26.349 INF ack event actionId=0a53d3df-d527-40bc-864d-ac37bf019e71 agentId=240f75a2-f073-4853-a53c-eb839a915579 ecs.version=1.6.0 fleet.access.apikey.id=DRL5RIwB93ftHy-WdwfN fleet.agent.id=240f75a2-f073-4853-a53c-eb839a915579 http.request.id=01HH2FM23D038N5FVQSKGDRR5G n=0 nEvents=1 server.address= service.name=fleet-server service.type=fleet-server timestamp=2023-12-07T15:51:26.061Z
15:51:26.918 INF handleUnenroll invalidate API keys ecs.version=1.6.0 fleet.access.apikey.id=DRL5RIwB93ftHy-WdwfN fleet.agent.id=240f75a2-f073-4853-a53c-eb839a915579 fleet.policy.apiKeyIDsToRetire=[{"id":"DRL5RIwB93ftHy-WdwfN"},{"id":"DxL5RIwB93ftHy-WgwdB"},{"id":"yZv5RIwB4likW8KqgJ7g","output":"634bdfad-4bc5-4f9a-b7af-357049e0ab14"}] http.request.id=01HH2FM23D038N5FVQSKGDRR5G nEvents=1 server.address= service.name=fleet-server service.type=fleet-server
[...]

PR #3167, opened 2023-12-13
chore: use pinned versions

v1v:

Use pinned version. Otherwise, if breaking changes in the upstream, it will affect the whole project.

It was reverted accidentally, see https://github.com/elastic/fleet-server/pull/3154#discussion_r1425496027

## Design Checklist

## Checklist


juliaElastic (approved):

juliaElastic:
it seems the build still doesn't work with the pinned version, looks like a missing image?
```
Error response from daemon: manifest for docker.elastic.co/cloud-release/elastic-agent-cloud:8.13.0-jqmw70gd-SNAPSHOT not found: manifest unknown: manifest unknown
```

v1v:
> it seems the build still doesn't work with the pinned version, looks like a missing image?

That's something the Release team might be able to answer :/

PR #3177, opened 2023-12-19
added download_rate to openapi

juliaElastic:

Added `download_rate` to openapi spec

## Design Checklist

## Checklist

## Related issues

Relates https://github.com/elastic/kibana/issues/171943


model/openapi.yml
```diff
@@ -283,6 +283,10 @@ components:
           description: The artifact download progress as a percentage.
           type: number
           format: double
+        download_rate:
```
juliaElastic, line 286: download_rate seems optional in agent: https://github.com/elastic/elastic-agent/blob/cc81e4b376068eff7e15330f2c9fa0969d28c8f5/internal/pkg/agent/application/upgrade/details/details.go#L38

nchaulet (approved):
LGTM!

juliaElastic:
integration test fails with
`Error response from daemon: manifest for docker.elastic.co/elasticsearch/elasticsearch:8.13.0-jqmw70gd-SNAPSHOT-amd64 not found: manifest unknown: manifest unknown`

nchaulet:
Interesting looks like that was merged today https://github.com/elastic/fleet-server/blob/4d155c31683f2afef333f087061fc09b6311805e/dev-tools/integration/.env#L3 and this will not work well with our script that create the docker image

juliaElastic:
@v1v the pinned version was merged, though the [build failed](https://buildkite.com/elastic/fleet-server/builds/2756#018c8143-1942-4eae-ab1b-9936911c03ae). Any idea how to fix this?
It seems there is a missing elasticsearch image:
`Error response from daemon: manifest for docker.elastic.co/elasticsearch/elasticsearch:8.13.0-jqmw70gd-SNAPSHOT-amd64 not found: manifest unknown: manifest unknown`

model/openapi.yml
cmacknz, line 286: We also added two more download fields `retry_error_msg` and `retry_until`, let's add those while we are in here:

They came from https://github.com/elastic/elastic-agent/pull/3845

cmacknz: Relates https://github.com/elastic/kibana/issues/173370

juliaElastic: Added here: https://github.com/elastic/fleet-server/pull/3177/commits/e0a0188a5f90e979d1f659e28bd64e28cd952002

model/openapi.yml
```diff
@@ -283,6 +283,17 @@ components:
...
+          description: The artifact download rate as bytes per second.
+          type: number
+          format: double 
+        retry_error_msg:
+          description: The error message that is a result of a retryable upgrade.
```
cmacknz, line 291:
```suggestion
          description: The error message that is a result of a retryable upgrade download failure.
```

model/openapi.yml
```diff
@@ -283,6 +283,17 @@ components:
...
+          type: string 
+        retry_until:
+          description: The RFC3339 timestamp the upgrade is retried until.
```
cmacknz, line 294:
```suggestion
          description: The RFC3339 timestamp of the deadline the upgrade download is retried until.
```

PR #3178, opened 2023-12-19
replace policy_throttle with policy_limit

### LINKED ISSUE #3008: change policy throttle to use rate limiter
We should change the policy throttle to use a [rate.Limiter](https://pkg.go.dev/golang.org/x/time/rate#Limiter) to make it able to handle burst traffic faster and make it consistent with our other limiters.
We can replace the implementation with a limiter with a burst 1 size in order to keep the current behaviour.

This was found during scale testing.

michel-laterman:

policy_throttle mechanism is not consistent with other rate limit controls in fleet-server

Deprecate the old policy_throttle setting and replace it with policy_limit so that all of our endpoint/operation limits use similar config and mechanisms for rate control. If no limits are specified the throttle value will be used as the limit interval (with burst 1).

## Design Checklist

- ~~I have or intend to scale test my changes, ensuring it will work reliably with 100K+ agents connected.~~ no changes to default config.

## Checklist

## Related issues

- Closes #3008


nchaulet:
@michel-laterman I am not sure a rate limit will work as we want here from what I understand the policy monitor and agent checkin work in different goroutine, and here it will only rate limit the `dispatchPending` function that send a message to the  policy channel, but from what I understand that operation should almost be atomic, so it event with a rate limit of one it will dispatch all the policies change almost at the same time no? 

Happy to discuss this IRL if you want to avoid some comment roundtrip.

michel-laterman:
@nchaulet, the `dispatchPending` method used to dispatch the policy to the 1st subscription in the queue. the throttle-timer that I removed would tick to send the next one.
I recently pushed another change to make `dispatchPending` process the entire subscription queue (in a loop), the limiter is now used as part of the loop.

michel-laterman:
buildkite run perf-tests

michel-laterman:
buildkite run perf-tests

michel-laterman:
buildkite run perf-tests

nchaulet (approved):
🚀  Should we explicitly configure that limit in fleet-controller for serverless?

michel-laterman:
yes, we should try to change our configuration for serverles

PR #3184, opened 2023-12-21
Add observability-perf job trigger and pipeline

michel-laterman:

It's difficult to run scale tests on a fleet-server pr, especially when it comes to running a serverless deployment.

Add a job to build a custom fleet-server image and use it in a 10k agent observability-perf test when the comment "buildkite run perf-tests" is made in a PR.


nchaulet (approved):
LGTM 🚀

michel-laterman:
buildkite run perf-test

.buildkite/pipeline.perf-tests.yml
```diff
@@ -0,0 +1,30 @@
...
+steps:
+  - label: ":docker: Publish docker image"
+    key: "create-image"
+    command: ".buildkite/scripts/build_push_docker_image.sh"
+    agents:
+      provider: "gcp"
+
+  - label: "perf test"
+    key: "obs-perf-test"
+    trigger: observability-perf
```
michel-laterman, line 19: my comment failed to trigger the job, is this the correct value I should be using, or should we use `fleet-smoke-tests` or `fleet-scale-tests` here?

michel-laterman:
buildkite run perf-tests

catalog-info.yaml
```diff
@@ -143,3 +143,47 @@ spec:
...
+      cancel_intermediate_builds_branch_filter: '!main !7.* !8.*'
+      skip_intermediate_builds: true
+      skip_intermediate_builds_branch_filter: '!main !7.* !8.*'
+      teams:
+        ingest-fp:
+          access_level: MANAGE_BUILD_AND_READ
+        cloud-tooling:
+          access_level: MANAGE_BUILD_AND_READ
+        everyone:
+          access_level: READ_ONLY
```
mrodm, line 189: Would other users out of `ingest-fp` trigger this pipeline (commenting in a PR)? 

Currently, it looks good to just allow `ingest-fp`. Maybe at some point, it could be changed this setting for `everyone` to be BUILD_AND_READ?

catalog-info.yaml
```diff
@@ -143,3 +143,47 @@ spec:
...
+  implementation:
+    apiVersion: buildkite.elastic.dev/v1
+    kind: Pipeline
+    metadata:
+      name: fleet-server-perf-tests
+      description: fleet-server observability-perf tests
+    spec:
+      repository: elastic/fleet-server
+      pipeline_file: .buildkite/pipeline.perf-tests.yaml
+      branch_configuration: "main"
```
mrodm, line 169: Should it run also for other branches (7.* or 8.*) ?
NOt sure if this setting is affected when triggering a build using a comment.

catalog-info.yaml
```diff
@@ -143,3 +143,47 @@ spec:
...
+      skip_intermediate_builds: false
```
mrodm, line 170: This line is duplicated at line 181
It can be removed.

```suggestion
```

.buildkite/pipeline.perf-tests.yml
mrodm: Changes in `.buildkite/pull-request.json` or `catalog-info.yaml` files are just applied when the PR is merged into main.

So, for the time being your comment would not trigger any build :( And your new pipeline would not exist neither in Buildkite.

.buildkite/pull-requests.json
```diff
@@ -23,7 +23,20 @@
...
+      },
+      {
+        "enabled": true,
+        "pipeline_slug": "fleet-server-perf-tests",
+        "allow_org_users": true,
+        "allowed_repo_permissions": ["admin", "write"],
+        "allowed_list": ["mergify[bot]"],
+        "set_commit_status": false,
+        "build_on_commit": false,
+        "build_on_comment": true,
```
mrodm, line 36: :+1:

.buildkite/pipeline.perf-tests.yml
```diff
@@ -0,0 +1,30 @@
...
+  DOCKER_REGISTRY: "docker.elastic.co"
+  DOCKER_IMAGE: "${DOCKER_REGISTRY}/observability-ci/fleet-server" # needs to rename for rollback
+  DOCKER_IMAGE_GIT_TAG: "${BUILDKITE_BRANCH}" # needs to rename for rollback
+  DOCKER_IMAGE_LATEST_TAG: "latest" # needs to rename for rollback
+  DOCKER_IMAGE_SHA_TAG: "git-${BUILDKITE_COMMIT:0:12}" # needs to rename for rollback, should be "git-${BUILDKITE_COMMIT:0:12}"
+
+steps:
+  - label: ":docker: Publish docker image"
+    key: "create-image"
+    command: ".buildkite/scripts/build_push_docker_image.sh"
```
mrodm, line 13: As it is using `observability-ci` namespace, it is needed to add the required credentials in `.buildkite/hooks/pre-command`.

Something like this but using the keys for your new pipeline slug (`fleet-server-perf-tests`) and step keys (`create-image`):
https://github.com/elastic/fleet-server/blob/bf17db29514f0d90e22587929dc1469f830dd278/.buildkite/hooks/pre-command#L43-L49

And the same for the `pre-exit` hook:
https://github.com/elastic/fleet-server/blob/bf17db29514f0d90e22587929dc1469f830dd278/.buildkite/hooks/pre-exit#L7-L11

catalog-info.yaml
michel-laterman: I don't think we typically run these types of PR scale tests manually on backports

catalog-info.yaml
michel-laterman: I'm not sure, i think we can decide on this later as well

mrodm (approved):

PR #3195, opened 2023-12-27
Deprecate fleet.agent.logging.level attribute

### LINKED ISSUE #3126: command line logging override is ignored
Seen on 8.11.1.

The elastic-agent starts fleet-server with `"-E","logging.level=debug"`

However the initial configuration fleet-server states is:
```
{"log.level":"info","@timestamp":"2023-11-27T09:40:32.997Z","message":"initial server configuration",
...
"new":{"Fleet":{"Agent":{"ID":"","Logging":{"Level":"info"}
...
"Logging":{"Files":null,"Level":"info","Pretty":false,"ToFiles":true,"ToStderr":true},
```
We expect the override to replace the logging level.
This is likely caused by the fleet-server's preference to use the `fleet.agent.logging.level` field if it is present (and I think we use a default):
https://github.com/elastic/fleet-server/blob/12a119ccfafb4b09ca479424a964954a4fbdd6f3/internal/pkg/logger/logger.go#L110-L114

It looks like the v2 compatibility update will set this on both:
https://github.com/elastic/fleet-server/blob/a244bbefbc6830b097ac4c57e717e98fe36af4e3/internal/pkg/server/agent.go#L386-L421

@cmacknz, this may be a relic from before the v2 architecture shift, I think we can deprecate the `fleet.agent.logging.level` attribute and rely on the top level `logging.level` attribute instead

michel-laterman:

command line flag passed by agent is being ignored by fleet-server.

Deprecate the fleet.agent.logging.level attribute and rely on the top level logging.* attributes instead. Add a unit test to make sure that command line overrides are respected.

## Checklist

## Related issues

- Closes #3126

### DIFF
--- a/changelog/fragments/1703700493-deprecate-fleet.agent.logging.level.yaml
+++ b/changelog/fragments/1703700493-deprecate-fleet.agent.logging.level.yaml
@@ -1,0 +1,35 @@
+# Kind can be one of:
+# - breaking-change: a change to previously-documented behavior
+# - deprecation: functionality that is being removed in a later release
+# - bug-fix: fixes a problem in a previous version
+# - enhancement: extends functionality but does not break or fix existing behavior
+# - feature: new functionality
+# - known-issue: problems that we are aware of in a given version
+# - security: impacts on the security of a product or a user’s deployment.
+# - upgrade: important information for someone upgrading from a prior version
+# - other: does not fit into any of the other categories
+kind: bug-fix
+
+# Change summary; a 80ish characters long description of the change.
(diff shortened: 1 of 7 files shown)


nchaulet:
@michel-laterman does fleet server will still support the log level be changed from the Fleet UI with an action with that change?

michel-laterman:
Yes, I just tested. changing the log level to debug in the fleet UI will result in elastic-agent and the fleet-server outputting debug logs

nchaulet (approved):
🚀

PR #3201, opened 2024-01-03
Add mtls attributes to reference and unit tests

michel-laterman:

unit tests for mTLS support.

Add the mTLS attributes that fleet-server can use to connect to Elasticsearch, and that clients can use to connect to fleet-server in the fleet-server.reference.yml doc. Add unit tests to ensure that mTLS works as expected in the internal es and api packages.


juliaElastic (approved):
LGTM

PR #3206, opened 2024-01-05
fix multiple secret ref replace

juliaElastic:

Fix issue with package policy secrets not being replaced properly when multiple secrets were in a variable.

Improved replace logic to replace all secret refs in a variable value.

- Build akamai package with enabling secrets from here: https://github.com/elastic/integrations/pull/8725/files [akamai-2.22.0.zip](https://github.com/elastic/fleet-server/files/13840673/akamai-2.22.0.zip)
- upload package to local kibana
```
curl -XPOST -H 'content-type: application/zip' -H 'kbn-xsrf: true' http://localhost:5601/api/fleet/epm/packages -u elastic:changeme --data-binary @akamai-2.22.0.zip
```
- add Akamai integration to a new agent policy and populate secrets with dummy values
- enroll an agent to this agent policy
- download agent diagnostics, check the `computed-config.yml` file
- verify that the secret refs are replaced  in `header.XSignatureBase`

See the fixed value (before the fix: right, after the fix: left):
image

## Design Checklist

## Checklist

## Related issues

Relates https://github.com/elastic/integrations/issues/8610

### DIFF
--- a/internal/pkg/policy/secret.go
+++ b/internal/pkg/policy/secret.go
@@ -15,7 +15,7 @@ import (
 )
 
 var (
-	secretRegex = regexp.MustCompile(`\$co\.elastic\.secret{(.*)}`)
+	secretRegex = regexp.MustCompile(`\$co\.elastic\.secret{([^}]*)}`)
 )
 
 // read secret values that belong to the agent policy's secret references, returns secrets as id:value map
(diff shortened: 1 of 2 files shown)


kpollich (approved):
Code LGTM 🚀

internal/pkg/policy/secret.go
```diff
@@ -177,13 +177,17 @@ func ProcessOutputSecret(ctx context.Context, output smap.Map, bulker bulk.Bulk)
 }
 
 // replaceStringRef replaces values matching a secret ref regex, e.g. $co.elastic.secret{<secret ref>} -> <secret value>
+// and does this for multiple matches
 func replaceStringRef(ref string, secretValues map[string]string) string {
 	matches := secretRegex.FindStringSubmatch(ref)
-	if len(matches) > 1 {
+	for len(matches) > 1 {
```
kpollich, line 183: Would `ReplaceAll` be any simpler here? 

https://pkg.go.dev/strings#ReplaceAll

juliaElastic: I thought about that, but then we would have to rewrite the `FindStringSubmatch` logic too, to find all secret refs in one regex (currently it only finds one).

PR #3211, opened 2024-01-08
[Fleet] Test JWT

nchaulet:
POC of auth with JWT

### DIFF
--- a/internal/pkg/api/auth.go
+++ b/internal/pkg/api/auth.go
@@ -39,7 +39,7 @@ func authAPIKey(r *http.Request, bulker bulk.Bulk, c cache.Cache) (*apikey.APIKe
 	start := time.Now()
 
 	//  Hacky way to have test passing
-	if !strings.Contains(r.URL.Path, "/enroll") && testing.Testing() == false {
+	if !strings.Contains(r.URL.Path, "/enroll") && !testing.Testing() {
 		jwtToken, err := apikey.ExtractJWTAPIKey(r)
 		if err != nil {
 			return nil, err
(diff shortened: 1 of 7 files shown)


nchaulet:
buildkite run perf-tests

internal/pkg/apikey/apikey.go
```diff
@@ -124,6 +124,10 @@ func NewAPIKeyFromToken(token string) (*APIKey, error) {
 
 // Token returns the b64 encoded token of the APIKey.
 func (k APIKey) Token() string {
+	if strings.Contains(k.ID, "jwt:") {
```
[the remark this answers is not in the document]
nchaulet: We will have to change the `authAPIKey` to support both JWT and ES api key to make this compatible

internal/pkg/api/auth.go
```diff
@@ -35,6 +38,39 @@ func authAPIKey(r *http.Request, bulker bulk.Bulk, c cache.Cache) (*apikey.APIKe
...
 
+	//  Hacky way to have test passing
+	if !strings.Contains(r.URL.Path, "/enroll") && !testing.Testing() {
+		jwtToken, err := apikey.ExtractJWTAPIKey(r)
+		if err != nil {
+			return nil, err
+		}
+		token, err := jwt.Parse(jwtToken, func(token *jwt.Token) (interface{}, error) {
+			// Don't forget to validate the alg is what you expect:
+			if _, ok := token.Method.(*jwt.SigningMethodHMAC); !ok {
```
adamkasztenny, line 49: Is the token signature being validated with the private key that was used to sign the token?

adamkasztenny: Also, in addition to the algorithm, you'd want to check all of the [registered claim names](https://datatracker.ietf.org/doc/html/rfc7519#section-4.1) (except `iat` and `jti`).

PR #3218, opened 2024-01-11
[Fleet] Ignore not yet created .fleet-policies index

nchaulet:
## Description 

Ignore not yet created .fleet-policies index to avoid returning 500 when that index do not exists yet.

## How to tests

run a fleet server with `.fleet-policies` not yet created and static tokens 
```
inputs:
  - type: fleet-server
    policy.id: "${FLEET_SERVER_POLICY_ID:fleet-server-policy}"
    server:
      static_policy_tokens:
        enabled: true
        policy_tokens:
          - policy_id: 901b70f0-fefa-11ed-aa5e-5974b0535e80
            token_key: abcdefg
```

Try to enroll an agent  it should return a 400 instead of a 500 previously

```
curl --request POST \
  --url http://localhost:8220/api/fleet/agents/enroll \
  --header 'Authorization: ApiKey MDEyMzQ6YWJjZGVmZw==' \
  --header 'Content-Type: application/json' \
  --header 'User-Agent: elastic agent 8.9.0' \
  --header 'kbn-xsrf: xx' \
  --data '{
  "type": "PERMANENT",
  "metadata": {
    "local": {
      "host": {
        "hostname": "TEST"
      },
      "elastic": {
        "agent": {
          "version": "8.6.0",
          "upgradeable": true
        }
      }
    }
  }
}'
```

### DIFF
--- a/internal/pkg/bulk/bulk_integration_test.go
+++ b/internal/pkg/bulk/bulk_integration_test.go
@@ -282,6 +282,29 @@ func TestBulkSearch(t *testing.T) {
 	}
 }
 
+func TestBulkSearchWithIgnoreUnavailable(t *testing.T) {
+	ctx, cn := context.WithCancel(context.Background())
+	defer cn()
+	ctx = testlog.SetLogger(t).WithContext(ctx)
+
+	_, bulker := SetupIndexWithBulk(ctx, t, testPolicy)
+
+	// Search
+	dsl := fmt.Sprintf(`{"query": { "term": {"kwval": "%s"}}}`, "random")
+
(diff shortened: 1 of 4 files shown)


michel-laterman (approved):
LGTM, do we want to return a 503 as we expect that the index will be created in the future?

nchaulet:
@michel-laterman I do not think we want to return a server error (it is similar on how we handle those .fleet indexes in kibana) the `ignore_unavailable` will make the search request to not throw if the index is not present that is expected in fleet server standalone until the first policy is created.

PR #3232, opened 2024-01-22
Add e2e test for timeout and 429 on startup

### LINKED ISSUE #3205: Fleet server standalone crash on startup if es return 429
## Description 

It seems Fleet server standalone crash during startup if elasticsearch return a 429 (could happen in serverless if the proxy refuse the connetion)

It's probably something we should handle, and recover from.

Screenshot 2024-01-04 at 4 42 32 PM

michel-laterman:

We have observed the fleet-server crashing if a 429 is returned when fleet-server initializes.

Add e2e tests to for this behaviour to ensure no regressions occur.

## Design Checklist

## Checklist

## Related issues

- Closes #3205

### DIFF
--- a/testing/e2e/stand_alone_test.go
+++ b/testing/e2e/stand_alone_test.go
@@ -10,6 +10,7 @@ import (
 	"context"
 	"fmt"
 	"html/template"
+	"net/http/httptest"
 	"os"
 	"os/exec"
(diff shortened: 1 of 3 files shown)


juliaElastic (approved):
LGTM

PR #3233, opened 2024-01-22
Stricter validation for required checkin API attributes

### LINKED ISSUE #2420: Improve request validation on checkin API
The fields that are supposed to be required in the checkin API are not actually validated by the API layer. This can break status calculations in the UI when these fields are not present in the agent documents. This was found while investigating an issue with the horde client implementation of this API.

https://github.com/elastic/fleet-server/blob/e96b46cd622c645b9be1786660eadfb556f26390/model/openapi.yml#L220

michel-laterman:

Checkin endpoint does not check for required attributes

Add validation to make sure that status and message are present in the checkin request body.

## Design Checklist

## Checklist

## Related issues

- Closes #2420

### DIFF
--- a/internal/pkg/api/handleCheckin.go
+++ b/internal/pkg/api/handleCheckin.go
@@ -181,6 +181,13 @@ func (ct *CheckinT) validateRequest(zlog zerolog.Logger, w http.ResponseWriter, 
 	}
 	cntCheckin.bodyIn.Add(readCounter.Count())
 
+	if req.Status == CheckinRequestStatus("") {
+		return val, fmt.Errorf("checkin status missing")
+	}
+	if len(req.Message) == 0 {
+		return val, fmt.Errorf("checkin message missing")
+	}
+
 	var pDur time.Duration
 	var err error
 	if req.PollTimeout != nil {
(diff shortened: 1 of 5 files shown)


nchaulet (approved):
LGTM 🚀 agents were always sending those attributes correct?

internal/pkg/api/handleCheckin.go
```diff
@@ -181,6 +181,13 @@ func (ct *CheckinT) validateRequest(zlog zerolog.Logger, w http.ResponseWriter,
 	}
 	cntCheckin.bodyIn.Add(readCounter.Count())
 
+	if req.Status == CheckinRequestStatus("") {
+		return val, fmt.Errorf("checkin status missing")
+	}
+	if len(req.Message) == 0 {
+		return val, fmt.Errorf("checkin message missing")
+	}
```
jsoriano, line 189: On what cases these fields were empty? Is it possible that we are breaking checkin in some case?

michel-laterman: @cmacknz, would the agent ever try to check in with an empty message?

cmacknz: It shouldn't, but if there were a bug that caused this, would it be better to accept the checkin instead of marking the agent offline and updating the component information?

michel-laterman: I've changed this to log a warning instead

michel-laterman:
buildkite test this

Issue #3239, opened 2024-01-29
logs consistency


michel-laterman:
fleet-server has log lines with inconsistent attributes.

For example:

https://github.com/elastic/fleet-server/blob/5e8a1e9f9f3cf9d7b97e8c41a7c6611b414eca55/internal/pkg/policy/policy_output.go#L54

https://github.com/elastic/fleet-server/blob/5e8a1e9f9f3cf9d7b97e8c41a7c6611b414eca55/internal/pkg/policy/self.go#L265

Both refer to policy outputs but are named different things. 
We should go through the codebase and try to make these more consistent (following ecs)

juliaElastic:
Few more places to make consistent:
The action type is logged under different fields:
```
ack event
actionId

Action delivered to agent on checkin
id
```

Found that the policy id is also logged under different names:
```
New revision of policy received and added to the queue
fleet.policy.id

Policy coordinator added a new policy revision
policy_id
```

juliaElastic:
Revision_idx fields are changed in this pr https://github.com/elastic/fleet-server/pull/3234

juliaElastic:
Found one more place where revision and coordinator idx is logged under a different name:
https://github.com/elastic/fleet-server/blob/618952027110aae6fe4369ff625f178fcf17f836/internal/pkg/api/handleCheckin.go#L768-L769

PR #3247, opened 2024-01-31
make str log fields consistent

juliaElastic:

Making log fields consistent to make it easier to check the logs.

## Design Checklist

## Checklist

## Related issues

Closes https://github.com/elastic/fleet-server/issues/3239


michel-laterman:
thanks for getting a start on this!

internal/pkg/coordinator/monitor.go
```diff
@@ -431,7 +431,7 @@ func runCoordinatorOutput(ctx context.Context, cord Coordinator, bulker bulk.Bul
 			if err != nil {
 				s.Err(err).Msg("Policy coordinator failed to add a new policy revision")
 			} else {
-				s.Info().Int64("revision_id", p.RevisionIdx).Msg("Policy coordinator added a new policy revision")
```
juliaElastic, line 434: removed redundant log field (already there in line 429)

internal/pkg/api/handleCheckin.go
```diff
@@ -765,8 +766,8 @@ func processPolicy(ctx context.Context, zlog zerolog.Logger, bulker bulk.Bulk, a
 	defer span.End()
 	zlog = zlog.With().
 		Str("fleet.ctx", "processPolicy").
-		Int64("fleet.policyRevision", pp.Policy.RevisionIdx).
-		Int64("fleet.policyCoordinator", pp.Policy.CoordinatorIdx).
+		Int64(dl.FieldRevisionIdx, pp.Policy.RevisionIdx).
```
michel-laterman, line 769: Should we define these as constants in `logger/ecs.go`?

juliaElastic: done

michel-laterman (approved):

PR #3243, opened 2024-01-30
[Buildkite] Ignore non successful exit code in cloud-deploy target (test-cloude2e)

mrodm:
## What is the problem this PR solves?

Cloud deployments are not cleaned in case of failure.
Example of build: https://buildkite.com/elastic/fleet-server/builds/2953#018cfe36-6262-440e-82a4-549363ded677

## How does this PR solve the problem?

Ignore non successful exit codes when running cloud-deploy target.

## Design Checklist

- [ ] I have ensured my design is stateless and will work when multiple fleet-server instances are behind a load balancer.
- [ ] I have or intend to scale test my changes, ensuring it will work reliably with 100K+ agents connected.
- [ ] I have included fail safe mechanisms to limit the load on fleet-server: rate limiting, circuit breakers, caching, load shedding, etc.

## Checklist

- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation
- [ ] I have made corresponding change to the default configuration files
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] I have added an entry in `./changelog/fragments` using the [changelog tool](https://github.com/elastic/fleet-server#changelog)

PR #3120, opened 2023-11-22
Fix cloude2e tests

michel-laterman:
Fix issue that prevented cloud e2e test from running.

Output on previous tests had an error:
```
make[2]: Entering directory '/opt/buildkite-agent/builds/bk-agent-prod-gcp-1700554025504218876/elastic/fleet-server/dev-tools/cloud'
""
make[2]: Leaving directory '/opt/buildkite-agent/builds/bk-agent-prod-gcp-1700554025504218876/elastic/fleet-server/dev-tools/cloud'
FLEET_SERVER_URL=make[1]: Entering directory '/opt/buildkite-agent/builds/bk-agent-prod-gcp-1700554025504218876/elastic/fleet-server/dev-tools/cloud' "" make[1]: Leaving directory '/opt/buildkite-agent/builds/bk-agent-prod-gcp-1700554025504218876/elastic/fleet-server/dev-tools/cloud' go test -v -tags=cloude2e -count=1 -race -p 1 ./testing/cloude2e
bash: Entering: command not found
make[1]: *** [Makefile:396: test-cloude2e-set] Error 127
make[1]: Leaving directory '/opt/buildkite-agent/builds/bk-agent-prod-gcp-1700554025504218876/elastic/fleet-server'
make: [Makefile:389: test-cloude2e] Error 2 (ignored)
```
This change correctly runs the test

PR #3249, opened 2024-02-01
[Buildkite] Update terraform ec provider for Cloud E2E tests step

mrodm:

In some CI builds, deployments created in Cloud e2e steps are not deleted. 
Example:
https://buildkite.com/elastic/fleet-server/builds/3075#018d6497-91b1-4b3d-908d-33c2b77bd316/111-492

Update the terraform provider `elastic/ec` up to 0.9.0. In that release, there is a bugfix that could be related to this issue:
https://github.com/elastic/terraform-provider-ec/blob/master/CHANGELOG.md#090-september-22-2023

## Design Checklist

## Checklist

## Related issues

- Relates https://github.com/elastic/fleet-server/pull/3243


mrodm:
buildkite test this

mrodm:
buildkite test this

dev-tools/cloud/terraform/main.tf
```diff
@@ -4,7 +4,7 @@ terraform {
   required_providers {
     ec = {
       source  = "elastic/ec"
-      version = "0.5.1"
+      version = "0.9.0"
```
mrodm, line 7: Updated the version following the changes in https://github.com/elastic/fleet-server/pull/3120

dev-tools/cloud/terraform/outputs.tf
```diff
@@ -16,16 +16,16 @@ output "elasticsearch_password" {
...
 
 output "kibana_url" {
-  value       = ec_deployment.deployment.kibana.0.https_endpoint
+  value       = ec_deployment.deployment.kibana.https_endpoint
   description = "The secure Kibana URL"
 }
 
 output "fleet_url" {
-  value       = ec_deployment.deployment.integrations_server.0.fleet_https_endpoint
+  value       = ec_deployment.deployment.integrations_server.endpoints != null ? ec_deployment.deployment.integrations_server.endpoints.fleet : ""
```
mrodm, line 29: Currently, `ec_deployment.deployment.integrations_server.endpoints` is reported as null, so it needs to be taken into account here.

dev-tools/cloud/terraform/main.tf
```diff
@@ -58,12 +58,16 @@ resource "ec_deployment" "deployment" {
     "creator" = var.creator
 }
 
-  elasticsearch {}
+  elasticsearch = {
+    hot = {
+      autoscaling = {}
+    }
```
jsoriano, line 64: Is this empty configuration required? Or are we disabling autoscaling or something? It may be worth of a comment.

dev-tools/cloud/terraform/outputs.tf
jsoriano: Was it always a string before?

jsoriano (approved):

dev-tools/cloud/terraform/main.tf
mrodm: It comes from the basic example in the terraform provider:
https://registry.terraform.io/providers/elastic/ec/latest/docs/resources/deployment#basic

According to that:
- `hot` is a required parameter of `elasticsearch`
- `autoscaling` is a required parameter of `hot`

So from what I understand, this is the minimum necessary settings for `elasticsearch`.

Issue #2765, opened 2023-07-05
Add CI pipeline for OSX


jsoriano:
Even if we don't generate fleet-server artifacts for OSX, many of our developers use it on a daily basis. We need to ensure that we can always reliably run tests locally. For this we should have a CI pipeline that run tests on OSX and detect regressions there.

Context: On a recent change on E2E tests (https://github.com/elastic/fleet-server/pull/2710), I broke them for OSX, this was not detected because neither I or CI use OSX. We discovered it while testing a different change and we had to dedicate some time to investigate and fix it, as part of an unrelated task. This kind of issues slow down development.

jlind23:
@pazone i am for now assigning you this as it is also blocked on the Mac OS worker availability.

pazone:
@jsoriano We are ready to start implementation on that since the ephemeral macos instances are available for us. 
However, we have a quite limited amount of macOS VMs capacity, and as we discovered recently they are quite slow because of the lack of CPUs. 
So what kind of tests do we need to execute on MacOS? Keeping in mind that might be slower we can probably have some subsets of tests for MacOS

jsoriano:
@pazone, this is great. I think that we can start with the [check and build](https://github.com/elastic/fleet-server/blob/8daafefd267cd26c8f6f09cfd1f7c96986e50db1/.buildkite/pipeline.yml#L15) and [unit test](https://github.com/elastic/fleet-server/blob/8daafefd267cd26c8f6f09cfd1f7c96986e50db1/.buildkite/pipeline.yml#L73) steps. They should be light and a good starting point.

Later we can add integration and e2e tests if there is enough capacity. The issue that originated this request was with e2e tests.

cc @michel-laterman

Thanks!

pazone:
@jsoriano Do we also need MacOS 13 arm?

jsoriano:
> Do we also need MacOS 13 arm?

Not sure about the versions that we need. The original intention of this issue was to keep things working for developers that use to work in MacOS. Do you have stats on the current usage share? cc @michel-laterman in case he has more details.

> Should we proceed with e2e tests?

Sounds good to me if we have capacity.

PR #3253, opened 2024-02-05
[CI] MacOS 13 Unit tests

pazone:
 We need to ensure that we can always reliably run tests locally. For this, we should have a CI pipeline that runs tests on OSX and detects regressions there.

Makes buildkite run unit tests on each PR and push to `main`. 

Added steps: 
 - Run local build: MacOS
 - Run unit tests: MacOS 13

Made unit test artifacts and sonarqube scan ready for cross-platform builds

Relates https://github.com/elastic/fleet-server/issues/2765


jsoriano (approved):

.buildkite/pipeline.yml
```diff
@@ -31,6 +31,13 @@ steps:
           cpu: "8"
           memory: "4G"
 
+      - label: ":building_construction: Run local build: MacOS"
```
v1v, line 34: @pazone, what's the reason we build this? IIUC, the tests step does not use these artifacts? Or am i seeing this wrongly?

Issue #3254, opened 2024-02-05
Policy monitor load issues


michel-laterman:
## Overview

The fleet-server performance tests have been failing for a while.
The specific test case that fails is multiple policy changes (submitted before they are fully acked). 
This is reproducible on an 8GB ess instance with 5k agents.

Currently we do not have reports from customers or the community where this is an issue.

There is some discussion and investigation of this issue in the [policy monitor debounce pr](https://github.com/elastic/fleet-server/pull/3234) but I believe we should move the discussion of the issue here.

Our current proposal for a work around is to increase the default policy-monitor debounce time and extend/remove the policy dispatch timeout (currently 5s).

## Issues

I believe that the issue with this scale test is created by the following components:
- policy monitor limiter (previously the throttle)
- policy monitor has a single mutex
- no early cancelations

The limiter introduces a delay (default 5ms) in between dispatches to subscriptions. This is done to avoid having a policy change that effects a lot of agents trigger all checkins from writing at once (and allocating expensive gzip writers).

However, this delay increases the time it takes for the `dispatchPending` method, and since there is a single mutex other operations on the policy monitor are blocked during this, including processing new subscriptions, unsubscribes, and updating the policies.

There is also no way to cancel a dispatch early; specifically if we are dispatching 10k subscriptions on rev:2, and rev:3 is detected, we should stop dispatching rev:2, update the policy reference to rev:3 then dispatch the newest revision to the remaining subscriptions.

## Proposed solutions

I think we can use a combination of solutions to fix the policyMonitor issues under load.

### Unify action+policy limits in checkinT

The action limit (introduced in #2994) and the policy limit are both used to stop the checkin endpoint from writing every response to awaiting agents at once (so that expensive gzip writer allocations can be avoided).

I think that we should remove the limits from the actionDispatcher and policyMonitor and use a unified limiter in checkT (to control access to the gzip pool in writeResponse).
This would allow the policyMonitor to dispatch pending changes much faster.

This is the easiest to implement and likely has the most impact.

### Interruptible dispatches

If we refactor the policy monitor so that `dispatchPending` may be canceled early when an updated policy is received by the underlying monitor we can avoid dispatching a policy that is outdated to agents.

This solution requires a lot more work in terms of implementation/testing and can be done along with any other solution.

### Change subscription tracking

Currently the policy subscriptions are tracked as a linked list  associated with a policy id as well as the pendingQ of the policy monitor (another linked list) that tracks subscriptions with pending changes.
[...]

Issue #3087, opened 2023-11-07
context.TODO and loggers


michel-laterman:
https://github.com/elastic/fleet-server/pull/3078 changed how we pass loggers to mainly use the context.
as part of the change we added `context.TODO()` where contexts are not available.

We should see if we can replace these calls by altering the function calls.
We can also investigate and decide if we want to explicitly pass the logger object in a few places (such as in the api module).

This could be a good issue for on-week

PR #3255, opened 2024-02-05
Replace policy and action limiters with a checkin limiter

michel-laterman:

Scale tests for multiple policy changes are failing. A contributing factor is the policy limiter which increases the time it takes for policies to be dispatched (and the policy mutex lock to be held).

Replace the separate policy and action limiters with a unified limiter in the checkinT struct that is used if a response action (which includes policy change actions that are generated by the policy monitor) is detected in the checkin response, and gzip in enabled.

This means that the policyMonitor will dispatch pending policies much faster and release the lock so a policy may be updated/new subscriptions may be processed, but our checkin responses are still rate limited so we can reuse our gzip pool.

Note that the action_limit settings will be used and the policy_limit settings are ignored.

## Design Checklist

## Checklist

- [x] I have made corresponding changes to the documentation [pr here](https://github.com/elastic/ingest-docs/pull/895)

## Related issues

- Relates #3254


michel-laterman:
buildkite run perf-tests

michel-laterman:
serverless perf tests have failed due to ongoing AZ issues forcing containers to go oom.
ECS perf test run is here: https://buildkite.com/elastic/observability-perf/builds/2367#018d7f6d-39a4-4480-90f7-09c8e788ab95 and it looks to have succeeded.

michel-laterman:
latest ecs perf-test: https://buildkite.com/elastic/observability-perf/builds/2370#018d8079-17f9-4085-afd5-29bc28fe0433
looks like it's succeeding

internal/pkg/api/handleCheckin_test.go
```diff
@@ -267,12 +267,12 @@ func TestResolveSeqNo(t *testing.T) {
 			ctx, cancel := context.WithCancel(context.Background())
 			defer cancel()
 			verCon := mustBuildConstraints("8.0.0")
-			cfg := &config.Server{}
+			cfg := &config.Server{Limits: config.ServerLimits{ActionLimit: config.Limit{Interval: 5 * time.Millisecond, Burst: 1}}}
```
juliaElastic, line 270: should the `Interval` and `Burst` values be moved to constants?

internal/pkg/api/handleCheckin.go
```diff
@@ -90,6 +92,13 @@ func NewCheckinT(
 	tr *action.TokenResolver,
 	bulker bulk.Bulk,
 ) *CheckinT {
+	rt := rate.Every(cfg.Limits.ActionLimit.Interval)
+	if cfg.Limits.ActionLimit.Interval == 0 && cfg.Limits.PolicyThrottle != 0 {
+		rt = rate.Every(cfg.Limits.PolicyThrottle)
+	} else if cfg.Limits.ActionLimit.Interval == 0 && cfg.Limits.PolicyThrottle == 0 {
+		rt = rate.Inf
+	}
+	zerolog.Ctx(context.TODO()).Debug().Any("event_rate", rt).Int("burst", cfg.Limits.ActionLimit.Burst).Msg("checkin response gzip limiter")
```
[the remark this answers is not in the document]
michel-laterman: Ideally yes, but we would need to make sure that the context is tied to the function instead of `checkinT`'s lifecycle.
We also use context.TODO in a few other places similar to this.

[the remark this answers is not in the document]
michel-laterman: There's an existing issue to address this: https://github.com/elastic/fleet-server/issues/3087

juliaElastic (approved):
Code LGTM

internal/pkg/api/handleCheckin.go
```diff
@@ -99,6 +108,7 @@ func NewCheckinT(
 		gcp:    gcp,
 		ad:     ad,
 		tr:     tr,
+		limit:  rate.NewLimiter(rt, cfg.Limits.ActionLimit.Burst),
```
juliaElastic, line 111: There is no `ActionLimit.Max` setting, does it mean only the `Burst` is being limited? What does an `Interval: 0.25ms` mean?

michel-laterman: interval and burst are used to configure the rate limiter (interval is the time it takes for 1 token to be added to the rate limit pool, burst is max pool size)
the max attributes was used to add limits to the total number of connections allowed on an endpoint (here we would use the checkin endpoint setting).

michel-laterman:
buildkite test this

michel-laterman:
buildkite run perf-tests

Issue #3263, opened 2024-02-09
Agent `upgraded_at` field keeps updating to current time 


juliaElastic:
Stack version 8.12.1 and possibly others.

There seems to be an issue of Agents `upgraded_at` field keep being updated to current time, and this results in Fleet UI now showing `Upgrade available` when it should, and `Upgrade agent` action being disabled, because Fleet UI doesn't consider agent upgradeable if the agent was updated in the last 10 minutes.

It's not clear yet if the issue is on fleet-server or agent side.

Reproduced on a fresh 8.12.1 cluster, by enrolling a 8.11.4 agent, upgrade to 8.12.0 and wait 10 minutes.
The agent is still not allowed to be upgraded again to 8.12.1, and the `upgraded_at` field looks recent, event though the last upgrade happened more than 10m ago.

Workaround: 
- force upgrade agents with the API `"force": true` flag, or
- manually delete the `upgrade_details:null` value from agent docs

jlind23:
@juliaElastic does it mean a newly install agent on 8.12.0 would have successfully be upgraded to 8.12.1?
@amolnater-qasource do you have a scenario where an agent of the previous minor 8.11 is upgrading to the next minor and then all the patches? 8.12.0 then 8.12.1?
@kpollich @juliaElastic is there any way for us to automatically test this?

kpollich:
> does it mean a newly install agent on 8.12.0 would have successfully be upgraded to 8.12.1?

An agent on 8.12.0 cannot be upgraded to 8.12.1 via Fleet UI currently without the workaround Julia drafted here: https://github.com/elastic/fleet-server/pull/3264#issuecomment-1936216485. 

> is there any way for us to automatically test this?

We need an automated test on all release branches where an upgrade from an agent on the latest available patch release for that branch is upgraded to the build for the current HEAD of that release branch. e.g. on the `8.12` branch, we'd run an upgrade for an agent running the released `8.12.0` agent binary to the current `8.12.0-SNAPSHOT` build built off the release branch.

Additionally, we could have a daily run that does the same using the daily snapshot build instead of a PR build.

jlind23:
@pierrehilbert @cmacknz regarding Kyle's comment above isn't this something we test already in the elastic agent testing framework?

amolnater-qasource:
Hi @jlind23 

[...]

PR #3264, opened 2024-02-09
fix process upgrade details when null

juliaElastic:

Agent was never upgradeable again after an upgrade to 8.12.0

When `upgrade_details: null` comes from the agent doc, the logic incorrectly supposed it has a value, and set `upgraded_at: now`. This resulted the agent never being upgradeable again on the UI (as it is not allowed to upgrade again within 10 minutes of an upgrade).
Upgraded the schema, so that the `null` value is correctly translated to `nil`, and the condition in `handleCheckin` can check for `nil`.

- Start a stack at least 8.12.1 or latest.
- enroll an agent 8.11.4 or older
- upgrade to 8.12.0
- wait 10 minutes so that the watcher period ends
- expect the agent to be upgradeable again to 8.12.1

image

## Checklist

## Related issues

Closes https://github.com/elastic/fleet-server/issues/3263

### DIFF
--- a/internal/pkg/model/schema.go
+++ b/internal/pkg/model/schema.go
@@ -196,7 +196,7 @@ type Agent struct {
 	UpdatedAt string `json:"updated_at,omitempty"`
 
 	// Additional upgrade status details.
-	UpgradeDetails json.RawMessage `json:"upgrade_details,omitempty"`
+	UpgradeDetails *UpgradeDetails `json:"upgrade_details,omitempty"`
 
 	// Date/time the Elastic Agent started the current upgrade
 	UpgradeStartedAt string `json:"upgrade_started_at,omitempty"`
@@ -496,4 +496,8 @@ type ToRetireAPIKeyIdsItems struct {
 
 	// Date/time the API key was retired
 	RetiredAt string `json:"retired_at,omitempty"`
+}
+
+// UpgradeDetails Additional upgrade status details.
+type UpgradeDetails struct {
 }
--- a/changelog/fragments/1707481003-fix-upgraded-at.yaml
+++ b/changelog/fragments/1707481003-fix-upgraded-at.yaml
@@ -1,0 +1,32 @@
+# Kind can be one of:
+# - breaking-change: a change to previously-documented behavior
+# - deprecation: functionality that is being removed in a later release
+# - bug-fix: fixes a problem in a previous version
+# - enhancement: extends functionality but does not break or fix existing behavior
+# - feature: new functionality
+# - known-issue: problems that we are aware of in a given version
+# - security: impacts on the security of a product or a user’s deployment.
+# - upgrade: important information for someone upgrading from a prior version
+# - other: does not fit into any of the other categories
+kind: bug-fix
+
+# Change summary; a 80ish characters long description of the change.
+summary: Fixed a bug where agents were stuck in non-upgradeable state after upgrade.
+
+# Long description; in case the summary is not enough to describe the change
+# this field accommodate a description without length limits.
+# NOTE: This field will be rendered only for breaking-change and known-issue kinds at the moment.
+# description: 
+
+# Affected component; usually one of "elastic-agent", "fleet-server", "filebeat", "metricbeat", "auditbeat", "all", etc.
+component:
+
+# PR URL; optional; the PR number that added the changeset.
+# If not present is automatically filled by the tooling finding the PR where this changelog fragment has been added.
+# NOTE: the tooling supports backports, so it's able to fill the original PR number instead of the backport PR number.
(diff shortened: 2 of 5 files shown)


michel-laterman (approved):
Would we need to add this as a known issue to some of the 8.12.x releases?
Is there a work-around for customers?

internal/pkg/model/schema.go
```diff
@@ -497,3 +497,7 @@ type ToRetireAPIKeyIdsItems struct {
 	// Date/time the API key was retired
 	RetiredAt string `json:"retired_at,omitempty"`
 }
+
+// UpgradeDetails Additional upgrade status details.
+type UpgradeDetails struct {
```
nchaulet, line 502: should we use the upgradedetails model from the agent-lib here? or it is something different?

juliaElastic:
> Would we need to add this as a known issue to some of the 8.12.x releases? Is there a work-around for customers?

How can we add to the known issues?
I guess a workaround is to either force upgrade those agents, or manually delete the `upgrade_details:null` value from agent docs.

kpollich:
> How can we add to the known issues?

We can file an issue to ingest-docs summarizing the issue and @kilfoyle should be able to help (providing he has some time of course!) with getting a known issue into the release notes retroactively. 

> I guess a workaround is to either force upgrade those agents

This sounds good to me as a recommendation. Force upgrading is hopefully not overly cumbersome for users, and should be fairly straightforward for us to document as an API request if I understand correctly?

juliaElastic:
> We can file an issue to ingest-docs summarizing the issue and @kilfoyle should be able to help (providing he has some time of course!) with getting a known issue into the release notes retroactively.

Thanks, raised https://github.com/elastic/ingest-docs/issues/907

lucabelluccini:
Hello,
as suggested by @michel-laterman:

- [ ] Create a known issue in the release notes of the affected versions (it seems both 8.12.0 and 8.12.1)
- [ ] We need to determine an easy to use workaround for users.
    - [ ] Bulk upgrade via APIs? How?
    - [ ] Manual patching of `.fleet-agent` docs? Risky and also not guaranteed over time
- [ ] Write a known issue for support portal (ongoing)
- [ ] Ensure this is merged for 8.12.2 or 8.13.0 (possibly assess if it might entitle to something more urgent?)

juliaElastic:
Updated the doc request with an example of how to use `force` flag with `bulk_upgrade` API.

juliaElastic:
Oh it seems that the workaround to use the force flag doesn't work: https://github.com/elastic/infosec/issues/16411

juliaElastic:
Unfortunately it looks like the `force` flag doesn't work as a workaround here, the [10m validation](https://github.com/elastic/kibana/blob/main/x-pack/plugins/fleet/server/routes/agent/upgrade_handler.ts#L88) is there even then, we have to come up with another workaround.
I'm thinking of an `update_by_query` command which would remove `upgrade_details:null` values from agent doc, unless there is a better idea.

lucabelluccini:
Update by query would persist until the next check-in?
Also, we would need to be super careful on the filtering section of the update by query - the risk of breaking other agents is not negligible.

juliaElastic:
> Update by query would persist until the next check-in? Also, we would need to be super careful on the filtering section of the update by query - the risk of breaking other agents is not negligible.

`upgrade_details` is not changed on checkin, only if another upgrade is attempted

I've put together these steps, which are updating those docs where `upgrade_details` is null or not defined. It has to be run as superuser.
Tested locally that updating the docs makes the logic stop updating `upgraded_at` field, so the agent should be available for upgrade after 10m.

```
 POST _security/role/fleet_superuser
 {
    "indices": [
        {
            "names": [".fleet*",".kibana*"],
            "privileges": ["all"],
            "allow_restricted_indices": true
        }
    ]
  }
        
POST _security/user/fleet_superuser 
 {
    "password": "password",
    "roles": ["superuser", "fleet_superuser"]
 }

curl -sk -XPOST --user fleet_superuser:password -H 'content-type:application/json' \
  -H'x-elastic-product-origin:fleet' \
  http://localhost:9200/.fleet-agents/_update_by_query \
  -d '{
  "script": {
    "source": "ctx._source.remove(\"upgrade_details\")",
    "lang": "painless"
  },
  "query": {
    "bool": {
        "must_not": {
          "exists": {
            "field": "upgrade_details"
          }
        }
      }
    }
}'

DELETE _security/user/fleet_superuser
DELETE _security/role/fleet_superuser
```

lucabelluccini:
Thank you @juliaElastic

WiegerElastic:
https://github.com/elastic/infosec/issues/16411

PR #3273, opened 2024-02-12
Add optional subscription channel size

michel-laterman:

Daily 10k scale tests seems to still be flakey on the multiple policy change test.
This is meant as another simple solution before we try to add early cancelations to dispatching policy changes.

Add an optional arg when subscribing to a subscription monitor to specify the channel size. Min val of 1 is enforced. Use in policy monitor to allow more buffered changes to be queued as mutex might be help and block other dispatches.

## Design Checklist

## Checklist

## Related issues

- Relates #3254

### DIFF
--- a/internal/pkg/monitor/mock/monitor.go
+++ b/internal/pkg/monitor/mock/monitor.go
@@ -41,7 +41,7 @@ func NewMockMonitor() *MockMonitor {
 	return &MockMonitor{}
 }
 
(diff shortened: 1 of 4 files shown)


michel-laterman:
buildkite run perf-tests

michel-laterman:
the 10k serverless tests succeeded, but one of the pods went into a crashloop; this is not a good approach to resolve the issue.

PR #3277, opened 2024-02-13
Consume APMConfig from input units in agent mode

### LINKED ISSUE #2868: Consume APM configuration over the control protocol
In order to enable APM tracing when Fleet Server is managed by Elastic Agent, Fleet Server needs to apply APM configuration settings when they're passed over the control protocol.

This may require that we change how the `apm.Tracer` object is created and passed to internal objects as right now it is only created if APM is enabled on initial boot of Fleet Server here: https://github.com/elastic/fleet-server/blob/f76576d2e0e3c38efe20a2c2c35b2ef80d7f8f12/internal/pkg/server/fleet.go#L549-L552

We may instead want to rely on `apm.DefaultTracer` instead of manually passing a tracer around and then use `apm.SetDefaultTracer` when the config change comes over the control protocol.

References:
- elastic-agent-client changes: https://github.com/elastic/elastic-agent-client/pull/82
- https://github.com/elastic/elastic-agent/issues/2612

michel-laterman:

elastic-agent-client is now able to send APMConfig settings separately from expected config (similar to how the log level can be separated), but fleet-server is unable to use these settings.

Consume and prefer APMConfig from the expected input component while operating in agent mode over config in inputs[0].server.instrumentation.

## Design Checklist

## Checklist

## Related issues

- Closes #2868

### DIFF
--- a/changelog/fragments/1707852625-Prefer-elastic-agent-client-APMConfig-in-agent-mode.yaml
+++ b/changelog/fragments/1707852625-Prefer-elastic-agent-client-APMConfig-in-agent-mode.yaml
@@ -1,0 +1,36 @@
+# Kind can be one of:
+# - breaking-change: a change to previously-documented behavior
+# - deprecation: functionality that is being removed in a later release
+# - bug-fix: fixes a problem in a previous version
+# - enhancement: extends functionality but does not break or fix existing behavior
+# - feature: new functionality
+# - known-issue: problems that we are aware of in a given version
+# - security: impacts on the security of a product or a user’s deployment.
+# - upgrade: important information for someone upgrading from a prior version
+# - other: does not fit into any of the other categories
+kind: enhancement
+
+# Change summary; a 80ish characters long description of the change.
+summary: Prefer elastic-agent-client APMConfig in agent mode
+
+# Long description; in case the summary is not enough to describe the change
+# this field accommodate a description without length limits.
+# NOTE: This field will be rendered only for breaking-change and known-issue kinds at the moment.
+description: |
+  While running in agent-mode fleet-server will use the APMConfig
+  settings of expected input if it's set over the settings in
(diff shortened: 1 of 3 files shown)


internal/pkg/server/agent.go
```diff
@@ -420,10 +421,54 @@ func (a *Agent) configFromUnits() (*config.Config, error) {
...
+			TLS:          iTLS,
+			Environment:  apmest.GetEnvironment(),
+			APIKey:       apmest.GetApiKey(),
+			SecretToken:  apmest.GetSecretToken(),
+			Hosts:        apmest.GetHosts(),
+			GlobalLabels: apmest.GetGlobalLabels(),
+		}
+		return cfg, nil
+	}
+	return config.Instrumentation{}, fmt.Errorf("unable to transform APMConfig to instrumentatiopn")
```
juliaElastic, line 473:
```suggestion
	return config.Instrumentation{}, fmt.Errorf("unable to transform APMConfig to instrumentation")
```

changelog/fragments/1707852625-Prefer-elastic-agent-client-APMConfig-in-agent-mode.yaml
```diff
@@ -0,0 +1,36 @@
...
+  managing agents to inject APM config data.
+
+# Affected component; a word indicating the component this changeset affects.
+component:
+
+# PR URL; optional; the PR number that added the changeset.
+# If not present is automatically filled by the tooling finding the PR where this changelog fragment has been added.
+# NOTE: the tooling supports backports, so it's able to fill the original PR number instead of the backport PR number.
+# Please provide it if you are adding a fragment for a different PR.
+#pr: https://github.com/owner/repo/1234
```
juliaElastic, line 32:
```suggestion
pr: 3277
```

juliaElastic (approved):
LGTM

PR #3278, opened 2024-02-14
return 429 error if ES api key auth returned 429

juliaElastic:

Misleading 401 unauthorized response when ES API key auth returns 429 error. 

Return 429 from Fleet Server API if the root cause is ES auth 429 error.

Tested by running a scale test with a custom image with changes from this pr: https://buildkite.com/elastic/observability-perf/builds/2421#018da31f-8f18-43c5-aac0-5d559f0152cb

Errors show up as 429:
image

## Design Checklist

## Checklist

### DIFF
--- a/internal/pkg/api/auth.go
+++ b/internal/pkg/api/auth.go
@@ -7,7 +7,6 @@ package api
 
 import (
 	"errors"
-	"fmt"
 	"net/http"
 	"time"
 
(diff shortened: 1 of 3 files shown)


jsoriano (commented):
Not tested myself, but change makes sense, thanks!

Would it be possible to add a test for this?

juliaElastic:
> Would it be possible to add a test for this?

I didn't see existing unit tests on this, I'll try to add.

jsoriano (approved):
Thanks for adding tests :+1:

jsoriano (approved):

michel-laterman (approved):

juliaElastic:
buildkite, test this

juliaElastic:
buildkite test this

michel-laterman:
@juliaElastic, Should we backport this to 8.13?

PR #3283, opened 2024-02-14
Fix file delivery order

pzl:

Chunked file delivery would occasionally be delivered out-of-order for certain file sizes

Chunk identifiers (not binary contents, just a few metadata fields) are just manually sorted in code. The positional order is not stored in the chunk documents, only contained in a doc's `_id` field, which is not a sortable result. Given a upper limit of file size (100mb) and fixed chunk size (4mb), the upper bound for number of documents here is 25. Software sort is reasonable enough

Update `securitySolution.maxUploadResponseActionFileBytes` in kibana.yml to be `104857600`

spin up an endpoint, use the responder console to upload a ~90mb file. The bug reproducing-error is a file hash mismatch. With this fix, hashes will match and the action will be marked success

## Design Checklist

## Checklist


juliaElastic (approved):
LGTM

pzl:
buildkite test this

PR #3293, opened 2024-02-20
Disable all e2e tests that require elastic-agent

michel-laterman:
Disable the test suites that require elastic-agent and stop building the custom test image for e2e tests.

### DIFF
--- a/dev-tools/cloud/Makefile
+++ b/dev-tools/cloud/Makefile
@@ -33,7 +33,7 @@ cloud-deploy: build-and-push-cloud-image  ## Create a new cloud deployment
 
 .PHONY: cloud-clean
 cloud-clean: ## Clean cloud deployment
-	@cd ${TERRAFORM_PATH}; terraform destroy -auto-approve
+	@cd ${TERRAFORM_PATH}; terraform destroy -auto-approve -var="elastic_agent_docker_image=${CI_ELASTIC_AGENT_DOCKER_IMAGE}:${CUSTOM_IMAGE_TAG}"
 
 .PHONY: cloud-get-fleet-url
 cloud-get-fleet-url:  ## Get Fleet URL from this deployment
--- a/dev-tools/cloud/Makefile
+++ b/dev-tools/cloud/Makefile
@@ -33,7 +33,7 @@ cloud-deploy: build-and-push-cloud-image  ## Create a new cloud deployment
 
 .PHONY: cloud-clean
 cloud-clean: ## Clean cloud deployment
-	@cd ${TERRAFORM_PATH}; terraform destroy -auto-approve -var="elastic_agent_docker_image=${CI_ELASTIC_AGENT_DOCKER_IMAGE}:${CUSTOM_IMAGE_TAG}"
+	@cd ${TERRAFORM_PATH}; terraform destroy -auto-approve
 
 .PHONY: cloud-get-fleet-url
 cloud-get-fleet-url:  ## Get Fleet URL from this deployment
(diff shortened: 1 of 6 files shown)


jlind23 (approved):

dev-tools/cloud/Makefile
```diff
@@ -33,7 +33,7 @@ cloud-deploy: build-and-push-cloud-image  ## Create a new cloud deployment
 
 .PHONY: cloud-clean
 cloud-clean: ## Clean cloud deployment
-	@cd ${TERRAFORM_PATH}; terraform destroy -auto-approve
+	@cd ${TERRAFORM_PATH}; terraform destroy -auto-approve -var="elastic_agent_docker_image=${CI_ELASTIC_AGENT_DOCKER_IMAGE}:${CUSTOM_IMAGE_TAG}"
```
michel-laterman, line 36: @mrodm, I think we need to specify a default here

The [previous build](https://buildkite.com/elastic/fleet-server/builds/3263#018dc7f2-f851-4915-97bc-2f0e5682f871) has an issue where it was hanging awaiting the value for `elastic_agent_docker_image`.

Does this make sense to you? Or is there an alternative approach we should take?

mrodm: According to the docs, these commands should be using files ending with `.auto.tfvars` automatically

https://developer.hashicorp.com/terraform/language/values/variables#variable-definitions-tfvars-files

Not sure why it didn't get that value from it.

> The previous build has an issue where it was hanging awaiting the value for elastic_agent_docker_image.

I will cancel this build to avoid running more time this step.

mrodm: I think the issue here is that the Makefile target `build-and-push-cloud-image` has failed, and therefore no `terraform` command has been executed:
- `docker_image.auto.tfvars` is not created
- `terraform init` has not been run.

So if in this case terraform destroy is executed (from the cleanup) and it fails because of that.

Probably, this can be detected in the cleanup function by checking if `docker_image.auto.tfvars` exists or if the terraform lock file exists. If 

For instance:
```diff
 $ git diff
diff --git dev-tools/cloud/launch_cloud_e2e_tests.sh dev-tools/cloud/launch_cloud_e2e_tests.sh
index 0f5fd2dc..e77c5ec6 100755
--- dev-tools/cloud/launch_cloud_e2e_tests.sh
+++ dev-tools/cloud/launch_cloud_e2e_tests.sh
@@ -5,12 +5,15 @@ set -euo pipefail
 CLOUD_TESTING_BASE="$(dirname $0)"
 
 cleanup() {
-  r=$?
+    r=$?
 
-  echo "--- Cleaning deployment"
-  make -C "${CLOUD_TESTING_BASE}" cloud-clean
-
-  exit $r
+    if [ -f ${CLOUD_TESTING_BASE}/terraform/.terraform.lock.hcl ] ; then
+        echo "--- Cleaning deployment"
+        make -C "${CLOUD_TESTING_BASE}" cloud-clean
+    else
+        echo "+++ Skipped cleaning deployment, no terraform files"
+    fi
+    exit $r
 }
 trap cleanup EXIT INT TERM
```

WDYT ? @michel-laterman

dev-tools/cloud/launch_cloud_e2e_tests.sh
```diff
@@ -7,9 +7,12 @@ CLOUD_TESTING_BASE="$(dirname $0)"
...
   r=$?
 
-  echo "--- Cleaning deployment"
-  make -C "${CLOUD_TESTING_BASE}" cloud-clean
-
+  if [ -f ${CLOUD_TESTING_BASE}/terraform/.terraform.lock.hcl ] ; then
+    echo "--- Cleaning deployment"
+    make -C "${CLOUD_TESTING_BASE}" cloud-clean
+  else
+      echo "+++ Skipped cleaning deployment, no terraform files"
```
mrodm, line 14: Just taking a look to the build , probably it's better to just show the echo wihtout any Buildkite formatting 

```suggestion
      echo "Skipped cleaning deployment, no Terraform files"
```

PR #3295, opened 2024-02-21
Make benchmarks more consistent

### LINKED ISSUE #3152: [Performance] Improve fleet server benchmarks
Thanks to https://github.com/elastic/fleet-server/pull/3141 we now have fleet server benchmarks running on every PR.
As of today this is not a blocking check as we found some flakiness in these benchmarks.

Looking at the comparison results [here](https://buildkite.com/elastic/fleet-server/builds/2661) and in the screenshot below, the performance difference is higher than the 10% threshold we defined. We need to understand why there is so much and fix the flakiness.
Screenshot 2023-12-07 at 14 13 17

## Outcome

- [ ] The outcome of this issue is to make the benchmark run a blocker for the Buildkite pipeline

michel-laterman:

Automated benchstats comparison is inconsistent

Call `b.ResetTimer()` directly before main benchmarks are run in order to ignore any overhead in setting up the benchmarks
Restructure some of the benchmarks for more consistent performance.
Use a nop logger is some benchmarks.
Run `benchstat` after the individual benchmark jobs to get a summery of the benchmarks for the pr/main.

## Related issues

- Closes #3152

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -46,6 +46,7 @@ steps:
         command: ".buildkite/scripts/run_benchmark.sh pr"
         artifact_paths:
           - build/next.out
+          - build/next.stat
         agents:
           provider: "gcp"
           machineType: "c2-standard-8"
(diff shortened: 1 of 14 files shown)


juliaElastic (approved):
LGTM

PR #3314, opened 2024-02-23
Use fleet-server binary version for cloude2e images

michel-laterman:

Bumping fleet-server versions causes the cloude2e test to fail

Use the fleet-server binary version with snapshot flag when building the cloude2e test image instead of the stack version so fleet-server builds will not fail on version bumps.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -286,7 +286,7 @@ release-manager-release: ## - Builds a snapshot release. The Go version defined 
 ## get-version : Get the Fleet server version
 .PHONY: get-version
 get-version:
-	@echo $(DEFAULT_VERSION)
+	@echo $(VERSION)
 
 ##################################################
 # Integration testing targets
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -286,7 +286,7 @@ release-manager-release: ## - Builds a snapshot release. The Go version defined
 ## get-version : Get the Fleet server version
 .PHONY: get-version
 get-version:
-	@echo $(DEFAULT_VERSION)
+	@echo $(VERSION)
```
michel-laterman, line 289: `$VERSION` differs from `$DEFAULT_VERSION` as it may include snapshot or other version indicators.
The only other target (besides cloude2e) that uses `get-version` is the buildkite dra target, which should filter out the snapshot indicator

nchaulet (approved):
LGTM 🚀

Issue #3334, opened 2023-12-28
[Fleet]: Unhealthy agent output badge is not removed on editing incorrect output when agent is not connected.


amolnater-qasource:
**Kibana Build details:**
```
VERSION: 8.12.0 BC3
BUILD: 69985
COMMIT: 2a8afed8572a4c709aa1c64216748197eeb9b18f
Artifact Link: https://staging.elastic.co/8.12.0-61156bc6/summary-8.12.0.html
```

**Host OS:**  All

**Preconditions:**
1. 8.12.0 BC3 Kibana cloud environment should be available.
2. Agent should be installed.

**Steps to reproduce:**
1. Navigate to Fleet Settings.
2. Create invalid Remote Elasticsearch output.
3. Under Agent policy settings, select this invalid Remote Elasticsearch output.
4. Navigate to Fleet settings and observe agent output is Unhealthy.
5. Now remove the connected agent and update the agent output with correct configuration.
6. Observe output status still remains unhealthy.

**Screen Recording:**

https://github.com/elastic/kibana/assets/77374876/fff5f2f2-6b8e-4fb1-a371-9ab51030b8e5

https://github.com/elastic/kibana/assets/77374876/00cff7af-ee3d-42af-a192-46e2b4782f24

**Expected Result:**
Unhealthy agent output badge should be removed on editing incorrect output when agent is not connected and new status should be updated once agent gets connected.

**Feature:**
https://github.com/elastic/kibana/issues/104986

amolnater-qasource:
Hi Team,

We have revalidated this issue on latest 8.13.0 BC7 kibana cloud environment and found it **fixed now**.

**Observations:**
- Unhealthy agent output badge is removed on editing incorrect output to correct when agent is not connected and doesn't appear again after sometime.


**Screen Recording:**

https://github.com/elastic/fleet-server/assets/77374876/3e7a1476-3954-4098-a9d8-40ba951541b9




![image](https://github.com/elastic/fleet-server/assets/77374876/67cc521a-0dc9-4cbf-beb5-7c338d116239)


**Build details:**
VERSION: 8.13.0 BC7
BUILD: 72069
COMMIT: 2e3a5cd43e835baa1d596b1aa54735992259ecb9

Hence, we are marking this issue as **QA:Validated**.
Thanks!!

PR #3335, opened 2024-03-11
Stop self monitor output health reporting if output config is not ack…

juliaElastic:
…ed by agents

Output health kept reporting incorrect state after output config was updated, but the new config was not yet acked by any agents.

Self monitor compares the current bulker output config with the latest output config from `.fleet-policies`. If the bulker doesn't use the latest config, the self monitor will stop reporting output health until an agent acks it.

- Create a remote es output that points to a second ES
- Enroll an agent that uses the remote es
- Update the remote es config to point to an invalid host
- The output health UI should show the connection error
- Stop the agent
- Update the remote es config to point to a valid host again
- The output health UI should clear
- Re-start the agent
- The output health UI should show healthy status

https://github.com/elastic/fleet-server/assets/90178898/94a7f681-ff32-476d-b869-01e50045a68e

## Design Checklist

## Checklist

## Related issues

Closes https://github.com/elastic/fleet-server/issues/3334


nchaulet (approved):
code LGTM 🚀

PR #3337, opened 2024-03-11
Update elastic agent client to v7.8.0

cmacknz:
The elastic-agent-client was several versions out of date. This brings it up to date with the latest changes.


internal/pkg/server/agent.go
```diff
@@ -71,8 +71,8 @@ func NewAgent(cliCfg *ucfg.Config, reader io.Reader, bi build.Info, reloadables
 		reloadables: reloadables,
 	}
 	a.agent, _, err = client.NewV2FromReader(reader, client.VersionInfo{
-		Name:    kFleetServer,
-		Version: bi.Version,
+		Name:      kFleetServer,
+		BuildHash: bi.Commit,
```
cmacknz, line 75: We replaced the version with the commit, because with the early agent releases project the version may not be unique anymore but the commit always is.

juliaElastic (approved):
LGTM

michel-laterman:
I think you need to run `go mod tidy` in `./testing`

cmacknz:
Ah I forgot that the testing module had a separate `go.mod` and couldn't figure out why it was complaining, thnaks.

PR #3350, opened 2024-03-13
Allow level: trace in settings action

michel-laterman:

Allow fleet to set trace level for logging

Allow the trace level to be set in settings actions.

## Design Checklist

## Checklist


airatgizza1:
@michel-laterman Hi, your commit breaks git clone on windows. Remove : from the file name in changelog/fragments/1710319964-Allow-level:-trace-for-settings-actions.yaml

PR #3353, opened 2024-03-13
Add go.mod to pkg/api

michel-laterman:

Allow consumers to use autogenerate API client code without importing the full list of fleet-server dependencies.

Release the autogenerated pkg/api client and type definitions as a submodule so consumers do not have to import fleet-server's entire dependency tree. Update e2e tests to use the new module name. Update oapi runtime to v1.1.1 across all go.mod files.


michel-laterman:
buildkite test this

nchaulet (approved):
LGTM 🚀

PR #3388, opened 2024-03-21
Fix push docker image script: only push to latest tag if on main

juliaElastic:

We found that using the [perf-tests](https://github.com/elastic/fleet-server/blob/main/.buildkite/pipeline.perf-tests.yaml) unintentionally overrides the image `docker.elastic.co/observability-ci/fleet-server:latest`, which has an impact on everyone using the latest image (this broke security team's cypress tests).

The script had a condition on `BUILDKITE_TAG` which is empty for pr and main builds. Instead, changing the condition to only publish to latest image name if `DOCKER_IMAGE_GIT_TAG:main`.
For prs, the envvar has the value of the current branch, e.g. `DOCKER_IMAGE_GIT_TAG="juliaElastic:unhealthy_reason"`

## Design Checklist

## Checklist


nchaulet (approved):
LGTM 🚀

PR #3407, opened 2024-03-27
3110 return 4xx errors instead of 5xx errors for bad requests payloads

kaanyalti:

Some of the fleet server endpoints return 500 errors when the handler runs into an error while validating the incoming request body. Handler should respond with 400 error code.

Added a new error type "BadRequestErr" to encapsulate the errors that may occur while validating a request body. Updated the error table to return a 400 status code HTTPErrResp.

## Design Checklist

## Checklist

### DIFF
--- a/internal/pkg/api/handleUpload.go
+++ b/internal/pkg/api/handleUpload.go
@@ -185,7 +185,7 @@ func (ut *UploadT) validateUploadCompleteRequest(r *http.Request, id string) (st
 
 	var req UploadCompleteRequest
 	if err := json.NewDecoder(r.Body).Decode(&req); err != nil {
-		return "", &BadRequestErr{msg: "unable to parse upload complete request body"}
+		&BadRequestErr{msg: "unable to decode upload complete request"}
 	}
 
 	hash := strings.TrimSpace(req.Transithash.Sha256)
--- a/internal/pkg/api/handleUpload.go
+++ b/internal/pkg/api/handleUpload.go
@@ -77,7 +77,7 @@ func (ut *UploadT) validateUploadBeginRequest(ctx context.Context, reader io.Rea
 		if errors.Is(err, io.EOF) {
 			return nil, "", fmt.Errorf("%w: %w", ErrFileInfoBodyRequired, err)
 		}
-		return nil, "", &BadRequestErr{msg: "error decoding upload begin request", nextErr: err}
+		return nil, "", &BadRequestErr{msg: "unable to decode upload begin request", nextErr: err}
 	}
 
 	// check API key matches payload agent ID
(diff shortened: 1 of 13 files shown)


michel-laterman (commented):
Can you also make sure that the other endpoints behave the same: enroll, ack, uploads, and uploads/id?

internal/pkg/api/handleUpload.go
```diff
@@ -77,7 +77,7 @@ func (ut *UploadT) validateUploadBeginRequest(ctx context.Context, reader io.Rea
 		if errors.Is(err, io.EOF) {
 			return nil, "", fmt.Errorf("%w: %w", ErrFileInfoBodyRequired, err)
 		}
-		return nil, "", err
+		return nil, "", &BadRequestErr{msg: "error decoding upload begin request", nextErr: err}
```
michel-laterman, line 80:
```suggestion
		return nil, "", &BadRequestErr{msg: "unable to decode upload begin request", nextErr: err}
```

internal/pkg/api/handleUpload.go
```diff
@@ -185,7 +185,7 @@ func (ut *UploadT) validateUploadCompleteRequest(r *http.Request, id string) (st
 
 	var req UploadCompleteRequest
 	if err := json.NewDecoder(r.Body).Decode(&req); err != nil {
-		return "", errors.New("unable to parse request body")
+		return "", &BadRequestErr{msg: "unable to parse upload complete request body"}
```
michel-laterman, line 188:
```suggestion
		&BadRequestErr{msg: "unable to decode upload complete request"}
```

internal/pkg/api/handleUpload_test.go
```diff
@@ -934,6 +942,56 @@ func TestUploadCompleteIncorrectTransitHash(t *testing.T) {
...
+		},
+	})
+
+	rec := httptest.NewRecorder()
+	req := httptest.NewRequest(http.MethodPost, "/api/fleet/uploads/"+mockUploadID, strings.NewReader(`{"transithash": {"sha256": `))
+
+	hr.ServeHTTP(rec, req)
+
+	assert.Equal(t, http.StatusBadRequest, rec.Code)
+	assert.Equal(t, rec.Body.String(), "{\"statusCode\":400,\"error\":\"BadRequest\",\"message\":\"Bad request: unable to parse upload complete request body\"}")
```
michel-laterman, line 992:
```suggestion
	assert.Equal(t, rec.Body.String(), "{\"statusCode\":400,\"error\":\"BadRequest\",\"message\":\"Bad request: unable to decode upload complete request\"}")
```

michel-laterman (approved):
Think it looks good, I would just like to have more consistent error messages across the endpoints

PR #3411, opened 2024-03-27
agent-mode loads output from policy

michel-laterman:

When running in agent-mode fleet-server can never connect to any additional Elasticsearch hosts.

From the discussions in the initial attempt to fix this issue in the elastic-agent (https://github.com/elastic/elastic-agent/pull/4473) we have decided to treat the output config that the agent passes to the fleet-server more like a "bootstrap" block that the fleet-server can use to form an initial connection to ES and pull more output config from the associated policy.

This approach does not solve for the edge-case where the URL used in bootstrapping (specified when enrolling/installing) goes down and the fleet-server needs to be restarted.
In this case fleet-server will not be healthy as it can't gather output config from the bootstrap ES host. 
Additionally we still expect that the bootstrap es host will never be removed as a valid output.

Allow the self-monitor that fleet-server starts when in agent mode to send a config.Config struct through the server's config channel. This struct only has the `Output` and (new) `RevisionIdx` attributes set from values retrieved from the policy output.
When fleet receives new config it will handle the output only config as a special case and merge it with the previous 
 output config in order to get an up to date complete config. 
When merging config, non-default values from the policy are preferred.
For an output block to be used, at least one host must be reachable.

1. Enroll an elastic-agent instance running fleet-server and add another host to the elasticsearch output in fleet (ui)
  - the new host does not need to be a valid elasticsearch host
2. Gather a diagnostics bundle from the elastic-agent
3. The `fleet-server.yml` in the bundle should have multiple hosts in the output block, logs in the bundle should also indicate that the output block from the latest revision is being used

## Design Checklist

## Checklist

-  ~~I have made corresponding change to the default configuration files~~

## Related issues

- Closes https://github.com/elastic/elastic-agent/issues/2784


michel-laterman:
buildkite test this

michel-laterman:
buildkite test this

internal/pkg/server/fleet_integration_test.go
```diff
@@ -379,6 +381,7 @@ func TestServerConfigErrorReload(t *testing.T) {
 		cancel()
 	}).Return(nil)
 	mReporter.On("UpdateState", client.UnitStateStopping, mock.Anything, mock.Anything).Return(nil)
+	mReporter.On("UpdateState", client.UnitStateFailed, mock.MatchedBy(func(err error) bool { return errors.Is(err, context.Canceled) }), mock.Anything).Return(nil).Maybe()
```
michel-laterman, line 384: I found that without this in the mocked calls list the test will panic because the `cancel()` call when the state updates to healthy interrupts something and returns the context cancelled error triggering a fail-state detection instead of a stopping call

internal/pkg/config/output_test.go
```diff
@@ -382,3 +382,170 @@ func setTestEnv(t *testing.T, env map[string]string) {
...
+			ServiceToken:     "token",
+			Timeout:          2 * time.Second,
+			MaxRetries:       2,
+			MaxConnPerHost:   3,
+			MaxContentLength: 4,
+		},
+	}}
+	for _, tc := range tests {
+		t.Run(tc.name, func(t *testing.T) {
+			res := MergeElasticsearchFromPolicy(cfg, tc.pol)
```
nchaulet, line 467: @michel-laterman why do we need to merge the policy here? cannot use what come from the config? the proxy and tls settings should be configured there too no? otherwise there is no way for user to change `proxy` and `tls` from Fleet right?

changelog/fragments/1712108631-Use-policy-outputs-when-running-in-agent-mode.yaml
```diff
@@ -0,0 +1,38 @@
...
+summary: Use policy outputs when running in agent-mode
+
+# Long description; in case the summary is not enough to describe the change
+# this field accommodate a description without length limits.
+# NOTE: This field will be rendered only for breaking-change and known-issue kinds at the moment.
+description: |
+  Fleet-server will retrieve and use the output from the policy when running in agent-mode.
+  This allows the fleet-server to connect to multiple Elasticsearch hosts if it is successful when
+  connecting to the host provided at enrolment/installation.
+  We expect that the host provided during enrollment/installation is never removed as a valid output.
```
juliaElastic, line 23: Should we raise an ingest-docs issue to document this?

internal/pkg/config/config.go
```diff
@@ -39,12 +39,13 @@ const kRedacted = "[redacted]"
...
-	Inputs  []Input `config:"inputs"`
-	Logging Logging `config:"logging"`
-	HTTP    HTTP    `config:"http"`
-	m       sync.Mutex
+	Fleet       Fleet   `config:"fleet"`
+	Output      Output  `config:"output"`
+	Inputs      []Input `config:"inputs"`
+	Logging     Logging `config:"logging"`
+	HTTP        HTTP    `config:"http"`
+	RevisionIdx int64   `config:",ignore"`
```
juliaElastic, line 47: what does it mean to `ignore` RevisionIdx?

internal/pkg/config/output.go
```diff
@@ -173,6 +181,108 @@ func (c *Elasticsearch) ToESConfig(longPoll bool) (elasticsearch.Config, error)
...
+		ProxyDisable:     cfg.ProxyDisable,
+		ProxyHeaders:     cfg.ProxyHeaders,
+		TLS:              mergeElasticsearchTLS(cfg.TLS, pol.TLS), // tls can be a special case
+		MaxRetries:       cfg.MaxRetries,
+		MaxConnPerHost:   cfg.MaxConnPerHost,
+		Timeout:          cfg.Timeout,
+		MaxContentLength: cfg.MaxContentLength,
+	}
+	// If policy has a non-default Hosts value use it's values for Protocol and hosts
+	if pol.Hosts != nil && !(len(pol.Hosts) == 1 && pol.Hosts[0] == "localhost:9200") {
```
juliaElastic, line 203: should the `localhost:9200` moved to a constant?

internal/pkg/policy/self.go
```diff
@@ -32,6 +34,10 @@ const DefaultCheckTime = 5 * time.Second
 // DefaultCheckTimeout is the default timeout when checking for policies.
 const DefaultCheckTimeout = 30 * time.Second
 
+var errInvalidOutput = fmt.Errorf("policy output invalid")
```
juliaElastic, line 37:
```suggestion
var ErrInvalidOutput = fmt.Errorf("policy output invalid")
```

internal/pkg/policy/self.go
```diff
@@ -174,9 +183,35 @@ func (m *selfMonitorT) processPolicies(ctx context.Context, policies []model.Pol
...
+					Output: config.Output{
+						Elasticsearch: esOut,
+					},
+					RevisionIdx: m.lastRev,
+				}
+			}
 			break
 		} else if m.policyID == "" && policy.DefaultFleetServer {
 			m.policy = &policy
+			esOut, err := m.getPolicyOutput()
```
juliaElastic, line 202: this logic could be moved to a function to avoid duplication with lines 186-198

internal/pkg/policy/self.go
```diff
@@ -187,6 +222,49 @@ func (m *selfMonitorT) groupByLatest(policies []model.Policy) map[string]model.P
 	return groupByLatest(policies)
 }
 
+// getPolicyOutput will return the Elasticsearch output block of m.policy if there is a new revision.
+func (m *selfMonitorT) getPolicyOutput() (config.Elasticsearch, error) {
+	var policyES config.Elasticsearch
+	// policy revision has not changed
+	if m.policy.RevisionIdx == m.lastRev {
+		return policyES, ErrNoPolicyUpdate
```
juliaElastic, line 230: nit: it's not really an error if the policy hasn't changed, the function could return an extra boolean e.g. `hasChanged` to separate this scenario from errors

internal/pkg/policy/self.go
```diff
@@ -187,6 +222,49 @@ func (m *selfMonitorT) groupByLatest(policies []model.Policy) map[string]model.P
...
+	}
+	// always copy revisionIdx
+	m.lastRev = m.policy.RevisionIdx
+
+	// Find elasticsearch output in the policy
+	// TODO figure out how to get output name from policy in order not to scan outputs?
+	for name, data := range m.policy.Data.Outputs {
```
juliaElastic, line 237: so the logic takes the first elasticsearch output, right? If there are more than one elasticsearch outputs, and the first one is not accessible, fleet-server will never try the second one, is that right?

internal/pkg/config/config.go
michel-laterman: does not serialize

internal/pkg/config/output_test.go
michel-laterman: Currently the elastic-agent injection will not replace keys that are already present in local config:

https://github.com/elastic/elastic-agent/blob/2873f66e40fbbb490dc1367cb4cf2c9ccc937ae3/internal/pkg/agent/application/fleet_server_bootstrap.go#L67-L77

So the agent can set `tls.CAs` from policy if it was not an enrollment arg (i'm not sure if it serializes locally)
but the bigger issue is that the agent only ever sends a single host in the output block due to this behaviour

internal/pkg/policy/self.go
michel-laterman: correct, i want to be able to use the (output) name associated with the integration directly but i'm not sure where I can find the value

changelog/fragments/1712108631-Use-policy-outputs-when-running-in-agent-mode.yaml
michel-laterman: yes, we should

internal/pkg/policy/self.go
```diff
@@ -174,9 +183,35 @@ func (m *selfMonitorT) processPolicies(ctx context.Context, policies []model.Pol
 		policy := latest[i]
 		if m.policyID != "" && policy.PolicyID == m.policyID {
 			m.policy = &policy
+			esOut, err := m.getPolicyOutput()
+			if err != nil {
+				if !errors.Is(err, ErrNoPolicyUpdate) {
+					m.log.Warn().Err(err).Str(logger.PolicyID, m.policyID).Msg("Failed to get fleet-server output")
```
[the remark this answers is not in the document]
michel-laterman: `updateState` would get called directly after this log line, and as that's implemented now it would immediately change the state back to healthy and i'm hesitant to add more state-tracking within the self-monitor (especially for parsing errors)

michel-laterman:
buildkite test this

michel-laterman:
the integration tests are currently failing, many at the enrollment step; i think what's going on is:
1) fleet-server goes online using local config
2) test-case starts and begins an agent enrollement
3) at the same time the policy self-monitor detects the policy output and sends it to the config channel
4) config channel triggers a reload before the enrollment is completed
5) root context gets canceled as part of the reload; the enrollment request responds with a 499 due to early cancelation 

I can think of a couple of different approaches to work around this
* replace the output es client separately from the normal config reload process so it does not trigger a restart of all sub-processes in the fleet-server
* give some time (1-3s?) to allow fleet-server to drain connections once the root context is cancelled

michel-laterman:
buildkite test this

michel-laterman:
It looks like the integration tests are caused by the logger; this is only occuring in integration tests and i'm not entirely sure why

michel-laterman:
A lot of the integration test failures may have been caused by the tests not waiting for a fleet-server shutdown before terminating; this led to the logger in the server trying to use `tb.Log` after the test has ended.

The logger we pass to the server has a hook that checks the output policy reload related messages so the tests start after the server has reloaded the output block using the policy.
This is needed so we don't reload when a test agent is enrolling.

juliaElastic (approved):
LGTM

kaanyalti (approved):

internal/pkg/server/fleet.go
```diff
@@ -193,8 +194,54 @@ LOOP:
...
+				// work around to get a new cfg object based off curCfg
+				// we override the output with esOutput and have a complete config with a new mutex
+				tmp, err := ucfg.NewFrom(curCfg, config.DefaultOptions...)
+				if err != nil {
+					log.Error().Err(err).Int64(logger.RevisionIdx, rev).Msg("Unable to convert config")
+					continue
+				}
+				err = tmp.Unpack(cfg, config.DefaultOptions...)
+				if err != nil {
+					log.Error().Err(err).Int64(logger.RevisionIdx, rev).Msg("Unable to unpack config")
```
michel-laterman, line 237: When running under a real agent unpack results in an error: `verification mode must be an identifier accessing 'inputs.0.server.ssl.verification_mode' accessing 'inputs.0.server.ssl'`
I'll try to do something else to resolve this; but it might need a change to elastic-agent-libs

michel-laterman: pr: https://github.com/elastic/elastic-agent-libs/pull/196

michel-laterman: another pr: https://github.com/elastic/elastic-agent-libs/pull/197

I made a custom build of the fleet-server and ran it with a real agent under a real cluster and found we now see the successful "Using output from policy" log message, but do not see the correct output in the diagnostics bundle. I'm going to see if there is a straightforward way to pass the config to a method in `agent.go` instead of `fleet.go` so we can get functional diagnostics

cmacknz:
> This approach does not solve for the edge-case where the URL used in bootstrapping (specified when enrolling/installing) goes down and the fleet-server needs to be restarted.
In this case fleet-server will not be healthy as it can't gather output config from the bootstrap ES host.
Additionally we still expect that the bootstrap es host will never be removed as a valid output.

As long as agent has received a policy from Fleet server at least once, it will have persisted it locally on disk. This should mean that except for the very first agent checks in, there is a cached set of fleet server outputs that are going to arrive via the control protocol.

Are the control protocol outputs also added to the set of Fleet server hosts when they are receive? They are the same as the ones Fleet Server could query itself, they just come in via a different path.

Removing the bootstrap broker is interesting, it might make sense to model the bootstrap broker as an explicit thing in Fleet if somebody asks us for this. Its purpose is to get the policy the very first time before there are policy outputs to read from. At the same time, once you have gotten the policy at least once, there shouldn't ever be a need to rely only on the bootstrap output again.

cmacknz:
Modelling the bootstrap broker explicitly in diagnostics as something separate from the outputs in the policy might also be a way to account for it diagnostics.

michel-laterman:
@cmacknz, just to be clear I have not adjusted any bootstrap behaviour in fleet-server or the elastic-agent.
I've just allowed the policy self-monitor to send updates to the config reload channel

We know that the (elastic-agent) component modifier adjusts the es hosts it sends to fleet-server to use only what is specified during enrollment when elastic-agent is not restarted if the desciption of agent persistency is correct:
> As long as agent has received a policy from Fleet server at least once, it will have persisted it locally on disk. This should mean that except for the very first time agent checks in, there is a cached set of fleet server outputs that are going to arrive via the control protocol.

So we can have an edge case that occurs as follows:
- ES hosts are set to `A, B`
- fleet-server is enrolled with host `A`
  - policy monitor will send `A, B` to reload
- fleet-server policy is adjusted by user (i.e., max_agent has changed)
- elastic-agent will send updated policy with only host `A` (if the agent has not been restarted)
  - policy monitor will then sent `A, B` to reload again

I'm going to test what output is sent to a fleet-server after an agent restarts
My plan is to run elasticsearch + kibana locally, then enroll an agent running fleet-server and explicitly pass the es ca.
After it is healthy i'll adjust the output to have a different path to the (same) ca and an additional es host then restart the agent

cmacknz:
> We know that the (elastic-agent) component modifier adjusts the es hosts it sends to fleet-server to use only what is specified during enrollment when elastic-agent is not restarted if the description of agent persistency is correct:

Ugh right we are overriding the outputs in the policy. We'd need fleet server to cache the outputs list to get around this.

I don't want to blow up the scope of this change too much, but I wonder if all of this would be simpler if we defined Fleet server to always have two outputs, one is explicitly the bootstrap output(s) and the other is the set of standard outputs in the policy (which may overlap with bootstrap). The bootstrap output only gets used if the standard outputs don't exist yet. 

That way you'd benefit from the agent's cache of the most recent policy. Basically we'd want agent to keep injecting an output it would just be an additional output instead of overwriting the single output that exists today.

michel-laterman:
I tried the following for a test:

Created 8.14-SNAPSHOT on qa
Built an agent using the main branch (did **not** include fleet-server from this pr)
- enrolled a new fleet-server/agent instance in fleet using a custom policy and default output with `--es-certificate-authorities=/etc/ssl/certs/ca-certificates`
- copied ca file to another location (`/tmp`)
- added `ssl.certificate_authorities: ["/tmp/ca-certificates.crt"]` to es output in kibana

the new `ssl.certificate_authorities` value did not appear in fleet-server.yml from any diagnostics i collected (before/after restarting the agent)

I also then tried uninstalling the agent, then re-installing without the `--es-certificate-authorities` flag and found that fleet-server's output includes
```
        tls:
            verification_mode: full
            certificate_authorities:
                - /tmp/ca-certificates.crt
            renegotiation: never
```
and the  ca path appears in fleet-server's  "initial server configuration" message.

The main finding is that the elastic-agent will never replace an attribute that was present during enrolment with a value from a policy when sending config to fleet-server.
As a result we still are going to have the existing edge cases around restarts

michel-laterman:
buildkite test this

internal/pkg/policy/self.go
```diff
@@ -187,6 +204,89 @@ func (m *selfMonitorT) groupByLatest(policies []model.Policy) map[string]model.P
...
+		Output: config.Output{
+			Elasticsearch: policyES,
+		},
+		RevisionIdx: m.lastRev,
+	}
+	return nil
+}
+
+// getFleetOutputName returns the output name that the fleet-server input of the policy uses
+func getFleetOutputName(p *model.Policy) (string, bool) {
```
michel-laterman, line 264: @juliaElastic I've added this method so we use the `use_output` attribute of the fleet-server input to get the output name directly instead of scanning for the 1st `elasticsearch` type output

lucabelluccini:
Hello!

- Would it be possible to cover/describe the behavior of the ES hosts for Fleet Server integration in the documentation especially if there is some subtle consequence on first enrollment?
- Can we make sure the EA diagnostic reflect the real hosts being used?

michel-laterman:
@cmacknz I created https://github.com/elastic/fleet-server/issues/3464 to handle output bootstrap tracking separately to limit how large this pr grows. Should I make another issue in fleet server to discuss/track my findings [above](https://github.com/elastic/fleet-server/pull/3411#issuecomment-2052638414)?
> The main finding is that the elastic-agent will never replace an attribute that was present during enrolment with a value from a policy when sending config to fleet-server.
> As a result we still are going to have the existing edge cases around restarts

@lucabelluccini I can confirm that the `fleet-server.yml` will be accurate in collected diagnostics.
I'll add a section that outlines expected behaviour for 8.14+ in our support docs
I'll also work with our docs team to get our public docs updated as well.

greicefaustino:
Hey team, in which version we will have this fixed? There's a tentative release version? Are there other dependencies to be merged?

cmacknz:
8.14.0 would be the first version with the fix.

lucabelluccini:
PLEASE NOTE: THIS FIX HAS BEEN REVERTED IN https://github.com/elastic/fleet-server/pull/3496

PR #3473, opened 2024-04-17
Update beats import to v7.17.20

### LINKED ISSUE #3435: Fail to bootstrap Fleet Server with self-signed certificates in 7.17
Fleet Server has started to fail to bootstrap with self-signed certificates in 7.17 branch. ~~It doesn't affect any released version at the moment.~~ It affects 7.17.20.

It fails with:
```
Error: fail to enroll: fail to execute request to fleet-server: remote error: tls: bad certificate
```
The very same certificates work with 7.17.19 and any other version tested, including 8.14.0-SNAPSHOT.

The only significant change since 7.17.19 is https://github.com/elastic/fleet-server/pull/3391, which updates beats dependency from 7.11.2 to 7.17.18.

One change included in Beats between these versions completely rewrites certificate validation: https://github.com/elastic/beats/pull/22495/files#diff-6225b0191feaa98542380b95e652c6f1f6805ee9a141330788011681cce0e487R153

There is a change in elastic-agent that solves an issue with certificate validation when bootstrapping, that was not backported to 7.x. Not sure if could be related: https://github.com/elastic/elastic-agent/pull/1867/files#diff-7efa04d4079650519d21a6e8ba217a7874130cecf8adc92dc2a78d4cfd2aee09R556

For confirmed bugs, please report:
- Version: 7.17 branch (doesn't affect 7.17.19)
- Steps to Reproduce:
  - Can be reproduced with `elastic-package stack up -v -d --version 7.17-SNAPSHOT`. That under the hood: 
    - Generates self-signed certificates.
    - Installs the CA system-wide in the containers (in `/etc/ssl/certs`).
    - Bootstraps Fleet Server using these certificates.

michel-laterman:

beats v7.12.0 introduces major changes to how TLS is handled and upgrading causes bootstapping to fail

Update beats import to v7.17.20 and fix how the api tls server is initiated to prevent https://github.com/elastic/fleet-server/issues/3435.

bootstrap fleet-server built from this pr with an elastic-agent + cluster running v7.17.21-SNAPSHOT

## Related issues

- Closes #3435


cmacknz:
What was the explanation for why this broke self-signed certificates? Why this fixes it still isn't clear to me even though the change is small.

juliaElastic (approved):
LGTM, could the changes be tested locally?

michel-laterman:
@juliaElastic, I've tested by running fleet-server under agent in a VM against a QA snapshot deployment, do you think that is sufficient or should I run with a locally managed instance of ES as well?

michel-laterman:
Yes,I tested and also got @kaanyalti to test locally as well

cmacknz (approved):

cmd/fleet/server.go
```diff
@@ -105,11 +105,11 @@ func runServer(ctx context.Context, router *httprouter.Router, cfg *config.Serve
 		ln = wrapConnLimitter(ctx, ln, cfg)
 
 		if cfg.TLS != nil && cfg.TLS.IsEnabled() {
-			commonTlsCfg, err := tlscommon.LoadTLSServerConfig(cfg.TLS)
+			commonTLSCfg, err := tlscommon.LoadTLSServerConfig(cfg.TLS)
 			if err != nil {
 				return err
 			}
-			server.TLSConfig = commonTlsCfg.ToConfig()
+			server.TLSConfig = commonTLSCfg.BuildServerConfig(cfg.Host)
```
cmacknz, line 112: https://github.com/elastic/fleet-server/blob/4e3c8b0e571e96921ab7c69c30ab8a857a4690d3/cmd/fleet/server.go#L106-L118

This is now the same as main so 👍 

```go
			commonTLSCfg, err := tlscommon.LoadTLSServerConfig(cfg.TLS)
			if err != nil {
				return err
			}
			server.TLSConfig = commonTLSCfg.BuildServerConfig(cfg.Host)

			// Must enable http/2 in the configuration explicitly.
			// (see https://golang.org/pkg/net/http/#Server.Serve)
			server.TLSConfig.NextProtos = []string{"h2", "http/1.1"}

			ln = tls.NewListener(ln, server.TLSConfig)
```

michel-laterman: `BuildServerConfig` function was introduced in beats 7.12.x when we updated the required version of go to 1.15.x
It specifies a custom `VerifyConnection` attribute that is need for the changes with certificate verification introduced in that version of go

Issue #3464, opened 2024-04-15
separate "bootstrap" from retrieved config when running in agent mode


michel-laterman:
Follow up for https://github.com/elastic/fleet-server/pull/3411

fleet-server has been altered so that in agent-mode the policy self-monitor can send partial config updates to change output settings based off the output section of the fleet-server policy.

However this can introduce an edge case when the server is running where a policy-change unrelated to the output will result in the "bootstrap" output sent from the elastic-agent being used temporarily.
The sequence of events may be:
- ES hosts are set to A, B
- fleet-server is enrolled with host A
  - policy monitor will send A, B to reload
- fleet-server policy is adjusted by user (i.e., max_agent has changed)
- elastic-agent will send updated policy with only host A (if the agent has not been restarted)
  - policy monitor will then sent A, B to reload again

The issue here is the final reload where the fleet-server because the reload operation will restart the API server.

When receiving new output fleet-server may need to determine how to insert/inject output (and the revision idx value) into the config so that the output from policy does not trigger a second reload.

cmacknz:
I think the root of our problems is that agent [overwrites](https://github.com/cmacknz/elastic-agent/blob/cb558c59af0b2555a1a22ce9865020c2393275fa/internal/pkg/agent/application/fleet_server_bootstrap.go#L67-L77) the outputs from the policy with the bootstrap output. This makes the bootstrap process less complex from Fleet Server's perspective.

We need agent to always pass the outputs from the policy and just augment them with the bootstrap output, because there is no guarantee that the initial bootstrap output remains valid over the complete life of fleet server since it is set only at install time.

We could include the bootstrap output as a second output unit separate from the ones in the agent policy. Agent would always inject it so that Fleet server always receives two output units. 

Another options if adding a second output unit is too complex is to create a `bootstrap:` sub-section of the existing output block that contains the bootstrap output. Something like:

```yaml
output:
  elasticsearch:
    hosts: '${ELASTICSEARCH_HOSTS:https://localhost:9200}'
    service_token: '${ELASTICSEARCH_SERVICE_TOKEN}'
    ssl.ca_trusted_fingerprint: '${ELASTICSEARCH_CA_TRUSTED_FINGERPRINT}'
    bootstrap:
      hosts: 'https://bootstrap:9200'
      service_token: '${ELASTICSEARCH_SERVICE_TOKEN}'
[...]

PR #3495, opened 2024-04-24
Revert "agent-mode loads output from policy"

michel-laterman:
Reverts elastic/fleet-server#3411

Due to ongoing issues with the current 8.14 BC we should revert this change and reimplement once we have better defined what fleet-server should expect as bootstrap configuration (see https://github.com/elastic/fleet-server/issues/3464)


cmacknz (approved):

nchaulet (approved):

michel-laterman:
@kilfoyle @lucabelluccini we're reverting the reload from output changes from main and the 8.14 release due to instability

lucabelluccini:
Thanks @michel-laterman  for the heads up - I'll take care of the Knowledge article. I've already pinged @greicefaustino & @Maxwellk-Gib who were tracking this.

kilfoyle:
>  we're reverting the reload from output changes from main and the 8.14 release due to instability

Thanks @michel-laterman! I don't recall working on any docs for this, so I think there's nothing to undo. Let me know if that doesn't sound right.

michel-laterman:
@kilfoyle https://github.com/elastic/ingest-docs/pull/1034 will need to be reverted

PR #3506, opened 2024-04-30
Agent mode support for bootstrap output

### LINKED ISSUE #3464: separate "bootstrap" from retrieved config when running in agent mode

michel-laterman:

When running under the elastic-agent fleet-server is not able to use output settings from it's policy.

This PR requires https://github.com/elastic/elastic-agent/pull/4643 to work.

The elastic-agent will inject enrollment configuration options in `output.elasticsearch.bootstrap` instead of overwriting matching keys in `output.elasticsearch`.
When running under agent mode, fleet-server will inject specific keys in `bootstrap` that are not in `output.elasticsearch`, then test the resulting output to see if it can connect to Elasticsearch, if so it is used. If not then `bootstrap` is used instead and the output is periodically retested in case the failure was caused by a temporary network issue.

Create an elastic-agent package from: https://github.com/elastic/elastic-agent/pull/4643
replace the fleet-server component using one generated from this pr.

If testing with docker images `docker.elastic.co/observability-ci/elastic-agent:8.15.0-SNAPSHOT-dd4c89e-1715633206` can be used as the `BASE_IMAGE` for generating a new image/deployment with the `make cloud-deploy` target in `dev-tools/cloud`. 
Or the `docker.elastic.co/observability-ci/elastic-agent:8.15.0-SNAPSHOT-laterman-1715705581` image can be used as it contains the changes from both PRs.

I've verified that the following behaviours work:

When deployed to ESS:
- fleet-server is available/healthy
- logging level for fleet-server can be changed with no issues
- diagnostics can be collected
- e2e cypress tests for defender succeed (thanks @tomsonpl!) 

When deployed locally:
- fleet-server is available/healthy
- logging level for fleet-server can be changed with no issues
- diagnostics can be collected
- multiple hosts can be added to Elasticasearch output (in Kibana) and show up in fleet-server.yml when a diagnostics bundle is collected

## Design Checklist

## Checklist

## Related issues

- Closes #3464 
- Relates https://github.com/elastic/elastic-agent/issues/2784


internal/pkg/server/agent.go
```diff
@@ -472,3 +491,59 @@ func apmConfigToInstrumentation(src *proto.APMConfig) (config.Instrumentation, e
 	}
 	return config.Instrumentation{}, fmt.Errorf("unable to transform APMConfig to instrumentation")
 }
+
+// injectMissingOutputAttributes will insert any kv pairs that are present in bootstrap but are not in outMap.
+// If the value of an existing key in outMap and bootstrap is map[string]interface{} the same injection is called recursively.
+func injectMissingOutputAttributes(outMap, bootstrap map[string]interface{}) {
```
cmacknz, line 497: Why do we need to do this?

michel-laterman: this will inject `service_token` and any tls or proxy settings into the output if they are not specified in the policy output

cmacknz: If we can fall back to the bootstrap output, do we actually need to do this? Can't those things be configured in the Fleet UI? Having the user enter them again is better than auto-magically updating the outputs with parameters users can't see.

If Fleet Server always has the outputs defined in the policy, and the bootstrap output it was installed with, isn't there always a working output available? Or is there something fleet server specific that we don't allow configuring in the Fleet UI?

I am just double checking that we are not adding complexity that we don't need.

michel-laterman: I think there's a bit of a misunderstanding here.

Bootstrap config/output is whatever enrolment flags are used when elastic-agent is set to start a fleet-server instance (es host, tls options, proxy options, service token).
While these values are initially populated using the policy output they are not updated in the agent after they are used.

Currently the elastic-agent will overwrite any keys in the output block before sending the config to fleet-server.
This causes the agent to pass the singular host it was enrolled with instead of the hosts list that may be in the real policy output.
The PR changes it so that all bootstrap options appear under a new key and do not overwrite any existing values.
so the output hosts would be the real policy's list, and bootstrap.hosts is the one that agent was enrolled with.

Injection is used primarily to ensure the service_token value gets to fleet-server, as the fleet-server policy never contains the service_token.

Currently the bootstrap config is inserted with every config expected output unit change that elastic-agent sends to fleet-server as it was the simplest to implement/test.

cmacknz: > Injection is used primarily to ensure the service_token value gets to fleet-server, as the fleet-server policy never contains the service_token.

Got it so we need to do this for some keys all of the time.

Can you reduce the scope of this function to inject only the things you know will always be missing? And log obviously that this has happened.

What I want us to avoid is writing a generic function that unconditionally injects any missing keys, regardless of if it is correct or desirable to do that. I am worried about us obsoleting a parameter, or having us inject a now invalid or expired TLS attribute, and breaking connectivity to ES inexplicibly.

I am a bit paranoid about generic configuration manipulation after watching the investigation that went into finding https://github.com/elastic/fleet-server/pull/3508 :)

cmacknz: > A bad output option is added or there's a config change immediately followed by a network issue and fleet-server fails the (es) output check

> at this point the bootstrap values with the expired CA is used causing fleet-server to fail until the file is replaced and the agent is restarted.

The point about a transient network issue causing the output check to fail is interesting. This makes me think the output check needs to be periodic and not a one time event.

> For the 1st issue, i think we can alter the injection to remove verification_mode: none from being injected if the policy output has CAs or a fingerprint specified.

This makes sense but it will still seem a bit magical from the user's perspective, so we need to clearly document the behavior.

Both of the cases you listed above need to become tests.

michel-laterman:
Testing progress: https://github.com/elastic/elastic-agent/pull/4643#issuecomment-2108891499

i'll now try to implement the areas of improvement above:
1) skip injecting `verification_mode: none` if a CA or CA fingerprint is in the retrieved policy.
2) async period output testing if the bootstap block has been passed

pchila (approved):
Just a couple of comments, looks good overall

cmacknz:
I think we need an automated test proving that Elastic Agent can bootstrap Fleet Server in one of the repositories before this or https://github.com/elastic/elastic-agent/pull/4643 are merged.

If the coordination of the two PRs with the test is annoying enough I'd be fine with the test being added a separate PR, but not closing the implementation issue until it exists.

michel-laterman:
buildkite test this

Issue #3490, opened 2024-04-24
[Fleet] Remote Elasticsearch output do not support custom ssl certificate


nchaulet:
## Description 

It seems we do not support remote Elasticsearch output with custom certificate in Fleet server. When creating a remote output with an ES using self signed certificate and providing the ssl configuration in the advanced settings, the output is mark as unhealthy and agent using that policy are not sending data to the new remote ES.

It seems we do not use the whole output configuration when creating the Elasticsearch client to connect to the remote Elastic Search in Fleet Server. This cause Fleet server being unable to verify the output health or to create API keys for other Elastic agent.

Relevant code path, looks like we create the remote output client only with the host and service token, this is probably what need to be fixed.  https://github.com/elastic/fleet-server/blob/main/internal/pkg/bulk/engine.go#L200

## Workaround

I do not see any workaround, and this probably should be fixed with a new version.

nchaulet:
I started to take a look, and I think here https://github.com/elastic/fleet-server/blob/f1340f39ff52c4175efe00c14e5d3edb0181388c/internal/pkg/bulk/engine.go#L201 we can create a `config.Elasticsearch` structure from the outputMap entry and then use the `ToESConfig`  method to create the client for the remote output @michel-laterman @juliaElastic does it make sense to you?

juliaElastic:
> we can create a `config.Elasticsearch` structure from the outputMap entry and then use the `ToESConfig` method to create the client for the remote output @michel-laterman @juliaElastic does it make sense to you?

Sounds good to me.

PR #3522, opened 2024-05-06
Fix remote ES client to use ssl.* settings

nchaulet:
## Description 

Resolve https://github.com/elastic/fleet-server/issues/3490

Fix remote ES client creation. the previous creation was only using the hosts and service token fields, making it impossible to create a client for a remote ES using self signed certificate. 

That PR fix that by creating a client with the correct config 

## Manual test

Create a remote ES to a self signed certificate and check the output is healthy and can generate API keys.

Screenshot 2024-05-06 at 11 47 14 AM
Screenshot 2024-05-06 at 11 47 11 AM

Screenshot 2024-05-06 at 12 20 21 PM

## Automated tests

I updated the integration test to run a remote ES with ssl enabled.

### DIFF
--- a/internal/pkg/bulk/engine.go
+++ b/internal/pkg/bulk/engine.go
@@ -208,6 +208,12 @@ func (b *Bulker) createRemoteEsClient(ctx context.Context, outputName string, ou
 	if err != nil {
 		return nil, err
 	}
+	if len(esOutput.Hosts) == 0 {
+		return nil, fmt.Errorf("failed to get hosts from output: %v", outputName)
+	}
+	if esOutput.ServiceToken == "" {
+		return nil, fmt.Errorf("failed to get service token from output: %v", outputName)
+	}
 	cfg := config.Config{
 		Output: config.Output{
 			Elasticsearch: esOutput,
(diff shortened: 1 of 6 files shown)


cmacknz:
It should be possible to add a test proving this was fixed.

nchaulet:
> It should be possible to add a test proving this was fixed.

Yes just updated the integration test to use a remote ES with ssl enabled.

internal/pkg/bulk/engine.go
```diff
@@ -198,29 +199,18 @@ func (b *Bulker) CreateAndGetBulker(ctx context.Context, zlog zerolog.Logger, ou
...
-		}
+	var esOutput config.Elasticsearch
+	esConfig, err := ucfg.NewFrom(outputMap[outputName], config.DefaultOptions...)
+	if err != nil {
+		return nil, err
 	}
-	serviceToken, ok := outputMap[outputName]["service_token"].(string)
-	if !ok {
-		return nil, fmt.Errorf("failed to get service token from output: %v", outputName)
+	err = esConfig.Unpack(&esOutput)
```
juliaElastic, line 207: should there be validation that the required fields are in the config, or is that done by the schema?

juliaElastic (approved):
Code LGTM

internal/pkg/bulk/engine.go
nchaulet: Added back the validation, the scema will only validate that data is correctly formed

PR #3535, opened 2024-05-10
Pass namespaces to Fleet Server managed documents

### LINKED ISSUE #3505: [Fleet] Support spaces in fleet object created by Fleet Server
Support spaces in fleet object created by Fleet Server

We are going to make Fleet space aware  => Fleet server need to populate the namespace accordingly tin the document it manage.

Pass space ID through from:
 `.fleet-enrollment-api-keys` to `.fleet-agents` documents.
 `.fleet-actions` to `.fleet-actions-results` documents.

The namespace property has already been added to all of the `.fleet*` mappings managed in the fleet 

## Details 

`.fleet-policies`

The namespace should be populated in Kibana base on the saved object namespace.

`.fleet-enrollment-api-keys`

The namespace should be populated in Kibana based on the related agent policy.

.fleet-agents

The namespace should be populated on enrolment, we should also update the value when assigning a new policy to the agent, this is done in Kibana. 
`.fleet-actions`
We should populate the namespace when creating the action in Kibana

` .fleet-actions-results`
We should populate the namespace when creating the action result it should be the same as the corresponding .fleet-actions.

## Open questions

Space will not work correctly if Fleet server is not updated, how do we communicate that? Cc @kpollich

nchaulet:
## Description 

Resolve #3505 

Set the namespaces property in the different `.fleet-*` document managed by fleet server.

Feature flag? do we need to release that feature under a feature flag, in my opinion if the namespaces are not populated in Kibana fleet server behavior do not change so it should be safe to release this without feature flag.

## Test

I added a new integration test to cover the change

## Todo

- [x] Pass namespace during enrollments
- [x] Pass namespace for actions
- [x] Pass namespace during file upload
- [x] Add integrations test for namespace

## Questions

- [ ] Policy change should we support changing namespace here ? @kpollich what do you think?


kpollich:
> Policy change should we support changing namespace here ? @kpollich what do you think?

We were initially planning to allow for changing an agent policy's space directly from Fleet UI, e.g.  

image

However, it's not clear if this is the direction we'll go. We may direct users to stack management instead, to avoid having multiple sources of truth for object/space management. When the space of an agent policy changes, though, we probably do need to dispatch an action to update the `namespaces` property on its relevant policy components. I wonder if we could register a callback with stack management that detects when a policy is moved to another space so we can fire off the policy change operation. We might be able to also handle this in a background task that listens for a certain saved object save event - if that's a thing.

michel-laterman (approved):
Looks good so far.
It's good that namespace is kept as an ES document detail and not added to the API that clients use.

Please add a changelog fragment as well

nchaulet:
> We were initially planning to allow for changing an agent policy's space directly from Fleet UI, e.g.

I think we probably can handle the whole namespace change directly from Kibana has we do when reassigning to a policy by updating the `.fleet-agents` doc

kaanyalti (approved):

PR #3587, opened 2024-05-29
WIP adding TLS diagnostics from elastic-agent-libs

michel-laterman:

TLS information provided by fleet-server in diagnostics bundles is lacking.

Add custom hooks to use diag hooks added in https://github.com/elastic/elastic-agent-libs/pull/207 to provide additional files that contain information about the TLS certs used by the server's API, TLS infomation used when connecting to elasticsearch, and a full trace to each specified elasticsearch host.

For example, when enrolling into a cloud deployment with no explicit cert/key for fleet-server, the new files the hooks add are :

fleet-server-output-request.txt
```
2024/05/29 00:18:41.571489 No TLS settings
2024/05/29 00:18:41.571492 Proxy disable=false url=<nil>
2024/05/29 00:18:41.571541 Request 0 to https://1df578882a2d4d0190aabbbe4ac77268.us-central1.gcp.qa.cld.elstc.co:443 starting
2024/05/29 00:18:41.571576 GetConn called for "1df578882a2d4d0190aabbbe4ac77268.us-central1.gcp.qa.cld.elstc.co:443"
2024/05/29 00:18:41.710378 Connection started to tcp:34.71.9.202:443
2024/05/29 00:18:41.834875 Connection to tcp:34.71.9.202:443 done, err: <nil>
2024/05/29 00:18:41.984139 TLS handshake starting
2024/05/29 00:18:41.984148 TLS handshake done. state={Version:772 HandshakeComplete:true DidResume:false CipherSuite:4865 NegotiatedProtocol: NegotiatedProtocolIsMutual:true ServerName:1df578882a2d4d0190aabbbe4ac77268.us-central1.gcp.qa.cld.elstc.co PeerCertificates:[0xc00021b600 0xc00021bb80 0xc00021c100] VerifiedChains:[] SignedCertificateTimestamps:[] OCSPResponse:[] TLSUnique:[] ekm:0x5557d33c1f20} err=<nil>
2024/05/29 00:18:41.984267 GotConn for "34.71.9.202:443"
2024/05/29 00:18:41.984371 Wrote request headers
2024/05/29 00:18:41.984385 Wrote request err=<nil>
2024/05/29 00:18:42.111030 Response started
2024/05/29 00:18:42.111184 request 0 successful.
```
fleet-server-api-tls.txt
```
tlscommon.ServerConfig: Start diagnostics 2024-05-29 00:18:41.569890423 +0000 UTC
tlscommon.ServerConfig: verification_mode=full
tlscommon.ServerConfig: client_auth=none
tlscommon.ServerConfig: ca_sha256=[]
tlscommon.ServerConfig: CertificateSettings: checking certificate keypair
tlscommon.ServerConfig: CertificateSettings: certificate keypair OK.
tlscommon.ServerConfig: CertificateSettings: cert 0 - Subject: CN=fleet-server-dev,O=elastic-fleet
	Issuer: CN=localhost,O=elastic-fleet
	NotBefore: 2024-05-29 00:18:15 +0000 UTC
	NotAfter: 2034-05-29 00:18:15 +0000 UTC
	Fingerprint: EvAr8erNJZJ0BoXor1jnCl5X2WCU79ISMgwMITOEJBs=
	SAN IP: []
	SAN DNS: [fleet-server-dev]
tlscommon.ServerConfig: CertificateAuthorities: certificate_authorities provided.
tlscommon.ServerConfig: CertificateAuthorities: - cert 0 Subject: CN=localhost,O=elastic-fleet
	IsCa: true
	BasicConstraintsValid: true
	NotBefore: 2024-05-29 00:18:15 +0000 UTC
	NotAfter: 2034-05-29 00:18:15 +0000 UTC
	Fingerprint: cSawoNEa2cSPV4Mwce4mRysrcBGtHBGcSeoBL+Cr48Q=
```

fleet-server-output-tls.txt
```
error: nil tlscommon.Config
```

## Design Checklist

## Checklist


michel-laterman:
@lucabelluccini, this is an example of the hooks in use; do you think we should add one for apm instrumentation so fleet-server would produce another request-trace?

changelog/fragments/1718034684-Add-tlscommon-and-httpcommon-hooks.yaml
```diff
@@ -0,0 +1,36 @@
...
+# - breaking-change: a change to previously-documented behavior
+# - deprecation: functionality that is being removed in a later release
+# - bug-fix: fixes a problem in a previous version
+# - enhancement: extends functionality but does not break or fix existing behavior
+# - feature: new functionality
+# - known-issue: problems that we are aware of in a given version
+# - security: impacts on the security of a product or a user’s deployment.
+# - upgrade: important information for someone upgrading from a prior version
+# - other: does not fit into any of the other categories
+kind: feature
```
blakerouse, line 11:
```suggestion
kind: enhancement
```

changelog/fragments/1718034684-Add-tlscommon-and-httpcommon-hooks.yaml
```diff
@@ -0,0 +1,36 @@
...
+
+# Change summary; a 80ish characters long description of the change.
+summary: Add tlscommon and httpcommon hooks
```
blakerouse, line 14:
```suggestion
summary: Add tlscommon and httpcommon diagnostic information
```

internal/pkg/config/output.go
```diff
@@ -228,3 +231,45 @@ func makeURL(defaultScheme string, defaultPath string, rawURL string, defaultPor
...
+		headers.Set(k, v)
+	}
+
+	reqs := make([]*http.Request, 0, len(c.Hosts))
+	ctx, cancel := context.WithCancel(ctx)
+	defer cancel()
+	for _, host := range c.Hosts {
+		u, err := url.Parse(host)
+		if err != nil {
+			zerolog.Ctx(ctx).Warn().Err(err).Str("host", host).Msg("Unable to transform host to url.URL")
```
blakerouse, line 260: Would be much better to provide these errors in the diagnostic bundle, so its clear just reading that one file.

internal/pkg/config/output.go
```diff
@@ -228,3 +231,45 @@ func makeURL(defaultScheme string, defaultPath string, rawURL string, defaultPor
...
+			continue
+		}
+		if u.Scheme == "" {
+			u.Scheme = c.Protocol
+		}
+		req, err := http.NewRequestWithContext(ctx, http.MethodGet, u.String(), nil)
+		if err != nil {
+			zerolog.Ctx(ctx).Warn().Err(err).Str("host", host).Msg("Unable to create request to host")
```
blakerouse, line 268: Same as above.

internal/pkg/server/agent.go
```diff
@@ -115,6 +115,42 @@ func (a *Agent) Run(ctx context.Context) error {
 		}
 		return p
 	})
+	a.agent.RegisterDiagnosticHook("fleet-server api tls diag", "fleet-server's API TLS config", "fleet-server-api-tls.txt", "text/plain", func() []byte {
+		if a.srv == nil {
+			log.Warn().Msg("Diagnostics hook failure fleet-server is nil.")
```
blakerouse, line 120: If the content of the file is plain, why not just add this message into the file output.

internal/pkg/server/agent.go
```diff
@@ -115,6 +115,42 @@ func (a *Agent) Run(ctx context.Context) error {
...
+			return nil
+		}
+		cfg := a.srv.GetConfig()
+		if cfg == nil || len(cfg.Inputs) == 0 {
+			log.Warn().Msg("Diagnostics hook failure config is nil.")
```
blakerouse, line 125: Same as above.

internal/pkg/server/agent.go
```diff
@@ -115,6 +115,42 @@ func (a *Agent) Run(ctx context.Context) error {
...
+			return nil
+		}
+		return cfg.Inputs[0].Server.TLS.DiagCerts()()
+	})
+	a.agent.RegisterDiagnosticHook("fleet-server output tls diag", "fleet-server's output TLS config", "fleet-server-output-tls.txt", "text/plain", func() []byte {
+		if a.srv == nil {
+			log.Warn().Msg("Diagnostics hook failure fleet-server is nil.")
```
blakerouse, line 132: Same as above.

internal/pkg/server/agent.go
```diff
@@ -115,6 +115,42 @@ func (a *Agent) Run(ctx context.Context) error {
...
+			return nil
+		}
+		cfg := a.srv.GetConfig()
+		if cfg == nil {
+			log.Warn().Msg("Diagnostics hook failure config is nil.")
```
blakerouse, line 137: Same as above.

internal/pkg/server/agent.go
```diff
@@ -115,6 +115,42 @@ func (a *Agent) Run(ctx context.Context) error {
...
+			return nil
+		}
+		return cfg.Output.Elasticsearch.TLS.DiagCerts()()
+	})
+	a.agent.RegisterDiagnosticHook("fleet-server output request diag", "fleet-server output request trace diagnostics", "fleet-server-output-request.txt", "text/plain", func() []byte {
+		if a.srv == nil {
+			log.Warn().Msg("Diagnostics hook failure fleet-server is nil.")
```
blakerouse, line 144: Same as above.

internal/pkg/server/agent.go
```diff
@@ -115,6 +115,42 @@ func (a *Agent) Run(ctx context.Context) error {
...
+			return nil
+		}
+		cfg := a.srv.GetConfig()
+		if cfg == nil {
+			log.Warn().Msg("Diagnostics hook failure config is nil.")
```
blakerouse, line 149: Same as above.

blakerouse (changes requested):

internal/pkg/config/output.go
```diff
@@ -228,3 +232,52 @@ func makeURL(defaultScheme string, defaultPath string, rawURL string, defaultPor
...
+		u, err := url.Parse(host)
+		if err != nil {
+			zerolog.Ctx(ctx).Warn().Err(err).Str("host", host).Msg("Unable to transform host to url.URL")
+			res.WriteString(fmt.Sprintf("Unable to transform host %q to url.URL: %v\n", host, err))
+			continue
+		}
+		if u.Scheme == "" {
+			u.Scheme = c.Protocol
+		}
+		ctx, cancel := context.WithTimeout(ctx, time.Second*30) // TODO(michel-laterman): Configurable timeout; should this be part of config, or part of a diagnostics action?
```
michel-laterman, line 270: As per the discussion in https://github.com/elastic/elastic-agent/issues/4880 i'm adding a timeout to each request so the (diagnostics) action does not indefinitely block.
We can add a `duration` value to the fleet-server action definition in `openapi.yml`
@blakerouse, What do you think about adding a duration attribute to the action definition in elastic-agent-client spec (where we can also use it for the optional CPU trace)?

model/openapi.yml
```diff
@@ -557,6 +557,7 @@ components:
             type: string
             enum:
               - CPU
+              - CONN
```
michel-laterman, line 560: From the discussion in https://github.com/elastic/elastic-agent/issues/4880, we want to have the HTTP connection request diagnostics as an optional value that is enabled by default.

internal/pkg/config/output.go
```diff
@@ -228,3 +232,50 @@ func makeURL(defaultScheme string, defaultPath string, rawURL string, defaultPor
...
+			Headers: httpcommon.ProxyHeaders(c.ProxyHeaders),
+		},
+	}
+	headers := http.Header{}
+	for k, v := range c.Headers {
+		headers.Set(k, v)
+	}
+
+	reqs := make([]*http.Request, 0, len(c.Hosts))
+	ctx, cancel := context.WithTimeout(ctx, time.Second*30) // TODO(michel-laterman): Configurable timeout; should this be part of config, or part of a diagnostics action?
```
blakerouse, line 256: Good question here. Does the `DiagRequest` already send a context with a deadline?

internal/pkg/server/agent.go
```diff
@@ -115,6 +115,42 @@ func (a *Agent) Run(ctx context.Context) error {
...
+		if a.srv == nil {
+			log.Warn().Msg("Diagnostics hook failure fleet-server is nil.")
+			return []byte(`Diagnostics hook failure fleet-server is nil`)
+		}
+		cfg := a.srv.GetConfig()
+		if cfg == nil {
+			log.Warn().Msg("Diagnostics hook failure config is nil.")
+			return []byte(`Diagnostics hook failure config is nil`)
+		}
+		return cfg.Output.Elasticsearch.DiagRequests(ctx)
```
blakerouse, line 152: Where is this `ctx` coming from? Doesn't seem to be inside of this hook call. If the context is outside of the request hook that seems like that could be incorrect.

michel-laterman: This context is the "root" context for fleet-server running agent.
It can be cancelled when the fleet-server is shutdown (by signal); the request-hook will abort in this case

internal/pkg/config/output.go
michel-laterman: Nope, the parent hook is cancelable in case the server needs to shut down

cmacknz: The timeout makes the most sense as part of the diagnostics action, that would lead to us exposing it in Fleet when the remote diagnostics request is sent, and not in a totally separate part of the UI users are unlikely to see.

michel-laterman: I've made https://github.com/elastic/fleet-server/issues/3648 and https://github.com/elastic/elastic-agent-client/issues/113 to track the work needed

internal/pkg/server/agent.go
blakerouse: This should add a timeout to it then, so use that context but add a timeout. We should ensure that there is atleast a maximum amount of time that the `DiagRequests` will take.

internal/pkg/config/output.go
blakerouse: Until that happens a sensible default should be okay.

michel-laterman:
buildkite test this

jlind23:
@michel-laterman let me know once this is ready to be merged and i'll bypass the sonarcloud check.

PR #3611, opened 2024-06-06
[Fleet] Replace findByApiKeyID with a get call when possible

nchaulet:
## Description 

Related to https://github.com/elastic/ingest-dev/issues/3274

Replace findByApiKeyID with a get document call when possible,

@michel-laterman before I wrote more tests for that I am curious to get your though on that and if you see any potential issues


internal/pkg/api/handleCheckin.go
```diff
@@ -855,6 +855,27 @@ func processPolicy(ctx context.Context, zlog zerolog.Logger, bulker bulk.Bulk, a
...
 }
 
+func getAgentAndVerifyAPIKeyID(ctx context.Context, bulker bulk.Bulk, agentID string, apiKeyID string) (*model.Agent, error) {
+	span, ctx := apm.StartSpan(ctx, "getAgentAndVerifyAPIKeyID", "read")
+	defer span.End()
+	agent, err := dl.GetAgent(ctx, bulker, agentID)
+	if err != nil {
+		if errors.Is(err, dl.ErrNotFound) {
+			err = ErrAgentNotFound
+		} else if errors.Is(err, dl.ErrNotFound) {
```
michel-laterman, line 865: Did you mean do have the same error twice?

nchaulet: No that's a typo :)

nchaulet:
buildkite run perf-tests

internal/pkg/api/auth.go
```diff
@@ -116,7 +116,12 @@ func authAgent(r *http.Request, id *string, bulker bulk.Bulk, c cache.Cache) (*m
 			Msg("authApiKey slow")
 	}
 
-	agent, err := findAgentByAPIKeyID(ctx, bulker, key.ID)
+	var agent *model.Agent
+	if id != nil {
+		agent, err = getAgentAndVerifyAPIKeyID(ctx, bulker, *id, key.ID)
+	} else {
+		agent, err = findAgentByAPIKeyID(ctx, bulker, key.ID)
+	}
```
michel-laterman, line 124: The basic approach should work, we have a 1-1 map of agents and keys because we update the keys instead of revoking+reissuing when there are changes.
We actually do the API key check below (line 151) so this method may need to be refactored to avoid redundant checks

michel-laterman (approved):

internal/pkg/api/auth.go
nchaulet: > We actually do the API key check below (line 151) so this method may need to be refactored to avoid redundant checks

I think we probably still want to verify api key id in a different way depending of the entry point it do not have the same meaning in one case the doc is probably corrupted in the other it's probably someone impersonating an agent.

internal/pkg/api/auth.go
```diff
@@ -116,7 +116,12 @@ func authAgent(r *http.Request, id *string, bulker bulk.Bulk, c cache.Cache) (*m
 			Msg("authApiKey slow")
 	}
 
-	agent, err := findAgentByAPIKeyID(ctx, bulker, key.ID)
+	var agent *model.Agent
+	if id != nil {
```
cmacknz, line 120: This block needs some comments explaining why we why do this.

PR #3677, opened 2024-07-03
[Fleet] Support download rate in string with unit format

nchaulet:
## Description 

Relates https://github.com/elastic/ingest-dev/issues/3446

Elastic-agent are currently not sending download rate as a float but as a string with unit, while this is fixed on the agent side and for older agent, Fleet server should accept both format.

That PR does that, and return a 400 instead of a 500 when an error parsing meta happen.

## Todo

- [x] Changelog

### DIFF
--- a/changelog/fragments/1720021515-fix_download_rate.yaml
+++ b/changelog/fragments/1720021515-fix_download_rate.yaml
@@ -1,0 +1,31 @@
+# Kind can be one of:
+# - breaking-change: a change to previously-documented behavior
+# - deprecation: functionality that is being removed in a later release
+# - bug-fix: fixes a problem in a previous version
+# - enhancement: extends functionality but does not break or fix existing behavior
+# - feature: new functionality
+# - known-issue: problems that we are aware of in a given version
+# - security: impacts on the security of a product or a user’s deployment.
+# - upgrade: important information for someone upgrading from a prior version
+# - other: does not fit into any of the other categories
+kind: feature
+
+# Change summary; a 80ish characters long description of the change.
+summary: fix_download_rate
+
+# Long description; in case the summary is not enough to describe the change
+# this field accommodate a description without length limits.
+# NOTE: This field will be rendered only for breaking-change and known-issue kinds at the moment.
+description: Fix checkin API to API to support download_rate as string with unit per second (example 123MBps)
+
+# Affected component; usually one of "elastic-agent", "fleet-server", "filebeat", "metricbeat", "auditbeat", "all", etc.
+component: fleet-server
+
+# PR URL; optional; the PR number that added the changeset.
+# If not present is automatically filled by the tooling finding the PR where this changelog fragment has been added.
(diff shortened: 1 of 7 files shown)


internal/pkg/api/metadata.go
```diff
@@ -0,0 +1,81 @@
...
+
+import (
+	"encoding/json"
+	"fmt"
+	"strings"
+
+	"github.com/docker/go-units"
+)
+
+func parseDownloadRate(jsonDownloadRate json.RawMessage) (*float64, error) {
```
michel-laterman, line 15: We should adjust the openapi spec for the download rate to be:
```yaml
oneOf:
  - type: string
  - type: integer
```

michel-laterman (changes requested):

internal/pkg/api/metadata.go
nchaulet: I am wondering if we should change the openAPISpec, as this mostly a temporary fix to support a client that did not sent the correct value.

cmacknz: Don't make the error an officially accepted value. Tolerate it but don't encourage it's use.

nchaulet: Yes I will add one with a random string input already had one with `MBps`

changelog/fragments/1720021515-fix_download_rate.yaml
```diff
@@ -0,0 +1,31 @@
...
+# - breaking-change: a change to previously-documented behavior
+# - deprecation: functionality that is being removed in a later release
+# - bug-fix: fixes a problem in a previous version
+# - enhancement: extends functionality but does not break or fix existing behavior
+# - feature: new functionality
+# - known-issue: problems that we are aware of in a given version
+# - security: impacts on the security of a product or a user’s deployment.
+# - upgrade: important information for someone upgrading from a prior version
+# - other: does not fit into any of the other categories
+kind: feature
```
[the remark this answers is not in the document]
nchaulet: bug fix 👍

nchaulet:
@michel-laterman I think it will make sense to backport this 8.15 no?

internal/pkg/api/metadata_test.go
```diff
@@ -0,0 +1,52 @@
...
+		name:          "download rate as float",
+		raw:           json.RawMessage(`1000000`),
+		expectedValue: 1000000.00,
+	}, {
+		name:          "download rate as MBps",
+		raw:           json.RawMessage(`"1MBps"`),
+		expectedValue: 1000000.00,
+	}, {
+		name:           "download rate random string",
+		raw:            json.RawMessage(`"toto"`),
```
michel-laterman, line 33: My suggestion for an additional test case would be to include just the suffix that is checked:
```json.RawMessage(`"MBps"`)``` as a test case

michel-laterman (approved):
Please backport and add a an additional test case

PR #3818, opened 2024-08-13
Add `/api/fleet/agents/:id/audit/unenroll` endpoint

michel-laterman:

Uninstalled agents appear as offline in the fleet UI.

Add `/api/fleet/agents/:id/audit/unenroll` API that an elastic-agent or Endpoint process may use to annotate the agent document so the agent may appear with a different status.

### Changes from RFC

[RFC here](https://docs.google.com/document/d/1gYbsGfvjc7NhbURwYNqEl25ouar81nZ_8bkpsi0Dc6Y/edit#heading=h.pfm1mse74cxb)

Changes are:
- 409 responses will include a JSON body to be consistent with other errors
- Successfull requests will alter the following agent document attributes:
  - `audit_unenrolled_time` - time from request
  - `audit_unenrolled_reason` - reason from request
  - `unenrolled_at` - server current time UTC
  - `updated_at` - server current time UTC (new)
  - ~~`active` - set to `false` (new)~~ removed change, was causing the api key auth to fail breaking integration and e2e tests
 
`make test-e2e` will run e2e tests as we need to change the elastic-agent or other test tools in order to use the endpoint

## Design Checklist

## Checklist

## Related issues

- Relates https://github.com/elastic/elastic-agent/issues/484

### DIFF
--- a/internal/pkg/api/error.go
+++ b/internal/pkg/api/error.go
@@ -481,6 +481,16 @@ func NewHTTPErrResp(err error) HTTPErrResp {
 				zerolog.InfoLevel,
 			},
 		},
+		// audit unenroll
+		{
+			ErrAuditUnenrollReason,
+			HTTPErrResp{
+				http.StatusConflict,
+				"ErrAuditReasonConflict",
+				"agent document contaiins audit_unenroll_reason",
+				zerolog.InfoLevel,
+			},
+		},
 	}
 
 	for _, e := range errTable {
--- a/internal/pkg/api/error.go
+++ b/internal/pkg/api/error.go
@@ -487,7 +487,7 @@ func NewHTTPErrResp(err error) HTTPErrResp {
 			HTTPErrResp{
 				http.StatusConflict,
 				"ErrAuditReasonConflict",
-				"agent document contaiins audit_unenroll_reason",
+				"agent document contains audit_unenroll_reason",
 				zerolog.InfoLevel,
 			},
 		},
--- a/internal/pkg/api/handleAudit.go
+++ b/internal/pkg/api/handleAudit.go
@@ -5,14 +5,24 @@
 package api
 
 import (
-	"errors"
+	"context"
+	"encoding/json"
+	"fmt"
 	"net/http"
+	"time"
 
 	"github.com/elastic/fleet-server/v7/internal/pkg/bulk"
 	"github.com/elastic/fleet-server/v7/internal/pkg/cache"
 	"github.com/elastic/fleet-server/v7/internal/pkg/config"
+	"github.com/elastic/fleet-server/v7/internal/pkg/dl"
+	"github.com/elastic/fleet-server/v7/internal/pkg/model"
+
+	"github.com/miolini/datacounter"
 	"github.com/rs/zerolog"
+	"go.elastic.co/apm/v2"
 )
+
+var ErrAuditUnenrollReason = fmt.Errorf("agent document contains audit_unenroll_reason attribute")
 
 type AuditT struct {
 	cfg   *config.Server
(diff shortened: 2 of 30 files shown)


internal/pkg/config/defaults/lte10000_limits.yml
```diff
@@ -53,3 +53,7 @@ server_limits:
     interval: 1ms
     burst: 2000
     max: 4000
+  audit_unenroll_limit:
```
michel-laterman, line 56: The default/env limits for the audit/unenroll endpoint are set to the same values as the ack endpoint (excluding body which is set to 1 kb)

michel-laterman (commented):
I think I'll also need to make a change to the ES templates in order to add `audit_unenrolled_time` and `audit_unenrolled_reason` to the agent documents.

I'll also make a followup PR to change the behaviour of the checkin API to clear the attributes that are set by the audit/unenroll API to allow Endpoint to mark itself as "orphaned" temporarily.

internal/pkg/api/handleAudit.go
```diff
@@ -0,0 +1,122 @@
...
+		dl.FieldAuditUnenrolledTime:   req.Timestamp,
+		dl.FieldAuditUnenrolledReason: req.Reason,
+	}
+	body, err := doc.Marshal()
+	if err != nil {
+		return fmt.Errorf("auditUnenroll marshal: %w", err)
+	}
+
+	if err := audit.bulk.Update(ctx, dl.FleetAgents, agent.Id, body, bulk.WithRefresh(), bulk.WithRetryOnConflict(3)); err != nil {
+		return fmt.Errorf("auditUnenroll update: %w", err)
```
blakerouse, line 117: Seems this will return a 500 error back to the caller, is that the correct error in the case its a conflict? Should it instead be the same conflict error in the case the field is set? I don't know if we want the caller to error.

internal/pkg/api/server_test.go
```diff
@@ -349,7 +349,7 @@ key: %s`,
 		cfg.TLS = tlsCFG
 
 		st := NewStatusT(cfg, nil, nil)
-		srv := NewServer(addr, cfg, nil, nil, nil, nil, st, sm, fbuild.Info{}, nil, nil, nil, nil, nil)
+		srv := NewServer(addr, cfg, nil, nil, nil, nil, st, sm, fbuild.Info{}, nil, nil, nil, nil, nil, nil)
```
blakerouse, line 352: This has become a ton of parameters will most being `nil`. Switching to a struct as a parameter would be better and cleaner. (don't do it in this PR, but an overall improvement)

internal/pkg/config/defaults/lte10000_limits.yml
blakerouse: +1

internal/pkg/server/fleet.go
```diff
@@ -540,9 +540,10 @@ func (f *Fleet) runSubsystems(ctx context.Context, cfg *config.Config, g *errgro
 	ut := api.NewUploadT(&cfg.Inputs[0].Server, bulker, monCli, f.cache) // uses no-retry client for bufferless chunk upload
 	ft := api.NewFileDeliveryT(&cfg.Inputs[0].Server, bulker, monCli, f.cache)
 	pt := api.NewPGPRetrieverT(&cfg.Inputs[0].Server, bulker, f.cache)
+	auditT := api.NewAuditT(&cfg.Inputs[0].Server, bulker, f.cache)
 
 	for _, endpoint := range (&cfg.Inputs[0].Server).BindEndpoints() {
-		apiServer := api.NewServer(endpoint, &cfg.Inputs[0].Server, ct, et, at, ack, st, sm, f.bi, ut, ft, pt, bulker, tracer)
+		apiServer := api.NewServer(endpoint, &cfg.Inputs[0].Server, ct, et, at, ack, st, sm, f.bi, ut, ft, pt, auditT, bulker, tracer)
```
blakerouse, line 546: struct would be much better, this is a lot...

blakerouse (approved):
Well done!

Just a few comments, no blockers.

pchila (approved):
Just a typo and a question about an error message (non-blocking as it can also be fixed in a follow-up PR), looks good otherwise...

+1 on refactoring long lists of parameters into structs though... in those places the code is not very readable :sweat_smile:

internal/pkg/api/server_test.go
michel-laterman: Created https://github.com/elastic/fleet-server/issues/3823 to track

internal/pkg/api/handleAudit.go
michel-laterman: It's detected much earlier on line 54 of `handleAudit.go`
https://github.com/elastic/fleet-server/pull/3818/files/#diff-bca31366226def7a73d6e6b142a9e17bbb4c2375a8f0fa16c5c81d4e6a374f13R54-R56

internal/pkg/api/error.go
```diff
@@ -481,6 +481,16 @@ func NewHTTPErrResp(err error) HTTPErrResp {
 				zerolog.InfoLevel,
 			},
 		},
+		// audit unenroll
+		{
+			ErrAuditUnenrollReason,
+			HTTPErrResp{
+				http.StatusConflict,
+				"ErrAuditReasonConflict",
+				"agent document contaiins audit_unenroll_reason",
```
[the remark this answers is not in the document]
michel-laterman: yes; though i may remove this as a 409 response as @intxgo has suggested that it may be useful to allow Endpoint to keep pinging the API so the `updated_at` timestamp gets updated

PR #3827, opened 2024-08-20
Alter checkin API to remove attributes set by audit/unenroll API

michel-laterman:

Follow up to https://github.com/elastic/fleet-server/pull/3818, alters the checkin API to remove attributes that the audit/unenroll endpoint sets

Alter the checkin wrapper's flush method to include bulk update calls to remove the attributes for agents that have a non-empty `audit_unenrolled_reason`.

1. Enroll an agent
2. Use the `audit/unenroll` API to annotate the agent document.
3. Use the `checkin` api to signal the agent is healthy

Agent document in Elasticsearch should no longer have the `audit_unenrolled_reason` or associated attributes

## Design Checklist

## Checklist

## Related issues

- Relates https://github.com/elastic/elastic-agent/issues/484


michel-laterman:
buildkite test this

dev-tools/integration/docker-compose.yml
michel-laterman, line 1: My laptop failed to run the integration tests due to something unrelated to these changes, however I think that including these changes (having all ES config specified in env vars) makes it easier to determine how the elasticsearch containers are configured

michel-laterman: I'll move this to another PR

internal/pkg/checkin/bulk.go
```diff
@@ -167,7 +176,8 @@ func (bc *Bulk) flush(ctx context.Context) error {
 		return nil
 	}
 
-	updates := make([]bulk.MultiOp, 0, len(pending))
+	// make cap pending*2 in case all checked in agents have audit/unenroll attributes that must be removed.
+	updates := make([]bulk.MultiOp, 0, len(pending)*2)
```
blakerouse, line 180: What is the odds that all of them have it? Could making this larger initial be taking more memory then really will be needed? I know in worst case it will take that much memory, but in a normal case I don't think this should be raised.

internal/pkg/checkin/bulk.go
```diff
@@ -240,6 +249,15 @@ func (bc *Bulk) flush(ctx context.Context) error {
 			if body, err = fields.Marshal(); err != nil {
 				return err
 			}
+
+			// If deleteAudit is set on the agent make sure the script is part of the request.
+			if pendingData.extra.deleteAudit {
+				updates = append(updates, bulk.MultiOp{
+					ID:    id,
+					Body:  deleteAuditAttributesScript,
+					Index: dl.FleetAgents,
```
blakerouse, line 258: This is running the script for each agent individually? Would it be possible to do this in a single script?

michel-laterman: Yes, it runs the script for each agent ID in the batch.
My original approach had the script in an `update_by_query` request, but it kept getting version conflict errors (and there does not appear to be any retry on conflict option).

I can see if the update action is able to take both a `script` and `doc` as the payload, or we can change it so that the other attributes are set as part of the script so only one thing is ran (per agent)

michel-laterman: You can't provide both a script and doc for bulk update actions.

blakerouse: I see. Looks like you updated it to just use a script to do both at the same time in the case those attribute needs to be removed. Looks good.

blakerouse (approved):

michel-laterman:
buildkite run perf-tests

michel-laterman:
buildkite run perf-tests

michel-laterman:
I realized that the perf-tests here don't actually use the changed endpoint; i'll work on altering them to do so in other prs

PR #3857, opened 2024-09-01
Bump version to 9.0.0

pierrehilbert:

## Design Checklist

## Checklist

### DIFF
--- a/version/version.go
+++ b/version/version.go
@@ -6,4 +6,4 @@ package version
 
 // DefaultVersion is the current release version of Fleet-server, this version must match the
 // Elastic Agent version.
-const DefaultVersion = "8.16.0"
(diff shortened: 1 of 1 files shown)


ycombinator:
@michel-laterman could you please help figure out why E2E tests are failing on this PR. The error looks like this:

```
lstat /var/lib/docker/tmp/buildkit-mount2666411982/cover/fleet-server-8.16.0-SNAPSHOT-linux-x86_64: no such file or directory
```

But I'm not sure why that path doesn't exist.

jlind23 (approved):

jlind23:
@ycombinator E2E tests will fail until we have a 9.0 unified release build so we are safe to forge merge this.

PR #3878, opened 2024-09-09
mergify: support for backport-major.minor

v1v:

We have agreed to use `backport-<major>.<minor>` GitHub labels rather than `backport-<major>.<minor>.0` so the same GH label strategy works for the ingest and observability projects.

I decided to keep the current minor versions with the support for both labels.

## Design Checklist

## Checklist

### DIFF
--- a/.mergify.yml
+++ b/.mergify.yml
@@ -50,7 +50,7 @@ pull_request_rules:
     conditions:
       - merged
       - base=main
-      - label ~= ^(backport-v7.17.0|backport-7.17)$
+      - label~=^(backport-v7.17.0|backport-7.17)$
     actions:
       backport:
         assignees:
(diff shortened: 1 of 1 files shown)


.mergify.yml
```diff
@@ -50,7 +50,7 @@ pull_request_rules:
     conditions:
       - merged
       - base=main
-      - label=backport-v7.17.0
+      - label ~= ^(backport-v7.17.0|backport-7.17)$
```
v1v, line 53:
```suggestion
      - label~=^(backport-v7.17.0|backport-7.17)$
```

.mergify.yml
```diff
@@ -236,7 +236,7 @@ pull_request_rules:
   - name: backport patches to 8.15 branch
     conditions:
       - merged
-      - label=backport-v8.15.0
+      - label ~= ^(backport-v8.15.0|backport-8.15)$
```
v1v, line 239:
```suggestion
      - label~=^(backport-v8.15.0|backport-8.15)$
```

jlind23:
buildkite test it

v1v:
Failures are unrelated to this PR. This is about the Mergify file, there is nothing about testing the changes in the CI

jlind23:
Force merging this as it is failing due to the lack of 9.0 snapshot version available.

PR #3889, opened 2024-09-10
Fix version source when making e2e agent image

### LINKED ISSUE #3886: Fix version reference in Dockerfile for E2E tests
In the [Dockerfile used for E2E tests](https://github.com/elastic/fleet-server/blob/main/dev-tools/e2e/Dockerfile), there are two commands that copy the _built_ Fleet Server artifact into the appropriate Agent components folder:

https://github.com/elastic/fleet-server/blob/8eb65da26e24f5058fdd178cef70645d60380cb6/dev-tools/e2e/Dockerfile#L6-L7

https://github.com/elastic/fleet-server/blob/8eb65da26e24f5058fdd178cef70645d60380cb6/dev-tools/e2e/Dockerfile#L12-L13

The value of `${STACK_VERSION}` mentioned in these commands should be the version of the built Fleet Server artifact, which should be the same as: https://github.com/elastic/fleet-server/blob/8eb65da26e24f5058fdd178cef70645d60380cb6/version/version.go#L9

However, this isn't always the case, e.g. https://github.com/elastic/fleet-server/pull/3857#issuecomment-2339363879

We should fix this reference to contain the correct version value.

michel-laterman:

e2e tests fail on version change

use the version from `version/version.go` instead of the docker image label when searching for the cover enabled binary to replace the fleet-server component in the image.

## Related issues

- Closes #3886

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -336,7 +336,7 @@ test-int-set: ## - Run integration tests without setup
 .PHONY: build-e2e-agent-image
 build-e2e-agent-image: docker-cover-e2e-binaries ## - Build a custom elastic-agent image with fleet-server binaries with coverage enabled injected
 	@printf "${CMD_COLOR_ON} Creating test e2e agent image\n${CMD_COLOR_OFF}"
-	FLEET_VERSION=$(VERSION) GOARCH=amd64 ./dev-tools/e2e/build.sh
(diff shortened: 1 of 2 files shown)


michel-laterman:
Thanks @swiatekm, the issue in this case is we build `fleet-server-9.0.0-SNAPSHOT-*` but the image is looking for `fleet-server-9.0.0-*`;

michel-laterman:
Looks like there are version errors between fleet-server and es: `Error - failed version compatibility check with elasticsearch (Agent: 9.0.0-SNAPSHOT, Elasticsearch: 8.16.0)`

PR #3901, opened 2024-09-12
Fix config deprecation warnings

michel-laterman:

We have deprecated config options throughout the 8.x lifecycle, but no warning are ever displayed

Specify the correct attributes that should result in a warning.

## Design Checklist

## Checklist

### DIFF
--- a/internal/pkg/config/config.go
+++ b/internal/pkg/config/config.go
@@ -10,12 +10,13 @@ import (
 	"errors"
 	"sync"
 
+	"github.com/gofrs/uuid"
+	"github.com/rs/zerolog"
+
 	"github.com/elastic/fleet-server/v7/version"
 	"github.com/elastic/go-ucfg"
 	"github.com/elastic/go-ucfg/flag"
 	"github.com/elastic/go-ucfg/yaml"
-	"github.com/gofrs/uuid"
-	"github.com/rs/zerolog"
 )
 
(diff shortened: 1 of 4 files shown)


internal/pkg/config/config.go
```diff
@@ -55,7 +57,9 @@ type Config struct {
 }
 
 var deprecatedConfigOptions = map[string]string{
-	"inputs[0].limits.max_connections": "max_connections has been deprecated and will be removed in a future release. Please configure server limits using max_agents instead.",
+	"inputs.0.server.limits.max_connections": "max_connections has been deprecated and will be removed in a future release. Please configure server limits using max_agents instead.",
+	"inputs.0.server.limits.policy_throttle": "policy_throttle has been deprecated and will be removed in a future release. Please use policy_limits instead.",
+	"fleet.agent.logging.level":              "fleet.agent.logging.level has been deprecated and will be rmoved in a future release. Please set logging level with the root attribute logging.level instead.",
```
andrzej-stencel, line 62:
```suggestion
	"fleet.agent.logging.level":              "fleet.agent.logging.level has been deprecated and will be removed in a future release. Please set logging level with the root attribute logging.level instead.",
```

andrzej-stencel (approved):

pchila (approved):

Issue #3620, opened 2024-06-07
E2E agent install test fixes for darwin/arm64 and windows


michel-laterman:
E2E agent install tests function for Linux

For MacOS (darwin) to function:
- change the agent artifact download from `darwin/arm64` as  it may be labelled as `aarch64`

For Windows:
- run the tests's `install` without `sudo`
- fix `TearDownTest()` to uninstall the agent without `sudo`
- fix the suite's `TearDownSuite()` function to work across OS'

michel-laterman:
https://github.com/elastic/fleet-server/pull/3904 allows e2e tests to run on arm machines

PR #3904, opened 2024-09-13
Fix test setup to run on ARM instances

michel-laterman:

Allow tests to run off arm platforms

Remove hard-coded `amd64` values from test setup

`make test-int` or `make test-e2e`

## Related issues

- Relates #3620

### DIFF
--- a/Dockerfile.build
+++ b/Dockerfile.build
@@ -1,6 +1,6 @@
 ARG GO_VERSION
-ARG ARCH=main # ARCH should be main or arm
-FROM docker.elastic.co/beats-dev/golang-crossbuild:${GO_VERSION}-${ARCH}
+ARG TARGETARCH # Should be main or arm
+FROM docker.elastic.co/beats-dev/golang-crossbuild:${GO_VERSION}-${TARGETARCH}
 
 RUN \
   apt-get update \
--- a/Dockerfile.build
+++ b/Dockerfile.build
@@ -1,5 +1,6 @@
 ARG GO_VERSION
-FROM docker.elastic.co/beats-dev/golang-crossbuild:${GO_VERSION}-main-debian11
+ARG ARCH=main # ARCH should be main or arm
+FROM docker.elastic.co/beats-dev/golang-crossbuild:${GO_VERSION}-${ARCH}
 
 RUN \
   apt-get update \
--- a/Dockerfile.build
+++ b/Dockerfile.build
@@ -1,6 +1,6 @@
 ARG GO_VERSION
-ARG TARGETARCH # Should be main or arm
-FROM docker.elastic.co/beats-dev/golang-crossbuild:${GO_VERSION}-${TARGETARCH}
+ARG SUFFIX # Should be main-debian11 or arm
+FROM docker.elastic.co/beats-dev/golang-crossbuild:${GO_VERSION}-${SUFFIX}
 
 RUN \
   apt-get update \
--- a/Makefile
+++ b/Makefile
@@ -43,6 +43,11 @@ endif
 DOCKER_IMAGE_TAG?=${VERSION}
 DOCKER_IMAGE?=docker.elastic.co/fleet-server/fleet-server
 
+ifeq ($(shell uname -p),arm)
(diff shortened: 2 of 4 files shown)


Dockerfile.build
```diff
@@ -1,5 +1,6 @@
 ARG GO_VERSION
-FROM docker.elastic.co/beats-dev/golang-crossbuild:${GO_VERSION}-main-debian11
```
swiatekm, line 2: Speaking as someone seeing this for the first time, the fact that `ARCH` can be `main` is very surprising. Can we not use the built-in `TARGETARCH` and use `main` in the Dockerfile if the value is `amd64`?

andrzej-stencel: Or not use the built-in `TARGETARCH` as it seems misused in this case. Rename this ARG to `GO_ARCH` or another arbitrary name?

pchila: The `main` arch seems a bit weird actually... 
@michel-laterman Does this come from some quirk on golang-crossbuild docker tags?

michel-laterman: yep, it's defined in the crossbuild readme

jlind23:
@michel-laterman I just looked at a slack conversation between @v1v and @jmlrt concluding that there was a recent change on the elasticsearch side that removed those images suffixed with the arch value.
IIUC we should just default to the multi arch one just like below:
`docker pull docker.elastic.co/elasticsearch/elasticsearch:8.15.2-7f2cd899-SNAPSHOT
8.15.2-7f2cd899-SNAPSHOT: Pulling from elasticsearch/elasticsearch
b14b0d0a7610: Already exists
bc03f0deaf3c: Pull complete
9a114eab9584: Pull complete
4ca545ee6d5d: Pull complete
341b5ebe39a8: Pull complete
6b896993e0fc: Pull complete
301b398356a8: Pull complete
07428383d5cd: Pull complete
df1cde76a217: Pull complete
f7adb9f55788: Pull complete
Digest: sha256:6cbd256f19c67ee0e982064befb85b29770ec9bc71ab33d5bb623e9a2936d942
Status: Downloaded newer image for docker.elastic.co/elasticsearch/elasticsearch:8.15.2-7f2cd899-SNAPSHOT
docker.elastic.co/elasticsearch/elasticsearch:8.15.2-7f2cd899-SNAPSHOT
`

Dockerfile.build
swiatekm: > Or not use the built-in TARGETARCH arg at all, as it seems misused in this case. Use GO_ARCH (in line with GO_VERSION) or another arbitrary name?

I wanted to use TARGETARCH, because it lets us actually build a multiplatform image. But it looks like there isn't a way to conditionally use `main` in the crossbuild image tag based on this. It'd be best if the crossbuild image was multiplatform as well - then this would be completely transparent, and we'd just do `docker buildx build --platform=...` in the Makefile.

But as this cannot be done, I'm ok with the previous approach of using a custom variable for this.

pchila: @v1v is there a specific reason why `main` is used instead of x86/amd64 ? I see the list of the images in [golang-crossbuild README](https://github.com/elastic/golang-crossbuild?tab=readme-ov-file#build-tags) and the `main` arch sticks out while all the other archs follow more conventional naming (arm, mips, ppc etc.)

But at least I can see now why `main` is used as default arg here, to target x86/amd64 architecture

v1v: I believe `main` is a specific folder:
- https://github.com/elastic/golang-crossbuild/tree/main/go/main

See https://github.com/elastic/golang-crossbuild#gomain-docker-image

michel-laterman: I've changed the name variable name to `SUFFIX` to distance it from the `TARGETARCH/ARCH` concepts as it includes more information than just the architecture.
I also set it to use the same `main-debian11` suffix as our other targets use.

I'm more then happy to change to `TARGETARCH` if the upstream sources support it.

pchila (approved):
LGTM

swiatekm (approved):

PR #3908, opened 2024-09-16
Add secret_paths attributes to policies sent to agents

michel-laterman:

Agent cannot determine where secrets have been injected into a policy.

Add a new attribute `secret_paths` to the policy data sent to agents. This attribute is a list of keys where the fleet-server has repalced a reference with a secret value. The agent is expected to redact the values of these keys when outputting policy data.

## Design Checklist

## Checklist


model/openapi.yml
```diff
@@ -522,6 +522,11 @@ components:
         id:
           description: The policy's ID.
           type: string
+        secret_paths:
```
cmacknz, line 525: Since this is outside of the signed section of the policy, in theory it could be modified to cause agent to not redact secrets locally couldn't it?

To put it in the signed section of the policy, we'd have to determine which fields contain secrets in Kibana instead, because that is where the signing key resides.

Is there a reason why we can't put these in Kibana?

cmacknz:
In addition to my bigger picture question about being able to sign the secret_paths, SonarQube says there are some non-error paths in replaceSliceRefs that have no test coverage.

internal/pkg/policy/secret.go
```diff
@@ -39,57 +40,99 @@ func getSecretValues(ctx context.Context, secretRefs []model.SecretReferencesIte
...
+	keys := make([]string, 0)
+	result := make([]any, len(arr))
+	var r any
+
+	for i, v := range arr {
+		switch value := v.(type) {
+		case string:
+			ref, replaced := replaceStringRef(value, secrets)
+			if replaced {
+				keys = append(keys, fmt.Sprintf("[%d]", i))
```
michel-laterman, line 115: I'm going to delay merging this as I think go-ucfg uses `.` when accessing array values.

michel-laterman: https://github.com/elastic/elastic-agent/pull/5621
go-ucfg does use `.` when accessing arrays

model/openapi.yml
cmacknz: OK, thinking more deeply, I was thinking through a scenario where someone had intercepted the checkin response, and wanted to to force the agent to disclose secrets by removing the redaction hints.

However, if they have the checkin response after Fleet Server has touched it and added these keys, they already have the replaced secret values that Fleet Server inserted.

So I don't think there is a point in doing this because it doesn't protect us from anything.

PR #3924, opened 2024-09-19
Update go to 1.23.1

michel-laterman:
Update go to v1.23.1


ycombinator (approved):
Thanks for this PR but wouldn't our automation have put up the bump PR anyway? Just curious why you had to do this manually instead.

michel-laterman:
our automation is set to do updates to patch releases; we've been updating to the next minor releases manually (though there is probably a way to trigger the automation)

PR #3927, opened 2024-09-19
Add test collector plugin for Buildkite analytics

rowlandgeoff:
Add the test collector plugin to enable Buildkite test analytics. Test results will be shown here after main branch pipeline runs: https://buildkite.com/organizations/elastic/analytics/suites/fleet-server

Closes https://github.com/elastic/ingest-dev/issues/3916

### DIFF
--- a/.buildkite/hooks/pre-command
+++ b/.buildkite/hooks/pre-command
@@ -55,6 +55,13 @@ if [[ "$BUILDKITE_PIPELINE_SLUG" == "fleet-server" && "$BUILDKITE_STEP_KEY" == "
   export EC_API_KEY=${EC_API_KEY_SECRET}
 fi
 
+if [[ "$BUILDKITE_PIPELINE_SLUG" == "fleet-server" && "$BUILDKITE_STEP_KEY" == "int-test" ]]; then
+  # BK analytics
+  echo "--- Prepare BK test analytics token :vault:"
+  BUILDKITE_ANALYTICS_TOKEN=$(vault kv get -field token kv/ci-shared/platform-ingest/buildkite_analytics_tokens/fleet-server)
+  export BUILDKITE_ANALYTICS_TOKEN
+fi
+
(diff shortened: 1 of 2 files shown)


.buildkite/hooks/pre-command
```diff
@@ -55,6 +55,15 @@ if [[ "$BUILDKITE_PIPELINE_SLUG" == "fleet-server" && "$BUILDKITE_STEP_KEY" == "
   export EC_API_KEY=${EC_API_KEY_SECRET}
 fi
 
+# BK analytics
+if [[ "$BUILDKITE_PIPELINE_SLUG" == "fleet-server" ]]; then
+  if [[ "$BUILDKITE_STEP_KEY" == "int-test" || "$BUILDKITE_STEP_KEY" == "e2e-test" ]]; then
+    echo "--- Prepare BK test analytics token :vault:"
+    BUILDKITE_ANALYTICS_TOKEN=$(vault kv get -field token kv/ci-shared/platform-ingest/buildkite_fleet_server_analytics_token)
```
alexsapran, line 62: I have no comment; we did in other repositories, but IMHO, it's not ideal either—except Beats, due to the nature of the pipelines.

The secret should be in a repository-scoped vault path rather than a shared vault path. This is a good idea because the secret makes sense only for this repository.

swiatekm (approved):

jlind23:
Looks good to me, were you able to have a successful run on your PR already?

.buildkite/hooks/pre-command
rowlandgeoff: Do we have repo-scoped vault paths that we use for CI?

rowlandgeoff:
> Looks good to me, were you able to have a successful run on your PR already?

@jlind23 Yes, had 2 successful runs on the PR before I switched it to only run on main.

.buildkite/hooks/pre-command
alexsapran: sure we do https://docs.elastic.dev/ci/using-secrets#using-repository-scoped-secrets

But we can move them afterwards.

jlind23:
@rowlandgeoff Let's merge it then :-)

PR #3935, opened 2024-09-20
Fix bad error handling in api key auth

### LINKED ISSUE #3929: Return 503 Service Unavailable when unable to authenticate with Elasticsearch instead of 401
- Relates https://github.com/elastic/elastic-agent/issues/5428

When we attempt to validate an agent's API key and cannot communicate with ES because Fleet Server's API key/service token is temporarily unavailable we return a 401 to Elastic Agents. This seems correct, but if it happens for 7 check ins in a row will cause the agent to automatically unenroll.

We should evaluate handling "cannot auth with ES" from "explicitly told API key was invalid" separately and returning a 503 when Fleet Server itself cannot auth with ES to avoid unintentional mass unenrollments.

The relevant code block is below: https://github.com/elastic/fleet-server/blob/6c144499dfacc53c45f2df1173345b8723e1db05/internal/pkg/apikey/auth.go#L34-L61

blakerouse:

Currently any error returned back to Fleet Server from elasticsearch that is not 401 or 429 is returned as a 401 error. This tells the calling client that the API key is invalidate. That is only true when the error is 401, otherwise the error means something different.

This solves the issue by returning the actual error back elasticsearch to the calling client. This means that a 500 error from elasticsearch will not result in a 401 back to the client, instead it will be a 500 error.

This uses the standard `es.ParseError` logic to determine the error and return something readable to the calling client.

- Bootstrap Fleet Server
- Turn-off elasticsearch
- See that 401 is not returned to the client

## Design Checklist

- ~~[ ] I have or intend to scale test my changes, ensuring it will work reliably with 100K+ agents connected.~~ (no effect on scale)
- ~~[ ] I have included fail safe mechanisms to limit the load on fleet-server: rate limiting, circuit breakers, caching, load shedding, etc.~~ (no effect)

## Checklist

## Related issues

- Closes #3929

### DIFF
--- a/changelog/fragments/1726845328-Fix-authentication-return-error-code.yaml
+++ b/changelog/fragments/1726845328-Fix-authentication-return-error-code.yaml
@@ -8,7 +8,7 @@
 # - security: impacts on the security of a product or a user’s deployment.
 # - upgrade: important information for someone upgrading from a prior version
 # - other: does not fit into any of the other categories
-kind: feature
+kind: bug-fix
 
 # Change summary; a 80ish characters long description of the change.
 summary: Fix authentication return error code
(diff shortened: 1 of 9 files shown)


blakerouse:
go-lint is very wrong here... idk

PR #3938, opened 2024-09-20
Add make multipass target

michel-laterman:

Simplify setup of a linux vm on apple arm machines as virtualbox (vagrant) is not usable.

Add a `make multipass` target to start a dev env with all tools needed for testing.


swiatekm (approved):
LGTM.

Minor nitpick: The file name with the multipass config should indicate it's a template. It'd be more informative for it to be named `dev-tools/multipass-cloud-init.yml.template` or `dev-tools/multipass-cloud-init.yml.envsubst`.

PR #3986, opened 2024-10-08
Add timeout to bulker flush, add default case in failQueue

juliaElastic:

Scale tests are often blocked at flushing the bulker queue.

Adding a context timeout to the bulker flush so it times out if it takes more time than the deadline.

Ran scale tests here: https://buildkite.com/elastic/observability-perf/builds?branch=increase-poll-action-retries-for-agent-upgrades

## Design Checklist

## Checklist

## Related issues

Relates https://github.com/elastic/ingest-dev/issues/3783


internal/pkg/api/handleCheckin.go
```diff
@@ -337,6 +337,9 @@ func (ct *CheckinT) ProcessRequest(zlog zerolog.Logger, w http.ResponseWriter, r
 	actions, ackToken = convertActions(zlog, agent.Id, pendingActions)
 
 	span, ctx := apm.StartSpan(r.Context(), "longPoll", "process")
+	ctx, cancel := context.WithTimeout(ctx, pollDuration)
```
cmacknz, line 340: Interesting, evaluating what the timeouts and lifetimes of all the requests in here is actually challenging and I'm not sure if they are right.

This context is not obviously tied to the actual network requests, also at this point, we are passed auth, which should probably also respect a timeout and it's not clear that it is tied to the poll duration either.

What this does is cause us to hit the `ctx.Done()` block below, which triggers us to hit the `ct.writeResponse` call. That call is a network operation that should also have a timeout, but it can't be tied to this context because it is expired, so we'd need a different one.

cmacknz: Looking closer, this context deadline actually changes the checkin logic without setting a deadline on any of the underlying network operations or interactions with ES.

There is already a ticker for the long poll duration: https://github.com/elastic/fleet-server/blob/7ecbda15c53ba0bcedb9507cc49dc35311513e0e/internal/pkg/api/handleCheckin.go#L316-L319

It causes us to hit the `CheckIn` method here: https://github.com/elastic/fleet-server/blob/7ecbda15c53ba0bcedb9507cc49dc35311513e0e/internal/pkg/api/handleCheckin.go#L374-L379

So all setting this context deadline does is get us to this block, but only if we aren't already in the CheckIn method: https://github.com/elastic/fleet-server/blob/7ecbda15c53ba0bcedb9507cc49dc35311513e0e/internal/pkg/api/handleCheckin.go#L345-L356

If both `<-ctx.Done()` and `<-tick.C` are pending at the same time the Go runtime will randomly choose which case is taken, so the behavior here isn't even deterministic.

juliaElastic: Thanks for looking into it. Do you think there is a way to enforce the pollDuration on the underlying network requests? Maybe someone from the Control Plane team can spend some time on this, it seems to be the reason of some drones stuck in a failed checkin state.

cmacknz: Looking closer, if we are in this loop, the only place where we interact directly with ES are in `processPolicy` and putting the deadline on the context solves that case: https://github.com/elastic/fleet-server/blob/7ecbda15c53ba0bcedb9507cc49dc35311513e0e/internal/pkg/api/handleCheckin.go#L364

If I follow this all the way down the call stack to where the actual search call happens, the context should prevent us from waiting forever for a response:

https://github.com/elastic/fleet-server/blob/7ecbda15c53ba0bcedb9507cc49dc35311513e0e/internal/pkg/bulk/engine.go#L568-L574

The one that is actually on the underlying ES network request is in the bulker run function: https://github.com/elastic/fleet-server/blob/7ecbda15c53ba0bcedb9507cc49dc35311513e0e/internal/pkg/bulk/engine.go#L356

The context for this appears to just be tied to `context.Background` in multiple places: https://github.com/elastic/fleet-server/blob/7ecbda15c53ba0bcedb9507cc49dc35311513e0e/internal/pkg/server/fleet.go#L374-L379

https://github.com/elastic/fleet-server/blob/7ecbda15c53ba0bcedb9507cc49dc35311513e0e/internal/pkg/bulk/engine.go#L166-L182

It does look like there may be a default 90s timeout on the underlying ES client, but I don't see this actually being called anywhere (possible I missed it): https://github.com/elastic/fleet-server/blob/7ecbda15c53ba0bcedb9507cc49dc35311513e0e/internal/pkg/config/output.go#L59-L62

There are lots of places that could be the problem, TBH I'd just add more logging or spans until we definitively know exactly where we get stuck when these 28+ minute checkins happen.

cmacknz: > Maybe someone from the Control Plane team can spend some time on this, it seems to be the reason of some drones stuck in a failed checkin state.

For now that person is me, let's avoid context switching in someone else while we narrow down what is actually wrong and can evaluate the effort to fix it.

internal/pkg/api/handleCheckin.go
```diff
@@ -337,15 +337,20 @@ func (ct *CheckinT) ProcessRequest(zlog zerolog.Logger, w http.ResponseWriter, r
...
 
 	span, ctx := apm.StartSpan(r.Context(), "longPoll", "process")
+	// ctx, cancel := context.WithTimeout(ctx, pollDuration)
+	// defer cancel()
+
 	if len(actions) == 0 {
 	LOOP:
 		for {
 			select {
 			case <-ctx.Done():
```
blakerouse, line 347: Currently looking at this the only time this case would be hit would be if the client closes there connection to Fleet Server. As the `span, ctx := apm.StartSpan(r.Context(), ...` is being used as the context here, so then this section writes the response. Looking at this code, it shouldn't even write the response. If the context is cancelled then that means the client is no longer connected.

internal/pkg/api/handleCheckin.go
```diff
@@ -360,6 +365,7 @@ func (ct *CheckinT) ProcessRequest(zlog zerolog.Logger, w http.ResponseWriter, r
 				actions = append(actions, acs...)
 				break LOOP
 			case policy := <-sub.Output():
+				zlog.Debug().Str(logger.AgentID, agent.Id).Msg("SCALEDEBUG new policy")
 				actionResp, err := processPolicy(ctx, zlog, ct.bulker, agent.Id, policy)
```
blakerouse, line 369: Possible that it gets stuck here processing the policy, as it doesn't create its own context and timeout after a period of time. This is still using the same context of the request connection.

internal/pkg/api/handleCheckin.go
```diff
@@ -368,7 +374,7 @@ func (ct *CheckinT) ProcessRequest(zlog zerolog.Logger, w http.ResponseWriter, r
 				actions = append(actions, *actionResp)
 				break LOOP
 			case <-longPoll.C:
-				zlog.Trace().Msg("fire long poll")
+				zlog.Debug().Str(logger.AgentID, agent.Id).Msg("fire long poll")
 				break LOOP
 			case <-tick.C:
 				err := ct.bc.CheckIn(agent.Id, string(req.Status), req.Message, nil, rawComponents, nil, ver, unhealthyReason, false)
```
blakerouse, line 380: Doesn't seem likely that it gets stuck here as `ct.bc.CheckIn` just grabs a lock and adds to a map. But again it could be possible that there is a dead lock here and that lock is held and never freed. Shouldn't be ruled out.

internal/pkg/api/handleCheckin.go
juliaElastic: the logic of returning CheckinResponse here was added in this pr: https://github.com/elastic/fleet-server/pull/3165/files#diff-e0c02bac8d151e9941eedd5ef643441665ee3d2f78baf42c121edd45dee08ded

Can the context be cancelled only by the client here? I'm wondering if the `writeResponse` is successful, is it correct to return the `AckToken` without actions? I'm trying to find where the `AckToken` is persisted back to ES.

internal/pkg/bulk/engine.go
```diff
@@ -336,6 +336,9 @@ func (b *Bulker) Run(ctx context.Context) error {
 
 	w := semaphore.NewWeighted(int64(b.opts.maxPending))
 
+	ctx, cancel := context.WithTimeout(ctx, 5*time.Minute)
```
cmacknz, line 339: This causes each call to the bulker's Run() function to return unconditionally after 5 minutes, including the one in the fleet server main loop.

You can see this happening in the logs. Every 5 minutes the bulker aborts everything it is doing.

To avoid this unconditional exit you would create the context in doFlush https://github.com/juliaElastic/fleet-server/blob/a9a82eaae87b7c2d4b54d714d34943a970c8f3cd/internal/pkg/bulk/engine.go#L352

That would bound each execution of `doFlush` to 5 minutes which should have the same effect, but without the logging about constant exiting.

juliaElastic: I tried moving the timeout to `doFlush` but it doesn't work, I think the `doFlush` function exits too quickly and cancels the context.
See the logs from a 10k run: https://platform-logging.kb.us-west2.gcp.elastic-cloud.com/app/r/s/S0Wla

juliaElastic: I moved the timeout outside of `doFlush`, see this comment: https://github.com/elastic/ingest-dev/issues/3783#issuecomment-2459697129
Strangely I don't really see the flush timing out anymore, mostly seeing the upgrade step being stuck at 75k runs, 10k runs pass almost always.

internal/pkg/checkin/bulk.go
```diff
@@ -153,6 +153,7 @@ LOOP:
 		case <-tick.C:
 			if err = bc.flush(ctx); err != nil {
 				zerolog.Ctx(ctx).Error().Err(err).Msg("Eat bulk checkin error; Keep on truckin'")
+				break LOOP
```
juliaElastic, line 156: breaking the loop here makes the bulker exit and be restarted
it seems to help with the upgrade step, tested with a 51k run: https://buildkite.com/elastic/observability-perf/builds/3670#01930bb4-fb8d-47c2-8699-94c0cc471699
Logs: https://platform-logging.kb.us-west2.gcp.elastic-cloud.com/app/r/s/NiMhb

juliaElastic:
> @juliaElastic What is the status of this PR? I see you request another review, are you seeing success with that current state of the PR now? Can you provide a summary of those results?

I summarized here: https://github.com/elastic/ingest-dev/issues/3783#issuecomment-2467506043
I couldn't reproduce issues with policy change steps anymore with this pr, only upgrade seems to fail sometimes, which seems a different issue.
I would like to merge this if possible, unless we want to wait until the upgrade issue is tracked down too.

internal/pkg/bulk/engine.go
```diff
@@ -405,7 +406,10 @@ func (b *Bulker) Run(ctx context.Context) error {
 					Int("byteCnt", byteCnt).
 					Msg("Flush on threshold")
 
-				err = doFlush()
+				// deadline prevents bulker being blocked on flush
+				flushCtx, cancel := context.WithTimeout(ctx, defaultFlushContextTimeout)
+				defer cancel()
```
michel-laterman, line 411: These contexts are created in a loop, but the cancels are deffered until the function (`Bulker.Run`) returns. Should we move context creation to within the `doFlush` function to prevent this?

juliaElastic: I originally added it to `doFlush` but it didn't work, Fleet Server didn't come up successfully. I think `doFlush` finishes too quickly and would cancel the context.

cmacknz: It doesn't work because most of the work happening is asynchronous in a goroutine, so once you unblock from acquiring the semaphore if you cancel immediately you'll exit the goroutine.

https://github.com/elastic/fleet-server/blob/1a23838f0eb64ae2748b058e3839a60afed2725d/internal/pkg/bulk/engine.go#L458-L461

There probably needs to be two separate deadlines created in doFlush():

1. On the semaphore acquire call: https://github.com/elastic/fleet-server/blob/1a23838f0eb64ae2748b058e3839a60afed2725d/internal/pkg/bulk/engine.go#L446-L449
2. Inside the goroutine doing the work https://github.com/elastic/fleet-server/blob/1a23838f0eb64ae2748b058e3839a60afed2725d/internal/pkg/bulk/engine.go#L458-L472

juliaElastic: tried to add deadline in doFlush to those 2 places, but it doesn't' seem to work
logs: https://platform-logging.kb.us-west2.gcp.elastic-cloud.com/app/r/s/L3Nkr

internal/pkg/bulk/engine.go
```diff
@@ -458,6 +464,10 @@ func (b *Bulker) flushQueue(ctx context.Context, w *semaphore.Weighted, queue qu
 	go func() {
 		start := time.Now()
 
+		// deadline prevents bulker being blocked on flush
+		ctx, cancel := context.WithTimeout(ctx, defaultFlushContextTimeout)
```
michel-laterman, line 468: The context we are wrapping here has a timeout associated with it from line 448.
Is that what we want in this case?

juliaElastic: you're right, probably should be separate timeout

juliaElastic: made the change, I had a passing 10k and 50k run: https://buildkite.com/elastic/observability-perf/builds/3745

juliaElastic:
@cmacknz @michel-laterman Are you ok to merge?

michel-laterman (approved):
🚀

cmacknz (approved):
LGTM

PR #3991, opened 2024-10-10
Don't fail checkin if agent has upgrade details with no action

jillguyonnet:

Relates https://github.com/elastic/ingest-dev/issues/4223

We're investigating an issue in a serverless project where Fleet Server keeps throwing 500 on agent checkin because the agent has upgrade details but the action id they refer to is not found.

Modify `Checkin.processUpgradeDetails` so that checkin does not fail in this scenario and log a warning instead.

### DIFF
--- a/changelog/fragments/1728652985-Dont-fail-checkin-if-upgrade-action-not-found.yaml
+++ b/changelog/fragments/1728652985-Dont-fail-checkin-if-upgrade-action-not-found.yaml
@@ -1,0 +1,35 @@
+# Kind can be one of:
+# - breaking-change: a change to previously-documented behavior
+# - deprecation: functionality that is being removed in a later release
+# - bug-fix: fixes a problem in a previous version
+# - enhancement: extends functionality but does not break or fix existing behavior
+# - feature: new functionality
+# - known-issue: problems that we are aware of in a given version
+# - security: impacts on the security of a product or a user’s deployment.
+# - upgrade: important information for someone upgrading from a prior version
+# - other: does not fit into any of the other categories
+kind: feature
+
+# Change summary; a 80ish characters long description of the change.
+summary: Dont fail checkin if upgrade action not found
+
+# Long description; in case the summary is not enough to describe the change
+# this field accommodate a description without length limits.
+# NOTE: This field will be rendered only for breaking-change and known-issue kinds at the moment.
+description:
+  On agent checkin, when an agent has upgrade details, the corresponding agent action is looked up and linked with APM spans.
+  However, there may be cases where such an action does not exist, e.g. when running `elastic-agent upgrade` for a Fleet-managed agent.
+  This change ensures that agent checkin does not fail if the agent has upgrade details but no corresponding action.
+
+# Affected component; usually one of "elastic-agent", "fleet-server", "filebeat", "metricbeat", "auditbeat", "all", etc.
+component:
+
+# PR URL; optional; the PR number that added the changeset.
+# If not present is automatically filled by the tooling finding the PR where this changelog fragment has been added.
+# NOTE: the tooling supports backports, so it's able to fill the original PR number instead of the backport PR number.
+# Please provide it if you are adding a fragment for a different PR.
+pr: https://github.com/elastic/fleet-server/pull/3991
+
+# Issue URL; optional; the GitHub issue related to this changeset (either closes or is part of).
+# If not present is automatically filled by the tooling with the issue linked to the PR number.
+#issue: https://github.com/owner/repo/1234
--- a/testing/e2e/api_version/client_api_current.go
+++ b/testing/e2e/api_version/client_api_current.go
@@ -350,6 +350,26 @@ func (tester *ClientAPITester) TestCheckinWithBadRequest() {
 	tester.Require().Equal(http.StatusBadRequest, statusCode, "Expected status code 400 for bad request")
 }
 
+func (tester *ClientAPITester) TestCheckinWithActionNotFound() {
+	ctx, cancel := context.WithTimeout(context.Background(), 4*time.Minute)
+	defer cancel()
+
+	// enroll agent
+	tester.T().Log("test enrollment")
+	agentID, agentKey := tester.Enroll(ctx, tester.enrollmentKey)
+	tester.VerifyAgentInKibana(ctx, agentID)
+
+	tester.T().Logf("test checkin with no upgrade action: agent %s", agentID)
+	// checkin request with upgrade details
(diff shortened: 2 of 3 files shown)


jlind23:
After chatting about this with @cmacknz we would like to change this behaviour and not fail nor even return a 5xx whenever this is happening. At most, a warning should be logged but that would be it.

jillguyonnet:
Thanks @jlind23 - I'm not sure how to test this properly yet but I pushed a commit as it seems like a minimal change.

jlind23:
I would like at least to get @michel-laterman's eyes on this before merging.

michel-laterman (commented):
We should add a changelog entry for this.

Also we can add an e2e test, we can enrol the agent then checkin with upgrade_details where we specify the action id

cmacknz:
> After chatting about this with @cmacknz we would like to change this behaviour and not fail nor even return a 5xx whenever this is happening. At most, a warning should be logged but that would be it.

To add some context for why I suggested this, if you run `elastic-agent upgrade` for a Fleet managed agent there will be no action, and then all sub-sequent check in will fail with 500s. 

We need the system to keep operating normally if this happens, this should not be a way to block agent from successfully checking in.

Taking a larger step back, if an agent says it is updating, and Fleet doesn't expect that, the agent should probably be flagged with an "unexpected update" status or something in the Fleet UI. Most users are not going to look in the Fleet Server logs to see things like this.

We could also arguably prevent using `elastic-agent upgrade` at all for Fleet managed upgrades, but the system would still have to tolerate this "surprise upgrade details" situation well.

jillguyonnet:
> We should add a changelog entry for this.
>
> Also we can add an e2e test, we can enrol the agent then checkin with upgrade_details where we specify the action id

Thanks @michel-laterman - I've added a changelog entry. I've tried adding an e2e test but running into issues so far. I'll reach out to you about this.

jillguyonnet:
Thank you @cmacknz for the added context, it makes definite sense that a non existent agent action should not cause the checkin to fail.

On the serverless project where the issue is happening, the agentless agent version was updated as part of a new k8s deployment running a newer version of the agent. IIUC, because this is not a Fleet-managed upgrade process, it could have resulted in agents with upgrade details but no agent actions.

I'm waiting to check the proposed fix implementation with @michel-laterman.

It sounds like it would be worth capturing your last observation (handling non Fleet managed upgrades) somewhere else, perhaps in an ingest-dev issue?

cmacknz:
> On the serverless project where the issue is happening, the agentless agent version was updated as part of a new k8s deployment running a newer version of the agent. IIUC, because this is not a Fleet-managed upgrade process, it could have resulted in agents with upgrade details but no agent actions.

Containers don't trigger the upgrade process, and you can't send upgrade actions to agents in containers because they report themselves (or should) as non-upgradable. You can exec into a container and run `elastic-agent upgrade` though which may be where this came from.

I am a little bit hesitant to start forbidding use of the `elastic-agent upgrade` completely for Fleet managed agents as it is a useful escape hatch if something goes terribly wrong. For now I've written https://github.com/elastic/kibana/issues/196426 as an enhancement on the Fleet side to flag agents that are upgrading unexpectedly.

changelog/fragments/1728652985-Dont-fail-checkin-if-upgrade-action-not-found.yaml
```diff
@@ -0,0 +1,35 @@
...
+# - enhancement: extends functionality but does not break or fix existing behavior
+# - feature: new functionality
+# - known-issue: problems that we are aware of in a given version
+# - security: impacts on the security of a product or a user’s deployment.
+# - upgrade: important information for someone upgrading from a prior version
+# - other: does not fit into any of the other categories
+kind: feature
+
+# Change summary; a 80ish characters long description of the change.
+summary: Dont-fail-checkin-if-upgrade-action-not-found
```
michel-laterman, line 14:
```suggestion
summary: Don't fail checkin if upgrade action not found
```

changelog/fragments/1728652985-Dont-fail-checkin-if-upgrade-action-not-found.yaml
```diff
@@ -0,0 +1,35 @@
...
+
+# PR URL; optional; the PR number that added the changeset.
+# If not present is automatically filled by the tooling finding the PR where this changelog fragment has been added.
+# NOTE: the tooling supports backports, so it's able to fill the original PR number instead of the backport PR number.
+# Please provide it if you are adding a fragment for a different PR.
+pr: https://github.com/elastic/fleet-server/pull/3991
+
+# Issue URL; optional; the GitHub issue related to this changeset (either closes or is part of).
+# If not present is automatically filled by the tooling with the issue linked to the PR number.
+issue: https://github.com/elastic/ingest-dev/issues/4223
```
michel-laterman, line 35:
```suggestion
#issue:
```
We can't link to internal issues in public docs

testing/e2e/api_version/client_api_current.go
```diff
@@ -350,6 +350,26 @@ func (tester *ClientAPITester) TestCheckinWithBadRequest() {
...
+	tester.T().Log("test enrollment")
+	agentID, agentKey := tester.Enroll(ctx, tester.enrollmentKey)
+	tester.VerifyAgentInKibana(ctx, agentID)
+
+	tester.T().Logf("test checkin with no upgrade action: agent %s", agentID)
+	// checkin request with upgrade details
+	req := &api.AgentCheckinJSONRequestBody{
+		Status:         api.CheckinRequestStatusOnline,
+		Message:        "test checkin",
+		UpgradeDetails: &api.UpgradeDetailsStateUPGDOWNLOADING,
```
michel-laterman, line 367: The type of `UpgradeDetails` is `*api.UpgradeDetails`
so you probably want something like:
```go
UpgradeDetails: &api.UpgradeDetails{
    ActionId: "test-missing-id",
    State:    &api.UpgradeDetailsStateUPGDOWNLOADING,
},
```

jillguyonnet:
I'm using the CI to run my test as e2e tests for the time being are they are not running properly locally, this might take a few commits.

michel-laterman (approved):
lgtm

thanks for adding the e2e test!

pierrehilbert:
Should we have this fix also in 8.16?
cc @jillguyonnet

jillguyonnet:
@pierrehilbert Yes, it would make sense. Does the `backport-8.x` label not cover it?

pierrehilbert:
No, we already created the 8.16 branch, I'm adding the backport to 8.16, thanks for confirming :-)

PR #4012, opened 2024-10-16
[Integration testing] Remove AltErr and only expect 400 from ES bulk API

ycombinator:

This PR expects only a 400 HTTP response code from the Elasticsearch bulk API in integration tests when invalid index name is specified.  Concretely, it removes the `AltErr` field temporarily introduced in https://github.com/elastic/fleet-server/pull/4001 which was allowing the tests to accept an alternate error from Elasticsearch.

Related: https://github.com/elastic/elasticsearch/pull/114869

### DIFF
--- a/internal/pkg/bulk/bulk_integration_test.go
+++ b/internal/pkg/bulk/bulk_integration_test.go
@@ -28,11 +28,10 @@ func TestBulkCreate(t *testing.T) {
 	index, bulker := SetupIndexWithBulk(ctx, t, testPolicy, WithFlushThresholdCount(1))
 
 	tests := []struct {
-		Name   string
-		Index  string
-		ID     string
-		Err    error
-		AltErr error // FIXME: workaround until https://elasticco.atlassian.net/browse/ES-9711 is resolved
+		Name  string
+		Index string
+		ID    string
+		Err   error
 	}{
 		{
 			Name:  "Empty Id",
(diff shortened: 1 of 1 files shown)


blakerouse (approved):
+1

swiatekm (approved):

jlind23:
Looks like we have a failure on ` FAIL: TestBulkCreate/Invalid_utf-8 (0.26s)`
Shall we change this expected code to 400 instead - https://github.com/elastic/fleet-server/blob/deefa2d19aff6d4c75c2f7acae368a0387278391/internal/pkg/bulk/bulk_integration_test.go#L73?

ycombinator:
> Looks like we have a failure on ` FAIL: TestBulkCreate/Invalid_utf-8 (0.26s)` Shall we change this expected code to 400 instead -
> 
> https://github.com/elastic/fleet-server/blob/deefa2d19aff6d4c75c2f7acae368a0387278391/internal/pkg/bulk/bulk_integration_test.go#L73
> 
> ?

This change is already part of this PR.  So I'm not sure why this is test is failing. Investigating...

internal/pkg/bulk/bulk_integration_test.go
```diff
@@ -70,12 +69,8 @@ func TestBulkCreate(t *testing.T) {
 			Name:  "Invalid utf-8",
 			Index: string([]byte{0xfe, 0xfe, 0xff, 0xff}),
 			Err: es.ErrElastic{
-				Status: 500,
-				Type:   "json_parse_exception",
-			},
-			AltErr: es.ErrElastic{
 				Status: 400,
-				Type:   "parse_exception",
```
michel-laterman, line 78: This is the correct error type, not `json_parse_exception`

ycombinator: Thanks @michel-laterman! ❤️

pierrehilbert:
Should we backport this to 8.16 also?

ycombinator:
> Should we backport this to 8.16 also?

Doesn't look like the [ES fix](https://github.com/elastic/elasticsearch/pull/114869) was backported to `8.16`, so I don't think this PR should be backported to `8.16` either.

PR #4036, opened 2024-10-23
fix audit unenroll

juliaElastic:

audit unenroll should set the agent active flag to false

## Design Checklist

## Checklist

## Related issues

Closes https://github.com/elastic/kibana/issues/197180

### DIFF
--- a/internal/pkg/api/handleAudit.go
+++ b/internal/pkg/api/handleAudit.go
@@ -103,6 +103,7 @@ func (audit *AuditT) markUnenroll(ctx context.Context, zlog zerolog.Logger, req 
 
 	now := time.Now().UTC().Format(time.RFC3339)
 	doc := bulk.UpdateFields{
+		dl.FieldActive:                false,
 		dl.FieldUnenrolledAt:          now,
 		dl.FieldUpdatedAt:             now,
 		dl.FieldAuditUnenrolledTime:   req.Timestamp,


juliaElastic:
I need confirmation on how this feature is supposed to work, read through the comments in https://github.com/elastic/elastic-agent/issues/484 and the integration/e2e tests and it seems the feature is supposed to allow calling the audit unenroll API multiple times, and a checkin after. 

However, the existing unenroll action sets the active flag to false (and that is needed for the UI to correctly show unenrolled state), and by adding it to the audit unenroll API, the agent will no longer accept API calls (the API key will be invalidated).

internal/pkg/server/fleet_integration_test.go
```diff
@@ -1485,17 +1485,6 @@ func Test_SmokeTest_AuditUnenroll(t *testing.T) {
 	require.Equal(t, http.StatusOK, res.StatusCode)
 	res.Body.Close()
 
-	t.Log("Orphaned can replace uninstall")
```
michel-laterman, line 1488: We need to test/support this scenario.
It's important for Endpoint to be able to signal that it is orphaned even if the installed agent is removed

michel-laterman:
In the original implementation I tried setting agents to inactive (https://github.com/elastic/fleet-server/pull/3818) however that would cause the API key authentication we do on checkins to fail. It would require a change in our overall auth logic if we wanted to make this change

PR #4037, opened 2024-10-23
Support sampling rate in APM configuration

pchila:
This PR introduces support for APM sample rate using the values provided by elastic-agent though the control protocol

Added a check for a value of transacton sampling in `apmConfigToInstrumentation` (it's an optional float32) and formatted the vale as a string to be included in the instrumentation copnfig.

## Design Checklist

## Checklist

### DIFF
--- a/internal/pkg/server/agent_test.go
+++ b/internal/pkg/server/agent_test.go
@@ -262,6 +262,8 @@ func Test_Agent_configFromUnits(t *testing.T) {
 			},
 		})
 		require.NoError(t, err)
+
+		samplingRate := float32(0.5)
 		mockInClient := &mockClientUnit{}
 		mockInClient.On("Expected").Return(
 			client.Expected{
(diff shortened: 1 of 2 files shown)


internal/pkg/server/agent_test.go
```diff
@@ -449,6 +461,7 @@ func Test_Agent_configFromUnits(t *testing.T) {
 		assert.Equal(t, []string{"testhost:8080"}, cfg.Inputs[0].Server.Instrumentation.Hosts)
 		assert.Equal(t, "test", cfg.Inputs[0].Server.Instrumentation.GlobalLabels)
 		assert.Equal(t, "test-token", cfg.Output.Elasticsearch.ServiceToken)
+		assert.Equal(t, "0.01", cfg.Inputs[0].Server.Instrumentation.TransactionSampleRate)
```
michel-laterman, line 464: Should we keep a test case where the sampling rate is not specified?

michel-laterman (approved):
lgtm, we may want to adjust/add a test case

kaanyalti (approved):

internal/pkg/server/agent_test.go
pchila: I can remove the sampling rate from one of the testcases or just add a new one with minimal apm config, I'm not sure I have a preference between the two. What would you rather have ?

PR #4013, opened 2024-10-16
Bump the version on `8.x` to `8.17.0`

ycombinator:
Update version to next minor 8.17.0.

Merge after create the new release branch.

PR #4058, opened 2024-10-28
Fix `TestAgentInstallSuite`

ycombinator:

CI has been failing on the `8.x` branch since https://github.com/elastic/fleet-server/pull/4013 was merged.  It fails on `TestAgentInstallSuite` with the message: "unable to find package download".  This PR attempts to figure out why this failure is happening (it's not happening on the `main` branch) and fix it.

TBD - still investigating

## Design Checklist

## Checklist


ycombinator:
Looks like the root cause is https://github.com/elastic/fleet-server/issues/4059.

ycombinator:
The test that's failing in CI now will be fixed in https://github.com/elastic/fleet-server/pull/4030.  So we need both PRs to be merged for `8.x` to go green again.

PR #4142, opened 2024-11-22
Fix server.address field in HTTP logs

michel-laterman:

`server.address` is empty in HTTP request logs.

Populate `server.address` using the interface associated with `http.LocalAddrContextKey`.

start a fleet-server and make an HTTP request.

## Design Checklist

## Checklist

### DIFF
--- a/internal/pkg/logger/http.go
+++ b/internal/pkg/logger/http.go
@@ -216,7 +216,11 @@ func Middleware(next http.Handler) http.Handler {
 		w.Header().Set(HeaderRequestID, reqID)
 
 		// get the server bound addr from the req ctx
-		addr, _ := r.Context().Value(http.LocalAddrContextKey).(string)
+		var addr string
+		netAddr, ok := r.Context().Value(http.LocalAddrContextKey).(net.Addr)
+		if ok {
+			addr = netAddr.String()
+		}
 
 		// Add trace correlation fields
 		ctx := r.Context()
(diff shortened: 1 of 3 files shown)


internal/pkg/logger/http.go
```diff
@@ -250,13 +254,12 @@ func Middleware(next http.Handler) http.Handler {
...
 
-		if zlog.Debug().Enabled() || (wrCounter.statusCode < 200 && wrCounter.statusCode >= 300) {
+		if zlog.Info().Enabled() {
 			e.Uint64(ECSHTTPRequestBodyBytes, rdCounter.Count())
 			e.Uint64(ECSHTTPResponseBodyBytes, wrCounter.Count())
 			e.Int(ECSHTTPResponseCode, wrCounter.statusCode)
 			e.Int64(ECSEventDuration, time.Since(start).Nanoseconds())
-
-			e.Msgf("%d HTTP Request", wrCounter.statusCode)
+			e.Msg("HTTP Request")
```
michel-laterman, line 262: I've changed the middleware layer to log an info level message for each request if info is enabled.

The previous behaviour was to log an (info level) message if Debug is enabled, or the server responded in a non 2xx status

michel-laterman: It depends on the number of connected agents.
IIRC agents will long poll 10m by default, so we would expect a log every 10m per agent for checkins + additional entries for each action's acks

swiatekm: That sounds like potentially quite a bit. Could we rate limit or aggregate these based on return code?

swiatekm: FYI, I'm fine with the rest of this PR, it's just this change that seems potentially dangerous.

michel-laterman: actually the previous condition is broken; it would only output the (info level) message if debug is enabled.

michel-laterman: i'll revert & fix the condition; then make an issue for sampling logs

swiatekm (approved):

PR #4168, opened 2024-12-02
Replace most context.TODO calls, comment on other Background and TODOs

### LINKED ISSUE #3087: context.TODO and loggers

michel-laterman:

Replace use of context.TODO and context.Background from most use cases

## Design Checklist

## Checklist

## Related issues

- Closes #3087

### DIFF
--- a/changelog/fragments/1733164018-Replace-use-of-context.TODO.yaml
+++ b/changelog/fragments/1733164018-Replace-use-of-context.TODO.yaml
@@ -1,0 +1,32 @@
+# Kind can be one of:
+# - breaking-change: a change to previously-documented behavior
+# - deprecation: functionality that is being removed in a later release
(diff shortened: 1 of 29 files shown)


michel-laterman:
Looks like there is a flaky test: https://github.com/elastic/fleet-server/issues/4170

michel-laterman:
Integration test is flaky, fix in https://github.com/elastic/fleet-server/pull/4171

michel-laterman:
buildkite test this

PR #4171, opened 2024-12-03
Remove race condition when accessing remote bulker map

michel-laterman:
## What is the problem this PR solves?

Remove a race condition/bug that may occur when remote ES outputs are used.

## How does this PR solve the problem?

Use the `remoteOutputMutex` whenever accessing the `bulkerMap`.
Change `GetBulkerMap` to return a copy of the map so that remote output health will not conflict with adding/removing a bulker from the map.

## Design Checklist

- ~~I have ensured my design is stateless and will work when multiple fleet-server instances are behind a load balancer.~~
- ~~I have or intend to scale test my changes, ensuring it will work reliably with 100K+ agents connected.~~
- ~~I have included fail safe mechanisms to limit the load on fleet-server: rate limiting, circuit breakers, caching, load shedding, etc.~~

## Checklist

- [x] I have commented my code, particularly in hard-to-understand areas
- ~~I have made corresponding changes to the documentation~~
- ~~I have made corresponding change to the default configuration files~~
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] I have added an entry in `./changelog/fragments` using the [changelog tool](https://github.com/elastic/fleet-server#changelog)

## Related issues

- Closes #4170

Issue #4170, opened 2024-12-02
Test_Agent_Remote_ES_Output flaky due to race


michel-laterman:
## Flaky Test

* **Test Name:** Test_Agent_Remote_ES_Output
* **Link:** https://github.com/elastic/fleet-server/blob/main/internal/pkg/bulk/engine.go.
* **Branch:** main.
* **Notes:** remote bulker map can encounter a race between policy.Prepare output when doing a checkin or ack, and the policy self monitor (that checks for output health). 

https://buildkite.com/elastic/fleet-server/builds/5802#019388a2-40b1-42bf-8b6e-f1c2042fd80f

### Stack Trace

```
=== RUN   Test_Agent_Remote_ES_Output
--
  | remote_es_output_integration_test.go:147: Test fleet server port=37795
  | remote_es_output_integration_test.go:150: Create policy with remote ES output
  | remote_es_output_integration_test.go:181: Create API key and enrollment key for new policy
  | remote_es_output_integration_test.go:214: Enroll agent
  | fleet_integration_test.go:727: map[action:created item:map[access_api_key:X0lPemlKTUJqdThBeUFmVF9iSlU6TW5vQW8tN1VRQ0NjcGxVN3FJVXd6dw== access_api_key_id:_IOziJMBju8AyAfT_bJU actions:<nil> active:true enrolled_at:2024-12-02T18:48:37Z id:84c90f2e-67c0-4179-acfc-32cb048abdbf local_metadata:map[] policy_id:337d887c-628a-42cd-b99b-0c0e14384e68 status:online tags:[] type:PERMANENT user_provided_metadata:<nil>]]
  | remote_es_output_integration_test.go:40: Fake a checkin for agent 84c90f2e-67c0-4179-acfc-32cb048abdbf
  | ==================
  | WARNING: DATA RACE
  | Write at 0x00c0003f4360 by goroutine 3151:
  | runtime.mapassign_faststr()
  | /opt/buildkite-agent/.gvm/versions/go1.23.3.linux.amd64/src/runtime/map_faststr.go:223 +0x0
  | github.com/elastic/fleet-server/v7/internal/pkg/bulk.(*Bulker).updateBulkerMap()
  | /opt/buildkite-agent/builds/bk-agent-prod-gcp-1733164329918929620/elastic/fleet-server/internal/pkg/bulk/engine.go:148 +0xef
[...]

PR #4181, opened 2024-12-05
Prevent mapping explosion on logs

belimawr:

This commit remove some JSON objects that were added to the logs, thus preventing mapping explosion when those logs are ingested into Elasticsearch.

Some entries are converted to strings and others just removed to keep the log within a reasonable size.

Run Fleet-Server, watch the logs

## Design Checklist

## Checklist

- ~~[ ] I have added an entry in `./changelog/fragments` using the [changelog tool](https://github.com/elastic/fleet-server#changelog)~~

~~## Related issues~~


internal/pkg/server/fleet.go
```diff
@@ -254,25 +254,17 @@ func configCacheChanged(curCfg, newCfg *config.Config) bool {
...
-			Msg("fleet configuration has changed")
+		zlog.Info().Msg("fleet configuration has changed")
 	case !reflect.DeepEqual(curCfg.Output, newCfg.Output):
-		zlog.Info().
-			Interface("old", curCfg.Redact()).
-			Msg("output configuration has changed")
+		zlog.Info().Msg("output configuration has changed")
 	case !reflect.DeepEqual(curCfg.Inputs[0].Server, newCfg.Inputs[0].Server):
-		zlog.Info().
-			Interface("old", curCfg.Redact()).
```
michel-laterman, line 274: Can we log the new/old config as strings on the debug level instead?

belimawr:
CI is failing because of a known flaky test, https://github.com/elastic/fleet-server/issues/4170 that is addressed by https://github.com/elastic/fleet-server/pull/4171.

pkoutsovasilis:
> CI is failing because of a known flaky test, #4170 that is addressed by #4171.

Hey @belimawr 👋 I see that this https://github.com/elastic/fleet-server/pull/4171 is merged. Could you please merge the latest main in this PR to get a ✅ CI run? 🙂

belimawr:
> > CI is failing because of a known flaky test, #4170 that is addressed by #4171.
> 
> Hey @belimawr 👋 I see that this #4171 is merged. Could you please merge the latest main in this PR to get a ✅ CI run? 🙂

Done! Let's wait for CI.

internal/pkg/api/handleCheckin.go
```diff
@@ -1005,13 +1005,14 @@ func parseComponents(zlog zerolog.Logger, agent *model.Agent, req *CheckinReques
...
 		zlog.Trace().
-			RawJSON("oldComponents", agentComponentsJSON).
-			RawJSON("newComponents", *req.Components).
+			Str("oldComponents", string(agentComponentsJSON)).
+			Str("req.Components", string(reqComponentsJSON)).
 			Msg("local components data is not equal")
 
 		zlog.Info().
-			RawJSON("req.Components", *req.Components).
+			Str("req.Components", string(reqComponentsJSON)).
```
cmacknz, line 1015: Is there a reason for us to log the components JSON for every components change? Successful checkins will allow it to be queried from the .fleet-agents index at any time. This seems like something that would only be valuable to log whe there is an error.

At minimum this could be moved to the debug level, it doesn't make sense to log this at info by the unhealthy_reason at debug below, the unhealthy reason is much more interesting, but again it can be queried from .fleet-agents at any time.

belimawr: > Is there a reason for us to log the components JSON for every components change?

I'm not sure, honestly, I was just focusing on the mapping explosion. Someone from @elastic/fleet might be able to answer that.

The lines before already log the old and new components at trace level, so just removing them from this entry seems to be the best option, which will also avoid logging the same thing multiple times.

cmacknz (approved):

hp0620:
May I ask if this was packaged into 8.17.1? 
We have users running into the same issue with `elastic-cloud-logs` in the monitoring cluster which is in 8.17.1. 

I have checked the release notes, but I'm afraid I do not see this issue included in 8.16 nor 8.17, so I thought I would check here. 

Thank you!

belimawr:
No, this wasn't included in `v8.17.1` :/ It's not included in the `8.17` series.

However it is in the `v8.16.2` and following releases from the `8.16` series.

I'm backporting it to all missing active 8.x series.

PR #4256, opened 2024-12-30
Remove PGP endpoint auth requirement

### LINKED ISSUE #4255: pgp key retrieval requires auth
The PGP retrieval endpoint requires auth from the client:

https://github.com/elastic/fleet-server/blob/76abe7616100b629e5bbe5b8477a1912c418fb87/internal/pkg/api/handlePGPRequest.go#L54-L58

While this is defined behaviour as part of the OpenAPI spec

https://github.com/elastic/fleet-server/blob/76abe7616100b629e5bbe5b8477a1912c418fb87/model/openapi.yml#L1762-L1768

it should be removed as agents do not include an API key when attempting to get the PGP key

michel-laterman:

Air gapped agents are unable to retreive PGP key from fleet-server.

Remove auth key requirement from PGP retrieval endpoint

## Design Checklist

## Checklist

## Related issues

- Closes #4255

### DIFF
--- a/internal/pkg/api/handlePGPRequest.go
+++ b/internal/pkg/api/handlePGPRequest.go
@@ -15,11 +15,12 @@ import (
 	"os"
 	"path/filepath"
 
+	"github.com/rs/zerolog"
+	"go.elastic.co/apm/v2"
+
 	"github.com/elastic/fleet-server/v7/internal/pkg/bulk"
 	"github.com/elastic/fleet-server/v7/internal/pkg/cache"
 	"github.com/elastic/fleet-server/v7/internal/pkg/config"
-	"github.com/rs/zerolog"
-	"go.elastic.co/apm/v2"
 )
 
 const (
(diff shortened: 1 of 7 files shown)


internal/pkg/api/handlePGPRequest.go
```diff
@@ -52,10 +53,9 @@ func (pt *PGPRetrieverT) handlePGPKey(zlog zerolog.Logger, w http.ResponseWriter
 		return ErrTLSRequired
 	}
 	key, err := authAPIKey(r, pt.bulker, pt.cache)
-	if err != nil {
-		return err
+	if err == nil {
```
kaanyalti, line 56: Asking out of curiosity; what is the reason we are leaving the authAPIKey call in here?

kaanyalti (approved):
Changes look good to me. I just have a question about the authAPIKey call.

internal/pkg/api/handlePGPRequest.go
michel-laterman: in order to log the api key id if one is present

PR #4290, opened 2025-01-08
Add ability to enroll with a specific ID

### LINKED ISSUE #4226: Add ability to provide the agent-id in enroll API
**Describe the enhancement:**

Add the ability to specify the agent-id of the enrolling Elastic Agent.

On serverless, an Elastic Agent is static but the pod doesn't have any persistent storage so it cannot store the enrollment information between restarts of the Elastic Agent. There have also been other reports of this issue from customers where they do not need persistent storage from the integration and requiring the Elastic Agent to have it just for the enrollment information is not possible.

To provide a stable Elastic Agent in the agents list in Kibana, this would allow an Elastic Agent to enroll with the ID they want to have. This would also replace an existing Elastic Agent if one already has the same ID.

The new enrolled Elastic Agent will replace the previous Elastic Agent prevent it from being able to communicate with the Fleet Server any more.

**Describe any security issues:**

This does open the possibility that if a bad actor had the enrollment token and the ID of the Elastic Agent it would be able to enroll over top of it and prevent the communication of that current Elastic Agent as the other Elastic Agent would be come the newly communicating Elastic Agent.

To prevent this only an additional replace-token would be added to the enrollment API. This would be any unique value that is stored as a pbkdf2-sha512 hash on the Elastic Agent record. If an Elastic Agent is enrolled without this token then it doesn't allow any other Elastic Agent to enroll with the same ID (trying to enroll with the same ID would error). If an Elastic Agent is enrolled with the replace token and its the first enrollment then it would successfully enroll. On a second enrollment to replace the Elastic Agent the exact same replace token must be provided and if it matches (using pbkdf2-sha512 hash) then it would be considered the replacement of the Elastic Agent and allow the enrollment to complete.

blakerouse:

This solves an issue where an Elastic Agent is being replaced with a new Elastic Agent instance for the same host, pod, or workload. This allows the enrolling Elastic Agent to tell the ID that it wants to use, that ID can be currently in-use and this enrollment will take over the record of that Elastic Agent.

It solves the issue by taking a new `id` field in the enroll HTTP request. That `id` is then used as the Elastic Agent ID and determines if this is a new Elastic Agent or if it should take over an existing Elastic Agent record.

At the moment the integration tests are the best way to test this, as the ability to use this field has not been exposed yet on the Elastic Agent.

## Design Checklist

- [x] I have included fail safe mechanisms to limit the load on fleet-server: rate limiting, circuit breakers, caching, load shedding, etc. (already covered by enroll handle)

## Checklist

- ~~[ ] I have made corresponding change to the default configuration files~~ (no config changes)

## Related issues

- Closes #4226

### DIFF
--- a/internal/pkg/api/handleEnroll.go
+++ b/internal/pkg/api/handleEnroll.go
@@ -55,9 +55,10 @@ const kFleetAccessRolesJSON = `
 `
 
 var (
-	ErrUnknownEnrollType     = errors.New("unknown enroll request type")
-	ErrInactiveEnrollmentKey = errors.New("inactive enrollment key")
-	ErrPolicyNotFound        = errors.New("policy not found")
+	ErrUnknownEnrollType      = errors.New("unknown enroll request type")
+	ErrInactiveEnrollmentKey  = errors.New("inactive enrollment key")
+	ErrPolicyNotFound         = errors.New("policy not found")
+	ErrAgentIDInAnotherPolicy = errors.New("existing agent with ID in another policy")
 )
 
 type EnrollerT struct {
@@ -219,13 +220,6 @@ func (et *EnrollerT) _enroll(
 	}
 	now := time.Now()
 
-	// Generate an ID here so we can pre-create the api key and avoid a round trip
-	u, err := uuid.NewV4()
-	if err != nil {
-		return nil, err
-	}
-
-	agentID := u.String()
 	// only delete existing agent if it never checked in
 	if agent.Id != "" && agent.LastCheckin == "" {
 		zlog.Debug().
@@ -252,6 +246,68 @@ func (et *EnrollerT) _enroll(
 				Msg("Error when trying to delete old agent with enrollment id")
 			return nil, err
 		}
+		// deleted, so clear the ID so code below knows it needs to be created
+		agent.Id = ""
+	}
+
+	var agentID string
+	if req.Id != nil && *req.Id != "" {
+		agentID = *req.Id
+
+		// Possible this agent already exists.
+		vSpan, vCtx := apm.StartSpan(ctx, "checkAgentID", "validate")
+		var err error
+		agent, err = dl.FindAgent(vCtx, et.bulker, dl.QueryAgentByID, dl.FieldID, agentID)
+		if err != nil {
+			zlog.Debug().Err(err).
+				Str("ID", agentID).
+				Msg("Agent with ID not found")
+			if !errors.Is(err, dl.ErrNotFound) && !strings.Contains(err.Error(), "no such index") {
+				vSpan.End()
+				return nil, err
+			}
+		} else {
+			zlog.Debug().
+				Str("ID", agentID).
+				Msg("Agent with ID found")
+		}
+		vSpan.End()
+
+		if agent.Id != "" {
+			// confirm that its on the same policy
+			// it is not supported to have it the same ID enroll into different policies
+			if agent.PolicyID != policyID {
+				zlog.Warn().
+					Str("AgentId", agent.Id).
+					Str("PolicyId", policyID).
+					Str("CurrentPolicyId", agent.PolicyID).
+					Msg("Existing agent with same ID already enrolled into another policy")
+				return nil, ErrAgentIDInAnotherPolicy
+			}
+
+			// invalidate the previous api key
+			// this has to be done because its not possible to get the previous token
+			// so the other is invalidated and a new one is generated
+			zlog.Debug().
+				Str("AgentId", agent.Id).
+				Str("APIKeyID", agent.AccessAPIKeyID).
+				Msg("Invalidate old api key with same id")
+			err := invalidateAPIKey(ctx, zlog, et.bulker, agent.AccessAPIKeyID)
+			if err != nil {
+				zlog.Error().Err(err).
+					Str("AgentId", agent.Id).
+					Str("APIKeyID", agent.AccessAPIKeyID).
+					Msg("Error when trying to invalidate API key of old agent with same id")
+				return nil, err
+			}
+		}
+	} else {
+		// No ID provided so generate an ID.
+		u, err := uuid.NewV4()
+		if err != nil {
+			return nil, err
+		}
+		agentID = u.String()
 	}
 
 	// Update the local metadata agent id
@@ -271,47 +327,84 @@ func (et *EnrollerT) _enroll(
 		return invalidateAPIKey(ctx, zlog, et.bulker, accessAPIKey.ID)
 	})
 
-	agentData := model.Agent{
-		Active:         true,
-		PolicyID:       policyID,
-		Namespaces:     namespaces,
-		Type:           string(req.Type),
-		EnrolledAt:     now.UTC().Format(time.RFC3339),
-		LocalMetadata:  localMeta,
-		AccessAPIKeyID: accessAPIKey.ID,
-		ActionSeqNo:    []int64{sqn.UndefinedSeqNo},
-		Agent: &model.AgentMetadata{
+	// Existing agent, only update a subset of the fields
+	if agent.Id != "" {
+		agent.Active = true
+		agent.Namespaces = namespaces
+		agent.LocalMetadata = localMeta
+		agent.AccessAPIKeyID = accessAPIKey.ID
+		agent.Agent = &model.AgentMetadata{
 			ID:      agentID,
 			Version: ver,
-		},
-		Tags:         removeDuplicateStr(req.Metadata.Tags),
-		EnrollmentID: enrollmentID,
-	}
+		}
+		agent.Tags = removeDuplicateStr(req.Metadata.Tags)
+		agentField, err := json.Marshal(agent.Agent)
+		if err != nil {
+			return nil, fmt.Errorf("failed to marshal agent to JSON: %w", err)
+		}
+		// update the agent record
+		// clears state of policy revision, as this agent needs to get the latest policy
+		// clears state of unenrollment, as this is a new enrollment
+		doc := bulk.UpdateFields{
+			dl.FieldActive:                true,
+			dl.FieldNamespaces:            namespaces,
+			dl.FieldLocalMetadata:         json.RawMessage(localMeta),
+			dl.FieldAccessAPIKeyID:        accessAPIKey.ID,
+			dl.FieldAgent:                 json.RawMessage(agentField),
+			dl.FieldTags:                  agent.Tags,
+			dl.FieldPolicyRevisionIdx:     0,
+			dl.FieldAuditUnenrolledTime:   nil,
+			dl.FieldAuditUnenrolledReason: nil,
+			dl.FieldUnenrolledAt:          nil,
+			dl.FieldUnenrolledReason:      nil,
+			dl.FieldUpdatedAt:             time.Now().UTC().Format(time.RFC3339),
+		}
+		err = updateFleetAgent(ctx, et.bulker, agentID, doc)
+		if err != nil {
+			return nil, err
+		}
+	} else {
+		agent = model.Agent{
+			Active:         true,
+			PolicyID:       policyID,
+			Namespaces:     namespaces,
+			Type:           string(req.Type),
+			EnrolledAt:     now.UTC().Format(time.RFC3339),
+			LocalMetadata:  localMeta,
+			AccessAPIKeyID: accessAPIKey.ID,
+			ActionSeqNo:    []int64{sqn.UndefinedSeqNo},
+			Agent: &model.AgentMetadata{
+				ID:      agentID,
+				Version: ver,
+			},
+			Tags:         removeDuplicateStr(req.Metadata.Tags),
+			EnrollmentID: enrollmentID,
+		}
 
-	err = createFleetAgent(ctx, et.bulker, agentID, agentData)
-	if err != nil {
-		return nil, err
+		err = createFleetAgent(ctx, et.bulker, agentID, agent)
+		if err != nil {
+			return nil, err
+		}
+		// Register delete fleet agent for enrollment error rollback
+		rb.Register("delete agent", func(ctx context.Context) error {
+			return deleteAgent(ctx, zlog, et.bulker, agentID)
+		})
 	}
-
-	// Register delete fleet agent for enrollment error rollback
-	rb.Register("delete agent", func(ctx context.Context) error {
-		return deleteAgent(ctx, zlog, et.bulker, agentID)
-	})
 
 	resp := EnrollResponse{
 		Action: "created",
 		Item: EnrollResponseItem{
 			AccessApiKey:         accessAPIKey.Token(),
-			AccessApiKeyId:       agentData.AccessAPIKeyID,
-			Active:               agentData.Active,
-			EnrolledAt:           agentData.EnrolledAt,
+			AccessApiKeyId:       agent.AccessAPIKeyID,
+			Active:               agent.Active,
+			EnrolledAt:           agent.EnrolledAt,
 			Id:                   agentID,
-			LocalMetadata:        agentData.LocalMetadata,
-			PolicyId:             agentData.PolicyID,
+			LocalMetadata:        agent.LocalMetadata,
+			PolicyId:             agent.PolicyID,
 			Status:               "online",
-			Tags:                 agentData.Tags,
-			Type:                 agentData.Type,
-			UserProvidedMetadata: agentData.UserProvidedMetadata,
+			Tags:                 agent.Tags,
+			Type:                 agent.Type,
+			UserProvidedMetadata: agent.UserProvidedMetadata,
 		},
 	}
 
@@ -468,6 +561,17 @@ func updateLocalMetaAgentID(data []byte, agentID string) ([]byte, error) {
 	}
 
 	return data, nil
+}
+
+func updateFleetAgent(ctx context.Context, bulker bulk.Bulk, id string, doc bulk.UpdateFields) error {
+	span, ctx := apm.StartSpan(ctx, "updateAgent", "update")
+	defer span.End()
+
+	body, err := doc.Marshal()
+	if err != nil {
+		return err
+	}
+	return bulker.Update(ctx, dl.FleetAgents, id, body, bulk.WithRefresh(), bulk.WithRetryOnConflict(3))
 }
 
 func createFleetAgent(ctx context.Context, bulker bulk.Bulk, id string, agent model.Agent) error {
--- a/model/openapi.yml
+++ b/model/openapi.yml
@@ -139,6 +139,13 @@ components:
         - type
         - metadata
       properties:
+        id:
+          type: string
+          description: |
+            The ID of the agent.
+            This is the ID that will be used to reference this agent, if no ID is passed one will be generated.
+            If another agent is enrolled with the same ID the other agent will no longer be able to communicate,
+            this new agent is considered a replacement of the other agent.
         type:
           description: |
             The enrollment type of the agent.
--- a/internal/pkg/api/error.go
+++ b/internal/pkg/api/error.go
@@ -101,6 +101,15 @@ func NewHTTPErrResp(err error) HTTPErrResp {
 			},
 		},
 		{
+			ErrAgentIDInAnotherPolicy,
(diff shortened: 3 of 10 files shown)


jlind23:
After chatting with @blakerouse over slack we should fail the enrollment of this "new" agent if the policy has temper protection enabled.

kaanyalti:
Changes so far look good to me, waiting for updates relevant to @jlind23's comment above to approve

model/openapi.yml
```diff
@@ -139,6 +139,13 @@ components:
         - type
         - metadata
       properties:
+        id:
+          type: string
+          description: |
+            The ID of the agent.
+            This is the ID that will be used to reference this agent, if no ID is passed one will be generated.
+            If another agent is enrolled with the same ID the other agent will no longer be able to communicate,
+            this new agent is considered a replacement of the other agent.
```
michel-laterman, line 148: This is a pretty big change from our discussion.
Please also add a sentence saying the (replaced) agent will still be able to send data into ES

blakerouse: It is a slight change, because it is not possible to get an API key token again after the initial create. That made me have no choice to change the behavior.

blakerouse: I did add more as requested to the description of this field to inform that it will still allow data to flow.

blakerouse:
@jlind23 @michel-laterman @kaanyalti I have updated this PR based on the discuss I had with @jlind23 about security with this feature. This PR now includes an additional `replace_token` during the enrollment API. I have updated the PR description to describe this as well as the API specification describes it.

internal/pkg/api/handleEnroll.go
```diff
@@ -252,6 +260,105 @@ func (et *EnrollerT) _enroll(
...
+			// delete the current record as the new enrollment will overwrite this one
+			zlog.Debug().
+				Str("ID", agentID).
+				Msg("Inactive agent with ID found")
+			err = deleteAgent(ctx, zlog, et.bulker, agent.Id)
+			if err != nil {
+				zlog.Error().Err(err).
+					Str("AgentId", agent.Id).
+					Msg("Error when trying to delete old agent with same id")
+				return nil, err
```
pkoutsovasilis, line 294: missing `vSpan.End()` right?

pkoutsovasilis:
Except a small ending of a trace-span the code changes look good to me. I understand the potential pitfalls with this feature and definitely see how the `replace_token` helps in minimising some of them but still this feature to me serves only special-case scenarios and is not streamlined usage 🙂

internal/pkg/api/handleEnroll.go
```diff
@@ -197,10 +201,22 @@ func (et *EnrollerT) _enroll(
...
+	var replaceToken string
 
 	span, ctx := apm.StartSpan(ctx, "enroll", "process")
 	defer span.End()
 
+	now := time.Now()
+
+	if req.ReplaceToken != nil && *req.ReplaceToken != "" {
+		replaceTokenBytes, err := bcrypt.GenerateFromPassword([]byte(*req.ReplaceToken), bcrypt.DefaultCost)
+		if err != nil {
```
michalpristas, line 213: should we log error, looking at implementation of GenerateFromPassword it can be miriad of things

internal/pkg/api/handleEnroll.go
michalpristas: we could extract to separate function that finds agent id with defer vspan.End to decrease probability of this in a future

internal/pkg/api/handleEnroll.go
```diff
@@ -252,6 +260,105 @@ func (et *EnrollerT) _enroll(
...
+				return nil, ErrAgentNotReplaceable
+			}
+			if req.ReplaceToken == nil || *req.ReplaceToken == "" {
+				zlog.Warn().
+					Str("AgentId", agent.Id).
+					Msg("Existing agent with same ID already enrolled; no replace token given during enrollment")
+				return nil, ErrAgentNotReplaceable
+			}
+			err = bcrypt.CompareHashAndPassword([]byte(agent.ReplaceToken), []byte(*req.ReplaceToken))
+			if err != nil {
```
michalpristas, line 322: could we log error in case `err != bcrypt.ErrMismatchedHashAndPassword`

internal/pkg/api/handleEnroll.go
```diff
@@ -252,6 +260,105 @@ func (et *EnrollerT) _enroll(
...
+
+			// confirm that its on the same policy
+			// it is not supported to have it the same ID enroll into different policies
+			if agent.PolicyID != policyID {
+				zlog.Warn().
+					Str("AgentId", agent.Id).
+					Str("PolicyId", policyID).
+					Str("CurrentPolicyId", agent.PolicyID).
+					Msg("Existing agent with same ID already enrolled into another policy")
+				return nil, ErrAgentInAnotherPolicy
```
michalpristas, line 336: very minor: i wonder if this is not telling too much, attacker could try agentID and token. this would tell token is correct, policy is not. Maybe returning ErrAgentNotReplaceable with log unchanged would make more sense?

internal/pkg/api/handleEnroll.go
blakerouse: Added a log message.

internal/pkg/api/handleEnroll.go
blakerouse: Added log.

blakerouse:
@pkoutsovasilis @michalpristas I updated the PR with the request fixes. Thanks for the reviews.

internal/pkg/api/handleEnroll.go
```diff
@@ -271,47 +355,84 @@ func (et *EnrollerT) _enroll(
...
+			dl.FieldLocalMetadata:         json.RawMessage(localMeta),
+			dl.FieldAccessAPIKeyID:        accessAPIKey.ID,
+			dl.FieldAgent:                 json.RawMessage(agentField),
+			dl.FieldTags:                  agent.Tags,
+			dl.FieldPolicyRevisionIdx:     0,
+			dl.FieldAuditUnenrolledTime:   nil,
+			dl.FieldAuditUnenrolledReason: nil,
+			dl.FieldUnenrolledAt:          nil,
+			dl.FieldUnenrolledReason:      nil,
+			dl.FieldUpdatedAt:             time.Now().UTC().Format(time.RFC3339),
```
pkoutsovasilis, line 387: nit; here we use `time.Now()` while below we use the top declared variable `now`, is there a reason these two are using the same time?

blakerouse: Nice catch, missed that it was above. I updated it to use the `now` variable from above.

pkoutsovasilis (approved):
LGTM

internal/pkg/api/handleEnroll.go
```diff
@@ -252,6 +261,81 @@ func (et *EnrollerT) _enroll(
...
+		agentID = *req.Id
+
+		// check if the agent with this ID already exists
+		var err error
+		agent, err = et._checkAgent(ctx, zlog, agentID)
+		if err != nil {
+			return nil, err
+		}
+
+		if agent.Id != "" {
```
michalpristas, line 279: can it return `agent.id == ""`? if not this if statement is not needed.
if so. we dont have this case handled as this should not be the same as empty ID when `req.ID` is not used.
we whould probably not continue with empty id, probably we should fail. generating a new one breaks the purpose of providing it via req.id

michalpristas (approved):
Thanks for resolving comment. one question about behavior other than that I'm ok with the change. 
i let you decide how you want to address the point i raised in this iteration

internal/pkg/api/handleEnroll.go
blakerouse: It absolutely will return an `agent.id == ""`. This happens when we check to see if an existing agent already exists with that ID. The ErrNotFound will not be returned from `_checkAgent`, it will return `nil` error and this will be `agent.id == "".

kaanyalti (approved):

internal/pkg/api/handleEnroll.go
```diff
@@ -197,10 +200,24 @@ func (et *EnrollerT) _enroll(
...
 	var enrollmentID string
+	var replaceToken string
 
 	span, ctx := apm.StartSpan(ctx, "enroll", "process")
 	defer span.End()
 
+	now := time.Now()
+
+	if req.ReplaceToken != nil && *req.ReplaceToken != "" {
+		replaceTokenBytes, err := bcrypt.GenerateFromPassword([]byte(*req.ReplaceToken), bcrypt.DefaultCost)
```
cmacknz, line 211: I don't think the blowfish hash in the underlying implementation here meets FIPS 140-2. At least for the MS Go fork it wouldn't hit the FIPS certified OpenSSL so would be non-compliant. The list of what gets forwarded to it is at https://github.com/microsoft/go/blob/microsoft/main/eng/doc/fips/UserGuide.md

Whatever crypto primitives we use need to be FIPS compliant.

CC @michel-laterman

pkoutsovasilis: That's actually more than a valid point @cmacknz 👍 AFAIK the only FIPS compliant password hashing is PBKDF2. I had a look and all call in it are forwarded by MS Go fork. We already use PBKDF2 password hashing in [elastic-agent](https://github.com/elastic/elastic-agent/blob/main/internal/pkg/crypto/io.go#L376) and I believe we should do the same here as well

cmacknz (changes requested):
Requesting changes until we get a FIPS compliant implementation. We're blocked on security review anyway.

blakerouse:
@cmacknz @pkoutsovasilis Thanks for catching this now for FIPS. I have updated the implementation to use PBKDF2 and SHA512 like the Elastic Agent. Let me know what you think.

internal/pkg/api/handleEnroll.go
```diff
@@ -538,3 +711,56 @@ func validateRequest(ctx context.Context, data io.Reader) (*EnrollRequest, error
...
+		return false, ErrAgentCorrupted
+	}
+	encoded, err := base64.RawStdEncoding.DecodeString(tokens[4])
+	if err != nil {
+		// hash invalid format
+		zlog.Error().Err(err).Msg("replace_token hash failed to base64 decode encoded")
+		return false, ErrAgentCorrupted
+	}
+	key := pbkdf2.Key([]byte(token), salt, iterations, cfg.KeyLength, sha512.New)
+	return bytes.Equal(key, encoded), nil
```
pkoutsovasilis, line 749:
```suggestion
	return hmac.Equal(key, encoded), nil
```

internal/pkg/config/pbkdf2.go
```diff
@@ -0,0 +1,42 @@
...
+	"errors"
+)
+
+type PBKDF2 struct {
+	Iterations int `config:"iterations"`
+	KeyLength  int `config:"key_length"`
+	SaltLength int `config:"salt_length"`
+
+	// BlockSize must be a factor of aes.BlockSize
+	BlockSize int `config:"block_size"`
```
pkoutsovasilis, line 18:
```suggestion
```
is this actually used somewhere?

internal/pkg/config/pbkdf2.go
```diff
@@ -0,0 +1,42 @@
...
+	if p.SaltLength == 0 {
+		return errors.New("salt_length must be superior to 0")
+	}
+	return nil
+}
+
+// InitDefaults is the default options to use with PDKDF2, changing might decrease
+// the efficacy of the encryption.
+func (p *PBKDF2) InitDefaults() {
+	p.Iterations = 10000
```
pkoutsovasilis, line 38: fact: In 2023, OWASP recommended to use [210,000](https://cheatsheetseries.owasp.org/cheatsheets/Password_Storage_Cheat_Sheet.html#pbkdf2) for PBKDF2-HMAC-SHA512. The more iterations the more time it will take to compute the hash, but not quite sure if we should follow this advise or not 🙂

blakerouse: @pkoutsovasilis I was just using the defaults that the Elastic Agent uses so they can be inline. Should we also update the Elastic Agent.

https://github.com/elastic/elastic-agent/blob/main/internal/pkg/crypto/io.go#L58

internal/pkg/config/pbkdf2.go
blakerouse: I copied the configuration from Elastic Agent. I will remove that as that is not used here.

internal/pkg/api/handleEnroll.go
blakerouse: Why `hmac.Equal`? Whats the reason? Would like to add that in the code comment so its clear. Its not clear to my why that is used here.

internal/pkg/config/pbkdf2.go
pkoutsovasilis: yy I ve seen that, however, I am not entirely sure how easy of a change is for the elastic-agent without breaking anything. Need to look into that 🙂

blakerouse: Updated here at least.

internal/pkg/config/pbkdf2.go
blakerouse: Removed

internal/pkg/api/handleEnroll.go
blakerouse: Got it, thanks for the details. I have updated it to use `hmac.Equal` and added a comment.

pkoutsovasilis (approved):

model/openapi.yml
```diff
@@ -139,6 +139,14 @@ components:
         - type
         - metadata
       properties:
+        id:
+          type: string
+          description: |
+            The ID of the agent.
+            This is the ID that will be used to reference this agent, if no ID is passed one will be generated.
+            If another agent is enrolled with the same ID the other agent will no longer be able to communicate,
+            this new agent is considered a replacement of the other agent. The other agent will be able to continue
```
cmacknz, line 148: > The other agent will be able to continue sending data to ES.

How do the ES API keys eventually get revoked? This normally happens through unenrolling an agent via the UI, but with replacement of the agent, you lose that.

Should replacing an agent revoke it's ES API keys to treat it like a force unenroll but triggered through fleet server?

internal/pkg/api/handleEnroll.go
```diff
@@ -252,6 +253,85 @@ func (et *EnrollerT) _enroll(
...
+			}
+
+			// invalidate the previous api key
+			// this has to be done because it's not possible to get the previous token
+			// so the other is invalidated and a new one is generated
+			zlog.Debug().
+				Str("AgentId", agent.Id).
+				Str("APIKeyID", agent.AccessAPIKeyID).
+				Msg("Invalidate old api key with same id")
+			err = invalidateAPIKey(ctx, zlog, et.bulker, agent.AccessAPIKeyID)
```
cmacknz, line 319: Does any of the state in the .fleet-agents document need to be reset? For example the new agent is not going to have the same persisted action state or .fleet-action sequence number, and will not yet have a copy of the policy. 

If we do any state resets there, it has to happen after the API key confirmed invalidated and not just when it is requested to avoid race conditons.

internal/pkg/api/handleEnroll.go
```diff
@@ -252,6 +253,85 @@ func (et *EnrollerT) _enroll(
...
+				// provides no real reason as that would expose too much information
+				zlog.Debug().
+					Str("AgentId", agent.Id).
+					Msg("Existing agent with same ID already enrolled; replace token didn't match")
+				return nil, ErrAgentNotReplaceable
+			}
+
+			// confirm that its on the same policy
+			// it is not supported to have it the same ID enroll into different policies
+			if agent.PolicyID != policyID {
```
cmacknz, line 303: There is an unlikely race condition where a user assigns the agent to a new policy as this replacement is happening. The Fleet server state isn't guaranteed to be current and we don't have any kind of lock or transaction in the index to help us avoid this type of problem.

cmacknz: A user could unenroll the agent as this function is executing to enroll it again with a new API key is another weird situation that could arise.

cmacknz (commented):
The implementation LGTM, seems FIPS compliant AFAIK. You could use the Go 1.24 release candidate to double check this, it has a mode that causes non-FIPS crypto to return errors.

I think there are some interesting but unlikely concurrency problems that could be introduced here around:
1. The agent you are replacing being in the process of checking in as you are replacing it. Possibly this gives it a chance to update .fleet-agents information on you. I'm not sure this matters.
2. User actions on the UI side of the .fleet-agents index being in conflict with the re-enrollment here. The user unenrolling the agent you are replacing concurrently is probably the most interesting case.
3. The new, replaced agent having it's action state reset to zero.  Maybe this doesn't matter, but we should test it. Will the agent resetting it's action sequence number to 0 have any unexpected effects?

model/openapi.yml
cmacknz: I think I maybe misunderstood what happens the first time, the ES API keys are tied to the agent ID so they are kept alive because they are reused.

So this is documenting that this feature only works as expected if the agent you are replacing stops running. Otherwise you will have two agents ingesting data but one of them is unmanaged and has had its Fleet API key revoked. Correct?

levinebw:
noting, infosec security review discussion here, https://github.com/elastic/fleet-server/issues/4226

model/openapi.yml
blakerouse: That is correct. It will be able to keep ingesting data, which we want for a short period, but will not be able to communicate with Fleet. Then its up to the deployer to turn it off when the other one is working.

This works for Kubernetes, as it will allow a new pod to spawn while the previous version is still running. Once the new version is up and running and healthy then the other pod will be stopped.

internal/pkg/api/handleEnroll.go
blakerouse: The `.fleet-action` sequence number I do not reset, as I don't think we want this Elastic Agent to replay all actions of the previous Elastic Agent. If this is used outside of the container use-case, you would not expect it to roll through performing an upgrade action because it replaced a previous Elastic Agent.

I took this into account and didn't reset the sequence number for that exact reason. The policy and revision is not tied to the sequence number in anyway. It is actually stored directly on the `.fleet-agents` document, which is reset by this code here - https://github.com/elastic/fleet-server/pull/4290/files#diff-b3d2d3a49ef214fba08e3aa10772162cbb65c378e24ef6c4a0c5df65b051f71aR378

blakerouse:
Most of my responses to your inline comments are related here, but I will re-iterate.

> 
> 1. The agent you are replacing being in the process of checking in as you are replacing it. Possibly this gives it a chance to 
update .fleet-agents information on you. I'm not sure this matters.

Luckily the way Fleet Server works it only updates fields on a document and not a whole document. Because of this I am not worried as much of all the fields being reset to the wrong value for check-in. Looking at the code the following fields are updated on check-in:

https://github.com/elastic/fleet-server/blob/main/internal/pkg/checkin/bulk.go#L227

I don't see any conflicts there that would cause an issue for the newly replaced Elastic Agent to be able to successfully check-in and update that information. The previous Elastic Agent that just checked-in will not be able from that point forward.

> 2. User actions on the UI side of the .fleet-agents index being in conflict with the re-enrollment here. The user unenrolling the agent you are replacing concurrently is probably the most interesting case.

Not an issue in agentless, as it cannot be unenrolled. I do think outside of agentless there is a small window where this could get in a bad state. I think it is a very small window, but I don't know if the effort involved to prevent that window completely is worth it currently.

> 3. The new, replaced agent having it's action state reset to zero.  Maybe this doesn't matter, but we should test it. Will the agent resetting it's action sequence number to 0 have any unexpected effects?

This is not reseting the action state, see my inline comment. It is just reseting the policy revision to ensure that on its first check-in that it gets it latest policy. I think if we solve this issue here, we don't even need to reset this here - https://github.com/elastic/elastic-agent/issues/6446

cmacknz (approved):

internal/pkg/api/handleEnroll.go
blakerouse: I don't think we really document specific flows in Fleet Server. Those flows I believe are normally document publicly as Elastic Agent documentation. Let me know where you think I should put this information and I can do that.

PR #4318, opened 2025-01-13
FIPS compliant builds

michel-laterman:

Add ability to make FIPS compliant builds when `FIPS=true` is specified.

Specifying `FIPS=true` will change the following:
- `PLATFORMS` will default to `linux/amd64 linux/arm64`
- `make local`, `make release-*` - Binary will be build with `-tags=fipsrequired` and `GOEXPERIMENT=systemcrypto`
- `make build-releaser` - chaingaurd microsoft go image will be used as base
- `make multipass` - microsoft's go toolchain will be downloaded and installed to VM

## Design Checklist

## Checklist


Makefile
```diff
@@ -64,6 +64,14 @@ GOBIN=$(shell go env GOPATH)/bin/
 
 OS_NAME:=$(shell uname -s)
 
+# Set FIPS=true to force FIPS compliance when building
+FIPS?=
+ifeq "${FIPS}" "true"
+BUILDER_IMAGE=fleet-server-fips-builder:${GO_VERSION}
+PLATFORMS = linux/amd64 linux/arm64
```
michel-laterman, line 71: I'm working off an ARM macbook.

When I try to compile these targets in the image built by Dockerfile.fips (base stage) by running `FIPS=true make docker-release`, I get `gcc: error: unrecognized command-line option '-m64'` when it tries to compile for linux/amd64. (linux/arm64 compiles as expected if platforms are re-ordered)

However if I run the microsoft/golang base dockerfile with `--platform=linux/amd64` I can compile for linux/amd64 with no issue, and I can compile for `linux/arm64` if I add `CC=x86_64-linux-gnu-gcc`.

cmacknz: This is all expected if the host gcc was only built to target the host machine architecture, which I think is typical for gcc. `gcc -dumpmachine` will tell you the target triple, it'd be weird if it were also `x86_64-linux-gnu-gcc`. Probably there is some series of command line options that can dump out the supported architectures.

The work arounds are the ones you've listed here:
1. Run the container with --platform=linux/amd64 to have your ARM Mac emulate x86. This is the slowest option.
2. Explicitly specify a cross toolchain. This is the fastest option because you don't have to emulate. You can likely select the correct toolchain conditionally, this is what https://github.com/elastic/beats/pull/40330/files did to avoid emulation by default.

The CI machines will never have to cross compile because you can always use one with the correct host architecture.

blakerouse (approved):
Nothing stands out that worries me in this PR. Being it has to directly be enabled to build, this looks good.

docs/fips.md
```diff
@@ -0,0 +1,105 @@
+# FIPS support
```
michalpristas, line 1: 💙

michalpristas (approved):

PR #4328, opened 2025-01-16
Allow use of version qualifier for staging artifacts

pchila:
## What is the problem this PR solves?

This commits adds support for an optional $VERSION_QUALIFIER env var/build option in the packaging pipeline to allow building prerelease artifacts from staging.

## How does this PR solve the problem?

* Support VERSION_QUALIFIER optional env var for prerelease builds
* Refactor packaging pipeline to allow staging builds from main (that indirectly allow testing this PR too with the additional `DRA_BRANCH` env var and skip snapshot builds when VERSION_QUALIFIER is present
* Remove unused VERSION and INFRA_REPO variables.

## How to test this PR locally

Using this PR and the pipeline https://buildkite.com/elastic/fleet-server-package-mbp with `pull/4328/merge` and options: 

```
DRA_BRANCH="main"
VERSION_QUALIFIER="alpha1"
```

results in a successful 9.0.0-alpha1 artifacts, example: https://buildkite.com/elastic/fleet-server-package-mbp/builds/1526

image

Also a test without VERSION_QUALIFIER https://buildkite.com/elastic/fleet-server-package-mbp/builds/1525 produces the expected artifacts (note that when branch is main, i.e. not when testing using this PR, **and** there is no version qualifier defined, the staging dra publish is skipped, per design).

## Related issues

Closes:

- https://github.com/elastic/fleet-server/issues/4325
- https://github.com/elastic/ingest-dev/issues/4859

PR #4337, opened 2025-01-17
Don't run staging builds on main (unless it's a prerelease)

dliappis:

This commit is a follow-up to #4328 and fixes a bug for main where staging main artifacts will be products for every merge commit, thus overriding manual builds using VERSION_QUALIFIER.

## Related issues

https://github.com/elastic/ingest-dev/issues/4859


pchila (approved):
LGTM

dliappis:
Closing a no needed -- the DRA staging step is skipped in this case.

PR #4338, opened 2025-01-17
Allow staging build on main with version qualifier

dliappis:

This commit fixes a bug in #4328 running snapshot instead of staging for x86-64 and also ensures staging on main can be build when
 VERSION_QUALIFIER is set.

## Related issues

- Closes https://github.com/elastic/ingest-dev/issues/4859


pchila (approved):
LGTM

PR #4351, opened 2025-01-23
ci(buildkite): exclude files/folders that are not tested in Buildkite

v1v:

Run faster builds and avoid waste of CI cycles.

I decided to support the whole .github folder since some other files such as CODEOWNERS or .dependaot.yml and the rest are not actually used by Buildkite.

I took some of the files that we skipped in https://github.com/elastic/elastic-agent/pull/4836. I don't know the reason it was not configured in this repository, too, but I think it's the same case.

## Design Checklist

## Checklist

## Related issues

Relates https://github.com/elastic/beats/pull/42404


rowlandgeoff (approved):
👍

cmacknz (approved):

PR #4367, opened 2025-01-29
Bump ristretto to v0.2.0

swiatekm:

This fixes  CVE-2024-45339.

By updating ristretto to a version that doesn't depend on glog.


pkoutsovasilis (approved):
code changes LGTM just update the NOTICE file 🙂

PR #4402, opened 2025-02-03
Use chi/middleware.Throttle instead of custom conn limiter

### LINKED ISSUE #4200: Fleet-Server abruptly closes connection when limit is reached instead of returning TooManyRequests (429)
When `limits.max_connections` is set, Fleet-Server will abruptly close connections, which makes the clients (usually Elastic-Agent) to get a `read: connection reset by peer`. This is not helpful and does not enable the client to correctly adapt its behaviour.

The problem comes from the `limitListener` (L 69): https://github.com/elastic/fleet-server/blob/8ff01e35a155abde455443039616c5069848c488/internal/pkg/limit/listener.go#L55-L80

If there is a proxy in front of Fleet-Server this can cause a very hard situation to debug from the client's (e.g: Elastic-Agent) perspective: Fleet Server gets a connection that is over the limit, closes the connection, which results in an EOF for the proxy, which translates it into a 502, which the elastic-agent enrol command swallows and does not display.

michel-laterman:

fleet-server silently closes connections instead of returning a 429 when the connection limit is reached.

Use [chi/middleware.Throttle](https://pkg.go.dev/github.com/go-chi/chi/middleware#Throttle) to track in-flight requests and return a 429 response once the limit is reached.

This is different behaviour from our current implementation where the check occurs when the TCP connection is established, before the TLS handshake occurs.

## Design Checklist

## Checklist

## Related issues

- Closes #4200


michel-laterman:
buildkite run perf-tests

michel-laterman:
buildkite run perf-tests

blakerouse (approved):
Looks good, less code nice!

PR #4403, opened 2025-02-04
Use --develop flag for agent install e2e tests

michel-laterman:

It's hard to run e2e tests on a host that has elastic-agent installed.

Use the `--develop` flag to install an agent in the development namespace

Change version in `dev-tools/integration/.env` to `9.1-SNAPSHOT`

`make test-e2e`

## Design Checklist

### DIFF
--- a/dev-tools/integration/.env
+++ b/dev-tools/integration/.env
@@ -1,6 +1,6 @@
 # If you use change this version without a pinned one, please update
 # .ci/bump-elastic-stack-snapshot.yml or .github/workflows/bump-golang.yml
-ELASTICSEARCH_VERSION=9.0.0-a38910c2-SNAPSHOT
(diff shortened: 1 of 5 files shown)


cmacknz (approved):

dev-tools/integration/.env
```diff
@@ -1,6 +1,6 @@
 # If you use change this version without a pinned one, please update
 # .ci/bump-elastic-stack-snapshot.yml or .github/workflows/bump-golang.yml
-ELASTICSEARCH_VERSION=9.0.0-a38910c2-SNAPSHOT
+ELASTICSEARCH_VERSION=9.1.0-ae17f1a1-SNAPSHOT
```
michel-laterman, line 3: When backporting this PR we will need to remove this change

PR #4514, opened 2025-02-25
Add unit test for fips metadata attribute in enroll/checkin

michel-laterman:

Add unit tests as a sanity check to make sure that fleet-server is able to handle the new `fips` metadata attribute.

fleet-server uses `json.RawMessage` when interacting with local metadata (in the enroll and checkin endpoints) as it just needs to compare byte values to see if the agent doc needs to be changed. The new tests verify that local metadata changes are propagated as expected.

## Design Checklist


blakerouse (approved):
Looks good, thanks for the added testing.

swiatekm (approved):

Issue #4525, opened 2025-02-27
[Bug]🐛 Size computation for allocation may overflow


odaysec:
https://github.com/elastic/fleet-server/blob/ba68a24383abfa328cce0cc3eddce1b73b497d8a/internal/pkg/danger/buf.go#L36-L36

Performing calculations involving the size of potentially large strings or slices can result in an overflow (for signed integer types) or a wraparound (for unsigned types). An overflow causes the result of the calculation to become negative, while a wraparound results in a small (positive) number. This can cause further issues. If, for the result is then used in an allocation, it will cause a runtime panic if it is negative, and allocate an unexpectedly small buffer otherwise.

## POC
In the following vulnerable, assume that there is a function `encryptBuffer` that encrypts byte slices whose length must be padded to be a multiple of 16. The function `encryptValue` provides a convenience wrapper around this function: when passed an arbitrary value, it first encodes that value as JSON, pads the resulting byte slice, and then passes it to `encryptBuffer`.
```go
package main

import "encoding/json"

func encryptValue(v interface{}) ([]byte, error) {
	jsonData, err := json.Marshal(v)
	if err != nil {
		return nil, err
	}
	size := len(jsonData) + (len(jsonData) % 16)
	buffer := make([]byte, size)
	copy(buffer, jsonData)
	return encryptBuffer(buffer)
}
```
When passed a value whose JSON encoding is close to the maximum value of type `int` in length, the computation of `size` will overflow, producing a negative value. When that negative value is passed to `make`, a runtime panic will occur.

To guard against this, the function should be improved to check the length of the JSON-encoded value. here is a version of `encryptValue` that ensures the value is no larger than 64 MB, which fits comfortably within an `int` and avoids the overflow:
```go
package main

import (
	"encoding/json"
	"errors"
)

func encryptValueGood(v interface{}) ([]byte, error) {
	jsonData, err := json.Marshal(v)
	if err != nil {
		return nil, err
	}
	if len(jsonData) > 64*1024*1024 {
		return nil, errors.New("value too large")
	}
	size := len(jsonData) + (len(jsonData) % 16)
	buffer := make([]byte, size)
	copy(buffer, jsonData)
	return encryptBuffer(buffer)
}
```

## Recommendation
Always guard against overflow in arithmetic operations involving potentially large numbers by doing one of the following:
  - Validate the size of the data from which the numbers are computed.
[...]

PR #4527, opened 2025-02-27
Fix Size computation for allocation Integer overflow

odaysec:
Performing calculations involving the size of potentially large strings or slices can result in an overflow (for signed integer types) or a wraparound (for unsigned types). An overflow causes the result of the calculation to become negative, while a wraparound results in a small (positive) number. This can cause further issues. If, for the result is then used in an allocation, it will cause a runtime panic if it is negative, and allocate an unexpectedly small buffer otherwise.

#4525 

## Design Checklist

## Checklist

### DIFF
--- a/internal/pkg/danger/buf.go
+++ b/internal/pkg/danger/buf.go
@@ -33,7 +33,11 @@ func (b *Buf) Reset() {
 }
 
 func (b *Buf) grow(n int) {
-	buf := make([]byte, len(b.buf), 2*cap(b.buf)+n)
+	newCap := 2*cap(b.buf) + n
+	if newCap < 0 {
+		panic("danger.Buf.grow: size computation overflow")
+	}
+	buf := make([]byte, len(b.buf), newCap)
 	copy(buf, b.buf)
 	b.buf = buf
 }


internal/pkg/danger/buf.go
```diff
@@ -33,7 +33,11 @@ func (b *Buf) Reset() {
 }
 
 func (b *Buf) grow(n int) {
-	buf := make([]byte, len(b.buf), 2*cap(b.buf)+n)
+	newCap := 2*cap(b.buf) + n
+	if newCap < 0 {
+		panic("danger.Buf.grow: size computation overflow")
```
cmacknz, line 38: This is an internal method that only has 2 calls in this file, one that already explicitly handles a negative count the same way bytes.Buffer does in Go itself https://cs.opensource.google/go/go/+/refs/tags/go1.24.0:src/bytes/buffer.go;l=164-170

https://github.com/elastic/fleet-server/blob/ba68a24383abfa328cce0cc3eddce1b73b497d8a/internal/pkg/danger/buf.go#L41-L48

And another that uses an explicitly non-negative constant: https://github.com/elastic/fleet-server/blob/ba68a24383abfa328cce0cc3eddce1b73b497d8a/internal/pkg/danger/buf.go#L66-L74

This change seems unnecessary.

cmacknz:
Closing, no functional change from current behavior. Re-open if I am missing something.

PR #4528, opened 2025-02-28
Clear agent.upgrade_attempts on upgrade complete

jillguyonnet:

https://github.com/elastic/kibana/pull/212744 adds retry logic to the task that automatically ugprades agents. Agents that were upgraded through this task have their new `upgrade_attempts` property populated. It is missing a way to clear this property when the upgrade completes successfully.

The change in this PR clears `upgrade_attempts` when the upgrade details of the agent get into `UPG_WATCHING` state and are processed in `handleCheckin`.

This should be tested alongside https://github.com/elastic/kibana/pull/212744 (or after it is merged - this is fine, since automatic upgrades are currently behind the `enableAutomaticAgentUpgrades` feature flag). With this change, agents upgraded through the automatic upgrade task should have their `upgrade_attempts` property set to `null` when the upgrade is successful.

Testing should also validate that `upgrade_attempts` stays set if the upgrade failed, e.g. after requesting an upgrade to an invalid version.

## Design Checklist

## Checklist

## Related issues

Relates https://github.com/elastic/ingest-dev/issues/4720

### DIFF
--- a/changelog/fragments/1740760469-Clear-agent-upgrade-attempts-when-upgrade-complete.yaml
+++ b/changelog/fragments/1740760469-Clear-agent-upgrade-attempts-when-upgrade-complete.yaml
@@ -8,7 +8,7 @@
 # - security: impacts on the security of a product or a user’s deployment.
 # - upgrade: important information for someone upgrading from a prior version
 # - other: does not fit into any of the other categories
-kind: feature
+kind: enhancement
 
 # Change summary; a 80ish characters long description of the change.
 summary: Clear agent.upgrade_attempts when upgrade is complete.
--- a/changelog/fragments/1740760469-Clear-agent-upgrade-attempts-when-upgrade-complete.yaml
+++ b/changelog/fragments/1740760469-Clear-agent-upgrade-attempts-when-upgrade-complete.yaml
@@ -31,4 +31,4 @@ pr: https://github.com/elastic/fleet-server/pull/4528
 
 # Issue URL; optional; the GitHub issue related to this changeset (either closes or is part of).
 # If not present is automatically filled by the tooling with the issue linked to the PR number.
-issue: https://github.com/elastic/ingest-dev/issues/4720
+# issue: 
--- a/changelog/fragments/1740760469-Clear-agent-upgrade-attempts-when-upgrade-complete.yaml
+++ b/changelog/fragments/1740760469-Clear-agent-upgrade-attempts-when-upgrade-complete.yaml
@@ -1,0 +1,34 @@
+# Kind can be one of:
+# - breaking-change: a change to previously-documented behavior
+# - deprecation: functionality that is being removed in a later release
+# - bug-fix: fixes a problem in a previous version
+# - enhancement: extends functionality but does not break or fix existing behavior
+# - feature: new functionality
+# - known-issue: problems that we are aware of in a given version
+# - security: impacts on the security of a product or a user’s deployment.
+# - upgrade: important information for someone upgrading from a prior version
+# - other: does not fit into any of the other categories
+kind: feature
+
+# Change summary; a 80ish characters long description of the change.
+summary: Clear agent.upgrade_attempts when upgrade is complete.
+
+# Long description; in case the summary is not enough to describe the change
+# this field accommodate a description without length limits.
+# NOTE: This field will be rendered only for breaking-change and known-issue kinds at the moment.
+description: |
+  The new AutomaticAgentUpgradeTask Kibana task sets the upgrade_attempts property in agents it upgrades.
+  This property is used to track upgrade retries and should therefore be cleared when the upgrade is complete.
+
+# Affected component; usually one of "elastic-agent", "fleet-server", "filebeat", "metricbeat", "auditbeat", "all", etc.
+component: fleet-server
+
+# PR URL; optional; the PR number that added the changeset.
+# If not present is automatically filled by the tooling finding the PR where this changelog fragment has been added.
+# NOTE: the tooling supports backports, so it's able to fill the original PR number instead of the backport PR number.
+# Please provide it if you are adding a fragment for a different PR.
+pr: https://github.com/elastic/fleet-server/pull/4528
+
+# Issue URL; optional; the GitHub issue related to this changeset (either closes or is part of).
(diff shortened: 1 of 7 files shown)


internal/pkg/api/handleCheckin.go
```diff
@@ -219,7 +219,7 @@ func (ct *CheckinT) validateRequest(zlog zerolog.Logger, w http.ResponseWriter,
 		}
 
 		wTime := pollDuration + time.Minute
-		rc := http.NewResponseController(w) //nolint:bodyclose // we are working with a ResponseWriter not a Respons
+		rc := http.NewResponseController(w)
```
jillguyonnet, line 222: This caused the `nolintlint` linter to [error](https://github.com/elastic/fleet-server/actions/runs/13591267337/job/37997742959):
```
Error: internal/pkg/api/handleCheckin.go:222:39: directive `//nolint:bodyclose // we are working with a ResponseWriter not a Respons` is unused for linter "bodyclose" (nolintlint)
```

I installed `golangci-lint` (version 1.64.5) locally and removing this `nolint` directive works, so that seemed like a better option than also silencing `nolintlint`.

By the way, `golangci-lint run` yields a lot of errors across the codebase, most of them `errcheck`. It seems the CI only lints modified files.

pchila (commented):
Code looks sensible, holding off the approval until we have a green CI run (it seems we are getting some errors from ECH when creating a stack and 503s when trying to clean it up (not sure if it's related to the failed creation)

michel-laterman (dismissed):
lgtm, merge when CI is green

changelog/fragments/1740760469-Clear-agent-upgrade-attempts-when-upgrade-complete.yaml
```diff
@@ -0,0 +1,34 @@
...
+# - breaking-change: a change to previously-documented behavior
+# - deprecation: functionality that is being removed in a later release
+# - bug-fix: fixes a problem in a previous version
+# - enhancement: extends functionality but does not break or fix existing behavior
+# - feature: new functionality
+# - known-issue: problems that we are aware of in a given version
+# - security: impacts on the security of a product or a user’s deployment.
+# - upgrade: important information for someone upgrading from a prior version
+# - other: does not fit into any of the other categories
+kind: feature
```
blakerouse, line 11: Is feature the correct kind of change here? Seems more like a bug-fix or an enhancement.

jillguyonnet: Probably should be `enhancement` 👍 Thanks for flagging.

jillguyonnet:
OK, my original approach didn't do what is expected, i.e. only reset `upgrade_attempts` if the upgrade is complete and successful. This is because in `handleAck` the upgrade is considered complete when done retrying. I was also under the impression that `upgrade_details` were updated as part of it, but I see that's actually handled in `handleCheckin`, which makes sense.

I have changed the approach to reset `upgrade_attempts` when the agent reports an `upgrade_details.state` value of `UPG_WATCHING` in `handleCheckin`. I could be missing some edge cases, but it seems to meet the primary expectation. My testing so far looks good.

@michel-laterman Would you please be able to review this approach? If it looks OK, I will update tests.

internal/pkg/model/schema.go
```diff
@@ -228,6 +228,9 @@ type Agent struct {
 	// Date/time the Elastic Agent was last upgraded
 	UpgradedAt string `json:"upgraded_at,omitempty"`
 
+	// List of timestamps of attempts of Elastic Agent automatic upgrades
+	UpgradeAttempts []string `json:"upgrade_attempts,omitempty"`
```
jillguyonnet, line 232: Should I be adding this differently? [The CI fails](https://buildkite.com/elastic/fleet-server/builds/6977#01956b11-f88a-441b-b4a7-c20e9a23da34) with
```
internal/pkg/model/schema.go: needs update
make[1]: *** [Makefile:211: check-no-changes] Error 1
```

michel-laterman: I think you added this manually.
To uupdate this file you need to add the attribute in `model/schema.json` (you did),
then run `make generate`

internal/pkg/api/handleCheckin.go
```diff
@@ -500,6 +500,10 @@ func (ct *CheckinT) processUpgradeDetails(ctx context.Context, agent *model.Agen
 	doc := bulk.UpdateFields{
 		dl.FieldUpgradeDetails: details,
 	}
+	if agent.UpgradeAttempts != nil && details.State == UpgradeDetailsStateUPGWATCHING {
+		doc[dl.FieldUpgradeAttempts] = nil
```
michel-laterman, line 504: Just to be clear, this will set `upgrade_attempts` to `nil`, it does not remove the attribute from the agent doc. This is the intended behaviour, right?

changelog/fragments/1740760469-Clear-agent-upgrade-attempts-when-upgrade-complete.yaml
```diff
@@ -0,0 +1,34 @@
...
+
+# PR URL; optional; the PR number that added the changeset.
+# If not present is automatically filled by the tooling finding the PR where this changelog fragment has been added.
+# NOTE: the tooling supports backports, so it's able to fill the original PR number instead of the backport PR number.
+# Please provide it if you are adding a fragment for a different PR.
+pr: https://github.com/elastic/fleet-server/pull/4528
+
+# Issue URL; optional; the GitHub issue related to this changeset (either closes or is part of).
+# If not present is automatically filled by the tooling with the issue linked to the PR number.
+issue: https://github.com/elastic/ingest-dev/issues/4720
```
michel-laterman, line 34: This is a private repo, we shouldn't link into it

internal/pkg/api/handleCheckin.go
juliaElastic: we want to set it to `null`

michel-laterman (dismissed):
lgtm

juliaElastic (approved):

blakerouse (approved):
Looks good.

PR #4542, opened 2025-03-04
pbkdf2 settings validation is FIPS compliant

michel-laterman:

User pbkdf2 settings are not validated for FIPS compliance.

Validate pbkdf2 settings.

## Design Checklist

## Checklist


internal/pkg/config/pbkdf2.go
```diff
@@ -4,30 +4,12 @@
...
-)
-
 type PBKDF2 struct {
 	Iterations int `config:"iterations"`
 	KeyLength  int `config:"key_length"`
 	SaltLength int `config:"salt_length"`
 }
 
-// Validate the config options
-func (p *PBKDF2) Validate() error {
```
cmacknz, line 18: This was added in https://github.com/elastic/fleet-server/pull/4290 which isn't released yet.

I didn't catch this was added there. Can we just not expose this in the configuration? Under what circumstance would we expect to want to change this configuration and not just in code?

Why can't we just switch to an always FIPS compliant implementation by default?

michel-laterman: @blakerouse what do you think, do we have a use case for exposing these values?

The defaults are FIPS compliant already.

Right now a user changing these values would only cause an issue if they input a non-FIPS values when in FIPS mode.
With the microsoft/go toolchain this will result in the pbkdf2.Key calls succeeding - when they should fail.

blakerouse: I agree just make the validation always FIPS compliant.

blakerouse (approved):
Looks good.

kaanyalti (approved):

PR #4567, opened 2025-03-11
More relaxed retries in agent tests

michalpristas:
Just updating retries in agent tests from 120x100ms to 180x100ms. 
120 is on the fence with new default retry strategy


swiatekm (approved):
If we're using the same retry count everywhere now, there's value in putting it in a constant imo.

andrzej-stencel (approved):

PR #4572, opened 2025-03-11
Fix fleet-server URL population for cloud test

michel-laterman:

Cloud deployment tests silently fail.

use `integrations_server.https_endpoint` instead of `integrations_server.endpoints.fleet`

## Design Checklist

## Checklist


michel-laterman:
The integrations_server.https_endpoint attribute returns the URL for APM, not fleet

PR #4617, opened 2025-03-21
Add buildkite step to run unit tests that require fips provider

michel-laterman:

We need to be able to run tests on VMs that have a FIPS provider.

As a PoC, a buildkite step that runs `FIPS=true make test-unit` has been added to run on a new VM. Running this target requires msft/go (gathered by the new `with_msft_go` func added to common.sh) and a FIPS provider (supplied by VM).

## Design Checklist

## Checklist


.buildkite/pipeline.yml
```diff
@@ -98,6 +98,19 @@ steps:
           - build/*.xml
           - build/coverage*.out
 
+      - label: ":smartbear-testexecute: Run FIPS Provider tests"
+        key: provider-test-fips
+        command: ".buildkite/scripts/fips_provider_test.sh"
+        agents:
+          provider: "aws"
+          imagePrefix: "elasticcloud-fips-ubuntu-2204-amd64-dev-jammyf-"
```
v1v, line 106:
```suggestion
          imagePrefix: "platform-ingest-fleet-server-aws-linux-x86-64-fleet-server-ubuntu-2204-fips"
```

simitt (commented):
@michel-laterman can you explain what is the goal of this PR is?  Maybe I am not seeing what this setup would cover which isn't already covered with: 
- Unit tests running with the strict go-1.24 FIPS mode enabled
- Plan to test more involved settings, eg. upgrade, gpg, TLS functionality through system tests with the `platform-ingest-fleet-server-ubuntu-2204-fips` image. 

When running unit tests with go-microsoft, it would still fall back to std lib functionality rather than fail if non compliant algorithms are used.

michel-laterman:
This runs the unit tests with the microsoft/go toolchain that uses the FIPS enabled OpenSSL

michel-laterman:
These tests are not currently required.

PR #4622, opened 2025-03-24
Custom user for Docker builder

michalpristas:
Adding custom user for docker builder defined in Dockerfile.build


michel-laterman (dismissed):
Please add this to Dockerfile.fips as well

ebeahan (approved):

PR #4687, opened 2025-03-26
Cleanup debug logs

pchila:

Cleanup some debug logs

## Design Checklist

## Checklist

### DIFF
--- a/cmd/fleet/main.go
+++ b/cmd/fleet/main.go
@@ -54,7 +54,6 @@ func initLogger(cfg *config.Config, version, commit string) (*logger.Logger, err
 		Str("exe", os.Args[0]).
 		Strs("args", os.Args[1:]).
 		Msg("Boot fleet-server")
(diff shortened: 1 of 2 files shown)


swiatekm (approved):

pkoutsovasilis (approved):

pkoutsovasilis (commented):
code changes look to me but what about the CI failures, any idea @pchila ? They seem unrelated to this change `Error: The local pre-command hook exited with status 1`

pchila:
@pkoutsovasilis Not sure what happened in CI, it's unrelated to the changes made.
Maybe there has been some issue with authorization (I see some authorization headers in `curl --fail` commands which aborted the script.
I will retry the step and see if it's a transient error or a systematic one.

cmacknz:
I would backport this change to all active branches. This is a bug fix we want to ship in patch release IMO.

PR #4737, opened 2025-03-28
Bump golangci lint action

ycombinator:

Bumps up the version of the `golangci/golangci-lint-action` GitHub action that's run as part of CI checks.

This check is failing without explanation on PRs made to the `7.17` branch while it's passing on the same PRs made to `main` and other newer branches.  These other branches have a newer version of the action, so this PR makes the `7.17` branch have the same version.


ycombinator:
Oops, wrong bra

PR #4738, opened 2025-03-28
Bump up version of golangci/golangci-lint-action

ycombinator:

Bumps up the version of the `golangci/golangci-lint-action` GitHub action that's run as part of CI checks.

This check is failing without explanation on PRs made to the `7.17` branch while it's passing on the same PRs made to `main` and other newer branches.  These other branches have a newer version of the action, so this PR makes the `7.17` branch have the same version.


ebeahan:
@ycombinator do you think you'll follow up troubleshooting further on this one, or should we just close since it's targeting `7.17`?

ycombinator:
> @ycombinator do you think you'll follow up troubleshooting further on this one, or should we just close since it's targeting `7.17`?

Thanks for the ping.  I think we can just close this PR as it's targeting `7.17` and this is not a critical fix.

PR #4762, opened 2025-04-02
Clear upgrade_attempts on handleAck

juliaElastic:

`upgrade_attempts` were not cleared correctly when agent doesn't have `upgrade_details` (for example in horde or older versions).

Clear `upgrade_attempts` when upgrade is acked (at the same time when `upgrade_started_at` field is cleared).

Test with agent policy with auto upgrade config and a few horde agents enrolled. Verify that after the upgrade completed, `upgrade_attempts` is set to `null`.
The `upgrade_attempts` field is only cleared if there is no `upgrade_details`. Tested with a real agent upgraded to a non-existent version, the agent going to `UPG_FAILED` state.

## Design Checklist

## Checklist

## Related issues

Relates https://github.com/elastic/fleet-server/pull/4528
Relates https://github.com/elastic/kibana/pull/212744

### DIFF
--- a/internal/pkg/api/handleAck.go
+++ b/internal/pkg/api/handleAck.go
@@ -619,6 +619,7 @@ func (ack *AckT) handleUpgrade(ctx context.Context, zlog zerolog.Logger, agent *
 			dl.FieldUpgradeStartedAt: nil,
 			dl.FieldUpgradeStatus:    nil,
 			dl.FieldUpgradedAt:       now,
+			dl.FieldUpgradeAttempts:  nil,
 		}
 	}
 
--- a/internal/pkg/api/handleAck.go
+++ b/internal/pkg/api/handleAck.go
@@ -619,7 +619,9 @@ func (ack *AckT) handleUpgrade(ctx context.Context, zlog zerolog.Logger, agent *
 			dl.FieldUpgradeStartedAt: nil,
 			dl.FieldUpgradeStatus:    nil,
 			dl.FieldUpgradedAt:       now,
-			dl.FieldUpgradeAttempts:  nil,
+		}
+		if agent.UpgradeDetails == nil {
+			doc[dl.FieldUpgradeAttempts] = nil
 		}
 	}
 
--- a/internal/pkg/api/handleAck_test.go
+++ b/internal/pkg/api/handleAck_test.go
@@ -673,13 +673,25 @@ func TestAckHandleUpgrade(t *testing.T) {
 		name   string
 		event  UpgradeEvent
 		bulker func(t *testing.T) *ftesting.MockBulk
+		agent  *model.Agent
 	}{{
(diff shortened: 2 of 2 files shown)


juliaElastic:
@jillguyonnet I saw this [comment](https://github.com/elastic/fleet-server/pull/4528#issuecomment-2701686020), do you know the quickest way to reproduce the failed upgrade? I tried upgrading to a non-existent version, but the retries take a long time (about 2 hours).

jillguyonnet:
> @jillguyonnet I saw this https://github.com/elastic/fleet-server/pull/4528#issuecomment-2701686020, do you know the quickest way to reproduce the failed upgrade? I tried upgrading to a non-existent version, but the retries take a long time (about 2 hours).

Strange, IIRC that's also what I did. I don't remember the agent retrying for that long (essentially it went into `UPG_FAILED` pretty quickly, maybe ~10 min).

juliaElastic:
> > @jillguyonnet I saw this [#4528 (comment)](https://github.com/elastic/fleet-server/pull/4528#issuecomment-2701686020), do you know the quickest way to reproduce the failed upgrade? I tried upgrading to a non-existent version, but the retries take a long time (about 2 hours).
> 
> Strange, IIRC that's also what I did. I don't remember the agent retrying for that long (essentially it went into `UPG_FAILED` pretty quickly, maybe ~10 min).

Thanks, I'm seeing now the `UPG_FAILED` after 15 mins, I wasn't sure when it would happen as `upgrade_details` had `retry_until` in 2 hours.

jillguyonnet:
> Thanks, I'm seeing now the UPG_FAILED after 15 mins, I wasn't sure when it would happen as upgrade_details had retry_until in 2 hours.

Yeah, I did notice that as well, but it transitioned into `UPG_FAILED` much quicker. Not sure if that's expected, come to think of it.

internal/pkg/api/handleAck.go
```diff
@@ -620,6 +620,9 @@ func (ack *AckT) handleUpgrade(ctx context.Context, zlog zerolog.Logger, agent *
 			dl.FieldUpgradeStatus:    nil,
 			dl.FieldUpgradedAt:       now,
 		}
+		if agent.UpgradeDetails == nil {
```
michalpristas, line 623: do we need to check nil?
we are upgrading UpgradedAt 2 lines above. so we consider this done

juliaElastic: when there is `upgrade_details`, the `upgrade_attempts` is cleared when the state is `UPG_WATCHING` (see reasoning [here](https://github.com/elastic/fleet-server/pull/4528#issuecomment-2701686020))
this fix is mainly for horde, where there is no `upgrade_details`

michalpristas (approved):

PR #4776, opened 2025-04-07
Fix concurrent access to remote bulker config

michel-laterman:

Fix possible panic when using remote outputs

Use existing mutex for access controls when interacting with remote bulker configs.

## Design Checklist

## Checklist

- [x] I have added tests that prove my fix is effective or that my feature works (existing tests in `internal/pkg/bulk/bulk_remote_output_test.go`)

### DIFF
--- a/internal/pkg/bulk/engine.go
+++ b/internal/pkg/bulk/engine.go
@@ -85,15 +85,16 @@ type Bulk interface {
 const kModBulk = "bulk"
 
 type Bulker struct {
-	es                    esapi.Transport
-	ch                    chan *bulkT
-	opts                  bulkOptT
-	blkPool               sync.Pool
-	apikeyLimit           *semaphore.Weighted
-	tracer                *apm.Tracer
+	es          esapi.Transport
+	ch          chan *bulkT
+	opts        bulkOptT
+	blkPool     sync.Pool
+	apikeyLimit *semaphore.Weighted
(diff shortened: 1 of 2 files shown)


internal/pkg/bulk/engine.go
```diff
@@ -259,7 +260,9 @@ func (b *Bulker) Client() *elasticsearch.Client {
 }
 
 func (b *Bulker) RemoteOutputConfigChanged(zlog zerolog.Logger, name string, newCfg map[string]interface{}) bool {
+	b.remoteOutputMutex.RLock()
 	curCfg := b.remoteOutputConfigMap[name]
+	b.remoteOutputMutex.RUnlock()
```
ycombinator, line 265: With this implementation, don't you run the risk of another goroutine change the value of `b.remoteOutputConfigMap[name]` after `curCfg` has been set, thereby rendering the value of `curCfg` stale?  Would the following implementation make more sense, as it would bring `curCfg` into the critical section protected by the mutex as well?

```suggestion
	b.remoteOutputMutex.RLock()
	defer b.remoteOutputMutex.RUnlock()
	curCfg := b.remoteOutputConfigMap[name]
```

blakerouse (approved):
Looks good from me.

Please wait for @ycombinator approval as well, being he gave the code suggession.

PR #4784, opened 2025-04-09
fix: add grpcnotrace to go build tags

kruskall:

DCE is disabled in fleet-server

resolve blockers to enable DCE (deadcode elimination)

Pass grpcnotrace tag to remove x/net/trace dependency in grpc dependency causing DCE to be disabled

run `make local`

fleet-server binary size:

35M -> 23M

## Design Checklist

## Checklist

### DIFF
--- a/changelog/fragments/1744852233-enable-dce.yaml
+++ b/changelog/fragments/1744852233-enable-dce.yaml
@@ -1,0 +1,34 @@
+# Kind can be one of:
+# - breaking-change: a change to previously-documented behavior
+# - deprecation: functionality that is being removed in a later release
+# - bug-fix: fixes a problem in a previous version
+# - enhancement: extends functionality but does not break or fix existing behavior
+# - feature: new functionality
+# - known-issue: problems that we are aware of in a given version
+# - security: impacts on the security of a product or a user’s deployment.
+# - upgrade: important information for someone upgrading from a prior version
+# - other: does not fit into any of the other categories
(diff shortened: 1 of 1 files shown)


pkoutsovasilis (dismissed):
This change looks to me thank you @kruskall . Although I wouldn't deem it as a `bug` an `enhancement` feels more appropriate. @michel-laterman what backports should this get?

pkoutsovasilis:
ok first things first do you agree @kruskall that this is an enhancement and not a bug fix? 🙂

pkoutsovasilis:
>Sorry, didn't mean to ignore that 😅
> Personally I consider DCE being disabled in a production binary/artifact a bug but I don't feel super strongly about it, I don't mind if this is considered an enhancement. 🙂

ok I see the way you, I personally see that by having DCE disabled we are not having business-logic errors, no panics, doesn't open the code to CVE vulnerabilities, etc. So I call this a file-size enhancement. But let's roll with your choice a `bug`, then we need to backport this to all active branches and it would be great if this has the same effects in the other active branches. wdyt? 🙂

pkoutsovasilis:
ok I am gonna initiate the backports @kruskall , you write a changelog file please 🙂

michel-laterman (dismissed):

michel-laterman:
Please add a changelog entry before merging

pkoutsovasilis (approved):
LGTM, ty @kruskall

PR #4812, opened 2025-04-16
mergify: support 8.19 and remove support 8.x

v1v:

`8.x` branch has been renamed to `8.19`

## Design Checklist

## Checklist

### DIFF
--- a/.mergify.yml
+++ b/.mergify.yml
@@ -264,19 +264,6 @@ pull_request_rules:
         labels:
           - "backport"
         title: "[{{ destination_branch }}](backport #{{ number }}) {{ title }}"
-  - name: backport patches to 8.x branch
-    conditions:
-      - merged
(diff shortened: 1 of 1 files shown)


.mergify.yml
```diff
@@ -319,6 +306,14 @@ pull_request_rules:
         labels:
           - "backport"
         title: "[{{ destination_branch }}](backport #{{ number }}) {{ title }}"
+  - name: backport patches to 8.19 branch
+    conditions:
+      - merged
+      - label=backport-8.19
+    actions:
+      backport:
+        branches:
```
v1v, line 315: Use default values:
- https://github.com/elastic/fleet-server/blob/cc0972901ad62ab25b148f8efd619d1373b57aa6/.mergify.yml#L12-L19

In a follow-up we can reduce the explicit duplicated entries per backport rule

ev1yehor (approved):
LGTM

PR #4849, opened 2025-04-25
bk: add catalog info validation

v1v:

VAlidate the catalog-info.yaml file

Avoid introducing breaking changes that can affect the whole `terrazzo` thing

## Design Checklist

## Checklist

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -11,6 +11,21 @@ env:
   TERRAFORM_VERSION: "1.6.4"
 
 steps:
+
+  - label: "Validate catalog-info"
+    if: build.pull_request.id != null
+    plugins:
+      monorepo-diff#v1.3.0:
+        diff: "git diff --name-only origin/${GITHUB_PR_TARGET_BRANCH}...HEAD"
+        wait: true
+        watch:
+          # if our catalog-info.yaml is changed, make sure it's well-formed according to our internal standards as well as Backstage's validation
+          - path: "catalog-info.yaml"
+            config:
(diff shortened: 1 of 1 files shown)


pazone (approved):

.buildkite/pipeline.yml
```diff
@@ -11,6 +11,21 @@ env:
...
+      monorepo-diff#v1.3.0:
+        diff: "git diff --name-only origin/${GITHUB_PR_TARGET_BRANCH}...HEAD"
+        wait: true
+        watch:
+          # if our catalog-info.yaml is changed, make sure it's well-formed according to our internal standards as well as Backstage's validation
+          - path: "catalog-info.yaml"
+            config:
+              command: "/agent/check-catalog-info.sh"
+              agents:
+                image: "docker.elastic.co/ci-agent-images/pipelib:0.13.0@sha256:190464950de81cb3390bd5e07bb1cc171dd4a765c1ce5de58ebca476f9a6839f"
```
ycombinator, line 27: Just wondering if we need the specific version tag + SHA here.  Couldn't we always pull the latest as mentioned in https://github.com/elastic/ci-agent-images/tree/main/container-images/engprod/pipelib#validating-your-renovate-configuration-in-a-pipeline?

v1v: I think that's related to some security practice where dependencies are not using mutable tags but fixed versions with the sha256.

I took the configuration from the Onboarding templates that the platform eng prod team provides.

Let me ping you in the private GH PR

ycombinator (approved):

PR #4983, opened 2025-05-28
Add basic fips cloudE2E test

michel-laterman:

Verify that FIPS capable artifacts work in cloud deployments.
Correct openapi status endpoint return values.

Add buildkite step to run cloud e2e tests with a FIPS capable image.
Add a smoke test to the cloudE2E testsuite that ensures the deployed fleet-server has basic functionality, and that fleet-server's crypto features (the `ReplacementToken`) works as expected.
Change the `status` attribute in status response bodies to be in all caps.


kaanyalti (approved):
Looks good to me

PR #4999, opened 2025-06-03
Generate `NOTICE-fips.txt` instead of `NOTICE-FIPS.txt`

ycombinator:
For consistency with [Elastic Agent](https://github.com/elastic/elastic-agent/blob/main/NOTICE-fips.txt) and [APM Server](https://github.com/elastic/apm-server/blob/main/NOTICE-fips.txt).


swiatekm:
Shouldn't this PR also rename the actual generated file checked into the repo?

michel-laterman (approved):

PR #5000, opened 2025-06-03
Fail cloude2e step if tests can't run, fix image publication

michel-laterman:

cloude2e tests can silently fail.
fleet-server image publication broken.

Fail the pipeline if cloude2e tests can't run.
`mage docker:image` now uses the `DOCKER_IMAGE` env var.


magefile.go
```diff
@@ -968,6 +968,7 @@ func (Docker) Cover() error {
 // DEV creates a development image.
 // SNAPSHOT creates a snapshot image.
 // VERSION_QUALIFIER may be used to manually specify a version qualifer for the image tag.
+// DOCKER_IMAGE may be used to completly specify the image name.
```
ycombinator:
```suggestion
// DOCKER_IMAGE may be used to completely specify the image name.
```

PR #5001, opened 2025-06-03
Update command used by Dependabot to update NOTICE files

ycombinator:
Otherwise the Post-dependabot check on PRs fails like so:

```
Run make notice
make: *** No rule to make target 'notice'.  Stop.
Error: Process completed with exit code 2.
```

### DIFF
--- a/.github/workflows/post-dependabot.yml
+++ b/.github/workflows/post-dependabot.yml
@@ -22,6 +22,10 @@ jobs:
         with:
           go-version-file: .go-version
 
+      - uses: magefile/mage-action@v3
+        with:
+          install-only: true
+
       - name: update NOTICE.txt and NOTICE-FIPS.txt
         run: mage check:notice
 


.github/workflows/post-dependabot.yml
```diff
@@ -22,8 +22,8 @@ jobs:
         with:
           go-version-file: .go-version
 
-      - name: update NOTICE.txt
-        run: make notice
+      - name: update NOTICE.txt and NOTICE-FIPS.txt
+        run: mage check-notice
```
michel-laterman: Do we know if dependabot has mage installed? We might need to do `make mage` first

v1v (dismissed):
Good catch! Thanks

.github/workflows/post-dependabot.yml
ycombinator: I think 28ef18db should now address this. @v1v can you confirm, please?  Thanks.

v1v: aha! That's a good point, as far as I know it's not installed and that's something I missed from my review.

if `make mage` is a validate Make goal, then I'd say to run it as part of the run, otherwise,  you can use https://github.com/elastic/elastic-agent/blob/f1086184d87ed755b9275081c3b42f58c54feb63/.github/workflows/bump-beats-version.yml#L49-L53

ebeahan (approved):
LGTM.

PR #5040, opened 2025-06-13
QueryByLatest integration test at scale

michel-laterman:

## Design Checklist

## Checklist

### DIFF
--- a/internal/pkg/dl/policies_integration_test.go
+++ b/internal/pkg/dl/policies_integration_test.go
@@ -13,7 +13,6 @@ import (
 	"time"
 
 	"github.com/gofrs/uuid/v5"
-	"github.com/google/go-cmp/cmp"
 	"github.com/stretchr/testify/require"
 
 	"github.com/elastic/fleet-server/v7/internal/pkg/bulk"
@@ -42,15 +41,24 @@ func createRandomPolicy(id string, revisionIdx int) model.Policy {
 	}
 }
 
-func storeRandomPolicy(ctx context.Context, bulker bulk.Bulk, index string) (model.Policy, error) {
+func storeRandomPolicy(ctx context.Context, bulker bulk.Bulk, index string, maxRev int) (model.Policy, error) {
 	var rec model.Policy
 	id := uuid.Must(uuid.NewV4()).String()
-	for i := 1; i < 4; i++ {
+	ops := make([]bulk.MultiOp, 0, maxRev)
+	for i := 1; i < maxRev; i++ {
 		rec = createRandomPolicy(id, i)
-		_, err := CreatePolicy(ctx, bulker, rec, WithIndexName(index))
+		p, err := json.Marshal(&rec)
 		if err != nil {
 			return model.Policy{}, err
 		}
+		ops = append(ops, bulk.MultiOp{
+			Index: index,
+			Body:  p,
+		})
+	}
+	_, err := bulker.MIndex(ctx, ops, bulk.WithRefresh())
+	if err != nil {
+		return model.Policy{}, err
 	}
 	return rec, nil
 }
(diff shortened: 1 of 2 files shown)


prodsecmachine:

:white_check_mark: **security/snyk** check is complete. No issues have been found. [(View Details)](https://app.snyk.io/org/beats/pr-checks/93cd7b93-00b4-469f-bd15-6bba79aa2c5c?refs=9effa544f6affe34cb082c492ecb2ba7b45c22a5&source=prChecksComment)

:white_check_mark: **license/snyk** check is complete. No issues have been found. [(View Details)](https://app.snyk.io/org/beats/pr-checks/93cd7b93-00b4-469f-bd15-6bba79aa2c5c/license?refs=9effa544f6affe34cb082c492ecb2ba7b45c22a5&source=prChecksComment)

internal/pkg/dl/policies_integration_test.go
```diff
@@ -42,45 +41,81 @@ func createRandomPolicy(id string, revisionIdx int) model.Policy {
...
-	index, bulker := ftesting.SetupCleanIndex(ctx, t, FleetPolicies)
+	index, bulker := ftesting.SetupCleanIndex(ctx, t, FleetPolicies, bulk.WithFlushThresholdCount(1))
+
+	for i := 0; i < 4; i++ {
+		err := storeRandomPolicy(ctx, bulker, index, 4)
+		if err != nil {
+			t.Fatal(err)
+		}
+	}
+	time.Sleep(time.Second * 2) // FIXME ES does not refresh instantly?
```
michel-laterman: When these sleeps are not here, the query below will not always gather each policy_id, i'm not sure if there is another query optional arg I should pass instead in order to remove the wait

internal/pkg/dl/policies_integration_test.go
```diff
@@ -42,45 +41,81 @@ func createRandomPolicy(id string, revisionIdx int) model.Policy {
...
+	for i := 0; i < 4; i++ {
+		err := storeRandomPolicy(ctx, bulker, index, 100000) // aggregation has a size limit of 10k, let's go over it
 		if err != nil {
 			t.Fatal(err)
 		}
-		recs[rec.PolicyID] = rec
 	}
+	time.Sleep(time.Second * 2) // FIXME ES does not refresh instantly?
 
 	policies, err := QueryLatestPolicies(ctx, bulker, WithIndexName(index))
```
blakerouse: Maybe adding a `refresh` in the latest policies will result in it working with out the need for the sleep?

michel-laterman: Good suggestion, however it didn't work.

blakerouse: Dang.

blakerouse:
Have you tried to test this with more than 10,000 policies? Seems to test 4 policies with more than 10,000 revisions, maybe its the other way that causes it to break?

internal/pkg/dl/policies_integration_test.go
```diff
@@ -42,45 +41,111 @@ func createRandomPolicy(id string, revisionIdx int) model.Policy {
...
+	if len(policies) != 4 {
+		t.Errorf("Expected 4 policies, got %d", len(policies))
+	}
+}
+
+// TesyQueryLatestPolices11kUnique tests behaviour when 11k unique policies are used, there is a 10k size specification in the aggregation.
+//
+//nolint:dupl // test duplication
+func TestQueryLatestPolicies11kUnique(t *testing.T) {
+	t.Skip("Re-enable after policy load issues have been sorted: https://github.com/elastic/fleet-server/issues/3254")
```
michel-laterman, line 125: @blakerouse added the test for number of policies; currently it fails dues to the aggregation size (as expected).

I'm not sure in terms of priority where resolving the policy load issues is for us

blakerouse: I think the case that there are 10,000 unique active policies is rare. I think its more likely that some automation creates and deletes lots of policies. If I remember correctly deleting a policy doesn't actually remove it from the index, it just sets it to `active: false`.

I wonder if it would be possible to remove all policies with `active: false` first before ES does the aggregation, allowing the size limit to be based on the number of active policies. If that is not possible, we might want to see about having some cleaner job that removes `active: false` policies after a period of time.

michel-laterman: I think if we set the query step of the aggregation to require `active: true` that would do it

michel-laterman: Actually, policies do not have an active attribute in the ES template: https://github.com/elastic/elasticsearch/blob/main/x-pack/plugin/core/template-resources/src/main/resources/fleet-policies.json

blakerouse: I think I was confusing `active: true` with the `.fleet-agents` index. That is only used there. I guess policies are actually deleted then by Kibana or they use `data.enabled`? Not sure then.

michel-laterman: I'm pretty sure that the `data.enabled` attribute in the template controls if the policy data is indexed

michel-laterman: I just checked with a cloud deployment; when a policy is deleted Kibana will delete all revisions of the policy

blakerouse: Okay. Then this issue should only present itself in the case that user has 10,000+ active policies. That seems extremely unlikely.

michel-laterman: FYI: Our public docs state a 1k policy limit.

internal/pkg/dl/policies_integration_test.go
pkoutsovasilis: can't this become a require.eventually?

PR #5069, opened 2025-06-26
Include base error for json decode error responses

michel-laterman:

There is no information on why request bodies fail to decode as JSON

Include the base json decode error as part of the response.

## Design Checklist

## Checklist


cmacknz (approved):
Thanks, doing a quick paranoia check I don't see any places where `nextErr` could contain anything we wouldn't want to give back to a client: https://github.com/search?q=repo%3Aelastic%2Ffleet-server%20nextErr&type=code

michel-laterman:
Current e2e failure is dues to the version bump, the test needs `elastic-agent-9.2.0-SNAPSHOT-linux-x86_64.tar.gz` but can't find it

michel-laterman:
buildkite test this

PR #5075, opened 2025-07-02
Add multiplatform images

michel-laterman:

Multiplatform images are not being produced

Use multiplatform builds which were incorrectly removed when migrating to mage.

`mage docker:image`


prodsecmachine:

:white_check_mark: **security/snyk** check is complete. No issues have been found. [(View Details)](https://app.snyk.io/org/beats/pr-checks/077b47e9-2c71-4eaa-affc-4408e408b98e?refs=c2114a6f6d062e0ecbe1f5d3d5738bfe3f8f1533&source=prChecksComment)

:white_check_mark: **license/snyk** check is complete. No issues have been found. [(View Details)](https://app.snyk.io/org/beats/pr-checks/077b47e9-2c71-4eaa-affc-4408e408b98e/license?refs=c2114a6f6d062e0ecbe1f5d3d5738bfe3f8f1533&source=prChecksComment)

changelog/fragments/1751485154-Add-missing-multiplatform-builds.yaml
```diff
@@ -0,0 +1,32 @@
...
+# - enhancement: extends functionality but does not break or fix existing behavior
+# - feature: new functionality
+# - known-issue: problems that we are aware of in a given version
+# - security: impacts on the security of a product or a user’s deployment.
+# - upgrade: important information for someone upgrading from a prior version
+# - other: does not fit into any of the other categories
+kind: bug-fix
+
+# Change summary; a 80ish characters long description of the change.
+summary: Add missing multiplatform builds
```
blakerouse: Does this need a changelog? Is this user facing? Have we released this on a final build?

magefile.go
```diff
@@ -987,8 +992,13 @@ func (Docker) Image() error {
...
 		image = v
 	}
+	dockerEnv := map[string]string{"DOCKER_BUILDKIT": "1"}
+	if err := sh.RunWithV(dockerEnv, "docker", "buildx", "create", "--use"); err != nil {
+		return fmt.Errorf("docker buildx create failed: %w", err)
+	}
 
-	return sh.RunWithV(map[string]string{"DOCKER_BUILDKIT": "1"}, "docker", "build",
+	return sh.RunWithV(dockerEnv, "docker", "buildx", "build",
+		"--platform", strings.Join(platformsDocker, ","),
```
pkoutsovasilis: should this be filtered by the PLATFORMS env var (I think with the make approach we could that)? what will happen if somebody on an x86 machine only invokes this?

michel-laterman: Right now when someone runs this, it will build the amd64 and arm64 images, I'll add filtering but I think that would mostly apply for developers working locally

pkoutsovasilis: IMO since this was supported in the previous make approach and we do have developers that work on x86 machines I feel that such filtering might be handy

blakerouse: Yeah I agree being able to filter would be nice.

ebeahan: I'm in favor of keeping parity with any functionality previously supported by `make` to avoid any surprises.

If there's an argument to change, we can capture in a separate issue as future work.

changelog/fragments/1751485154-Add-missing-multiplatform-builds.yaml
michel-laterman: good point; this only got to 9.1 and 8.19, i'll remove the changelog

magefile.go
```diff
@@ -987,8 +993,22 @@ func (Docker) Image() error {
...
 
-	return sh.RunWithV(map[string]string{"DOCKER_BUILDKIT": "1"}, "docker", "build",
+	// get PLATFORMS from env and filter down into those supported by Docker
+	pEnv := getPlatforms()
+	platforms := make([]string, 0, len(platformsDocker))
+	for _, platform := range pEnv {
+		if slices.Contains(platformsDocker, platform) {
+			platforms = append(platforms, platform)
+		}
+	}
```
blakerouse: I think it would be cleaner to move this to its own function. Like `getDockerPlatforms()`.

michel-laterman: done

blakerouse (approved):
Looks good.

Thanks for fixing this and keeping the ability to filter.

PR #4890, opened 2025-05-06
github-action: add catalog-validate for GitHub actions

v1v:
## Details

⚠️ This PR was created by an automated tool. Please review the changes carefully. ⚠️ 

We want to enable the Catalog Info validation using GitHub actions.

## Why

We want to make sure that the catalog-info file is valid and does not break the Catalog Ecosystem.

If there are any questions, please reach out to the @elastic/observablt-ci

## Further details

Requires https://github.com/elastic/oblt-actions/pull/272

## Tasks
- [ ] Grant read access in [the private container](https://github.com/orgs/elastic/packages/container/observability-robots%2Fci-agent-images%2Fpipelib/settings).

PR #5145, opened 2025-07-11
bk: use latest FIPS VM images and stop using catalog-info step but GH actions validation

v1v:

* Use the latest VM image for FIPS - that's how it works with the rest of the images.
* Stop using the catalog-info step in BK - https://github.com/elastic/fleet-server/pull/4890 supported this in GH actions.

Run faster rather than always and delegate the responsibility to the existing tooling.
Align the same VM image strategy regardless of FIPS or non-fips.

## Design Checklist

## Checklist


.buildkite/pipeline.yml
```diff
@@ -3,7 +3,7 @@
 env:
   DOCKER_COMPOSE_VERSION: "1.25.5"
   TERRAFORM_VERSION: "1.6.4"
-  IMAGE_UBUNTU_X86_64_FIPS: "platform-ingest-fleet-server-ubuntu-2204-fips-1751684469"
+  IMAGE_UBUNTU_X86_64_FIPS: "platform-ingest-fleet-server-ubuntu-2204-fips"
```
v1v, line 6: See [here](https://buildkite.com/elastic/fleet-server/builds/9268#01981c8b-174a-44ac-8130-8f90dee14b28/1)

image

pkoutsovasilis (approved):

PR #4797, opened 2025-04-10
mergify: new way of backporting to active branches

ev1yehor:
## What is the problem this PR solves?

Adding new way of backporting to active branches using new github action.

## Why

No need to support the hardcoded branches in the `.mergify.yml` file anymore.

We use a dynamic approach by fetching https://storage.googleapis.com/artifacts-api/snapshots/branches.json and filtering the branches.

## Issues

Already enabled in Elastic Agent:
- https://github.com/elastic/elastic-agent/actions/workflows/backport-active.yml

PR #5168, opened 2025-07-21
mergify: simplify and honour the defaults

v1v:

Tidy up the mergify configuration by honouring https://github.com/elastic/fleet-server/pull/4797

Removing duplicated configuration and references for the unused apmmachine

## Design Checklist

## Checklist

### DIFF
--- a/.mergify.yml
+++ b/.mergify.yml
@@ -24,19 +24,13 @@ pull_request_rules:
       - label=forwardport-main
     actions:
       backport:
-        assignees:
-          - "{{ author }}"
         branches:
           - "main"
-        labels:
-          - "backport"
-        title: "[{{ destination_branch }}](backport #{{ number }}) {{ title }}"
   - name: ask to resolve conflict
     conditions:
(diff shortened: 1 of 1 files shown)


.mergify.yml
```diff
@@ -24,19 +24,13 @@ pull_request_rules:
...
           - "main"
-        labels:
-          - "backport"
-        title: "[{{ destination_branch }}](backport #{{ number }}) {{ title }}"
   - name: ask to resolve conflict
     conditions:
       - -merged
       - -closed
       - conflict
-      - -author=apmmachine
```
v1v, line 39: we don't use it anymore

.mergify.yml
```diff
@@ -53,7 +47,7 @@ pull_request_rules:
       - -merged
       - -closed
       - conflict
-      - author=apmmachine
+      - author=github-actions[bot]
```
v1v, line 50: let's use the current bot we use for automation through GitHub actions

PR #5169, opened 2025-07-22
Add OTel collector properties to policy schema

jsoriano:

Support provisioning of OTel collector configuration for hybrid agents.

Add the OTel collector properties to the schema.

## Design Checklist

## Checklist

## Related issues

- Required for https://github.com/elastic/kibana/pull/227673.


model/schema.json
```diff
@@ -870,6 +870,37 @@
             }
           }
         },
+        "receivers": {
```
cmacknz, line 878: We should add [connectors](https://opentelemetry.io/docs/collector/configuration/#connectors) as well if you weren't already planning to do that

jsoriano: Good point, thanks, adding exporters too.

cmacknz:
> Do you know if there is something else that would need to be modified in Fleet Server or Elastic Agent for the Agent to receive the OTel configuration? Or does the Agent need to be started in some special way to act as an hybrid agent?

Agent will (should) accept collector configuration keys without issue. The way to check if agent is responsible for stripping out the configuration is to collect diagnostics and look at the `pre-config.yml`, that should be the agent policy before agent processed any of it. 

@michel-laterman might have an idea of where these keys get dropped. Some other part of the internal fleet-server policy model or an index mapping or something.

michel-laterman:
It should also be added as part of the openapi definition: https://github.com/elastic/fleet-server/blob/20990468213ce5517bf986af1f5388a7ec01016e/model/openapi.yml#L561

jsoriano:
With current code, I see this error in agent logs:
```
{"log.level":"warn","@timestamp":"2025-08-13T14:06:30.979Z","log.origin":{"function":"github.com/elastic/elastic-agent/internal/pkg/agent/application/gateway/fleet.(*FleetGateway).doExecute","file.name":"fleet/fleet_gateway.go","file.line":199},"message":"Possible transient error during checkin with fleet-server, retrying","log":{"source":"elastic-agent"},"error":{"message":"could not decode the response, raw response: "},"request_duration_ns":774654481,"failed_checkins":2,"retry_after_ns":123230918117,"ecs.version":"1.6.0"}
```

cmacknz:
OK the problem is in Fleet server, we are getting a non-200 response back but it isn't telling us which one. It's trying to parse an explicit error but there isn't any so it just logs nothing.

https://github.com/elastic/elastic-agent/blob/92bebb22f7e862f6f58d479c1014f461942ecd3d/internal/pkg/fleetapi/checkin_cmd.go#L129-L131

This is the source of that log https://github.com/elastic/elastic-agent/blob/92bebb22f7e862f6f58d479c1014f461942ecd3d/internal/pkg/fleetapi/client/client.go#L128-L129

We can modify the error handling there to get the HTTP status code and that will help narrow down where in Fleet Server the problem is, though I'd hope Fleet Server has some logs about whatever is going wrong too.

cmacknz:
I think if you pull out the policy generated by Kibana we can write a test isolating the problem in fleet-server to speed up debugging.

https://github.com/elastic/fleet-server/blob/c5f7f375e85a7445ef3d248c34ada0828dd419c8/internal/pkg/server/fleet_secrets_integration_test.go#L174-L208 

That looks like a reasonable template for doing this, where you can create a policy and make requests to the /checkin API without relying on a real agent. This will take Kibana and Elastic Agent out of the equation.

You could probably remove elastic agent using curl if you could get it's API key but I suspect there will be several problems in Fleet Server to get past so writing an integration test for this will be the fastest way to flush them all out.

ebeahan:
@michel-laterman @michalpristas this one still needs a review from our side.

changelog/fragments/1756122604-Add-OTel-collector-properties-to-policy-schema.yaml
```diff
@@ -0,0 +1,34 @@
...
+
+# PR URL; optional; the PR number that added the changeset.
+# If not present is automatically filled by the tooling finding the PR where this changelog fragment has been added.
+# NOTE: the tooling supports backports, so it's able to fill the original PR number instead of the backport PR number.
+# Please provide it if you are adding a fragment for a different PR.
+pr: https://github.com/elastic/fleet-server/pull/5169
+
+# Issue URL; optional; the GitHub issue related to this changeset (either closes or is part of).
+# If not present is automatically filled by the tooling with the issue linked to the PR number.
+issue: Add OTel collector properties to policy schema
```
michel-laterman: This should be an issue link (or removed)

internal/pkg/model/ext.go
```diff
@@ -92,21 +93,46 @@ func ClonePolicyData(d *PolicyData) *PolicyData {
 		OutputPermissions: d.OutputPermissions,
 		Outputs:           cloneMap(d.Outputs),
 		Revision:          d.Revision,
-		SecretReferences:  make([]SecretReferencesItems, 0, len(d.SecretReferences)),
+		SecretReferences:  slices.Clone(d.SecretReferences),
```
michel-laterman, line 96: 👍

model/openapi.yml
```diff
@@ -600,6 +622,37 @@ components:
...
+              processors:
+                description: list of enabled processors for each pipeline.
+                type: array
+                items:
+                  type: string
+              receivers:
+                description: list of enabled receivers for each pipeline.
+                type: array
+                items:
+                  type: string
```
michel-laterman: You should also have `x-go-type-skip-optional-pointer: true` for these three properties; that way they match the definition in the JSON schema

internal/pkg/server/fleet_integration_test.go
```diff
@@ -1673,3 +1673,112 @@ func Test_SmokeTest_AuditUnenroll(t *testing.T) {
...
+		Connectors: map[string]any{
+			componentID("forward"): map[string]any{},
+		},
+		Exporters: map[string]any{
+			componentID("someexporter"): map[string]any{},
+		},
+		Service: &model.Service{
+			Pipelines: map[string]*model.PipelinesItem{
+				componentID("metrics"): &model.PipelinesItem{
+					Receivers:  []string{componentID("somereceiver")},
```
michel-laterman, line 1708: Jus to be clear, the componentID generated here is not the same as the top level `Recievers`, is that expected (will a real OTel policy do the same)?

jsoriano: It should be the same, `componentID` just joins the id passed as argument with the previously generated uuid. So given the same argument, it should give the same result.

michel-laterman: yes, my mistake; i misread and thought the uuid generation happend inside the `componentID` method

michel-laterman (approved):
LGTM

PR #5176, opened 2025-07-24
Fix download_rate as string on checkin handling

### LINKED ISSUE #5164: Handle schema incompatibility with Elastic Agent <8.16
Versions of agent before 8.16 may report the download rate as a string, but in ES 8.16+ it is mapped as a double. This was originally reported [here](https://github.com/elastic/ingest-dev/issues/3446). The conclusion in that issue was that the problem was transitory, only came up during upgrades, and resolved itself as the upgrade finished. This is true as long as the upgrade is successful, but if it fails in the downloading or extracting stage, elastic-agent will continue to include the invalid data in the check-in, resulting in an inability to manage the agent from Fleet.

Instead, Fleet Server should detect this problem and set the offending field to 0 if the format is incorrect. This way, older agents can successfully report information about their upgrade state.

It's worth noting that Fleet Server actually sends the invalid data to Elasticsearch instead of rejecting it. The reason for this is that the validation Fleet Server does [assumes](https://github.com/elastic/fleet-server/blob/a2e20909df28177199b8fd0823456fecf3e40f1b/internal/pkg/api/handleCheckin.go#L469) the upgrade details are a union type, so only the fields relevant to a particular state are validated. This, however, is not exactly what elastic-agent sends. This is a bug in elastic-agent, but we need to deal with this in Fleet Server regardless.

michel-laterman:

Fix fleet-server to properly handle when an agent sends the upgrade downloading detail's `download_rate` as a string instead of float64.

Fleet-server will parse the string into a float representing bps and re-marshal the metadata.

## Design Checklist

## Checklist

## Related issues

- Closes #5164


prodsecmachine:

:white_check_mark: **security/snyk** check is complete. No issues have been found. [(View Details)](https://app.snyk.io/org/beats/pr-checks/a51beba2-4012-4672-bf6d-1f2b5ad158a1?refs=27d6dd13ee86fdd1a07c1627dd7695035920694b&source=prChecksComment)

:white_check_mark: **license/snyk** check is complete. No issues have been found. [(View Details)](https://app.snyk.io/org/beats/pr-checks/a51beba2-4012-4672-bf6d-1f2b5ad158a1/license?refs=27d6dd13ee86fdd1a07c1627dd7695035920694b&source=prChecksComment)

internal/pkg/api/handleCheckin.go
```diff
@@ -1125,3 +1127,51 @@ func calcPollDuration(zlog zerolog.Logger, pollDuration, setupDuration, jitterDu
 
 	return pollDuration, jitter
 }
+
+// downloadingBytesFixed will attempt to unpack upgrade details as upgrade metadata
+// However it can encounter an error from earlier agent versions sending a string as the download rate instead of a float.
+// If it encounters this error, it will attemp to fix the download rate when unpacking the object
+// FIXME UpgradeMetadataDownloading has an UnmarshalJSON function that is not being called correctly?
```
michel-laterman: The downloadBytesFixed function I've added here basically reimplements: 
https://github.com/elastic/fleet-server/blob/8474293d301cc46d1821eb233c100e87426ac7c0/internal/pkg/api/metadata.go#L40-L54

But I'm really confused why our `UpgradeMetadataDownloading.UnmarshalJSON` isn't being used internally when `metadata.AsUpgradeMetadataDownloading()` is called:
https://github.com/elastic/fleet-server/blob/8474293d301cc46d1821eb233c100e87426ac7c0/internal/pkg/api/openapi.gen.go#L1691-L1695

My understanding is that the `json.Unmarshal` call should use the Unmarshaller interface

michel-laterman:
Marking this back as a draft, I wasn't confident in the fix and have added an e2e test that shows it's still failing

cmacknz:
Current example of this error in a support case:

> Error while parsing document for index [.fleet-agents-7]: [1:3875] failed to parse field [upgrade_details.metadata.download_rate] of type [double] in document with id '1046bb63-c8aa-4ea2-8690-9bc2bd378af1'. Preview of field's value: '320.822575kBps'

internal/pkg/api/metadata_test.go
```diff
@@ -28,6 +27,10 @@ func Test_ParseDownloadRate(t *testing.T) {
 		name:          "download rate as MBps",
 		raw:           json.RawMessage(`"1MBps"`),
 		expectedValue: 1000000.00,
+	}, {
+		name:          "download rate has decimal",
+		raw:           json.RawMessage(`"320.822575kBps"`),
+		expectedValue: 320822.0, // Decimal is truncated
```
michel-laterman, line 33: @cmacknz, I've added the example as a unit-test.

michel-laterman:
buildkite test this

internal/pkg/api/handleCheckin.go
```diff
@@ -471,11 +471,15 @@ func (ct *CheckinT) processUpgradeDetails(ctx context.Context, agent *model.Agen
...
 		}
-		_, err := details.Metadata.AsUpgradeMetadataDownloading()
+		upgradeDetails, err := details.Metadata.AsUpgradeMetadataDownloading()
 		if err != nil {
 			vSpan.End()
 			return fmt.Errorf("%w %s: %w", ErrInvalidUpgradeMetadata, UpgradeDetailsStateUPGDOWNLOADING, err)
 		}
+		if err := details.Metadata.FromUpgradeMetadataDownloading(upgradeDetails); err != nil {
+			vSpan.End()
+			return fmt.Errorf("%w %s: unable to repack metadata: %w", ErrInvalidUpgradeMetadata, UpgradeDetailsStateUPGDOWNLOADING, err)
```
cmacknz, line 481: Sorry for the late review, but the root cause one layer up from this is we consider parsing upgrade details to be a fatal error and it isn't. Upgrade details are nice to have. 

- https://github.com/elastic/fleet-server/issues/5199

PR #5180, opened 2025-07-28
Changelog for 9.1.0

ebeahan:
In addition to cutting the changelog, also cleaning up many stale fragments from when the `9.1` branch was originally created.


ebeahan:
Generated with `elastic-agent-changelog-tool`. Manual cleanup still WIP.

ebeahan:
> Is it normal that removing 77 changelog fragments, we end up with only 14 entries in the asciidoc and 18 entries in the yaml

Good catch - no, I should have noted in the description. I had to "clean up" many stale fragments from when the branch was originally forked that no longer apply exclusively to 9.1.

pchila (approved):

PR #5232, opened 2025-08-11
Fix enroll to return 503 when ES returns 503.

### LINKED ISSUE #5197: Propogate 503 errors from elasticsearch inside of Fleet Server back to the client
**Describe the issue:**

At the moment when a 503 error occurs during enrollment with an Elastic Agent to Fleet Server that shows as a 500 error back to the Elastic Agent. This is causes the Elastic Agent to not handle the retries, and then enrollment just fails.

**Describe a specific use case for the issue:**

There are many cases where elasticsearch can be temporarily unhealthy which then causes Fleet Server to be unhealthy. It is good to let the caller know this and most likely a retry will cause it to succeed.

blakerouse:

If elasticsearch is returning at error back to Fleet Server, it currently masks it during enrollment and returns a 500 error.

It translates the status code from the elasticsearch response into a proper elasticsearch error and allows the fleet-server error handling to return the proper error back to the client.

1. Place elasticsearch behind a proxy
2. Start fleet-server connected to that proxy
3. Stop elasticsearch so the proxy returns a 503 error
4. Perform enrollment and see that a 503 error is returned

## Design Checklist

- [x] I have or intend to scale test my changes, ensuring it will work reliably with 100K+ agents connected. (no effect to scale)
- ~~[ ] I have included fail safe mechanisms to limit the load on fleet-server: rate limiting, circuit breakers, caching, load shedding, etc.~~ (does not apply)

## Checklist

## Related issues

Closes #5197

### DIFF
--- a/internal/pkg/apikey/invalidate.go
+++ b/internal/pkg/apikey/invalidate.go
@@ -46,7 +46,7 @@ func Invalidate(ctx context.Context, client *elasticsearch.Client, ids ...string
 	defer res.Body.Close()
 
 	if res.IsError() {
-		return fmt.Errorf("fail InvalidateAPIKey: %s", es.TranslateError(res.StatusCode, nil))
+		return fmt.Errorf("fail InvalidateAPIKey: %w", es.TranslateError(res.StatusCode, nil))
 	}
 
 	return nil


prodsecmachine:

:white_check_mark: **security/snyk** check is complete. No issues have been found. [(View Details)](https://app.snyk.io/org/beats/pr-checks/3fc56584-67b3-41b9-b06b-4268b118e1f6?refs=54551edb22fe0c4487afd61377925aba28d150c5&source=prChecksComment)

michel-laterman (approved):

PR #5272, opened 2025-08-19
fix(gsutil): fix installation on 7.17

dliappis:
This commit enhances the gsutil installation process by ensuring that it is installed via snap and removing any previous installations if they don't work.

This PR fixes packaging failures e.g. seen in https://buildkite.com/elastic/fleet-server-package-mbp/builds/2231#0198ae14-0848-4ee7-b666-59f49dc05f8a/65-169
that are related to a mismatch of a distribution (Ubuntu 20.04) that comes with an older Python version (3.8) that `gsutil` doesn't support anymore.

Installs a more modern (non EOL'ed) Python (`3.10`) and after uninstalling the old non-working gsutil via `apt`, installs `gsutil` inside a virtualenv env with the support Python.
It doesn't alter the default system Python.

Can only be tested in BK:

I triggered a build of the packaging pipeline and it passed https://buildkite.com/elastic/fleet-server-package-mbp/builds/2243#0198c283-3c6c-4fb6-b53b-ec94d68107d7/305-401
this should be compared vs earlier failures like: https://buildkite.com/elastic/fleet-server-package-mbp/builds/2231#0198ae14-0848-4ee7-b666-59f49dc05f8a/65-169

## Checklist

### DIFF
--- a/.buildkite/scripts/common.sh
+++ b/.buildkite/scripts/common.sh
@@ -135,6 +135,7 @@ fix_gsutil() {
         sudo chown -R "$(id -u)":"$(id -g)" "$CONFIG_DIR"
 
         # Install a modern Python alongside (3.10) -- this doesn't affect the system Python
+        export DEBIAN_FRONTEND=noninteractive
         sudo apt-get update -y
         sudo apt-get install -y software-properties-common curl ca-certificates gnupg
         if ! command -v python3.10 >/dev/null 2>&1; then
@@ -172,7 +173,7 @@ EOF
 
         # Handy symlinks (works even if some shells ignore /etc/profile.d)
         for b in gcloud gsutil bq; do
-            sudo ln -sf "/opt/google-cloud-sdk/bin/\$b" "/usr/local/bin/\$b"
+            sudo ln -sf "/opt/google-cloud-sdk/bin/$b" "/usr/local/bin/$b"
         done
 
         # Verify (new shell or source the file)


dliappis:
I triggered a build of the packaging pipeline and it passed https://buildkite.com/elastic/fleet-server-package-mbp/builds/2241#0198c24e-998d-4b46-8887-1228e031a35f
this should be compared vs earlier failures like: https://buildkite.com/elastic/fleet-server-package-mbp/builds/2231#0198ae14-0848-4ee7-b666-59f49dc05f8a/65-169

ebeahan (approved):

ebeahan (approved):

.buildkite/scripts/common.sh
```diff
@@ -95,6 +95,94 @@ with_Terraform() {
...
+        # Google Cloud CLI env
+        export CLOUDSDK_PYTHON="$VENV/bin/python"
+        export CLOUDSDK_CONFIG="$CONFIG_DIR"
+        case ":\$PATH:" in *:/opt/google-cloud-sdk/bin:*) ;; *) export PATH="/opt/google-cloud-sdk/bin:\$PATH";; esac
+EOF
+        sudo chmod 644 /etc/profile.d/gcloud.sh
+
+        # Handy symlinks (works even if some shells ignore /etc/profile.d)
+        for b in gcloud gsutil bq; do
+            sudo ln -sf "/opt/google-cloud-sdk/bin/\$b" "/usr/local/bin/\$b"
```
pkoutsovasilis: nit: since this isn't a heredoc do we need to escape like `\$b`?

.buildkite/scripts/common.sh
```diff
@@ -95,6 +95,94 @@ with_Terraform() {
...
+                return 1
+                ;;
+        esac
+
+        # Ensure config dir exists and is writable by the current user
+        sudo mkdir -p "$CONFIG_DIR"
+        sudo chown -R "$(id -u)":"$(id -g)" "$CONFIG_DIR"
+
+        # Install a modern Python alongside (3.10) -- this doesn't affect the system Python
+        sudo apt-get update -y
```
pkoutsovasilis, line 139: nit: I would also export `DEBIAN_FRONTEND=noninteractive` just to be safe. In theory `-y` does the same but I have faint recollections from past experience that there are some cases that the env var is still needed. Feel free to ignore

pkoutsovasilis (approved):
left some nits, but lgtm

.buildkite/scripts/common.sh
dliappis: Happy to admit this was an AI introduced bug :) fixed in b036864c92bfa733471cebdbd514a54964d94ba3

.buildkite/scripts/common.sh
dliappis: Better be safe than sorry, thanks! addressed in b036864c92bfa733471cebdbd514a54964d94ba3

ebeahan (approved):

PR #5317, opened 2025-08-21
Reset trace headers for bulk items when returning to pool

michel-laterman:

Bulker items can have a previous item's trace links.

Clear the reference to the trace link in the block's reset method. This method is called before a block is put back in the sync.Pool.

Note that this fix would only apply to functions using the bulker without a transaction tied to their context.

### DIFF
--- a/changelog/fragments/1755810145-Reset-trace-headers-on-bulk-items-when-returning-to-pool.yaml
+++ b/changelog/fragments/1755810145-Reset-trace-headers-on-bulk-items-when-returning-to-pool.yaml
@@ -1,0 +1,32 @@
+# Kind can be one of:
+# - breaking-change: a change to previously-documented behavior
+# - deprecation: functionality that is being removed in a later release
+# - bug-fix: fixes a problem in a previous version
+# - enhancement: extends functionality but does not break or fix existing behavior
+# - feature: new functionality
+# - known-issue: problems that we are aware of in a given version
+# - security: impacts on the security of a product or a user’s deployment.
+# - upgrade: important information for someone upgrading from a prior version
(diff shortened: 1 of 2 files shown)


prodsecmachine:

:white_check_mark: **security/snyk** check is complete. No issues have been found. [(View Details)](https://app.snyk.io/org/beats/pr-checks/47069274-b98f-4d1c-a521-a947236f9a30?refs=9b66d7eb80920112e8f3cfab367a52bd6a7641cb&source=prChecksComment)

changelog/fragments/1755810145-Reset-trace-headers-on-bulk-items-when-returning-to-pool.yaml
```diff
@@ -0,0 +1,32 @@
...
+# - enhancement: extends functionality but does not break or fix existing behavior
+# - feature: new functionality
+# - known-issue: problems that we are aware of in a given version
+# - security: impacts on the security of a product or a user’s deployment.
+# - upgrade: important information for someone upgrading from a prior version
+# - other: does not fit into any of the other categories
+kind: bug-fix
+
+# Change summary; a 80ish characters long description of the change.
+summary: Reset trace links on bulk items when returning to pool
```
ycombinator, line 14: Correct me if I'm wrong but this change is related to getting APM traces for Fleet Server.  If so, there is no user impact from this change, right?  In that case, do we even need a CHANGELOG entry?

michel-laterman: Most likely, but IIRC a self-managed user can enable tracing

PR #5372, opened 2025-08-28
Restore connection-level limiter

michel-laterman:

Use of only the throttle middleware can result in OOM incidents under high load.

Restore connection level limiter to prevent OOM incidents. (removed in #4402)
This limiter is used in addition to the request-level throttle so that once
our in-flight requests reaches max_connections a 429 is returned, but if the
total connections the server uses is over max_connections*1.1 the server drops
the connection before the TLS handshake.

## Design Checklist

## Checklist

### DIFF
--- a/NOTICE-fips.txt
+++ b/NOTICE-fips.txt
@@ -4595,6 +4595,43 @@ THE SOFTWARE.
 
 
 --------------------------------------------------------------------------------
+Dependency : golang.org/x/net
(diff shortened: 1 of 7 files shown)


blakerouse (commented):
Overall this looks good, but unit tests for the listener would be great. Could you add those?

blakerouse (approved):
Thanks for the added unit tests. Looks good.

PR #5396, opened 2025-09-03
Fix configFromUnit test, add logging for e2e tests

michel-laterman:
Fix flakey tests seen in https://github.com/elastic/fleet-server/pull/5384 and https://github.com/elastic/fleet-server/pull/5383.
Add logging to help debug agent container test issues.

### DIFF
--- a/internal/pkg/server/agent.go
+++ b/internal/pkg/server/agent.go
@@ -282,50 +282,48 @@ func (a *Agent) unitModified(ctx context.Context, unit *client.Unit) error {
 			// not our input unit; would have been marked failed in unitAdded; do nothing
 			return nil
 		}
-		if exp.State == client.UnitStateHealthy {
+		switch exp.State {
+		case client.UnitStateHealthy:
 			if a.outputUnit == nil {
 				// still no output unit; would have been marked starting already; do nothing
 				return nil
 			}
-
 			// configuration modified (should still be running)
 			return a.reconfigure(ctx)
-		} else if exp.State == client.UnitStateStopped {
+		case client.UnitStateStopped:
 			// unit should be stopped
 			a.stop()
 			return nil
+		default:
+			return fmt.Errorf("unknown unit state %v", exp.State)
 		}
-		return fmt.Errorf("unknown unit state %v", exp.State)
 	}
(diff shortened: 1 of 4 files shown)


internal/pkg/server/agent.go
```diff
@@ -726,8 +724,7 @@ func (a *Agent) esOutputCheck(ctx context.Context, cfg map[string]interface{}) e
 func (a *Agent) esOutputCheckLoop(ctx context.Context, delay time.Duration, cfg map[string]interface{}) {
 	for {
 		if err := sleep.WithContext(ctx, delay); err != nil {
-			zerolog.Ctx(ctx).Debug().Msg("Async output check context cancelled")
```
ycombinator, line 729: Why are we removing this debug log line?

internal/pkg/ver/check.go
```diff
@@ -28,17 +29,26 @@ var (
 )
 
 // CheckCompatiblility will check the remote Elasticsearch version retrieved by the Elasticsearch client with the passed fleet version.
-// Versions are compatible when Elasticsearch's version is greater then or equal to fleet-server's version
+// Versions are compatible when Elasticsearch's version is greater then or equal to fleet-server's version/
```
ycombinator: Typo?

```suggestion
// Versions are compatible when Elasticsearch's version is greater then or equal to fleet-server's version
```

internal/pkg/ver/check.go
```diff
@@ -28,17 +29,26 @@ var (
...
 func CheckCompatibility(ctx context.Context, esCli *elasticsearch.Client, fleetVersion string) (string, error) {
-	zerolog.Ctx(ctx).Debug().Str("fleet_version", fleetVersion).Msg("check version compatibility with elasticsearch")
+	// Version checks may run concurrently with other operations
+	// This can causee some flakeyness with tests so we need to get the logger from the context before its cancelled
```
ycombinator:
```suggestion
	// This can cause some flakiness with tests so we need to get the logger from the context before its cancelled
```

internal/pkg/server/agent.go
michel-laterman: causes a race condition in tests

PR #5455, opened 2025-09-06
fix: switch Dockerfile of serverless image to debian trixie

pkoutsovasilis:

PR [#5400](https://github.com/elastic/fleet-server/pull/5400) introduced building serverless images with `golang-crossbuild`.  
This caused failures in the `Publish docker image` step ([example](https://buildkite.com/elastic/fleet-server/builds/10400#01991afe-4fb0-4396-aa98-06474a3fc47b)) on `main` because the `docker:publish` mage target never sets the `SUFFIX` build-arg.

This PR switches serverless images to `golang:${GO_VERSION}-trixie`, which removes the need for a suffix.  
I avoided `golang-crossbuild` because `docker:publish` relies on buildx for multi-platform builds, while `golang-crossbuild` uses image tags per arch. That mismatch could lead to hidden issues.  

I’m still unsure how the FIPS variant works with `docker:publish`, but it doesn’t appear to be actively used. I’ll sync with @michel-laterman on this.

The target to run is `mage docker:publish` but this may end up pushing an image for you, so I would be very cautious before issuing that (remove the "--push" arg [here](https://github.com/elastic/fleet-server/blob/main/magefile.go#L1066)).

## Design Checklist

## Checklist

## Related issues

N/A

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,6 +1,5 @@
 ARG GO_VERSION
-ARG SUFFIX
-FROM --platform=${BUILDPLATFORM:-linux} docker.elastic.co/beats-dev/golang-crossbuild:${GO_VERSION}-${SUFFIX} AS builder
+FROM --platform=${BUILDPLATFORM:-linux} golang:${GO_VERSION}-trixie AS builder
 
 WORKDIR /fleet-server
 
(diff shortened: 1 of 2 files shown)


prodsecmachine:

:white_check_mark: **security/snyk** check is complete. No issues have been found. [(View Details)](https://app.snyk.io/org/beats/pr-checks/2cfe9ea6-5176-488d-933d-daaca7a6541a?refs=0d82dd482de25bf1c793fbe4551095c91d7fc198&source=prChecksComment)

ebeahan (approved):

PR #5478, opened 2025-09-09
Make File Transfer Storage Size Configurable

pzl:

Adds a configuration parameter to fleet server for the maximum size of transferred files via actions

Hard coded limit (100MiB previously) is now a config.

Default is also changed to be unlimited. The file transfer feature launched with _very_ conservative limits on request rates and file sizes.  File size itself is primarily an impact on Elasticsearch at rest, at does not effect fleet server, aside from the time to transfer the bytes. Requests are still chunked, chunk size is not changing.

## Design Checklist

## Checklist

### DIFF
--- a/changelog/fragments/1757423719-Make-file-storage-size-configurable.yaml
+++ b/changelog/fragments/1757423719-Make-file-storage-size-configurable.yaml
@@ -1,0 +1,32 @@
+# Kind can be one of:
+# - breaking-change: a change to previously-documented behavior
+# - deprecation: functionality that is being removed in a later release
+# - bug-fix: fixes a problem in a previous version
+# - enhancement: extends functionality but does not break or fix existing behavior
+# - feature: new functionality
+# - known-issue: problems that we are aware of in a given version
+# - security: impacts on the security of a product or a user’s deployment.
+# - upgrade: important information for someone upgrading from a prior version
+# - other: does not fit into any of the other categories
+kind: enhancement
+
+# Change summary; a 80ish characters long description of the change.
+summary: Makes file storage size configurable
+
+# Long description; in case the summary is not enough to describe the change
+# this field accommodate a description without length limits.
+# NOTE: This field will be rendered only for breaking-change and known-issue kinds at the moment.
+#description:
+
+# Affected component; a word indicating the component this changeset affects.
+component: fleet-server
+
+# PR URL; optional; the PR number that added the changeset.
+# If not present is automatically filled by the tooling finding the PR where this changelog fragment has been added.
+# NOTE: the tooling supports backports, so it's able to fill the original PR number instead of the backport PR number.
+# Please provide it if you are adding a fragment for a different PR.
+#pr: https://github.com/owner/repo/1234
+
+# Issue URL; optional; the GitHub issue related to this changeset (either closes or is part of).
+# If not present is automatically filled by the tooling with the issue linked to the PR number.
+#issue: https://github.com/owner/repo/1234
--- a/internal/pkg/api/handleUpload_test.go
+++ b/internal/pkg/api/handleUpload_test.go
@@ -11,6 +11,7 @@ import (
 	"crypto/sha256"
 	"encoding/hex"
 	"encoding/json"
+	"fmt"
 	"io"
 	"io/ioutil"
 	"net/http"
@@ -63,18 +64,6 @@ func TestUploadBeginValidation(t *testing.T) {
 			`{
 				"file": {
 					"size": 200,
-					"name": "foo.png",
-					"mime_type": "image/png"
-				},
-				"agent_id": "foo",
-				"action_id": "123",
-				"src": "agent"
-			}`,
-		},
-		{"Oversized file should be rejected", http.StatusBadRequest, "size",
-			`{
-				"file": {
-					"size": ` + strconv.Itoa(maxFileSize+1024) + `,
 					"name": "foo.png",
 					"mime_type": "image/png"
 				},
@@ -346,6 +335,48 @@ func TestUploadBeginBadRequest(t *testing.T) {
 	assert.Equal(t, http.StatusBadRequest, rec.Code)
 }
 
+func TestUploadBeginFileSize(t *testing.T) {
+
+	mockFile := func(size int64) string {
+		return fmt.Sprintf(`{
+				"file": {
+					"size": %d,
+					"name": "foo.png",
+					"mime_type": "image/png"
+				},
+				"agent_id": "foo",
+				"action_id": "123",
+				"src": "agent"
+			}`, size)
+	}
+
+	// now test various body contents
+	tests := []struct {
+		Name         string
+		MaxSize      int64
+		ExpectStatus int
+		InputSize    int64
+	}{
+		{"MaxSize 0 allows uploads", 0, http.StatusOK, 1000},
+		{"MaxSize 0 allows large uploads", 0, http.StatusOK, 1024 * 1024 * 1024 * 2},
+		{"MaxSize 0 does not allow 0-length files", 0, http.StatusBadRequest, 0},
+		{"Sizes larger than MaxSize are denied", 1024, http.StatusBadRequest, 2048},
+		{"Sizes smaller than MaxSize are allowed", 1024, http.StatusOK, 900},
+	}
+
+	for _, tc := range tests {
+		t.Run(tc.Name, func(t *testing.T) {
+
+			hr, _, _, _ := configureUploaderMock(t, tc.MaxSize)
+			rec := httptest.NewRecorder()
+			req := httptest.NewRequest(http.MethodPost, RouteUploadBegin, strings.NewReader(mockFile(tc.InputSize)))
+			hr.ServeHTTP(rec, req)
+			assert.Equal(t, tc.ExpectStatus, rec.Code)
+		})
+	}
+
+}
+
 /*
   Chunk data upload route
 */
@@ -377,7 +408,7 @@ func TestChunkUploadRouteParams(t *testing.T) {
 			mockUploadInfoResult(fakebulk, file.Info{
 				DocID:     "bar.foo",
 				ID:        mockUploadID,
-				ChunkSize: maxFileSize,
+				ChunkSize: file.MaxChunkSize,
 				Total:     file.MaxChunkSize + 1,
 				Count:     2, // this is a 2-chunk "file" based on size above
 				Start:     time.Now(),
@@ -410,7 +441,7 @@ func TestChunkUploadRequiresChunkHashHeader(t *testing.T) {
 	mockUploadInfoResult(fakebulk, file.Info{
 		DocID:     "bar.foo",
 		ID:        mockUploadID,
-		ChunkSize: maxFileSize,
+		ChunkSize: file.MaxChunkSize,
 		Total:     10,
 		Count:     1,
 		Start:     time.Now(),
@@ -458,7 +489,7 @@ func TestChunkUploadStatus(t *testing.T) {
 			mockUploadInfoResult(fakebulk, file.Info{
 				DocID:     "bar.foo",
 				ID:        mockUploadID,
-				ChunkSize: maxFileSize,
+				ChunkSize: file.MaxChunkSize,
 				Total:     10,
 				Count:     1,
 				Start:     time.Now(),
@@ -509,7 +540,7 @@ func TestChunkUploadExpiry(t *testing.T) {
 			mockUploadInfoResult(fakebulk, file.Info{
 				DocID:     "bar.foo",
 				ID:        mockUploadID,
-				ChunkSize: maxFileSize,
+				ChunkSize: file.MaxChunkSize,
 				Total:     10,
 				Count:     1,
 				Start:     tc.StartTime,
@@ -547,7 +578,7 @@ func TestChunkUploadWritesTimestamp(t *testing.T) {
 	mockUploadInfoResult(fakebulk, file.Info{
 		DocID:     "bar.foo",
 		ID:        mockUploadID,
-		ChunkSize: maxFileSize,
+		ChunkSize: file.MaxChunkSize,
 		Total:     10,
 		Count:     1,
 		Start:     time.Now(),
@@ -597,7 +628,7 @@ func TestUploadCompleteRequiresMatchingAuth(t *testing.T) {
 			mockInfo := file.Info{
 				DocID:     "bar." + tc.AgentInFileRecord,
 				ID:        mockUploadID,
-				ChunkSize: maxFileSize,
+				ChunkSize: file.MaxChunkSize,
 				Total:     10,
 				Count:     1,
 				Start:     time.Now().Add(-time.Minute),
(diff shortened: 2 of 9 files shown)


internal/pkg/config/limits.go
```diff
@@ -16,9 +16,10 @@ type Limit struct {
 }
 
 type ServerLimits struct {
-	MaxAgents         int `config:"max_agents"`
-	MaxHeaderByteSize int `config:"max_header_byte_size"`
-	MaxConnections    int `config:"max_connections"`
+	MaxAgents          int   `config:"max_agents"`
+	MaxHeaderByteSize  int   `config:"max_header_byte_size"`
+	MaxConnections     int   `config:"max_connections"`
+	MaxFileStorageSize int64 `config:"max_file_storage_size"`
```
ycombinator: Could we make this `MaxFileStorageByteSize` to clarify the units (similar to `MaxHeaderByteSize` above)?
```suggestion
	MaxFileStorageByteSize int64 `config:"max_file_storage_byte_size"`
```

internal/pkg/config/env_defaults.go
```diff
@@ -57,6 +57,8 @@ const (
 	defaultStatusMax      = 50
 	defaultStatusMaxBody  = 0
 
+	defaultFileStorageSize = 0 // no limit
```
ycombinator: Should we change this to match the previous default, `104857600`?

internal/pkg/file/uploader/upload.go
```diff
@@ -73,7 +73,7 @@ func (u *Uploader) Begin(ctx context.Context, namespaces []string, data JSDict)
 	}
 
 	size, _ := data.Int64("file", "size")
-	if size > u.sizeLimit {
+	if u.sizeLimit != 0 && size > u.sizeLimit {
```
ycombinator: I understand the `u.sizeLimit != 0` clause here is for allowing uploads of any size.  So `0` is the sentinel value for signaling to the code that uploads of any size are allowed.  

I'm wondering if we should use `0` to signal that uploads are not allowed and `-1` (or any other value < `0`) to signal that uploads of any size are allowed?

pzl: Disallowing in just this manner would lead to some wacky and hard-to-debug errors in kibana features that rely on the file transfer. Kibana features (especially in Defend) would still appear enabled and allow user actions to take place that would fail when the agent (or endpoint) receives an action and then proceeds to upload.

I do see the arbitrary nature of 0's and negatives as not particularly meaningful to whichever disabled/unlimited mode here. I guess they happen to be flipped here. -1 (or any negative value) will fail to upload, as file size will always be larger than that. Negatives seem to be the disabler here (not by design, just by happenstance).  `0` seemed to be a common unset/default/unlimited value elsewhere in these configs (and go) and I went with that by convention.  I think the unset/default value of the language probably shouldn't lead to the feature being _disabled_.

internal/pkg/config/env_defaults.go
pzl: The specific onus for this PR is removing the file size limit. It was initially put there as a precautionary measure during initial development of the feature, but upcoming feature work in the stack will see much larger file sizes in default use cases. 

Rather than deleting this control altogether, it seemed sensible to keep it as an optional limit (either as cost-control for the user, or to prevent "oopsies" if they don't intend to use the larger-file features coming). If we'd rather the control just be eliminated as complexity, that seems fine too.

pchila: @pzl has any benchmark been performed to check the behaviour of fleet server when uploading large(r) files? Do we have an idea of the size of files in the upcoming feature ?

pzl: Yes - a modified horde was used to have 8k drones sending >2GiB files (randomized range was 1.6Gib to 16 GiB) through fleet server with disabled file size limit, without issue. 

The primary control here that maintains stability (in the face of all that traffic) is the request rate limits (the file ops have their own limit settings), and the architecture of file chunking.   An entire 16GiB file is not continuously streamed in one request, but rather individual 4MiB chunks. Those chunk requests have several rate controls on them, as well as limits on how many "upload jobs" may be ongoing (i.e. there will not be 8k agents all trying to upload chunks, fleet server controls how many in-progress file jobs there are).  So memory will not be exhausted by too many parallel uploads, nor by buffering large files themselves (the 4MiB chunk data as well, is immediately streamed to elasticsearch with minimal processing, so even that is effectively `io.Copy`ied from network-in to net-out.

Fleet server is then able to handle its "max parallel upload jobs" amount -- essentially continually and indefinitely. A chunk request comes, is copied to elasticsearch, and closes. There is a (still very conservative!) limit on how many chunk requests can be handled at once.  So large files is not currently increasing any of that handling in parallel, but only how many chunk calls will happen in sequence until the whole "upload job" has concluded.   And then of course, at rest, fleet server does nothing with these files, they just sit as documents in elasticsearch. No load or impact to fleet server there.

changelog/fragments/1757423719-Make-file-storage-size-configurable.yaml
```diff
@@ -0,0 +1,32 @@
...
+
+# PR URL; optional; the PR number that added the changeset.
+# If not present is automatically filled by the tooling finding the PR where this changelog fragment has been added.
+# NOTE: the tooling supports backports, so it's able to fill the original PR number instead of the backport PR number.
+# Please provide it if you are adding a fragment for a different PR.
+#pr: https://github.com/owner/repo/1234
+
+# Issue URL; optional; the GitHub issue related to this changeset (either closes or is part of).
+# If not present is automatically filled by the tooling with the issue linked to the PR number.
+#issue: https://github.com/owner/repo/1234
```
kaanyalti, line 32:
```suggestion
issue: https://github.com/elastic/fleet-server/pull/5478
```

internal/pkg/api/handleUpload_test.go
```diff
@@ -346,6 +335,48 @@ func TestUploadBeginBadRequest(t *testing.T) {
...
+	// now test various body contents
+	tests := []struct {
+		Name         string
+		MaxSize      int64
+		ExpectStatus int
+		InputSize    int64
+	}{
+		{"MaxSize 0 allows uploads", 0, http.StatusOK, 1000},
+		{"MaxSize 0 allows large uploads", 0, http.StatusOK, 1024 * 1024 * 1024 * 2},
+		{"MaxSize 0 does not allow 0-length files", 0, http.StatusBadRequest, 0},
```
nkvoll: This seems a little odd? If I set MaxSize 1, would I allow 0-length files?

changelog/fragments/1757423719-Make-file-storage-size-configurable.yaml
```diff
@@ -0,0 +1,32 @@
...
+#description:
+
+# Affected component; a word indicating the component this changeset affects.
+component: fleet-server
+
+# PR URL; optional; the PR number that added the changeset.
+# If not present is automatically filled by the tooling finding the PR where this changelog fragment has been added.
+# NOTE: the tooling supports backports, so it's able to fill the original PR number instead of the backport PR number.
+# Please provide it if you are adding a fragment for a different PR.
+#pr: https://github.com/owner/repo/1234
```
kaanyalti, line 28:
```suggestion
pr: https://github.com/elastic/fleet-server/pull/5478
```

changelog/fragments/1757423719-Make-file-storage-size-configurable.yaml
kaanyalti, line 32: small nit, if there is a parent issue it would be nice to have it here

internal/pkg/api/handleUpload_test.go
pzl: nope! This test was just to make sure a size limit of "0" is not interpreted as a literal value, making "0 length files" suddenly succeed, when the limit is 0.

0-length files should always fail, but this makes sure we don't trip on an edge case when the limit is also `0`.

internal/pkg/config/env_defaults.go
nkvoll: Would users / customers be accustomed to this limiting / preventing certain files from being sent? I'm asking to try to get a handle what would be the "principle of least surprise" for users. Whether that would be "very large files are sent (even if multiple TBs)", or "I need to configure a max if above a threshold", and whether our previous limit of 100 MiB would have set an expectation.

internal/pkg/file/uploader/upload.go
nkvoll: Did we consider making an explicit `[something]Enabled: true/false` flag (and require the max*Size to unsigned) to differentiate and avoid magic numbers (at least in the user-facing interface)?

internal/pkg/config/env_defaults.go
pzl: This feature is only in use in two places:

- agent diagnostics bundle (very small zips, not going to change here)
- defend/endpoint file retrieval or distribution.

The defend feature is very much manual user interaction to perform. Users  _have_ run into size limits when attempting to inspect larger binaries, and have had to resort to other means to retrieve those files. Defend needs to unblock a lot of those common workflows.

We're not aware of any case of  _relying_ on the size limit as a feature, but more cases of users being hampered by it.

internal/pkg/file/uploader/upload.go
pzl: I'm not sure I understand the suggestion here.

change the config's type to `uint64`, and then:

```
// with other the consts
var fileSizeLimitDisabled = 0

...

// when setting defaults
defaultFileSizeLimit = fileSizeLimitDisabled
```

?

That's good with me, I'm happy with that. 

---

Just to restate the general case here: the end goal is to _remove_ the file size cap, that is the specific need. I'd like to leave it configurable in case some user has strange edge cases I cannot predict and wants to bound it for their own reasons. But it is not essential to keep this a config. We can simplify the PR if we just want to remove the cap. The actual protections in place are the rate limits.  If having a sentinel value here at `0` is the blocker, should we go the other way?

internal/pkg/config/env_defaults.go
```diff
@@ -141,8 +143,9 @@ type limit struct {
 }
 
 type serverLimitDefaults struct {
-	PolicyThrottle time.Duration `config:"policy_throttle"` // deprecated: replaced by policy_limit
-	MaxConnections int           `config:"max_connections"`
+	PolicyThrottle         time.Duration `config:"policy_throttle"` // deprecated: replaced by policy_limit
+	MaxConnections         int           `config:"max_connections"`
+	MaxFileStorageByteSize int64         `config:"max_file_storage_size"`
```
pchila: if negative values are not allowed maybe this should be defined as
```suggestion
	MaxFileStorageByteSize uint64         `config:"max_file_storage_size"`
```
or maybe add some validation using ucfg

internal/pkg/file/uploader/upload.go
pchila: I was about to add a comment about the config type to be changed to uint64, then I saw this comment so I'll try to keep the number of open threads to a minimum.

I don't want to hijack @nkvoll comment and will leave it to him to get back here, I would love to see the `uint64` type if negative values are not allowed and a constant that will read something like wrote in your comment.

About the sentinel value/explicit enabled/disabled flag I am not sure: if we have such a flag we could leave the default limit to 100MB as today (still configurable using `MaxFileStorageByteSize`)  and just turn off the limit using the `FileStorageLimitEnabled: false` (name still open to change :wink: ) but I am not 100% sure that this is what @nkvoll envisioned

internal/pkg/config/limits.go
```diff
@@ -16,9 +16,10 @@ type Limit struct {
 }
 
 type ServerLimits struct {
-	MaxAgents         int `config:"max_agents"`
-	MaxHeaderByteSize int `config:"max_header_byte_size"`
-	MaxConnections    int `config:"max_connections"`
+	MaxAgents              int   `config:"max_agents"`
+	MaxHeaderByteSize      int   `config:"max_header_byte_size"`
+	MaxConnections         int   `config:"max_connections"`
+	MaxFileStorageByteSize int64 `config:"max_file_storage_size"`
```
nkvoll: I wonder, for consistency, if this should have been `max_file_storage_byte_size`, (even if my preference would have been to have the unit come last, if we include it at all) to align with `max_header_byte_size`, two lines above, and align with the variable name.

nkvoll: Naming is hard and begs some questions, so please pardon me for asking a question with maybe an obvious answer, but: Why is it not `max_file_size` (e.g why is `storage` part of it -- could it be simplified but still describing exactly what it is)? It raises some "storage size vs display size"-connotations to me, which may be completely irrelevant.

internal/pkg/file/uploader/upload.go
nkvoll: Thanks for the general-case explanation @pzl, much appreciated!

What I'm thinking is pretty much what @pchila is saying, but maybe with a slight variant to avoid discussing the naming, by having it be a pointer. If the pointer is `nil`, there is no max.

```
MaxFileStorageByteSize *int64 `config:"max_file_storage_size"`
```

If I understand correctly, we want to default to "no maximum configured", and this being a pointer, usually defaults to `nil`, and is different from the any other legal value, which intuitively would be compared to the file size on disk?

internal/pkg/config/limits.go
pzl: I tried to avoid an overly-broad name in case fleet server ever needs to deal with files in another context (e.g. supporting files on the host its on, cache files, or any other situation involving "files" in fleet). I thought of this feature (file transfer to/from an agent and elasticsearch) as "file storage" to give it a narrower namespace. But it can be whatever, no attachment to the name

internal/pkg/file/uploader/upload.go
pzl: in re: negative numbers being nonsensical values here -- do we want this to be `*uint64`?

Also what should happen if it _is_ configured and set to 0? Feature disabled?

ebeahan:
@pzl do you plan to continue work on this PR? If not, I'll plan to close, and you can reopen/recreate when ready to work on it again.

internal/pkg/file/uploader/upload.go
ycombinator: I'd be in favor of what's been proposed in the last two comments above (from @nkvoll and @pzl):
- Make this field's type `*uint64`.
- When the value is `nil`, it means there is no maximum configured == no limit
- When the value is `0`, it means a maximum of 0 bytes is allowed == feature is disabled
- For all other values (`> 0`), that is the maximum size in bytes allowed

nkvoll:
I'm on board with what @ycombinator wrote above, that makes sense to me 👍

pchila:
> > > continue work on this PR?
> > 
> > 
> > Thanks @ebeahan , we're actually picking this back up now!
> > Seems like where this left off, was discussion around what type the storage size variable should be, and meanings on edge case numbers (0, negative values), for some concept of "unlimited".
> > @ycombinator @nkvoll @pchila Can we reach a consensus on what we want that to be?
> 
> Apologies for missing this ping (thanks @ebeahan). I think we're pretty close / aligned; here's my take: [#5478 (comment)](https://github.com/elastic/fleet-server/pull/5478#discussion_r2499030785)

@ycombinator proposal works for me as well :+1:

nkvoll (approved):

internal/pkg/config/env_defaults.go
```diff
@@ -141,8 +141,9 @@ type limit struct {
 }
 
 type serverLimitDefaults struct {
-	PolicyThrottle time.Duration `config:"policy_throttle"` // deprecated: replaced by policy_limit
-	MaxConnections int           `config:"max_connections"`
+	PolicyThrottle         time.Duration `config:"policy_throttle"` // deprecated: replaced by policy_limit
+	MaxConnections         int           `config:"max_connections"`
+	MaxFileStorageByteSize *uint64       `config:"max_file_storage_size"`
```
michel-laterman, line 146:
```suggestion
	MaxFileStorageByteSize *uint64       `config:"max_file_storage_byte_size"`
```

this is mostly a nitpick, but can you also (manually) add this to `fleet-server.reference.yml`?

changelog/fragments/1757423719-Make-file-storage-size-configurable.yaml
michel-laterman, line 28: please link the pr

PR #5493, opened 2025-09-10
Refactor the bulk checkin handler to use optional args

michel-laterman:

`checkin.Bulk.CheckIn` function is hard to extend or understand at a glance.

Refactor the bulk checkin handler to use optional args so the list is easier to read and so that we can more easily add policy ID/revision idx identifiers as a follow up.

`mage test`

## Design Checklist

## Checklist

## Related issues

- Relates https://github.com/elastic/elastic-agent/issues/6446


internal/pkg/api/handleCheckin.go
```diff
@@ -382,7 +382,7 @@ func (ct *CheckinT) ProcessRequest(zlog zerolog.Logger, w http.ResponseWriter, r
 				zlog.Trace().Msg("fire long poll")
 				break LOOP
 			case <-tick.C:
-				err := ct.bc.CheckIn(agent.Id, string(req.Status), req.Message, nil, rawComponents, nil, ver, unhealthyReason, false)
+				err := ct.bc.CheckIn(agent.Id, checkin.WithStatus(string(req.Status)), checkin.WithMessage(req.Message), checkin.WithComponents(rawComponents), checkin.WithVer(ver), checkin.WithUnhealthyReason(unhealthyReason)) // FIXME If we change to properly handle empty strings we could stop passing optional args here.
```
michel-laterman, line 385: As this call includes rawComponents and ver, the bulker's serialization cache is not used

However, I think that all we need to do is bump the timestamp associated with an agent on each tick here so we don't actually need to send any args as the status/message/unhealthyreason should not change per tick.

Removing the extra args should result in a very minor performance improvement (repeated serialization takes much less time than the `MUpdate` network call).

internal/pkg/api/handleCheckin.go
```diff
@@ -327,7 +327,7 @@ func (ct *CheckinT) ProcessRequest(zlog zerolog.Logger, w http.ResponseWriter, r
 	// Initial update on checkin, and any user fields that might have changed
 	// Run a script to remove audit_unenrolled_* and unenrolled_at attributes if one is set on checkin.
 	// 8.16.x releases would incorrectly set unenrolled_at
-	err = ct.bc.CheckIn(agent.Id, string(req.Status), req.Message, rawMeta, rawComponents, seqno, ver, unhealthyReason, agent.AuditUnenrolledReason != "" || agent.UnenrolledAt != "")
+	err = ct.bc.CheckIn(agent.Id, checkin.WithStatus(string(req.Status)), checkin.WithMessage(req.Message), checkin.WithMeta(rawMeta), checkin.WithComponents(rawComponents), checkin.WithSeqNo(seqno), checkin.WithVer(ver), checkin.WithUnhealthyReason(unhealthyReason), checkin.WithDeleteAudit(agent.AuditUnenrolledReason != "" || agent.UnenrolledAt != ""))
```
pchila, line 330: How many of these parameters are mandatory? If data is always supposed to be there, those shouldn't be variadic arguments

michel-laterman: For the initial checkin, the status, message, and unhealthy reason are mandatory, but as I described in my other comment i don't think these should be required when updating the timestamps on each tick of the long-poll

EDIT: Note this is what the fleet-server requires, but I'm not sure if the UI expects other attributes to be required/updated or how it handles older agent versions if anything in `extraT` is required.

PR #5501, opened 2025-09-11
Add agent_policy_id and policy_revision_idx to checkin request 

michel-laterman:

Agent policy details (id + revision) may go out of sync with what fleet-server records.
This occurs when a VM running and agent with policy `X` on revision `N` is restored to an earlier time when it was running `X N-1`, or even a different policy. It can also occur if the ES cluster is restored to an earlier snapshot where the agent will be running policy N+1.

Allow the agents to add their currently running policy_id and
revision_idx attributes to the checkin request bodies. These attributes,
if included and different from the agent doc will be used when updating
the agent doc in the pre-poll checkin. If the agent's policy id does not
match the expected policy id from the server a reassign is detected and
a new policy change action will be sent. If the revision differs a policy
change action will also be sent. If an agent checks in with a different
policy/revision the api keys may be managed. Add a feature flag to disable
this behaviour and only use Acks + the fleet-agents doc as the source of
truth.

Added testing in the e2e API test suite, run:
```
mage test:e2e
```

## Design Checklist

## Checklist

## Related issues

- Relates https://github.com/elastic/elastic-agent/issues/6446

### DIFF
--- a/internal/pkg/api/handleCheckin.go
+++ b/internal/pkg/api/handleCheckin.go
@@ -1177,7 +1177,7 @@ func (ct *CheckinT) processPolicyDetails(ctx context.Context, zlog zerolog.Logge
 	// Check if the checkin revision_idx is greater than the latest available
 	latestRev := ct.pm.LatestRev(ctx, agent.PolicyID)
 	if latestRev != 0 && revisionIDX > latestRev {
-		return 0, opts, nil
+		revisionIDX = 0 // set return val to 0 so the agent gets latest available revision.
 	}
 
 	// Update API keys if the policy has changed, or if the revision differs.


internal/pkg/api/handleCheckin_test.go
```diff
@@ -1118,3 +1140,208 @@ func TestValidateCheckinRequest(t *testing.T) {
...
+func TestProcessPolicyDetails(t *testing.T) {
+	policyID := "policy-id"
+	revIDX2 := int64(2)
+	tests := []struct {
+		name             string
+		agent            *model.Agent
+		req              *CheckinRequest
+		getPolicyMonitor func() *mockPolicyMonitor
+		revIDX           int64
+		returnsOps       bool
```
ycombinator: Typo?

```suggestion
		returnsOpts       bool
```

internal/pkg/api/handleCheckin_test.go
```diff
@@ -1118,3 +1140,208 @@ func TestValidateCheckinRequest(t *testing.T) {
...
+		},
+		getPolicyMonitor: func() *mockPolicyMonitor {
+			pm := &mockPolicyMonitor{}
+			pm.On("LatestRev", mock.Anything, policyID).Return(int64(1)).Once()
+			return pm
+		},
+		revIDX:     0,
+		returnsOps: true,
+		err:        nil,
+	}, {
```
ycombinator, line 1231: I didn't understand this scenario.  I think it's modeling the situation where ES might've been rolled back to an older snapshot so Agent is running policy revision 2 while ES (via the policy monitor) is returning policy revision 1.  In this scenario why would we subscribe to policy updates > policy revision 0?  Shouldn't it be 2, since Agent already knows about policy revision 2?

michel-laterman: Subscribing with rev 2 would be incorrect as we would be referencing a policy that does not exist, if we subscribe to 0 we will be sent the latest revision (1) to make the agent policy consistent with the cluster

ycombinator: Thanks, I understand now why we can't subscribe with rev 2.  But if we subscribe with 0 and Agent is sent 1 while it's running 2, will Agent even apply the new policy revision since it's older?

michel-laterman: AFAIK it should, I don't think the agent compares revisions when applying policies.

ycombinator (approved):

internal/pkg/api/handleCheckin.go
```diff
@@ -1124,3 +1142,55 @@ func calcPollDuration(zlog zerolog.Logger, pollDuration, setupDuration, jitterDu
...
+	// Policy reassign, subscribe to policy with revision 0
+	if policyID != agent.PolicyID {
+		zlog.Debug().Str(dl.FieldAgentPolicyID, policyID).Str("new_policy_id", agent.PolicyID).Msg("Policy ID mismatch detected, reassigning agent.")
+		return 0, opts, nil
+	}
+
+	// Check if the checkin revision_idx is greater than the latest available
+	latestRev := ct.pm.LatestRev(ctx, agent.PolicyID)
+	if latestRev != 0 && revisionIDX > latestRev {
+		return 0, opts, nil
```
blakerouse: Why do you not update the API key here? Isn't possible that an Elastic Agent be at a higher revision because of a snapshot restore that went back to an older version? How do we know that the API keys in this state are not invalid at this point?

blakerouse (changes requested):

PR #5534, opened 2025-09-17
Add mage target and buildkite step to flag crypto imports for FIPS builds

michel-laterman:
Add best-effort detection for external crypto libraries when building a FIPS compatible binary.


.buildkite/pipeline.yml
```diff
@@ -42,6 +42,9 @@ steps:
         agents:
           provider: "gcp"
 
+      - label: "Detect FIPS crypto imports"
```
v1v, line 45: I see `check-ci` step and I wonder whether this should follow the same naming convention or being added to the above step since it runs `mage check:detectFIPSCryptoImports`? Or maybe rename the script with `check_detect_fips_crypto.sh`?

PR #5609, opened 2025-10-02
Fix/migrate watching

michalpristas:
In a context of agent migration.

When agent is migrated and still has some upgrade details or upgrade was triggered manually for some reason we should reflect that in UI. 

This PR makes action optional and updates upgrade details anyways. 

I'd like ACK handling of upgrade to be reworked, but ideally in a separate PR to lower risk as I'd like this one to be in 9.2


prodsecmachine:

:white_check_mark: **security/snyk** check is complete. No issues have been found. [(View Details)](https://app.snyk.io/org/beats/pr-checks/ab21e954-ed44-4899-aaea-470693b72ddc?refs=6331a01fcc6672275828772d1d42160c62f7f36c&source=prChecksComment)

:white_check_mark: **license/snyk** check is complete. No issues have been found. [(View Details)](https://app.snyk.io/org/beats/pr-checks/ab21e954-ed44-4899-aaea-470693b72ddc/license?refs=6331a01fcc6672275828772d1d42160c62f7f36c&source=prChecksComment)

blakerouse (commented):
Sorry I am having trouble following how this PR works.

Does this fix the issue where if you check-in to a new cluster for the first time after migrate that it will reflect the upgrade state of that agent? If so, could you walk me through how this is done in this code.

internal/pkg/api/handleCheckin.go
```diff
@@ -454,22 +455,21 @@ func (ct *CheckinT) processUpgradeDetails(ctx context.Context, agent *model.Agen
 		return nil
 	}
 	// update docs with in progress details
-
 	vSpan, vCtx := apm.StartSpan(ctx, "Check update action", "validate")
-	action, err := ct.verifyActionExists(ctx, vSpan, agent, details)
-	if err != nil {
+
+	var actionTraceparent string
```
michalpristas, line 460: @blakerouse normally when no actin is present we stop here
but we make loading optional and if action is not present we just set traceparent (only property used here in the code) to ""

then later we check if we have here when setting parent span and continue until we reach a point where we update details on line 543

cmacknz:
@michalpristas is this PR still relevant?

michalpristas:
Yes it is

internal/pkg/api/handleAck.go
```diff
@@ -612,22 +612,18 @@ func (ack *AckT) handleUpgrade(ctx context.Context, zlog zerolog.Logger, agent *
 			zlog.Info().Msg("marking agent upgrade as failed, agent logs contain failure message")
 			doc = bulk.UpdateFields{
 				dl.FieldUpgradeStartedAt: nil,
-				dl.FieldUpgradeStatus:    "failed",
```
michel-laterman: Why are these being removed?

internal/pkg/api/handleCheckin.go
```diff
@@ -454,22 +455,21 @@ func (ct *CheckinT) processUpgradeDetails(ctx context.Context, agent *model.Agen
...
+	if action, err := ct.verifyActionExists(ctx, vSpan, agent, details); err == nil && action != nil {
+		actionTraceparent = action.Traceparent
+	} else if !errors.Is(err, es.ErrNotFound) {
```
michel-laterman, line 463: Do we have a case for a non-nil error?

internal/pkg/api/handleCheckin.go
```diff
@@ -565,7 +565,6 @@ func (ct *CheckinT) markUpgradeComplete(ctx context.Context, agent *model.Agent)
 	doc := bulk.UpdateFields{
 		dl.FieldUpgradeDetails:   nil,
 		dl.FieldUpgradeStartedAt: nil,
-		dl.FieldUpgradeStatus:    nil,
```
michel-laterman: Why is this one being removed?

michel-laterman (changes requested):
The PR is doing a few things; the changes to allow for missing actions (in case of migrations) makes sense; however I don't understand why the ack endpoint is being changed, or why the status is not being set to nil if a checkin indicates an upgrade completes.

Also can you please add a changelog fragment

internal/pkg/api/handleAck.go
michalpristas: these are not used anywhere, or at least i could not find any usage nor here nor in fleet

internal/pkg/api/handleCheckin.go
michalpristas: i'm not sure i follow. case for not nil error is ES fetch issue right?

internal/pkg/api/handleCheckin.go
michalpristas: same as above. havent found any place where this is used

internal/pkg/api/handleCheckin.go
michel-laterman: yes, my mistake

internal/pkg/api/handleAck.go
michel-laterman: I would rather this be done in a separate PR

michalpristas:
code cleanup: moved updgrade status removal to a separate PR: #5970 
now this change should be cleaner and more readable

internal/pkg/api/handleAck.go
michalpristas: separate PR created #5970

internal/pkg/api/handleCheckin.go
michalpristas: separate PR created #5970

internal/pkg/api/handleAck_test.go
```diff
@@ -719,67 +719,6 @@ func TestAckHandleUpgrade(t *testing.T) {
 			ESDocument: model.ESDocument{Id: "ab12dcd8-bde0-4045-92dc-c4b27668d735"},
 			Agent:      &model.AgentMetadata{Version: "8.0.0"},
 		},
-	}, {
-		name: "retry signaled",
```
michel-laterman: Do we need to remove these tests in this PR?

internal/pkg/api/handleCheckin.go
```diff
@@ -454,22 +455,21 @@ func (ct *CheckinT) processUpgradeDetails(ctx context.Context, agent *model.Agen
 		return nil
 	}
 	// update docs with in progress details
-
 	vSpan, vCtx := apm.StartSpan(ctx, "Check update action", "validate")
-	action, err := ct.verifyActionExists(ctx, vSpan, agent, details)
-	if err != nil {
+
+	var actionTraceparent string
+	if action, err := ct.verifyActionExists(ctx, vSpan, agent, details); err == nil && action != nil {
```
cmacknz, line 461: Does Kibana tolerate having an action ID it doesn't recognize as well?

I think passing through the upgrade details unmodified is correct, but we need to avoid Kibana now or in the future trying to do this same validation or making assumption about this field.

cmacknz: > I think passing through the upgrade details unmodified is correct

Well a possibility here would be removing the action_id if it doesn't exist so kibana can't depend on it, or putting some marker in the details about whether the action exists or not (though kibana can also just check this).

cmacknz: @juliaElastic any opinions on if sending upgrade_details to Kibana for action_ids that don't exist or are from another cluster is a problem?

juliaElastic: It should be fine to have an `upgrade_details` with a non existent `action_id`, the UI doesn't validate the `action_id`. The only difference I can think of is that these actions won't appear in Agent activity.

michalpristas: actions not in activity is fine imho. this should make reporting upgrades possible, during migration or triggered from CLI

internal/pkg/api/handleAck_test.go
michalpristas: this was removed [here](https://github.com/elastic/fleet-server/pull/5970/files). not sure about diff

michel-laterman (approved):
lgtm

PR #5834, opened 2025-11-04
Fix data race in checkin API and improve overall performance

### LINKED ISSUE #5563: Improve checkin bulker performance
**Describe the enhancement:**

At the moment each long poll connection gets a goroutine. Inside of that goroutine there is a select that performs actions depending on the channel that is triggered. I believe we could remove the `case <-tick.C` channel from this goroutine, by changing how the `ct.bc.CheckIn` works.

At the moment the `case <-tick.C` happens every `ct.cfg.Timeouts.CheckinTimestamp`. Its the same for every connection, meaning every connection gets that timer adding overhead and memory. The call to `bt.bc.CheckIn` also sends an event over a channel and again this happens on every connection based on `ct.cfg.Timeouts.CheckinTimestamp`.

I believe all of this can be simplified and reduce the load of the Fleet Server. Instead of using the ticker and the checkin call. The code can just register the `agentID` as connected to the checkin handler and then based on the `ct.cfg.Timeouts.CheckinTimestamp` all connected `agentID` can be updated in a single request to elasticsearch. When the client disconnects it unregisters the `agentID` and stops updating it in that single request.

This should reduce the number of times by X connections, and the number of messages going over the channels by X connections.

I believe this could reduce the memory and CPU usage of the Fleet Server, especially when it is handling a large number of connections.

### LINKED ISSUE #5793: CheckIn bulker 10 second window is a problem
At the moment the flush interval for the `checkin.NewBulk` is `10 * time.Second`. That means to me that during check-in the Elastic Agent document will not be updated for at most 10 seconds. I believe this could be an issue with ensuring that the `.fleet-agents` document for that Elastic Agent is actually correct for the next check-in.

The following scenario that could occur is (especially when there is multiple Fleet Servers). Lets assume there is Fleet Server 1 and Fleet Server 2.

- Elastic Agent checks-in on Fleet Server 1.
  - It has a status that needs to be written and updated, but it also has actions that it needs to handle.
- Fleet Server 1 sends a response as it has an action.
  - At this point the check-in bulk has not occurred (as its every 10 seconds and it all depends on when Fleet Server started for that interval)
- Elastic Agent checks-in on Fleet Server 2 (round-robin).
  - It has a different status that needs to be written, no actions to handle so it stays connected.
  - Now another check-in bulk needs to occur, but that happens in its own 10 second interval (the interval is not in sync with Fleet Server 1)

This means it is possible that Fleet Server 2 performs the sync before Fleet Server 1, but then Fleet Server 1 performs it sync. This now means that the first check-in overwrites the second check-in, when it shouldn't because actually the second check-in should take preference over the first one.

I believe we might need to ensure that upon check-in that the document is written as soon as possible. This would ensure this doesn't happen. The constant writing of the document to show that the Elastic Agent is connected to the long-poll endpoint can still use the 10 second window, but the initial check-in should not.

blakerouse:

Fix data race in checkin API to prevent multiple Fleet Servers handling the same Elastic Agent from overwriting each other. Ensures that the Elastic Agent document is updated before a response back from check-in, to ensure that the state between Fleet and Elastic Agent is always consistent.

Improves the checkin logic to make it easier to understand and removes each moving parts. Removes the need for an extra ticker in each long poll connection select. Removes the need to compute unique Elastic Agent bodies to send in the MUpdate. This reduces the amount of memory allocations and the amount of data that is sent over the MUpdate.

It solves the issue by ensuring to write the Elastic Agent document before sending a response, which ensures that if the Elastic Agent quickly makes another connection to a different Fleet Server that it is working with the previous state that was written by the previous checkin.

## Design Checklist

- ~~[ ] I have included fail safe mechanisms to limit the load on fleet-server: rate limiting, circuit breakers, caching, load shedding, etc.~~ (no changes in this PR require this)

## Checklist

- ~~[ ] I have made corresponding changes to the documentation~~ (none related in this PR, all internal)
- ~~[ ] I have made corresponding change to the default configuration files~~ (none)

## Related issues

- Closes #5793
- Closes #5563


internal/pkg/api/handleCheckin.go
```diff
@@ -495,84 +493,60 @@ func (ct *CheckinT) processUpgradeDetails(ctx context.Context, agent *model.Agen
...
 	}
-	span, ctx := apm.StartSpan(ctx, "Mark update complete", "update")
-	span.Context.SetLabel("agent_id", agent.Agent.ID)
-	defer span.End()
-	// if the checkin had no details, but agent has details treat like a successful upgrade
-	doc := bulk.UpdateFields{
-		dl.FieldUpgradeDetails:   nil,
-		dl.FieldUpgradeStartedAt: nil,
-		dl.FieldUpgradeStatus:    nil,
-		dl.FieldUpgradedAt:       time.Now().UTC().Format(time.RFC3339),
```
pchila, line 569: I am probably missing something but I don't see the  equivalent of setting `upgradedAt` field in the new version of `processUpgradeDetails`. Is this handled elsewhere ?

internal/pkg/api/handleCheckin_test.go
```diff
@@ -744,7 +743,18 @@ func TestProcessUpgradeDetails(t *testing.T) {
...
+					doc := struct {
+						Doc map[string]interface{} `json:"doc"`
+					}{}
+					if err := json.Unmarshal(p, &doc); err != nil {
+						t.Logf("bulk match unmarshal error: %v", err)
+						return false
+					}
+					_, upgradeDetails := doc.Doc[dl.FieldUpgradeDetails]
+					upgradeAttempts, ok := doc.Doc[dl.FieldUpgradeAttempts]
+					return upgradeDetails && ok && upgradeAttempts == nil && doc.Doc[dl.FieldUpgradedAt] != ""
```
pchila: Why is the body matched with "upgrade details" being present, `nil` upgrade attempts and not empty upgradedAt ?

Looking at the old `body` variable (if the one in line 386 of the old version is the right one) that was being matched, there was a specific state and a specific actionId in the details and no upgradedAt

pchila (commented):
Only went through handleCheckin* and I left a couple of comments. May come back with more once I go over the rest of the PR.

I also think that @michel-laterman could have a look at this since he should have a better context around agent checkin in fleet-server

internal/pkg/server/fleet.go
```diff
@@ -521,7 +521,7 @@ func (f *Fleet) runSubsystems(ctx context.Context, cfg *config.Config, g *errgro
 	ad := action.NewDispatcher(am, cfg.Inputs[0].Server.Limits.ActionLimit.Interval, cfg.Inputs[0].Server.Limits.ActionLimit.Burst, bulker)
 	g.Go(loggedRunFunc(ctx, "Action dispatcher", ad.Run))
 
-	bc := checkin.NewBulk(bulker)
+	bc := checkin.NewBulk(bulker, checkin.WithFlushInterval(cfg.Inputs[0].Server.Timeouts.CheckinTimestamp))
```
michel-laterman, line 524: 👍 good catch!

internal/pkg/checkin/bulk.go
```diff
@@ -179,35 +264,68 @@ func parseOpts(opts ...Opt) optionsT {
...
+		bulkOpts = append(bulkOpts, bulk.WithRefresh())
+	}
+	err = bc.bulker.Update(ctx, dl.FleetAgents, id, body, bulkOpts...)
+	if err != nil {
+		return fmt.Errorf("failed to update document: %w", err)
+	}
+
+	// deleteAudit performs a second request (being that this should not happen very often
+	// it's safer to only use this script when it is needed).
+	if checkin.deleteAudit {
```
michel-laterman: One thing I tried to avoid with the initial implementation was a second ES request to run the script; network calls are slow and reducing calls made/load generally makes fleet more stable.

internal/pkg/api/handleCheckin.go
```diff
@@ -445,23 +439,27 @@ func (ct *CheckinT) verifyActionExists(vCtx context.Context, vSpan *apm.Span, ag
...
-			return err
+		if agent.UpgradeDetails == nil {
+			return checkinOpts, nil
 		}
-		return nil
+		checkinOpts = append(checkinOpts,
+			checkin.WithUpgradeDetails(nil),
+			checkin.WithUpgradeStartedAt(nil),
+			checkin.WithUpgradeStatus(nil),
+		)
```
michel-laterman, line 453: I think as @pchila mentioned, we need to set `dl.FieldUpgradedAt` here if the previous state was watching to mark a successful upgrade (iirc @ycombinator did the implementation and may know more, or at least where the state-diagram for this is)

internal/pkg/api/handleCheckin.go
```diff
@@ -1196,54 +1173,69 @@ func calcPollDuration(zlog zerolog.Logger, pollDuration, setupDuration, jitterDu
...
-		for outputName, output := range agent.Outputs {
-			if output.Type != policy.OutputTypeElasticsearch {
-				continue
-			}
-			if err := updateAPIKey(ctx, zlog, ct.bulker, agent.Id, output.APIKeyID, output.PermissionsHash, output.ToRetireAPIKeyIds, outputName); err != nil {
-				// Only returns ErrUpdatingInactiveAgent
-				return 0, nil, err
+	if effectivePolicyID != agent.PolicyID || effectiveRevIdx != agent.PolicyRevisionIdx {
+		// the agent is either on the wrong policy or the wrong revision of that policy
+		zlog.Info().
```
michel-laterman, line 1221: This should probably be a debug log

internal/pkg/api/handleCheckin_test.go
```diff
@@ -1152,6 +1203,121 @@ func TestValidateCheckinRequest(t *testing.T) {
...
+					},
+				},
+			},
+			expValid: validatedCheckin{
+				// don't update metadata
+				rawMeta: nil,
+			},
+		},
+		{
+			name: "local metadata is empty string and agent has different value",
```
michel-laterman, line 1268: 👍

internal/pkg/checkin/bulk.go
```diff
@@ -179,35 +264,68 @@ func parseOpts(opts ...Opt) optionsT {
...
-	// WARNING: Expects mutex locked.
-	now := time.Now()
-	if now.Unix() != bc.unix {
-		bc.unix = now.Unix()
-		bc.ts = now.UTC().Format(time.RFC3339)
-	}
-
-	return bc.ts
+// Add adds the agent to the connected list.
+func (bc *Bulk) Add(id string) {
```
michel-laterman, line 279: This is a great refactor

internal/pkg/api/handleCheckin.go
blakerouse: Yes that is automatically handled here https://github.com/elastic/fleet-server/pull/5834/files#diff-945b012d5b50c9ece3bce4550dfaa563337e1b905e1b69ed8d28486d39e2b832R178

michel-laterman (commented):
Overall, I really like the refactor.
My biggest concern is the possbile 2nd ES call to remove the audit/unenrolled attributes

internal/pkg/api/handleCheckin_test.go
blakerouse: Because the agent checked-in with UpgradeDetailsStateUPGWATCHING. So it should have the `upgradeDetails` present and it should also have the `upgradeAttempts` present, but should be `nil` (aka. clearing the value in the field).

internal/pkg/checkin/bulk.go
blakerouse: Except this is super rare and this change overall reduces the call count and the body size for checkin. This overall is an improvement. I think using a painless script that has already caused lots of headaches is not worth the risk just to save 1 API call, only when it has audit called on it. If this was 2 requests every time an Elastic Agent checks-in I would agree, but its only in that one case.

I think keeping the code path simpler and easier to understand is better in the long run. The issues with `local_metadata` haunted us for a while and it all related to this script trying to do much.

internal/pkg/api/handleCheckin.go
blakerouse: Can you show me where I removed this in the old code? It is not clear to me that I missed anything, and it was sure not caught by any unit tests (that I could see).

internal/pkg/api/handleCheckin.go
blakerouse: Why debug?

michel-laterman: I'm worried about the log volume.

blakerouse: Do we expect this to happen often? I would think this log present would be nice to see at a default setting.

internal/pkg/checkin/bulk.go
michel-laterman: I'd say it's uncommon, but not super rare (we've had orphaned endpoint misbehaviour reported before). We can try to use `bulker.MUpdate` directly to send both requests at once this way there's only 1 network call and the painless script remains simple

internal/pkg/api/handleCheckin.go
michel-laterman: good point

internal/pkg/api/handleCheckin.go
michel-laterman: It's related to @pchila's comment on line 569 where the `fl.FieldUpgradedAt` was removed (and `UpdatedAt` is injected when the body is marshalled as you have linked)

internal/pkg/api/handleCheckin.go
pchila: Unless I am looking at the wrong line from your link I see the `dl.FieldUpdatedAt` populated, not the `dl.FieldUpgradedAt` which is the same that @michel-laterman pointed out in his [comment](https://github.com/elastic/fleet-server/pull/5834#discussion_r2524560119)

internal/pkg/api/handleCheckin_test.go
pchila: I was referring to the fact that the matching of the body argument for the mock call has changed compared to the old version where:
- `dl.FieldUpgradedAt` was not set
- details.ActionId == 'test-action'
- details.State == WATCHING

My question was: why a different matching is needed for the mocked `bulk.Update` call ?

blakerouse: The match has to be different as its a much different call. I will look at this, I do believe I missed this. Looking to fix it. Glad you caught it.

internal/pkg/checkin/bulk.go
blakerouse: I will update it to use the `MUpdate` so its a single request.

internal/pkg/api/handleCheckin.go
blakerouse: When I did the code change my mind was thinking that `dl.FieldUpdatedAt` and `dl.FieldUpgradedAt` where the same. That was where it went wrong, glad you noticed.

The unit test was also not doing a good check because `!= ""` is always true, because when it is missing its actual `nil`. `nil != ""` is true :-). I have also fixed this.

Take a look again, I have this fixed properly.

internal/pkg/api/handleCheckin_test.go
blakerouse: Fixed, give it another review.

internal/pkg/checkin/bulk.go
blakerouse: It has been updated to use a `MUpdate` and now has a specific unit test coverage to check the call. That was also missing.

internal/pkg/api/handleCheckin.go
blakerouse: Fixed, give it another review.

internal/pkg/checkin/bulk.go
blakerouse: Thanks!

PR #5837, opened 2025-11-05
Replace secrets in `agent.download` section of policies

### LINKED ISSUE #4470: Support new SSL secrets fields
Part of https://github.com/elastic/ingest-dev/issues/3443

**Related Kibana issue** New secrets added to agent policies with the following tickets:
- https://github.com/elastic/kibana/issues/207322
- https://github.com/elastic/kibana/issues/207326
- https://github.com/elastic/kibana/issues/207324
- https://github.com/elastic/kibana/issues/208748

The new fields to support are:

- For fleet server hosts:`secrets.ssl.key` (under fleet server `inputs` section of agent policy) and `fleet.secrets.ssl.key`
- For agent binary source: `agent.download.secrets.ssl.key`

These fields will contain secret references, same as already happens with outputs. Fleet server needs to fetch the secret and insert the values into the mapped fieds before sending the policy to the agent.

**This change MUST be backwards compatible**

This change must be backwards compatible in two cases:

1. Where the stack is updated before fleet-server, fleet-server will not yet understand the secret fields and cannot fail to start or take agents offline unnecessarily until it is also upgrades.
2. Where the stack is updated but fleet-server is not, the same requirement applies. It is valid to continue to use older fleet-server instances with newer stack versions and this change cannot break this.

ycombinator:

The `agent.download` section of a policy is now allowed to contain secrets.  This PR ensures that Fleet Server correctly replaces such secrets with their values before the policy is sent to the Agent.

This PR builds upon the work done in #5852, evaluating the `agent.download` section of the policy received from Elasticsearch and replacing any secrets within it.

## Design Checklist

- [ ] ~I have or intend to scale test my changes, ensuring it will work reliably with 100K+ agents connected.~
- [ ] ~I have included fail safe mechanisms to limit the load on fleet-server: rate limiting, circuit breakers, caching, load shedding, etc.~

## Checklist

- [ ] ~I have made corresponding changes to the documentation~
- [ ] ~I have made corresponding change to the default configuration files~

## Related issues

Closes #4470

### DIFF
--- a/changelog/fragments/1764192755-ssl-secrets.yaml
+++ b/changelog/fragments/1764192755-ssl-secrets.yaml
@@ -1,0 +1,32 @@
+# Kind can be one of:
+# - breaking-change: a change to previously-documented behavior
+# - deprecation: functionality that is being removed in a later release
+# - bug-fix: fixes a problem in a previous version
+# - enhancement: extends functionality but does not break or fix existing behavior
+# - feature: new functionality
+# - known-issue: problems that we are aware of in a given version
+# - security: impacts on the security of a product or a user’s deployment.
+# - upgrade: important information for someone upgrading from a prior version
+# - other: does not fit into any of the other categories
+kind: feature
+
+# Change summary; a 80ish characters long description of the change.
+summary: Support secrets in agent.download section of policy
+
+# Long description; in case the summary is not enough to describe the change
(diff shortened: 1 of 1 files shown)


pchila (approved):
LGTM

Issue #5857, opened 2025-11-06
findAgentByApiKeyId: could not unmarshal ES document into model.Agent: json: cannot unmarshal string into Go struct field Agent.components of type []model.ComponentsItem


cmacknz:
- Relates https://github.com/elastic/fleet-server/issues/5674

When this happens, fleet-server will log an error message like the following:

```
findAgentByApiKeyId: could not unmarshal ES document into model.Agent: json: cannot unmarshal string into Go struct field Agent.components of type []model.ComponentsItem
```

This problem is prevented by the fix in https://github.com/elastic/fleet-server/pull/5824, which corrects the variant of the problem that affects the `local_metadata` field. However, it can also affect the `components` field, and in the case of the `components` field [the mapping](https://github.com/elastic/elasticsearch/blob/bcb1402212e6279819d98cbd9b6f173f840b5236/x-pack/plugin/core/template-resources/src/main/resources/fleet-agents.json#L67-L70
) is not as strict as allows incorrect data to be written to the index. The component field is mapped as object but with `enabled: false` which means Elasticsearch does not parse or validate the field to ensure it matches the object. I am unsure what the historical reason for this is but this is what it does today.

```json
        "components": {
          "type": "object",
          "enabled": false
        },
```

This causes a problem when fleet-server tries to load the .fleet-agents document for an agent during authentication, where it requires the document to be valid JSON but in this case it is not causing unmarshalling and the checkin to fail. 

https://github.com/elastic/fleet-server/blob/c54782d85b71e8565df1217abfc990be989e7f30/internal/pkg/dl/agent.go#L79-L82

In fleet-server, we need to not validate this field when we read it from the index to make it so that this problem is not fatal. The components field will be replaced on each checkin and is not important to the processing of the request.

Until that fix is available, the work around for affected agents is to replace the `components` field with `null`. This seems be rarer than the `local_metadata` variant of the problem as so far we have only detected it in internal serverless alerting and have no seen support cases with this symptom (yet). A query like the one below should correct this for affected agents.

```
POST /.fleet-agents/_update_by_query?conflicts=proceed
{
  "script": {
    "source": "if (ctx._source.components instanceof String) { ctx._source.components = null; }",
    "lang": "painless"
  },
  "query": {
    "exists": {
      "field": "components"
    }
  }
}
```

PR #5858, opened 2025-11-06
Handle malformatted JSON in `.fleet-agents` `components` field

blakerouse:

Handles a possible error in loading the `model.Agent` where the `components` field might be invalid JSON or a string (due to another bug). This changes the code to be more defensive, also should do less on parsing the components model when that is not needed.

Changes the `components` field to be a `json.RawMessage` that does two things. One it stops it from being parsed when the `model.Agent` is loaded from elasticsearch, which happens often and will reduce the time it takes to parse this field. Two, it stops that field from erroring to be unmarshalled when it is an invalid format.

`mage unitTest` does a great job of covering it.

## Checklist

## Related issues

- Closes https://github.com/elastic/fleet-server/issues/5857

### DIFF
--- a/internal/pkg/api/handleCheckin.go
+++ b/internal/pkg/api/handleCheckin.go
@@ -1040,7 +1040,9 @@ func parseMeta(zlog zerolog.Logger, agent *model.Agent, req *CheckinRequest) ([]
 	// Deserialize the agent's metadata copy. If it fails, it's ignored as it will just
 	// be replaced with the correct contents from the clients checkin.
 	var agentLocalMeta interface{}
-	_ = json.Unmarshal(agent.LocalMetadata, &agentLocalMeta)
+	if err := json.Unmarshal(agent.LocalMetadata, &agentLocalMeta); err != nil {
+		zlog.Warn().Err(err).Msg("local_metadata in document invalid; ignoring it")
+	}
 
 	var outMeta []byte
 
@@ -1099,7 +1101,9 @@ func parseComponents(zlog zerolog.Logger, agent *model.Agent, req *CheckinReques
 	// Deserialize the agent's components. If it fails, it's ignored as it will just
 	// be replaced with the correct contents from the clients checkin.
 	var agentComponents []model.ComponentsItems
-	_ = json.Unmarshal(agent.Components, &agentComponents) // error is ignored
+	if err := json.Unmarshal(agent.Components, &agentComponents); err != nil {
+		zlog.Warn().Err(err).Msg("components in document invalid; ignoring it")
+	}
 
 	var outComponents []byte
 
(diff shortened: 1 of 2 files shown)


internal/pkg/api/handleCheckin_test.go
```diff
@@ -1046,6 +1074,14 @@ func TestParseComponents(t *testing.T) {
 	}
 }
 
+func mustMarshalJSON(obj interface{}) json.RawMessage {
+	data, err := json.Marshal(obj)
+	if err != nil {
+		panic(err)
```
cmacknz: nit: this should probably be a `require.NoError(t, err)` instead of a panic to get a nicer error.

internal/pkg/api/handleCheckin.go
```diff
@@ -1037,11 +1037,10 @@ func parseMeta(zlog zerolog.Logger, agent *model.Agent, req *CheckinRequest) ([]
...
 	}
 
-	// Deserialize the agent's metadata copy
+	// Deserialize the agent's metadata copy. If it fails, it's ignored as it will just
+	// be replaced with the correct contents from the clients checkin.
 	var agentLocalMeta interface{}
-	if err := json.Unmarshal(agent.LocalMetadata, &agentLocalMeta); err != nil {
-		return nil, fmt.Errorf("parseMeta local: %w", err)
-	}
+	_ = json.Unmarshal(agent.LocalMetadata, &agentLocalMeta)
```
cmacknz: We should probably still log the error if this happens so we can detect it.

internal/pkg/api/handleCheckin.go
```diff
@@ -1102,13 +1096,18 @@ func parseComponents(zlog zerolog.Logger, agent *model.Agent, req *CheckinReques
 		return nil, &unhealthyReason, nil
 	}
 
+	// Deserialize the agent's components. If it fails, it's ignored as it will just
+	// be replaced with the correct contents from the clients checkin.
+	var agentComponents []model.ComponentsItems
+	_ = json.Unmarshal(agent.Components, &agentComponents) // error is ignored
```
cmacknz: Same here, we should still log the error even if we ignore it

cmacknz: We should probably have metrics counters for these errors but that's more work than I think we want to do right now.

internal/pkg/api/handleCheckin_test.go
blakerouse: Updated to use require.

internal/pkg/api/handleCheckin.go
blakerouse: Add warn level log message.

internal/pkg/api/handleCheckin.go
blakerouse: Add warn level log message. Hopefully this just go away with the fixes in the painless script usage.

cmacknz (approved):

PR #5975, opened 2025-11-27
Support agent available rollback fields

pchila:
This PR handles new `available_rollbacks` field sent by elastic-agent during checkin (implemented in PR https://github.com/elastic/elastic-agent/pull/11143).
This information will be consumed by the Fleet UI to allow a  manual rollback of recently upgraded agents.

This PR introduces new `available_rollbacks` field both in `CheckinRequest` and `pendingT` structs as a slice of `AvailableRollbacks` and a JSON array of objects respectively.
On Elasticsearch the documents in `.fleet-agents` will have a new field `upgrade.rollbacks` similar to:

```json
...
"upgrade": {
            "rollbacks": []
  },
..
```

Refer to PR https://github.com/elastic/elastic-agent/pull/11143 `How to test this PR locally` section.

## Design Checklist

## Checklist

## Related issues

- Relates https://github.com/elastic/elastic-agent/pull/11143

### DIFF
--- a/internal/pkg/api/handleCheckin.go
+++ b/internal/pkg/api/handleCheckin.go
@@ -1150,7 +1150,7 @@ func parseComponents(zlog zerolog.Logger, agent *model.Agent, req *CheckinReques
 // the correct value on the model). See ProcessRequest and checkin.WithAvailableRollbacks for reference.
 func parseAvailableRollbacks(zlog zerolog.Logger, upgradeInfo *model.Upgrade, req *CheckinRequest) ([]byte, error) {
 
-	reqUpgradeInfo := model.Upgrade{}
+	reqUpgradeInfo := model.Upgrade{Rollbacks: []model.AvailableRollback{}}
 	if len(req.Upgrade) > 0 {
 		err := json.Unmarshal(req.Upgrade, &reqUpgradeInfo)
 		if err != nil {
--- a/internal/pkg/api/handleCheckin_test.go
+++ b/internal/pkg/api/handleCheckin_test.go
@@ -1156,7 +1156,7 @@ func TestValidateCheckinRequest(t *testing.T) {
 		{
 			name: "Available rollbacks are correctly parsed",
 			req: &http.Request{
-				Body: io.NopCloser(strings.NewReader(`{"validJson": "test", "status": "test", "message": "test message", "available_rollbacks": [{"version": "1.2.3-SNAPSHOT", "valid_until": "2025-11-27T15:12:44Z"}]}`)),
+				Body: io.NopCloser(strings.NewReader(`{"validJson": "test", "status": "test", "message": "test message", "upgrade":{ "rollbacks": [{"version": "1.2.3-SNAPSHOT", "valid_until": "2025-11-27T15:12:44Z"}]}}`)),
 			},
 			cfg: &config.Server{
 				Limits: config.ServerLimits{
@@ -1173,7 +1173,7 @@ func TestValidateCheckinRequest(t *testing.T) {
 		{
 			name: "Available rollbacks are incorrectly formatted (string instead of array): no error returned but the rawAvailableRollbacks are set to nil",
 			req: &http.Request{
-				Body: io.NopCloser(strings.NewReader(`{"validJson": "test", "status": "test", "message": "test message", "available_rollbacks": "foobar"}`)),
+				Body: io.NopCloser(strings.NewReader(`{"validJson": "test", "status": "test", "message": "test message", "upgrade":{"rollbacks": "foobar"}}`)),
 			},
 			cfg: &config.Server{
 				Limits: config.ServerLimits{


internal/pkg/api/handleCheckin.go
```diff
@@ -167,12 +167,13 @@ func invalidateAPIKeysOfInactiveAgent(ctx context.Context, zlog zerolog.Logger,
...
-	rawComp         []byte
-	seqno           sqn.SeqNo
-	unhealthyReason *[]string
+	req                   *CheckinRequest
+	dur                   time.Duration
+	rawMeta               []byte
+	rawComp               []byte
+	seqno                 sqn.SeqNo
+	unhealthyReason       *[]string
+	rawAvailableRollbacks []byte
```
michalpristas, line 176: not sure this was RFCed but could this be part of upgrade meta?  like `scheduled_at` or `error_msg` so we don't pollute root object with upgrade specific stuff?

michalpristas: i imagine you don't need this info outside of upgrade window anyways

pchila: Not sure if you are referring to the checkin request body, the validatedCheckin struct or the ES document structure.

The main reason that I didn't mix this information with existing upgrade stuff is that there are assumptions about upgrade details and I didn't know if there are more around other other upgrade fields.

The information about available rollbacks must survive beyond the duration of a single upgrade (and fleet detects the conclusion of an upgrade by the `upgrade_details` field disappearing in the checkin body) and it's also not a stable intrinsic characteristic of agent so that it can go in something like `local_metadata`.

The RFC will be updated with the final schema we are gonna have in this PR.

internal/pkg/api/handleCheckin.go
```diff
@@ -1131,6 +1139,39 @@ func parseComponents(zlog zerolog.Logger, agent *model.Agent, req *CheckinReques
...
 
+func parseAvailableRollbacks(zlog zerolog.Logger, agent *model.Agent, req *CheckinRequest) ([]byte, error) {
+	var reqRollbacks []model.AvailableRollback
+	if req.AvailableRollbacks != nil {
+		reqRollbacks = make([]model.AvailableRollback, len(*req.AvailableRollbacks))
+		for i, rr := range *req.AvailableRollbacks {
+			reqRollbacks[i] = model.AvailableRollback{
+				ValidUntil: rr.ValidUntil.UTC().Format(time.RFC3339),
+				Version:    rr.Version,
+			}
```
michel-laterman: It looks like this is the only place where fleet-server uses `model.AvailableRollback`, it deserialize `[]byte` to `[]model.AvailableRollback` then does a compare to detect if there is a change.

can we compare the []byte arrays directly below (using `bytes.Equal()`) instead of deserializing and using `reflect.DeepEqual`?

pchila: IIUC you are proposing we use []byte as well in `agent.AvailableRollbacks` ?
Here we are mapping from `api.AvailableRollbacks` to `model.AvailableRollback` and then comparing it to what we have in `model.Agent`. 

Are you proposing to keep the available rollbacks in `model.Agent` as a []byte and compare changes using []byte on both the checkin request and the model?
Wouldn't that force the JSON format of the request and the model to always be the same ?

internal/pkg/api/handleCheckin.go
blakerouse: I also worry about placing this in upgrade_details. At the moment that is used to describe the current upgrade. Being that manual rollback will be around for like 14 days after upgrade, I worry that mixing these fields into the `upgrade_details` would be correct.

I understand its a detail about upgrades, but it is also an overall state of the agents ability to rollback.

internal/pkg/api/handleCheckin.go
blakerouse: @michel-laterman Is saying to add a check like this https://github.com/elastic/fleet-server/pull/5975/files#diff-e0c02bac8d151e9941eedd5ef643441665ee3d2f78baf42c121edd45dee08dedL1085

If the bytes are the same you can exit early, do less work as its already the same.

internal/pkg/api/handleCheckin.go
cmacknz: I can buy that available rollbacks is conceptually separate from reporting the state of an in progress upgrade, which is how upgrade_details is used today. We could in principle remove upgrade_details and report upgrade status a different way and still want to keep available_rollbacks separately.

Perhaps a better organization would have been to have a root upgrade object with a details/state and rollback sub-objects but the ship has sailed on that already.

internal/pkg/api/handleCheckin.go
```diff
@@ -251,13 +252,19 @@ func (ct *CheckinT) validateRequest(zlog zerolog.Logger, w http.ResponseWriter,
 		return val, err
 	}
 
+	rawRollbacks, err := parseAvailableRollbacks(zlog, agent, &req)
+	if err != nil {
+		return val, err
```
cmacknz: This makes an invalid available_rollbacks field fail the whole checkin with a 4xx doesn't it?

I don't think having an invalid available_rollback field should prevent agent from reporting status or more importantly from getting an upgrade action to fix the bug causing corrupted available_rollbacks or a policy change to disable the rollback feature remotely. We ran into this with the download rate field, see https://github.com/elastic/fleet-server/issues/5164 for part of the series of issues that caused.

Can we strip any invalid available_rollbacks and proceed as normal, logging the error in fleet-server instead?

cmacknz: It looks like the only error right now is if the parsed rollbacks can't be marshalled to JSON in which case we should behave as if available rollbacks weren't reported at all.

```go
		marshalled, err := json.Marshal(reqRollbacks)
		if err != nil {
			return nil, fmt.Errorf("marshalling available rollbacks: %w", err)
		}
```

Having some indication on agent would be useful, maybe returning an HTTP 202 could be one way to indicate something is up, but we'd need to make sure agents tolerate that properly. We don't need to do that though, I am happy as long as formatting issues don't prevent agents from using central management.

internal/pkg/api/handleCheckin.go
pchila: @cmacknz I can create a new root `upgrade` object for the agent document and have a `rollbacks` key under that if it helps making things clearer/cleaner. I just couldn't find a place where the rollback information fits in the existing structure but we can always create  such key to be used for new upgrade information

internal/pkg/api/handleCheckin.go
pchila: I agree that an error parsing available rollbacks should not prevent the checkin operation from succeeding. I will change the handling by logging a warning in case something goes wrong and proceed with the checkin code.

internal/pkg/api/handleCheckin.go
pchila: If I am looking at the correct line in the `parseComponent` function
```golang
    // Quick comparison first; compare the JSON payloads.
	// If the data is not consistently normalized, this short-circuit will not work.
	if bytes.Equal(req.Components, agent.Components) {
		zlog.Trace().Msg("quick comparing agent components data is equal")
		return nil, &unhealthyReason, nil
	}
```
it's possible to do a quick check on the components because those are represented as a []byte in both the checkin body and `model.Agent`. AvailableRollbacks are [defined as slice of structs](https://github.com/elastic/fleet-server/pull/5975/files#diff-fd546cd1ca0214259034f220b902f63b33fbc4666359e0f8e06d861962917e30R309-R315) based on the [openapi definition](https://github.com/elastic/fleet-server/pull/5975/files#diff-5ad174233c27653aa08c5ce213ddfc13cbf7bcfc10cdf724a547da655e5743c0R374-R390) so to do a byte comparison I would have to re-marshal them or define them as  a []byte and then unmarshal them/use them as-is for the elasticsearch model (which would somewhat couple what the agent is sending with what is stored on elasticsearch)

Edit:
With https://github.com/elastic/fleet-server/pull/5975/commits/01f125e3a02cbce7649ad0d01973e04279e16368 I have changed the request model to use `json.RawMessage` in the request but the model on ES still uses a slice of structs to represent those: we still don't have a cheap byte comparison there unless we lose the definition of AvailableRollbacks also on the ES schema.

internal/pkg/api/handleCheckin.go
pchila: Redefined the available rollbacks  as `json.RawMessage` in the openapi model and parse it lazily in the `parseAvailableRollbacks` function, otherwise fleet-server would always respond 400 while it's eagerly unmarshalling the request.
You can find this change in https://github.com/elastic/fleet-server/pull/5975/commits/01f125e3a02cbce7649ad0d01973e04279e16368

I am reusing the model.Available rollback since at this point request and model on ES have the same structure but I am reluctant about doing a complete passthrough from `json.RawMessage` to `json.RawMessage` (as it's currently done for components for example) since that would tightly couple what the agent produces as payload to what is supported/read/used by Fleet UI from elasticsearch.

internal/pkg/api/handleCheckin.go
cmacknz: If that is not too much work, then my vote is to have the available rollbacks under a new `upgrade` object, so that if we later add a 3rd piece of upgrade state we can group them together instead of adding another field at the root of the document.

`upgrade_details` will still be a inconsistent with this and that is fine.

cmacknz: If introducing a parent `upgrade` object risks having this for 9.3 I'd rather have rollbacks in 9.3

pchila: Implemented and tested in c28c35da9edf501289310aaf9a1890c56475f651

internal/pkg/checkin/bulk.go
```diff
@@ -358,6 +368,20 @@ func toUpdateBody(now string, pending pendingT) ([]byte, error) {
...
+		if pending.extra.availableRollbacks != nil {
+			if _, ok := fields[dl.FieldUpgradeInfo]; !ok {
+				fields[dl.FieldUpgradeInfo] = bulk.UpdateFields{}
+			}
+
+			if upgradeInfo, ok := fields[dl.FieldUpgradeInfo].(bulk.UpdateFields); ok {
+				upgradeInfo[dl.FieldAvailableRollbacks] = json.RawMessage(pending.extra.availableRollbacks)
+			} else {
+				// return an error as we have an unexpected value for the upgrade info container
+				return nil, fmt.Errorf("unexpected value %T for key %s in agent document", fields[dl.FieldUpgradeInfo], dl.FieldUpgradeInfo)
```
cmacknz: IMO it is better to correct the problem here, and just overwrite whatever invalid value was unexpectedly there.

I think this error is in a place that would cause the .fleet-agents doc to stop updating for an agent, though I'm not sure what circumstances would lead us here besides a coding error.

pchila: Done in f026157f4377ba64cbe67c70f7e02ce2b1753b6c
Went through another run of testing and everything looks good

cmacknz (dismissed):

internal/pkg/api/handleCheckin.go
pchila: Moved the rollbacks under `upgrade.rollbacks` key in checkin body with commit f9a6fa452ca977a8cba3f33fbdb32eaf69a4ef61 and agent PR https://github.com/elastic/elastic-agent/pull/11753
tested e2e and it works correctly

PR #6048, opened 2025-12-10
file upload: do not refresh on every chunk

pzl:

Improves elasticsearch performance when uploading large files

forcing `refresh` is somewhat costly, and doing that on _every_ chunk upload was unnecessary. For large files and many chunks, this was able to overwhelm elasticsearch. The chunk records only need to be searchable and found at the very end, so a single refresh call is used at the finalization step instead of every chunk.

Note that the final refresh call is still necessary. It was observed that the finalization step was not always able to locate the most-recently-uploaded chunks immediately prior to the finish api call, and validation would fail. 

One could either do the "get agent diagnostics" zip function, or otherwise use elastic defend, go to responder and do a "get-file" command to execute this path.

## Design Checklist

- [X] I have or intend to scale test my changes, ensuring it will work reliably with 100K+ agents connected.
- [X] I have included fail safe mechanisms to limit the load on fleet-server: rate limiting, circuit breakers, caching, load shedding, etc.

## Checklist

### DIFF
--- a/NOTICE-fips.txt
+++ b/NOTICE-fips.txt
@@ -4919,11 +4919,11 @@ Contents of probable licence file $GOMODCACHE/google.golang.org/grpc@v1.77.0/LIC
 
 --------------------------------------------------------------------------------
 Dependency : google.golang.org/protobuf
-Version: v1.36.10
+Version: v1.36.11
 Licence type (autodetected): BSD-3-Clause
 --------------------------------------------------------------------------------
 
-Contents of probable licence file $GOMODCACHE/google.golang.org/protobuf@v1.36.10/LICENSE:
+Contents of probable licence file $GOMODCACHE/google.golang.org/protobuf@v1.36.11/LICENSE:
 
 Copyright (c) 2018 The Go Authors. All rights reserved.
 
--- a/NOTICE.txt
+++ b/NOTICE.txt
@@ -4919,11 +4919,11 @@ Contents of probable licence file $GOMODCACHE/google.golang.org/grpc@v1.77.0/LIC
 
 --------------------------------------------------------------------------------
 Dependency : google.golang.org/protobuf
-Version: v1.36.10
+Version: v1.36.11
 Licence type (autodetected): BSD-3-Clause
 --------------------------------------------------------------------------------
 
-Contents of probable licence file $GOMODCACHE/google.golang.org/protobuf@v1.36.10/LICENSE:
+Contents of probable licence file $GOMODCACHE/google.golang.org/protobuf@v1.36.11/LICENSE:
 
 Copyright (c) 2018 The Go Authors. All rights reserved.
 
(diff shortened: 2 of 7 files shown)


internal/pkg/file/uploader/finalize.go
```diff
@@ -42,6 +42,12 @@ func (u *Uploader) Complete(ctx context.Context, id string, transitHash string)
 		return info, ErrStatusNoUploads
 	}
 
+	// complete may be called before most recent chunks are available for search yet
+	// this explicitly calls refresh once at the end, instead of refreshing on each chunk
+	if err := EnsureChunksIndexed(ctx, u.bulker.Client(), info.Source); err != nil {
+		return info, err
```
ycombinator: Would you mind wrapping this `err` using `fmt.Errorf` so there's a bit more context and we don't just get the lower-level error from the ES client passed through?

internal/pkg/file/uploader/es.go
```diff
@@ -186,3 +185,11 @@ func DeleteAllChunksForFile(ctx context.Context, bulker bulk.Bulk, source string
 	_, err = client.DeleteByQuery([]string{fmt.Sprintf(UploadDataIndexPattern, source)}, bytes.NewReader(q), client.DeleteByQuery.WithContext(ctx))
 	return err
 }
+
+func EnsureChunksIndexed(ctx context.Context, client *elasticsearch.Client, source string) error {
+	req := esapi.IndicesRefreshRequest{
+		Index: []string{fmt.Sprintf(UploadDataIndexPattern, source)},
+	}
+	_, err := req.Do(ctx, client)
```
michel-laterman: Should we also check for non 2xx status codes?

pzl: one could, but this is not _that_ crucial of a call. 

The flow here, is if a client (endpoint or agent) is uploading a file, sends a chunk-upload call, and subsequently, _immediately_ calls the finalize API call. If its quick enough, the finalize step will happen before that last chunk is available for search. Even a `time.Sleep` here instead of the refresh call would functionally work, since the chunk index will refresh itself soon enough.  Whether this call works or not is just.. meh.  If the subsequent code after this that retrieves the chunks succeeds, finds all the chunks, and they are valid, this call didn't need to succeed. It is preventing a just-in-case race condition (that I could trigger under certain conditions). But the final chunk(s) may have _already_ been refreshed by the native periodic (~1s) elasticsearch check. So weird statuses here is not _necessarily_ an actual failure.

cmacknz: I think we want to log non-2xx status codes at least, index refreshes are not totally free so we would want to see if this correlates to increases in 429s for example

pzl:
got it: this is more of a "here be dragons, don't break it unless you know the ramifications," situation.  Accepting some brittle tests to draw even more attention that changes to this need to be very intentional. Guarding against inadvertent changes that upend the learnings and testing here

I can add some comments and tests to that effect, sure

PR #6159, opened 2026-01-08
Enforce size limit on `POST /api/fleet/uploads`

ycombinator:

This PR prevents request bodies of arbitrarily large size to be sent to the `POST /api/fleet/uploads` API. 

This PR checks the size of the request body sent to the `POST /api/fleet/uploads` API and, if it exceeds the configured limit, rejects the request, responding with an HTTP 413 Request Entity Too Large status code.  By default, the limit is configured to 5MiB but can be overridden in the Fleet Server input configuration via the `server.limits.upload_start_limit.max_body_byte_size` setting.

## Design Checklist

## Checklist

### DIFF
--- a/changelog/fragments/1767789965-upload-limit.yaml
+++ b/changelog/fragments/1767789965-upload-limit.yaml
@@ -11,7 +11,7 @@
 kind: security
 
 # Change summary; a 80ish characters long description of the change.
-summary: Enforce UploadBegin request body size
+summary: Enforce maximum limits on UploadBegin and UploadComplete request body sizes
 
 # Long description; in case the summary is not enough to describe the change
 # this field accommodate a description without length limits.


changelog/fragments/1767789965-upload-limit.yaml
```diff
@@ -0,0 +1,32 @@
...
+# - enhancement: extends functionality but does not break or fix existing behavior
+# - feature: new functionality
+# - known-issue: problems that we are aware of in a given version
+# - security: impacts on the security of a product or a user’s deployment.
+# - upgrade: important information for someone upgrading from a prior version
+# - other: does not fit into any of the other categories
+kind: security
+
+# Change summary; a 80ish characters long description of the change.
+summary: Enforce payload size limit, when configured, for file uploads
```
pzl: "which limit" confusion here

internal/pkg/api/handleUpload_test.go
```diff
@@ -107,6 +108,10 @@ func TestUploadBeginValidation(t *testing.T) {
 				"src": "agent"
 			}`,
 		},
+		{
+			"file that is too large is rejected", http.StatusRequestEntityTooLarge, "payload is too large",
```
pzl: "which limit" confusion in test description

internal/pkg/api/handleUpload_test.go
```diff
@@ -1234,3 +1242,19 @@ func size_ptr(x int) *uint64 {
 	y := uint64(x) //nolint:gosec // disable G115
 	return &y
 }
+
+func generateLargeFileInput(paddingSize int) string {
```
pzl: helper fn name may also be confusing -- not a large _file_ per se,  but that is something tested elsewhere

pzl (commented):
looks good

pointed out a few places where names or messages were unclear if it was describing large _files_ (.i.e. a file itself being like 3GB with config set lower, and either rejecting file body bytes above that; or when reading this json file-description payload, looking at `file.size` property and rejecting for _that_ limit). Whereas this is the limit placed on the total JSON payload body of file metadata / header.


The diff in isolation is clear which is being addressed, but down the road I can see this being an easy mixup, which thing is "too large": the _file contents itself_ or the _start / header_ payload.

changelog/fragments/1767789965-upload-limit.yaml
michel-laterman:
```suggestion
summary: Enforce UploadBegin request body size
```

internal/pkg/api/handleUpload_test.go
michel-laterman: @pzl is correct, this is the upload start request

ycombinator: Addressed in 721b7df605c114bc90fb323e069f33d9cfc4ef9d

internal/pkg/api/handleUpload_test.go
ycombinator: Renamed in 721b7df605c114bc90fb323e069f33d9cfc4ef9d

internal/pkg/api/handleUpload.go
```diff
@@ -71,6 +73,11 @@ func (ut *UploadT) validateUploadBeginRequest(ctx context.Context, reader io.Rea
 		if errors.Is(err, io.EOF) {
 			return nil, "", fmt.Errorf("%w: %w", ErrFileInfoBodyRequired, err)
 		}
+
+		var maxBytesErr *http.MaxBytesError
+		if errors.As(err, &maxBytesErr) {
+			return nil, "", fmt.Errorf("payload is too large, %w", err)
```
michel-laterman:
```suggestion
			return nil, "", fmt.Errorf("payload is too large: %w", err)
```
(nitpick) we normally use a colon when wrapping errors

internal/pkg/api/handleUpload.go
```diff
@@ -83,9 +90,20 @@ func (ut *UploadT) validateUploadBeginRequest(ctx context.Context, reader io.Rea
...
+	}
+
 	// decode early to match agentID in the payload
 	payload, agentID, err := ut.validateUploadBeginRequest(r.Context(), r.Body)
 	if err != nil {
+		var maxBytesErr *http.MaxBytesError
+		if errors.As(err, &maxBytesErr) {
+			w.WriteHeader(http.StatusRequestEntityTooLarge)
+			return err
+		}
```
michel-laterman: We use `NewHTTPErrResp` in `internal/pkg/api/error.go` to translate an error into a specific response; we don't call WriteHeader directly in this level.

michel-laterman (changes requested):
Do we also want to use the config vars for upload finailize and chunks?

internal/pkg/api/handleUpload.go
ycombinator: Implemented in b71c1983b0c0a0ed482909b0490afeab13178e5f.

PR #6270, opened 2026-02-03
[OpAMP] Introduce `/v1/opamp` endpoint behind feature flag

ycombinator:

This PR allows OpAMP clients, e.g. OpenTelemetry Collectors, to connect to Fleet Server over an HTTP OpAMP endpoint.

By creating a new `/v1/opamp` HTTP API endpoint in Fleet Server.  By default, this endpoint is disabled and requests to it will receive a HTTP 501 Not Implemented response.  It can be enabled by setting the `feature_flags.enable_opamp` Fleet Server configuration setting to `true`.

Note: an E2E test to verify OpAMP connectivity from an OTel Collector to Fleet Server is being added in a follow up PR: #6289.

1. Follow the instructions in https://github.com/ycombinator/fleet-server/blob/opamp/docs/developers-guide.md#opamp.  

2. Verify that the OTel Collector is **NOT** able to connect to the `/v1/opamp` endpoint in Fleet Server due to a HTTP 501 Not Implemented error.  This is the expected behavior because the `enable_opamp` feature flag has not yet been turned on. You should see errors in the OTel Collector logs like so:
   ```
   2026-02-03T14:52:19.665-0800	error	opampextension@v0.142.0/logger.go:26	invalid response from server: 501	{"resource": {"service.instance.id": "9287f390-91b2-45f0-a122-604b51f4ee4c", "service.name": "otelcol-contrib", "service.version": "0.142.0"}, "otelcol.component.id": "opamp", "otelcol.component.kind": "extension", "client": "http"}
   ```
   And you should see corresponding errors in the Fleet Server logs like so:
   ```
   22:52:19.666 WRN HTTP request error error.message="OpAMP endpoint is disabled" ecs.version=1.6.0 error.type=*errors.errorString event.duration=69000 http.request.id=bc4495f3-5276-4e71-a5be-2d9970dfe14a http.response.status_code=501 server.address=[::1]:8220 service.name=fleet-server service.type=fleet-server
   ```
3. Stop the Fleet Server binary.
4. Edit the `fleet-server.dev.yml` file and turn on the `enable_opamp` feature flag.
   ```
         feature_flags:
        enable_opamp: true
   ```
5. Restart the Fleet Server binary.
   ```
   ./build/binaries/fleet-server-9.4.0-darwin-aarch64/fleet-server -c fleet-server.dev.yml
   ```
6. Ensure that the HTTP 501 errors are gone from the OTel Collector and Fleet Server logs.  Instead, verify that the Fleet Server logs periodically show messages like so:
   ```
   23:09:59.834 INF HTTP Request client.address=[::1]:59029 ecs.version=1.6.0 event.duration=167311000 fleet.apikey.id=Qv9GkpsBUj9LN71ztptR http.request.body.bytes=24 http.request.id=52e9465b-8a74-4206-b532-3eda6f566f8b http.request.method=POST http.response.body.bytes=0 http.response.status_code=200 http.version=1.1 server.address=[::1]:8220 service.name=fleet-server service.type=fleet-server tls.established=false url.full=/v1/opamp
   ```
7. Check the Fleet UI in Kibana and verify that an Agent has "enrolled" and is updating every few seconds.  Verify that the Agent's version matches the OTel Collector's version and the hostname matches the hostname where the OTel Collector is running.
   Screenshot 2026-02-03 at 15 11 43

## Design Checklist

## Checklist

## Related issues

- Resolves https://github.com/elastic/fleet-server/issues/6329

### DIFF
--- a/model/oapi-cfg.yml
+++ b/model/oapi-cfg.yml
@@ -5,6 +5,3 @@ generate:
   chi-server: true
 #  strict-server: true # We need to figure out where/how to handle auth if we want to use strict generation
 output: ./internal/pkg/api/openapi.gen.go
-output-options:
-  exclude-operation-ids:
-    - opamp
--- a/model/openapi.yml
+++ b/model/openapi.yml
@@ -2054,14 +2054,3 @@ paths:
           $ref: "#/components/responses/internalServerError"
         "503":
           $ref: "#/components/responses/unavailable"
-#  /v1/opamp:
-#    post:
-#      operationId: opamp
-#      summary: Receive OpAMP AgentToServer messages
-#      description: |
-#        Receive OpAMP AgentToServer messages and respond with ServerToAgent message,
-#        as described in https://opentelemetry.io/docs/specs/opamp/
-#      security:
-#        - agentApiKey: []
-#      requestBody:
-#        required: true


model/openapi.yml
```diff
@@ -2054,3 +2054,11 @@ paths:
           $ref: "#/components/responses/internalServerError"
         "503":
           $ref: "#/components/responses/unavailable"
+  /v1/opamp:
+    post:
+      operationId: opAMP
+      summary: Receive OpAMP AgentToServer messages
```
michel-laterman: We should at least link to the OpAMP spec here

docs/developers-guide.md
```diff
@@ -420,3 +420,59 @@ For more advanced scenario you can build a custom docker image that you could us
 make -C dev-tools/cloud build-and-push-cloud-image
 ```
 
+## OpAMP
```
michel-laterman: Should we make a separate .md for opamp in `docs/`? We'll need somewhere to document some implementation details as well; such as the fact that currently we only support the HTTP transport.
Additionally we can clearly label that this is still all under development in that file

internal/pkg/api/api.go
```diff
@@ -230,3 +237,13 @@ func (a *apiServer) Status(w http.ResponseWriter, r *http.Request, params Status
 		ErrorResp(w, r, err)
 	}
 }
+func (a *apiServer) OpAMP(w http.ResponseWriter, r *http.Request) {
+	zlog := hlog.FromRequest(r).With().
+		Str("mod", kOpAMPMod).
+		Logger()
+	w.Header().Set("Content-Type", "application/x-protobuf")
```
michel-laterman: We need to force this endpoint to use protobufs for the error responses

internal/pkg/api/handleOpAMP.go
```diff
@@ -0,0 +1,280 @@
...
+	semconv "go.opentelemetry.io/otel/semconv/v1.37.0"
+	"google.golang.org/protobuf/proto"
+)
+
+const (
+	kOpAMPMod = "opAMP"
+)
+
+var (
+	ErrOpAMPDisabled = errors.New("OpAMP endpoint is disabled")
```
michel-laterman:
```suggestion
	ErrOpAMPDisabled = errors.New("opamp endpoint is disabled")
```
best practice: errors do not start with capital letters

internal/pkg/api/handleOpAMP.go
```diff
@@ -0,0 +1,280 @@
...
+	defer ticker.Stop()
+
+	for {
+		select {
+		case <-ctx.Done():
+			zerolog.Ctx(ctx).Debug().Msg("stopping opAMP timers")
+			return
+		case <-ticker.C:
+			zerolog.Ctx(ctx).Debug().Msg("opAMP timer tick; setting flags")
+			oa.flags = uint64(protobufs.ServerToAgentFlags_ServerToAgentFlags_ReportAvailableComponents)
```
michel-laterman: Why are we doing this?

internal/pkg/api/handleOpAMP.go
```diff
@@ -0,0 +1,280 @@
...
+		return &BadRequestErr{msg: "failed to unmarshal AgentToServer message"}
+	}
+
+	instanceUID, err := uuid.FromBytes(aToS.InstanceUid)
+	if err != nil {
+		return &BadRequestErr{msg: "failed to parse instance_uid from AgentToServer message"}
+	}
+	zlog.Debug().
+		Str("instance_uid", instanceUID.String()).
+		Str("aToS", aToS.String()).
```
michel-laterman: We should remove this attribute once we're ready to merge

internal/pkg/api/handleOpAMP.go
```diff
@@ -0,0 +1,280 @@
...
+		Msg("received AgentToServer message from agent")
+
+	// Check if Agent is "enrolled"; if it is, update it; otherwise, enroll it.
+	agent, err := oa.findEnrolledAgent(zlog, instanceUID.String())
```
michel-laterman: 👍  - we should document that this is the workflow. Also how do we behave if there is an instanceUID collision?

internal/pkg/api/handleOpAMP.go
```diff
@@ -0,0 +1,280 @@
...
+		}
+	}
+
+	if err := oa.updateAgent(zlog, agent, &aToS); err != nil {
+		return fmt.Errorf("failed to update persisted Agent information: %w", err)
+	}
+
+	sToA := protobufs.ServerToAgent{
+		Flags: oa.flags,
+	}
```
michel-laterman: instance_uid must be set: https://github.com/open-telemetry/opamp-spec/blob/main/specification.md#servertoagentinstance_uid

internal/pkg/api/handleOpAMP.go
```diff
@@ -0,0 +1,280 @@
...
+
+	// Reset flags; timer will set them again
+	zlog.Debug().Msg("resetting flags")
+	oa.flags = 0
+
+	return err
+}
+
+func (oa *OpAMPT) findEnrolledAgent(_ zerolog.Logger, agentID string) (*model.Agent, error) {
+	ctx := context.TODO()
```
michel-laterman: We can grab the context associated with the request and pass it here

internal/pkg/api/handleOpAMP.go
```diff
@@ -0,0 +1,280 @@
...
+	if !oa.cfg.Features.EnableOpAMP {
+		zlog.Debug().Msg("opAMP endpoint is disabled")
+		return ErrOpAMPDisabled
+	}
+
+	apiKey, err := authAPIKey(r, oa.bulk, oa.cache)
+	if err != nil {
+		zlog.Debug().Err(err).Msg("unauthenticated opamp request")
+		return err
+	}
```
michel-laterman: One issue here is that any error that isn't `ErrOpAMPDisabled` should result in us responding with a protobuf ServerToAgent message with the error_response set: ErrOpAMPDisabled; we don't handle this correctly

internal/pkg/api/handleOpAMP.go
```diff
@@ -0,0 +1,280 @@
...
+			zerolog.Ctx(ctx).Debug().Msg("stopping opAMP timers")
+			return
+		case <-ticker.C:
+			zerolog.Ctx(ctx).Debug().Msg("opAMP timer tick; setting flags")
+			oa.flags = uint64(protobufs.ServerToAgentFlags_ServerToAgentFlags_ReportAvailableComponents)
+		}
+	}
+}
+
+func (oa *OpAMPT) handleOpAMP(zlog zerolog.Logger, r *http.Request, w http.ResponseWriter) error {
```
michel-laterman: I don't think that copying an endpoint handler for a fleet-server endpoint is the correct approach here. We should use the [opamp-go Server](https://pkg.go.dev/github.com/open-telemetry/opamp-go@v0.22.0/server#OpAMPServer) and have the implementation logic in the callbacks. The library handles boilerplate for us, additionally it would make supporting websockets trivial once we can support it.

ycombinator: Thanks for pointing me to https://pkg.go.dev/github.com/open-telemetry/opamp-go@v0.22.0/server#OpAMPServer.  Moving this PR into draft as I update it to use that package.

model/openapi.yml
```diff
@@ -2054,3 +2054,11 @@ paths:
           $ref: "#/components/responses/internalServerError"
         "503":
           $ref: "#/components/responses/unavailable"
+#  /v1/opamp:
+#    post:
+#      x-oapi-codegen-skip: true # Only document the API; don't generate code for it.
```
ycombinator: Haven't been able to get this `x-oapi-codegen-skip` directive working.  Code is still being generated with it.

model/oapi-cfg.yml
```diff
@@ -5,3 +5,6 @@ generate:
   chi-server: true
 #  strict-server: true # We need to figure out where/how to handle auth if we want to use strict generation
 output: ./internal/pkg/api/openapi.gen.go
+output-options:
+  exclude-operation-ids:
+    - opamp
```
ycombinator: FIXME: Haven't been able to get this working to prevent generation of code from the spec for the `/v1/opamp` endpoint.

model/openapi.yml
michel-laterman: I don't see this directive documented anywhere

ycombinator: Sorry, wrong directive, suggested to me by Gemini 😅.

model/openapi.yml
ycombinator: Done in c4e6e5cd.

docs/developers-guide.md
ycombinator: Done in d8bc76b8.

internal/pkg/api/handleOpAMP.go
ycombinator: Leftover from testing, removed in e4c35c6b.

internal/pkg/api/handleOpAMP.go
ycombinator: I think we have to treat an instanceUID collision as the same situation as receiving an update from the same Agent that has "enrolled" before.  I'm not sure how we can distinguish between the two situations: an Agent that's been "enrolled" before sending a new message vs. a different Agent that somehow has the same instanceUID. WDYT?

internal/pkg/api/handleOpAMP.go
ycombinator: Thanks. Fixed in 7bdba7cd.

internal/pkg/api/handleOpAMP.go
ycombinator: Fixed in e3fadb24.

internal/pkg/api/handleOpAMP.go
ycombinator: Take a look at the new implementation in `handleMessage` and let me know what you think of the error handling.

internal/pkg/api/api.go
ycombinator: Do you think this is still needed, given that we're now using `opamp-go` Server?  I deliberated changed one of the error cases in `handleMessage` so a `ServerToAgent` message with `ErrorResponse` set would be returned and it seemed to work correctly.  I saw the following in the OTel Collector logs:

```
2026-02-06T11:32:49.162-0800	error	opampextension@v0.142.0/opamp_agent.go:144	OpAMP server returned an error response	{"resource": {"service.instance.id": "3d279333-3ece-41a2-8b75-f06009db2f20", "service.name": "otelcol-contrib", "service.version": "0.142.0"}, "otelcol.component.id": "opamp", "otelcol.component.kind": "extension", "message": "failed to update persisted Agent information: %!w(<nil>)"}
```

(Ignore the `%!w(<nil>)`; that's because there wasn't actually an error, I just inverted the error check to generate this response message).

internal/pkg/api/handleOpAMP.go
ycombinator: Ended up not needing this error after all.  Now we just don't register the `/v1/opamp` route if the feature flag is disabled.

internal/pkg/api/handleOpAMP.go
```diff
@@ -0,0 +1,310 @@
...
+
+func (oa *OpAMPT) Init() error {
+	settings := oaServer.Settings{
+		Callbacks: types.Callbacks{
+			OnConnecting: func(request *http.Request) types.ConnectionResponse {
+				zlog := hlog.FromRequest(request).With().
+					Str("mod", kOpAMPMod).
+					Logger()
+
+				apiKey, err := authAPIKey(request, oa.bulk, oa.cache)
```
blakerouse, line 74: This is interesting, so its using the enrollment token as its authentication always? Its not a unique authentication per agent. That is very different then what we have now, its 1 api-key per agent. The enrollment token is used to get a specific api-key and then that key is used by the enrolled agent.

internal/pkg/api/handleOpAMP.go
```diff
@@ -0,0 +1,310 @@
...
+			}
+		}
+
+		zlog.Debug().
+			Str("instance_uid", instanceUID.String()).
+			Str("aToS", message.String()).
+			Msg("received AgentToServer message from agent")
+
+		// Check if Agent is "enrolled"; if it is, update it; otherwise, enroll it.
+		agent, err := oa.findEnrolledAgent(ctx, zlog, instanceUID.String())
```
blakerouse, line 157: This is where I noticed the enrollment key behavior above. Normally we use the API key and ensure that the ID is for that API key. That is not the case here, is this what we are expecting?

internal/pkg/api/handleOpAMP.go
ycombinator: Yeah, we definitely want to replace the enrollment token (common to multiple agents) with an agent-specific API key longer term.  But we'll need to do this in a follow up PR because the OpAMP extension doesn't currently support a capability that we need for this to work as recommended by the OpAMP spec.  

Going into a bit more detail, we'll need to use the `ServerToAgent.ConnectionSettingsOffers.OpAMPConnectionSettings.headers` [field](https://opentelemetry.io/docs/specs/opamp/#opampconnectionsettingsheaders) to "switch out" the initial enrollment token with an Agent-specific API key.  But, for that to work, the OpAMP extension needs to support the `AcceptsOpAMPConnectionSettings` capability, which [it doesn't at the moment](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/d354cf59d4d046390e38437037289710220494cc/extension/opampextension/config.go#L72-L90).

blakerouse: Good to know its on the radar and that will be handled in the future. Thanks for providing me the context and information.

internal/pkg/config/env_defaults.go
```diff
@@ -57,6 +57,11 @@ const (
 	defaultStatusMax      = 50
 	defaultStatusMaxBody  = 0
 
+	defaultOpAMPInterval = time.Millisecond
+	defaultOpAMPBurst    = 1000
+	defaultOpAMPMax      = 0
+	defaultOpAMPMaxBody  = 0
```
cmacknz: This should have a max body shouldn't it? To avoid the OOM before auth category of problem?

Probably it should match the checkin max body limit, though perhaps we want it to be a bit bigger because of effective config.

internal/pkg/api/handleOpAMP.go
```diff
@@ -0,0 +1,310 @@
...
+			}
+		}
+
+		zlog.Debug().
+			Bool("is_enrolled", agent != nil).
+			Str("agent_id", instanceUID.String()).
+			Msg("agent enrollment status")
+
+		if agent == nil {
+			if agent, err = oa.enrollAgent(zlog, instanceUID.String(), message, apiKey); err != nil {
```
cmacknz, line 173: If a collector changes instance_uid on each request, we'd be creating potentially quite a lot of agents. I wonder if this is accidental or intentional DDoS mechanism for us?

I think this is not that different for containers that enroll on creation, but I wonder if we may need to modify `findAgent` to match on identifying attributes in addition to instance_uid or something in the future.

cmacknz: If we do end up creating unique per agent API keys here (which I'm not totally convinced makes sense for the monitoring use case in this particular way), this would be a mechanism for DDoSing Elasticsearch for sure as that's one of our most expensive operations.

internal/pkg/api/server.go
```diff
@@ -41,12 +44,25 @@ func NewServer(addr string, cfg *config.Server, opts ...APIOpt) *server {
...
+		cfg:    cfg,
+		logger: zap.NewStub("api-server"),
+	}
+
+	handler := newRouter(&cfg.Limits, a, a.tracer)
+	// If OpAMP feature is enabled, add OpAMP route handler to router and
+	// let OpAMP server modify connection context (setup later when HTTP server
+	// object is constructed).
+	if a.oa != nil && a.oa.Enabled() {
+		handler = addOpAMPRouteHandler(handler, a.oa, &cfg.Limits)
```
cmacknz, line 61: We should likely log that we are enabling opamp when we do so

cmacknz:
I followed the test instructions and got the contrib collector to connect, it is repeatedly logging the following error though, where it looks like Fleet Server is closing the connection too early generating an EOF error.

```
2026-02-10T16:17:54.526-0500    error   opampextension@v0.145.0/opamp_agent.go:142      Failed to connect to the OpAMP server     {"resource": {"service.instance.id": "6896fe6a-d39b-4325-ac44-72a4680ec809", "service.name": "otelcol-contrib", "service.version": "0.145.0"}, "otelcol.component.id": "opamp", "otelcol.component.kind": "extension", "error": "Post \"http://localhost:8220/v1/opamp\": EOF"}
github.com/open-telemetry/opentelemetry-collector-contrib/extension/opampextension.(*opampAgent).Start.func3
        github.com/open-telemetry/opentelemetry-collector-contrib/extension/opampextension@v0.145.0/opamp_agent.go:142
github.com/open-telemetry/opamp-go/client/internal.(*HTTPSender).sendRequestWithRetries
        github.com/open-telemetry/opamp-go@v0.22.0/client/internal/httpsender.go:256
github.com/open-telemetry/opamp-go/client/internal.(*HTTPSender).makeOneRequestRoundtrip
        github.com/open-telemetry/opamp-go@v0.22.0/client/internal/httpsender.go:193
github.com/open-telemetry/opamp-go/client/internal.(*HTTPSender).Run
        github.com/open-telemetry/opamp-go@v0.22.0/client/internal/httpsender.go:150
github.com/open-telemetry/opamp-go/client.(*httpClient).runUntilStopped
        github.com/open-telemetry/opamp-go@v0.22.0/client/httpclient.go:145
github.com/open-telemetry/opamp-go/client/internal.(*ClientCommon).StartConnectAndRun.func1
        github.com/open-telemetry/opamp-go@v0.22.0/client/internal/clientcommon.go:249
```

The agent is also stuck in the updating state as the screenshot shows, which doesn't seem right. This is what the agent document contains, probably we are not setting the checkin state and message properly.

```json
{
  "id": "019b8d7a-2da8-7657-b52d-492a9de33319",
  "type": "",
  "active": true,
  "enrolled_at": "2026-02-10T21:16:52Z",
  "policy_id": "fb423ac3-9d3d-48fb-9482-77c71e2214dd",
  "last_checkin": "2026-02-10T21:19:58Z",
  "last_checkin_status": "",
  "last_checkin_message": "",
  "packages": [],
  "sort": [
    1770758212000
  ],
  "agent": {
    "id": "019b8d7a-2da8-7657-b52d-492a9de33319",
    "version": ""
  },
  "local_metadata": {
    "elastic": {
      "agent": {
        "id": "019b8d7a-2da8-7657-b52d-492a9de33319",
        "version": "0.145.0"
      }
    },
    "host": {
      "hostname": "Craigs-MacBook-Pro",
      "name": "Craigs-MacBook-Pro"
    }
  },
  "unhealthy_reason": null,
  "last_known_status": "updating",
  "status": "updating",
  "metrics": {}
}
```

internal/pkg/api/handleOpAMP.go
ycombinator: https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/46001

juliaElastic:
> The agent is also stuck in the updating state as the screenshot shows, which doesn't seem right. This is what the agent document contains, probably we are not setting the checkin state and message properly.

There are a few fields that are involved in [status calculation](https://docs.google.com/document/d/1haO56bBxyUx61a_C7vj34lnwi-EIW410GTUwzB4oqz0/edit?tab=t.0#heading=h.arofp2njw0ji), added changes [here](https://github.com/elastic/fleet-server/pull/6330/changes/359d1af33c88d89d19a6439f1b77f1539076bdbc#diff-8eae4f76576728f1316ae744b620c2dac1f0ea3964426990df7329dcb9fff745R338) to map fields to Agent

internal/pkg/api/handleOpAMP.go
```diff
@@ -0,0 +1,310 @@
...
+			Msg("received AgentToServer message from agent")
+
+		// Check if Agent is "enrolled"; if it is, update it; otherwise, enroll it.
+		agent, err := oa.findEnrolledAgent(ctx, zlog, instanceUID.String())
+		if err != nil {
+			return &protobufs.ServerToAgent{
+				InstanceUid: instanceUID.Bytes(),
+				ErrorResponse: &protobufs.ServerErrorResponse{
+					Type:         protobufs.ServerErrorResponseType_ServerErrorResponseType_Unavailable,
+					ErrorMessage: fmt.Errorf("failed to check if agent is enrolled: %w", err).Error(),
```
michel-laterman:
```suggestion
					ErrorMessage: fmt.Sprintf("failed to check if agent is enrolled: %v", err)
```

internal/pkg/api/handleOpAMP.go
```diff
@@ -0,0 +1,310 @@
...
+			Str("agent_id", instanceUID.String()).
+			Msg("agent enrollment status")
+
+		if agent == nil {
+			if agent, err = oa.enrollAgent(zlog, instanceUID.String(), message, apiKey); err != nil {
+				return &protobufs.ServerToAgent{
+					InstanceUid: instanceUID.Bytes(),
+					ErrorResponse: &protobufs.ServerErrorResponse{
+						Type:         protobufs.ServerErrorResponseType_ServerErrorResponseType_Unavailable,
+						ErrorMessage: fmt.Errorf("failed to enroll agent: %w", err).Error(),
```
michel-laterman:
```suggestion
						ErrorMessage: fmt.Sprintf("failed to enroll agent: %v", err),
```

internal/pkg/api/handleOpAMP.go
```diff
@@ -0,0 +1,310 @@
...
+				}
+			}
+		}
+
+		if err := oa.updateAgent(zlog, agent, message); err != nil {
+			return &protobufs.ServerToAgent{
+				InstanceUid: instanceUID.Bytes(),
+				ErrorResponse: &protobufs.ServerErrorResponse{
+					Type:         protobufs.ServerErrorResponseType_ServerErrorResponseType_Unavailable,
+					ErrorMessage: fmt.Errorf("failed to update persisted Agent information: %w", err).Error(),
```
michel-laterman:
```suggestion
					ErrorMessage: fmt.Sprintf("failed to update persisted Agent information: %v", err),
```

internal/pkg/api/handleOpAMP.go
```diff
@@ -0,0 +1,310 @@
...
+		bc:         bc,
+		srv:        oaServer.New(nil),
+		agentMetas: map[string]localMetadata{},
+	}
+
+	return oa
+}
+
+func (oa *OpAMPT) Init() error {
+	settings := oaServer.Settings{
```
michel-laterman, line 67: We should enable compression support: https://github.com/open-telemetry/opamp-go/blob/97981b320ea6adc0a027380847d6b566995430d1/server/server.go#L20
 as the fleet-server currently supports gzipped responses

internal/pkg/api/handleOpAMP.go
```diff
@@ -0,0 +1,310 @@
...
+
+				// Setup connection callbacks.
+				connectionCallbacks := types.ConnectionCallbacks{
+					OnConnected: func(ctx context.Context, conn types.Connection) {
+						zlog.Debug().Msg("opAMP client connected")
+					},
+					OnConnectionClose: func(conn types.Connection) {
+						zlog.Debug().Msg("opAMP client disconnected")
+					},
+					OnMessage: oa.handleMessage(zlog, apiKey),
```
michel-laterman, line 106: We should add `OnReadMessageError` and `OnMessageResponseError` callbacks to at least log the errors.

internal/pkg/api/router.go
```diff
@@ -144,6 +150,8 @@ func (l *limiter) middleware(next http.Handler) http.Handler {
 			l.status.Wrap("status", &cntStatus, zerolog.DebugLevel)(next).ServeHTTP(w, r)
 		case "audit-unenroll":
 			l.auditUnenroll.Wrap("audit-unenroll", &cntAuditUnenroll, zerolog.DebugLevel)(next).ServeHTTP(w, r)
+		case "opamp":
+			l.opAMP.Wrap("opamp", &cntOpAMP, zerolog.DebugLevel)(next).ServeHTTP(w, r)
```
michel-laterman, line 154: I think this is Ok for now; but this may need to be done in the `OnConnecting` callback - I'm not sure how the upstream client will handle our JSON response

internal/pkg/api/server.go
michel-laterman: 👍

internal/pkg/api/handleOpAMP.go
michel-laterman: Why would the collector do a unique instance_uid per request? It should be tied to process lifetime at the very least: https://github.com/open-telemetry/opamp-spec/blob/main/specification.md#agenttoserverinstance_uid

A client can use `RequestInstanceUid` to request the server generate an ID - I think this is the DDoS mechanism you're referencing.

cmacknz: I was thinking of something like a crash looping container or process that changed instance_uid on each request. Or an improperly implemented opamp client, not a collector behaving properly.

We have this problem with containerized Elastic Agents too, however all of our reference configs include a host path volume and Elastic Agent is stateful by default that makes re-enrolling in this situation less likely.

Collectors are stateless by default and the opamp extension doesn't include a way to use a storage extension to persist it, though you can specify a fixed value.

I am not super worried about this, it's not a new problem for fleet-server, and there is a rate limiter on this endpoint, but I think we are more likely to get into accidental mass enrollment situations with opamp so I wanted to make sure we had it in our minds.

internal/pkg/api/server.go
ycombinator: Done in 4ef51f81.

internal/pkg/api/handleOpAMP.go
ycombinator: Done in 14ec080f.

internal/pkg/config/env_defaults.go
ycombinator: Checkin max body limit is 1 MB.  Set the OpAMP max body limit to 3 MB in ea7712d829782bec91a078183db978a1b86cee15.

internal/pkg/api/handleOpAMP.go
ycombinator: Implemented in c3571cfa.

internal/pkg/api/handleOpAMP.go
ycombinator: Created upstream issue for allowing a way for the OpAMP extension to persist the `instance_uid` across collector restarts: https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/46064

internal/pkg/api/server.go
```diff
@@ -156,7 +158,14 @@ func addOpAMPRouteHandler(existingHandler http.Handler, oa *OpAMPT, cfg *config.
 	r.Use(Limiter(cfg).middleware)
 
 	opAMPHandler := oa.handler
-	r.HandleFunc("/v1/opamp", http.HandlerFunc(opAMPHandler))
+	r.HandleFunc("/v1/opamp", func(w http.ResponseWriter, r *http.Request) {
+		// Disable HTTP keep-alive for OpAMP requests. The OTel Collector polls
+		// every ~30s and the server's IdleTimeout is also 30s, creating a race
+		// where the server closes the idle connection just as the client tries
+		// to reuse it, resulting in EOF errors on the client side.
+		w.Header().Set("Connection", "close")
```
cmacknz: Wouldn't it be better to increase the server's idle timeout than to unnecessarily force a new TCP connection in each request?

internal/pkg/api/handleOpAMP.go
```diff
@@ -0,0 +1,317 @@
...
+				if err != nil {
+					zlog.Warn().Err(err).Msg("unauthenticated opamp request")
+					return types.ConnectionResponse{
+						Accept:         false,
+						HTTPStatusCode: http.StatusUnauthorized,
+						HTTPResponseHeader: map[string]string{
+							"Content-Type": "application/x-protobuf",
+						},
+					}
+				}
```
michel-laterman, line 94: We also need to handle 429s as we do with:
https://github.com/elastic/fleet-server/blob/17bd02a456e0c6d31229b45731a196d169821173/internal/pkg/api/error.go#L171-L176

internal/pkg/api/server.go
ycombinator: We could do that but that would be a server-wide setting (default is 30s). It could mean that Fleet Server would handle fewer concurrent connections, server-wide.  Are we OK with that?  Maybe we could bump this up to something just beyond the Collector's polling interval, so something like 35s?

cmacknz: 35s sounds reasonable to me. My main worry is misbehaving collectors polling us much faster than 30s and them creating a new TCP connection for each one of their too fast requests.

internal/pkg/api/handleOpAMP.go
ycombinator: Done in 17bfe2e913900e020252cb1bf226914ca0600da4.

internal/pkg/api/server.go
ycombinator: Done in 8da7bbf7.

docs/opamp.md
```diff
@@ -0,0 +1,56 @@
...
+This section describes how to connect a OpenTelemetry Collector instance to Fleet Server over OpAMP.
+
+1. Create a deployment in Elastic Cloud.  Integrations Server is not needed as we will instead be
+   using the Fleet Server instance built from this repository so it can "speak" OpAMP to the OpenTelemetry
+   Collector.
+2. Create an Elasticsearch service account token using Kibana > Dev Tools > Console.
+   ```
+   POST /_security/service/elastic/fleet-server/credential/token/opamp
+   ```
+3. Create a `fleet-server.dev.yml` configuration file as described in https://github.com/elastic/kibana/blob/main/x-pack/platform/plugins/shared/fleet/dev_docs/local_setup/developing_kibana_and_fleet_server.md.
```
michel-laterman: Can't we just use the default one: https://github.com/elastic/fleet-server/blob/main/fleet-server.yml?

We should also include step to have people set the `enable_opamp` flag

docs/opamp.md
```diff
@@ -0,0 +1,56 @@
...
+4. Build the Fleet Server binary for your platform.
+   ```
+   PLATFORMS=darwin/arm64 mage build:binary
```
michel-laterman: `mage build:local` builds a binary for the local system in `bin/`

internal/pkg/api/handleOpAMP.go
```diff
@@ -0,0 +1,327 @@
...
+	"github.com/elastic/fleet-server/v7/internal/pkg/dl"
+	"github.com/elastic/fleet-server/v7/internal/pkg/model"
+	"github.com/gofrs/uuid/v5"
+	"github.com/open-telemetry/opamp-go/protobufs"
+	oaServer "github.com/open-telemetry/opamp-go/server"
+	"github.com/open-telemetry/opamp-go/server/types"
+	"github.com/rs/zerolog"
+	"github.com/rs/zerolog/hlog"
+	"go.opentelemetry.io/otel/attribute"
+	semconv "go.opentelemetry.io/otel/semconv/v1.37.0"
```
michel-laterman: should we try to group the imports as well?

internal/pkg/api/handleOpAMP.go
```diff
@@ -0,0 +1,327 @@
...
+	cfg   *config.Server
+	bulk  bulk.Bulk
+	cache cache.Cache
+	bc    *checkin.Bulk
+
+	srv     oaServer.OpAMPServer
+	handler oaServer.HTTPHandlerFunc
+	connCtx oaServer.ConnContext
+
+	agentMetas map[string]localMetadata
```
michel-laterman: Is `agentMeta` used for anything?

internal/pkg/api/handleOpAMP.go
```diff
@@ -0,0 +1,327 @@
...
+				ErrorResponse: &protobufs.ServerErrorResponse{
+					Type:         protobufs.ServerErrorResponseType_ServerErrorResponseType_BadRequest,
+					ErrorMessage: "failed to parse instance_uid from AgentToServer message",
+				},
+			}
+		}
+
+		zlog.Debug().
+			Str("instance_uid", instanceUID.String()).
+			Str("aToS", message.String()).
```
michel-laterman: Is this just for development? I don't think we should log the complete message (or response)

internal/pkg/api/handleOpAMP.go
```diff
@@ -0,0 +1,327 @@
...
+				ErrorResponse: &protobufs.ServerErrorResponse{
+					Type:         protobufs.ServerErrorResponseType_ServerErrorResponseType_Unavailable,
+					ErrorMessage: fmt.Sprintf("failed to check if agent is enrolled: %v", err),
+				},
+			}
+		}
+
+		zlog.Debug().
+			Bool("is_enrolled", agent != nil).
+			Str("agent_id", instanceUID.String()).
```
michel-laterman: instance_uid is now agent_id.
We should be consistent on one, I think if we wanted to follow the opamp spec we should use something like opamp.agent.uid, however we have ecs.AgentID (fleet.agent.id) defined for this as well.
Do we want to have the same namespace/conventions in logs as the existing fleet-agents or should we introduce some new ones for opamp?

internal/pkg/api/handleOpAMP.go
```diff
@@ -0,0 +1,327 @@
...
+	if agent.Id == "" {
+		return nil, nil
+	}
+
+	return &agent, nil
+}
+
+func (oa *OpAMPT) enrollAgent(zlog zerolog.Logger, agentID string, aToS *protobufs.AgentToServer, apiKey *apikey.APIKey) (*model.Agent, error) {
+	zlog.Debug().
+		Str("agentID", agentID).
```
michel-laterman: instance_uid = agent_id = agentID 🥲

internal/pkg/api/handleOpAMP.go
```diff
@@ -0,0 +1,327 @@
...
+	meta.Elastic.Agent.ID = agentID
+	if aToS.AgentDescription != nil {
+		// Extract agent version
+		for _, ia := range aToS.AgentDescription.IdentifyingAttributes {
+			switch attribute.Key(ia.Key) {
+			case semconv.ServiceVersionKey:
+				meta.Elastic.Agent.Version = ia.GetValue().GetStringValue()
+			}
+		}
+		zlog.Debug().Str("agent_version", meta.Elastic.Agent.Version).Msg("extracted agent version")
```
michel-laterman: we should also try to follow a similar naming convention here (`opamp.agent.version` or `fleet.agent.version`)

michel-laterman (approved):
I think the initial work is good, but we should be consistent with what we want to log and determine if we want to identify opamp agents differently (in logs) from fleet agents.

I also think that we should not be logging full json blobs or request/response messages.

internal/pkg/api/handleOpAMP.go
ycombinator: I think new ones for OpAMP would be good.  Would make it easy to know we're dealing with an OpAMP message.  Will use `opamp.agent.uid` consistently in all log messages.

docs/opamp.md
ycombinator: Fixed in 11f98ffe.

docs/opamp.md
ycombinator: Thanks. Fixed in 11f98ffe.

internal/pkg/api/handleOpAMP.go
ycombinator: Done in 3f108aa0.

internal/pkg/api/handleOpAMP.go
ycombinator: Good catch.  Holdover from the POC.  Removed in 8da08d20.

internal/pkg/api/handleOpAMP.go
ycombinator: Yes, removed in 90431e60.

internal/pkg/api/handleOpAMP.go
ycombinator: Fixed in fcbbead2.

internal/pkg/api/handleOpAMP.go
ycombinator: Fixed in fcbbead2.

internal/pkg/api/handleOpAMP.go
ycombinator: Fixed in 4499978c.

internal/pkg/api/handleOpAMP.go
```diff
@@ -0,0 +1,323 @@
...
+			return &protobufs.ServerToAgent{
+				ErrorResponse: &protobufs.ServerErrorResponse{
+					Type:         protobufs.ServerErrorResponseType_ServerErrorResponseType_BadRequest,
+					ErrorMessage: "failed to parse instance_uid from AgentToServer message",
+				},
+			}
+		}
+
+		zlog.Debug().
+			Str("instance_uid", instanceUID.String()).
```
michel-laterman: This should also be `opamp.agent.uid`

internal/pkg/api/handleOpAMP.go
```diff
@@ -0,0 +1,323 @@
...
+		zlog.Debug().Str("hostname", meta.Host.Hostname).Msg("extracted hostname")
+	}
+
+	// Update local metadata if something has changed
+	data, err := json.Marshal(meta)
+	if err != nil {
+		return nil, fmt.Errorf("failed to marshal local metadata: %w", err)
+	}
+
+	zlog.Debug().RawJSON("meta", data).Msg("updating local metadata")
```
michel-laterman: We should probably get rid of the raw json here

internal/pkg/api/handleOpAMP.go
```diff
@@ -0,0 +1,323 @@
...
+		LocalMetadata: data,
+	}
+
+	data, err = json.Marshal(agent)
+	if err != nil {
+		return nil, err
+	}
+
+	zlog.Debug().
+		Str("agent document", string(data)).
```
michel-laterman: we should get rid of the document string

michel-laterman (dismissed):

internal/pkg/api/handleOpAMP.go
ycombinator: Removed in 43e8ecff.

internal/pkg/api/handleOpAMP.go
ycombinator: Removed in 43e8ecff.

internal/pkg/api/handleOpAMP.go
ycombinator: Fixed in 58e84287.

internal/pkg/api/server.go
cmacknz: We should allow keep alives again afterwards.

internal/pkg/api/handleOpAMP.go
```diff
@@ -0,0 +1,320 @@
...
+
+	return oa
+}
+
+func (oa *OpAMPT) Init() error {
+	settings := oaServer.Settings{
+		Callbacks: types.Callbacks{
+			OnConnecting: func(request *http.Request) types.ConnectionResponse {
+				zlog := hlog.FromRequest(request).With().
+					Str("mod", kOpAMPMod).
```
michel-laterman, line 71: (nitpick) I don't see this attribute anywhere else on our other endpoints

internal/pkg/api/handleOpAMP.go
```diff
@@ -0,0 +1,320 @@
...
+			return &protobufs.ServerToAgent{
+				ErrorResponse: &protobufs.ServerErrorResponse{
+					Type:         protobufs.ServerErrorResponseType_ServerErrorResponseType_BadRequest,
+					ErrorMessage: "failed to parse instance_uid from AgentToServer message",
+				},
+			}
+		}
+
+		zlog.Debug().
+			Str("opamp.agent.id", instanceUID.String()).
```
michel-laterman: nitpick: Should we just attach the uid to the logger with `zlog = zlog.With().Str("opamp.agent.id", instanceUID.String()).Logger()` so it's on all logs that `zlog` produces?

internal/pkg/api/handleOpAMP.go
ycombinator: It's there in a couple of endpoints:
* https://github.com/elastic/fleet-server/blob/04b159f818278ed99599ad07436d847598931565/internal/pkg/api/api.go#L224
* https://github.com/elastic/fleet-server/blob/04b159f818278ed99599ad07436d847598931565/internal/pkg/api/api.go#L100

internal/pkg/api/handleOpAMP.go
ycombinator: Good idea. Done in ea839065.

internal/pkg/api/server.go
cmacknz: Looks like we are still closing the connection and referencing the now obsolete 30s server side timeout.

ycombinator: Thanks, fixed in 854e369e.

michel-laterman (approved):
lgtm

PR #6280, opened 2026-02-04
Run individual E2E tests

ycombinator:

This PR allows developers to run individual E2E tests.  This is useful when iterating on such tests during development.

It introduces a new environment variable `TEST_RUN` that's accepted by the `mage:e2e` target.  The value of this environment variable should be an expression that's understood by the `go test -run` flag.  

1. Run a specific test.
   ```
   TEST_RUN='TestStandAloneRunningSuite/TestAPMInstrumentation' mage test:e2e
   ```

2. Look at the output and verify that only the specified test was run.

## Design Checklist

## Checklist

### DIFF
--- a/testing/e2e/README.md
+++ b/testing/e2e/README.md
@@ -8,7 +8,7 @@ mage test:e2e
 ```
 
 If you want to run specific tests, you can specify them via the `TEST_RUN` environment variable, as an expression that
-is understood by the `go test -name` flag.
+is understood by the `go test -run` flag.
 ```bash
 TEST_RUN='TestStandAloneRunningSuite/TestAPMInstrumentation' mage test:e2e
 ```


testing/e2e/README.md
```diff
@@ -7,6 +7,12 @@ The tests can be ran from the repo's root directory by running:
 mage test:e2e
 ```
 
+If you want to run specific tests, you can specify them via the `TEST_RUN` environment variable, as an expression that
+is understood by the `go test -name` flag.
```
cmacknz: It looks like we are passing the value of `TEST_RUN` to the `-run` argument, which is a bit more flexible and we should adjust the documentation to state that this is an expression for `-run`.

ycombinator: Yes, that's supposed to say `-run`, not `-name` 🤦.  Good catch.

ycombinator: Fixed in 274e0ca1db0331df8936b9e1a4da892b24c27b8a.

PR #6289, opened 2026-02-04
[OpAMP] Add E2E test

### LINKED ISSUE #6393: [OpAMP][E2E Test] Verify that contrib OTel Collectors can talk to Fleet over OpAMP
In https://github.com/elastic/fleet-server/pull/6270, we introduced a `/v1/opamp` endpoint in Fleet Server.  OTel Collectors, specifically a contrib release, We need to write an E2E test in Fleet Server that verifies that a contrib OTel Collector can successfully connect to Fleet Server's OpAMP endpoint.

ycombinator:

This PR ensures that an OTel Collector (from an upstream contrib release) is able to successfully connect to Fleet Server over OpAMP.  Preliminary OpAMP support was added in Fleet Server in #6270 so this PR here is a follow up to that work.

Note: There will be a follow up PR that adds a test (or extends this one) to ensure that EDOT Collector is able to successfully connect to Fleet Server over OpAMP (#6394) 

By adding a new E2E test, `TestStandAloneRunningSuite/TestOpAMP` that downloads and extracts the OTel Collector binary from an upstream contrib release, configures it with the `opamp` extension, configures Fleet Server to turn on the `feature_flags.enable_opamp` feature flag, runs the Collector, and verifies that the Collector is connecting to Fleet Server over OpAMP.

## Design Checklist

## Checklist

## Related issues

- Closes #6393

### DIFF
--- a/testing/e2e/stand_alone_test.go
+++ b/testing/e2e/stand_alone_test.go
@@ -706,4 +706,5 @@ func (suite *StandAloneSuite) TestOpAMP() {
 	// suite.Equal(otelVersion, agentDoc.Agent.Version, "expected agent.version to match otelcol-contrib binary version")
 	// suite.Equal(1, agentDoc.Revision, "expected policy_revision_idx to be 1")
 	// suite.Contains(agentDoc.Tags, "otelcontribcol", "expected tags to contain otelcontribcol")
+	// suite.Equal("online", agentDoc.Status, "expected status to be online")
 }
--- a/testing/e2e/scaffold/scaffold.go
+++ b/testing/e2e/scaffold/scaffold.go
@@ -228,19 +228,9 @@ func (s *Scaffold) FleetServerStatusCondition(ctx context.Context, url string, c
 	}
 }
 
-// AgentIsOnline will check Kibana if the agent specified by the passed id has the online status.
+// AgentIsOnline polls Kibana's Fleet API until the agent with the given ID has the online status.
 // The test is marked as failed if the passed context terminates before that.
 func (s *Scaffold) AgentIsOnline(ctx context.Context, id string) {
-	s.AgentHasStatus(ctx, id, "online")
-}
-
-func (s *Scaffold) AgentIsUpdating(ctx context.Context, id string) {
-	s.AgentHasStatus(ctx, id, "updating")
-}
-
-// AgentHasStatus polls Kibana's Fleet API until the agent with the given ID has one of the
-// specified statuses.
-func (s *Scaffold) AgentHasStatus(ctx context.Context, id string, statuses ...string) {
 	timer := time.NewTimer(time.Second)
 	for {
 		select {
@@ -273,10 +263,8 @@ func (s *Scaffold) AgentHasStatus(ctx context.Context, id string, statuses ...st
 			s.T().Logf("Kibana agent response: %s", string(p))
 			err = json.Unmarshal(p, &obj)
 			s.Require().NoError(err, "unmarshal failure")
-			for _, status := range statuses {
-				if obj.Item.Status == status {
-					return
-				}
+			if obj.Item.Status == "online" {
+				return
 			}
 			timer.Reset(time.Second)
 		}
@@ -324,6 +312,7 @@ type ESAgentDoc struct {
 	PolicyID      string   `json:"policy_id"`
 	AgentPolicyID string   `json:"agent_policy_id"`
 	Type          string   `json:"type"`
+	Status        string   `json:"status"`
 	Tags          []string `json:"tags"`
 	Agent         struct {
 		ID      string `json:"id"`


testing/e2e/const.go
```diff
@@ -7,3 +7,4 @@ package e2e
 const binaryName = "fleet-server"                //nolint:unused // work around to get platform specific binary name for tests
 const agentName = "elastic-agent"                //nolint:unused // work around to get platform specific binary name for tests
 const agentDevName = "elastic-development-agent" //nolint:unused // work around to get platform specific binary name for tests
+const otelColContribVersion = "0.144.0"          //nolint:unused // OTel Collector version used in e2e tests
```
ebeahan: Do we need a way to bump this routinely inline with what's used by Agent?

Doesn't need to be in scope for these changes, but, if yes, something to track in the backlog.

ycombinator: Yes, and we should have automation for it.  I'll wait for the review on this PR to see if there are any suggestions around the OTel Collector download and then, if it still makes sense, create an issue for automatically updating this version periodically in the Outcome issue.

cmacknz: The contrib collector is ~100 MB, we should consider just checking the whole thing into the repository and updating it automatically every 3 weeks when it releases. Possibly there are better ways to do this, there is likely some prior art out there to look at, but downloading a 100 MB file for every test run seems inefficient and seems like a source of flakiness.

Probably the most common work around is to test against a container which could work for this.

testing/e2e/stand_alone_test.go
```diff
@@ -579,3 +582,159 @@ func (suite *StandAloneSuite) TestAPMInstrumentation() {
...
+	err = otelCmd.Start()
+	suite.Require().NoError(err)
+	defer otelCmd.Wait()
+
+	// Verify that the OTel Collector was enrolled in Fleet. OpAMP agents communicate via the
+	// OpAMP protocol rather than Fleet's normal checkin/ack protocol, so they never acknowledge
+	// the initial policy change action. Kibana therefore shows them as "updating" rather than
+	// "online".
+	suite.T().Logf("Waiting for agent %s to be enrolled in Fleet", instanceUID)
+	suite.AgentIsUpdating(ctx, instanceUID)
```
cmacknz: Would it be better to fetch the entire .fleet-agent document so we can more exhaustively validate what it contains?

testing/e2e/const.go
cmacknz: We could also compile contrib from source or with ocb

blakerouse: Checking it into the repository I believe would still result in it being fetched, just when the source is fetched versus when this testing is ran. I would prefer not to have the collector committed into the repository.

cmacknz: Sure we can optimize how this test works separately, comments here are not blocking

ycombinator: Source is 12MB on the wire so building from it seems like a good way to go.  I will update this PR.

ycombinator: Implemented in 41597ba0.

testing/e2e/stand_alone_test.go
ycombinator: Implemented in 7d7b5619.

cmacknz (approved):

testing/e2e/stand_alone_test.go
```diff
@@ -579,3 +579,132 @@ func (suite *StandAloneSuite) TestAPMInstrumentation() {
...
+	makeCmd.Stdout = os.Stdout
+	makeCmd.Stderr = os.Stderr
+	err = makeCmd.Run()
+	suite.Require().NoError(err)
+
+	// The make target places the binary under bin/; move it to the expected path.
+	builtBinary := filepath.Join(cloneDir, "bin", fmt.Sprintf("otelcontribcol_%s_%s", runtime.GOOS, runtime.GOARCH))
+	otelBinaryPath := filepath.Join(dir, "otelcol-contrib")
+	err = os.Rename(builtBinary, otelBinaryPath)
+	suite.Require().NoError(err)
```
michel-laterman, line 664: Can we just use the API: https://docs.github.com/en/rest/releases/releases?apiVersion=2022-11-28#get-the-latest-release

to get the latest release, then get the asset for the GOOS/GOARCH of the tester, such as [otelcol-contrib_0.146.1_linux_amd64.tar.gz](https://github.com/open-telemetry/opentelemetry-collector-releases/releases/download/v0.146.1/otelcol-contrib_0.146.1_linux_amd64.tar.gz) instead of cloning/building the collector?

ycombinator: We decided in https://github.com/elastic/fleet-server/pull/6289#discussion_r2853464143 to build from source instead of downloading the artifact since the download size was significantly smaller on the wire (12 MB for source vs. ~100 MB for binary artifact tar.gz).

PR #4574, opened 2025-03-12
Add DRA FIPS artifacts to package pipeline

michel-laterman:
## What is the problem this PR solves?

Provide FIPS dra artifacts.

## How does this PR solve the problem?

Add `FIPS="true"` env flag to packageing pipeline.
Packaging pipline uses microsoft/go instead of golang/go to build DRA artifacts for FIPS.
Change package name from `fleet-server-*-fips` to `fleet-server-fips-*`.

## Design Checklist

- ~~I have ensured my design is stateless and will work when multiple fleet-server instances are behind a load balancer.~~
- ~~I have or intend to scale test my changes, ensuring it will work reliably with 100K+ agents connected.~~
- ~~I have included fail safe mechanisms to limit the load on fleet-server: rate limiting, circuit breakers, caching, load shedding, etc.~~

## Checklist

- [x] I have commented my code, particularly in hard-to-understand areas
- [x] I have made corresponding changes to the documentation
- ~~I have made corresponding change to the default configuration files~~
- ~~I have added tests that prove my fix is effective or that my feature works~~
- ~~I have added an entry in `./changelog/fragments` using the [changelog tool](https://github.com/elastic/fleet-server#changelog)~~

PR #6303, opened 2026-02-06
ci: use custom VM images

v1v:

Use custom VM images

Ensure we know what tools we use and pin the version - then we will be able to automatically bump the versions using the PR review process.

## Design Checklist

## Checklist

### DIFF
--- a/.buildkite/pipeline.package.mbp.yml
+++ b/.buildkite/pipeline.package.mbp.yml
@@ -2,7 +2,7 @@
 name: "fleet server package mbp"
 env:
   IMAGE_UBUNTU_X86_64: "platform-ingest-fleet-server-ubuntu-2204-1772547557"
-  IMAGE_UBUNTU_ARM_64: "platform-ingest-fleet-server-ubuntu-2204-aarch64-1772547557"
+  IMAGE_UBUNTU_ARM_64: "test-platform-ingest-fleet-server-ubuntu-2204-aarch64-1772547557"
 
 # This section is used to define the plugins that will be used in the pipeline.
 # See https://buildkite.com/docs/pipelines/integrations/plugins/using#using-yaml-anchors-with-plugins
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -4,8 +4,8 @@ env:
   DOCKER_COMPOSE_VERSION: "1.25.5"
   TERRAFORM_VERSION: "1.6.4"
   IMAGE_UBUNTU_X86_64: "platform-ingest-fleet-server-ubuntu-2204-1772547557"
-  IMAGE_UBUNTU_ARM_64: "platform-ingest-fleet-server-ubuntu-2204-aarch64-1772547557"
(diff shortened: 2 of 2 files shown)


.buildkite/pipeline.package.mbp.yml
fr4nc1sc0-r4m0n, line 65: @v1v Apparently the FIPS are using the same images, right?

v1v: That's only for the packaging, as far as I see at https://github.com/elastic/fleet-server/commit/8bb8c740a7f82809b50a559584b2831d0f3fe5d8

The tests run using the FIPS images.

fr4nc1sc0-r4m0n: And what's the reason for that? Wouldn't it make more sense to build and test over the same images?

Even with that, the ones that are failing are the packaging pipelines https://buildkite.com/elastic/fleet-server-package-mbp/builds/3243

v1v: I don't know, https://github.com/elastic/fleet-server/pull/4574 introduced it.

fr4nc1sc0-r4m0n:
Seems that are flaky test because in each execution a different ones fails. Retrying

.buildkite/pipeline.package.mbp.yml
```diff
@@ -1,8 +1,8 @@
 # yaml-language-server: $schema=https://raw.githubusercontent.com/buildkite/pipeline-schema/main/schema.json
 name: "fleet server package mbp"
 env:
-  IMAGE_UBUNTU_X86_64: "family/core-ubuntu-2204"
-  IMAGE_UBUNTU_ARM_64: "core-ubuntu-2204-aarch64"
+  IMAGE_UBUNTU_X86_64: "platform-ingest-fleets-ubuntu-2204-1770381074"
```
v1v: `Running agent pre-command hook`, I'm not sure what's the reason for using that particular pre-command hook, but IIRC, fleet-server does not need it.

Can you confirm if the VM Images used are not copying the pre-command hook?

fr4nc1sc0-r4m0n: I've been trying to find any reference from the fleet-server images but I found nothing, so I'd say that no copy of the `hooks/base/pre-command` is happening in this images

v1v: `1770381074` is from February (`2026-02-06 12:31:14 UTC`), but there are new commits in the ci-agent-images. Can you generate the new VM images and update the timestamp here?

fr4nc1sc0-r4m0n: Sure:

- Image generation:  https://buildkite.com/elastic/vm-images-platform-ingest/builds/1333#
- The new timestamp is `1772547557`
- Changes applied: https://github.com/elastic/fleet-server/pull/6303/commits/25476f2cfa9f529ddd79e1cd2092e2b68b851f74

Let's see if it works.

fr4nc1sc0-r4m0n: In this execution the E2E test have passed: 

- https://buildkite.com/elastic/fleet-server/builds/13195#

PR #6400, opened 2026-02-20
[OpAMP] Map fields from AgentToServer message to Agent fields

juliaElastic:

Map data from OpAMP AgentToServer message to Agent fields in `.fleet-agents`

Add OpAMP message fields to Agent document: 
- convert capabilities to string array
- convert effective config to json object and sanitize
- add health and set last_checkin_status, last_checkin_message, etc.
- add identifying and non-identifying attributes
- add sequence number

Follow instructions in https://github.com/elastic/fleet-server/blob/main/docs/opamp.md

Download and extract otel collector, e.g.: 
```
curl -L -O https://github.com/open-telemetry/opentelemetry-collector-releases/releases/download/v0.144.0/otelcol-contrib_0.144.0_darwin_arm64.tar.gz 
```

Create otel config to include system fields and internal telemetry, e.g. `otel-opamp.yaml`:

```
receivers:
  otlp:
    protocols:
      grpc:
        endpoint: 0.0.0.0:4317

processors:
  resourcedetection:
    detectors: ["system","env"]
    system:
      hostname_sources: ["os"]
      resource_attributes:
        host.name:
          enabled: true
        host.arch:
          enabled: true
        os.description:
          enabled: true
        os.type:
          enabled: true

exporters:
  debug:
    verbosity: detailed
  elasticsearch/otel:
    endpoints: [ "http://localhost:9200" ]
    api_key: ${env:ES_API_KEY} 
    mapping:
      mode: otel
  otlp:
    endpoint: "http://localhost:4317"
    tls:
      insecure: true

extensions:
  opamp:
    server:
      http:
        endpoint: http://localhost:8220/v1/opamp
        tls:
          insecure: true
        headers:
          Authorization: ApiKey ${env:FLEET_ENROLLMENT_TOKEN}
    instance_uid: ${env:INSTANCE_UID}
    capabilities:
      reports_effective_config: true

service:
  pipelines:
    logs:
      receivers: [otlp]
      exporters: [debug]
    metrics:
      receivers: [otlp]
      processors: [resourcedetection]
      exporters: [elasticsearch/otel]
  extensions: [opamp]

  # publish collector internal telemetry
  telemetry:
    metrics:
      level: detailed
      readers:
        - periodic:
            interval: 30000
            exporter:
              otlp:
                protocol: grpc
                endpoint: http://localhost:4317
```

Create API keys and start otel collector:

```
 cd ~/Downloads/otelcol-contrib_0.144.0_darwin_arm64
 
 export INSTANCE_UID=<uuid> # e.g. "519b8d7a-2da8-7657-b52d-492a9de33313"
 export OTEL_RESOURCE_ATTRIBUTES="service.instance.id=$INSTANCE_UID" # to include instance id in internal telemetry data
 export ES_API_KEY=<api_key> # ES API key from observability onboarding UI
 export FLEET_ENROLLMENT_TOKEN=<enrollment_token> 
 ./otelcol-contrib --config ./otel-opamp.yaml
```

## Design Checklist

## Checklist

## Related issues

Closes https://github.com/elastic/ingest-dev/issues/6982


internal/pkg/api/handleOpAMP.go
```diff
@@ -310,10 +369,141 @@ type localMetadata struct {
...
+		"secret",
+		"token",
+		"apikey",
+		"api_key",
+		"access_key",
+		"private_key",
+		"credential",
+		"credentials",
+	} {
+		if key == token || strings.Contains(key, token) {
```
ycombinator: Should this be a `strings.Contains` check or a `==` check?  With the former, we could match non-sensitive strings like "passthrough" or "bypass".

internal/pkg/api/handleOpAMP.go
```diff
@@ -310,10 +369,141 @@ type localMetadata struct {
...
+}
+
+func ProtobufKVToRawMessage(kv []*protobufs.KeyValue) (json.RawMessage, error) {
+	// 1. Build an intermediate map to represent the JSON object
+	data := make(map[string]interface{}, len(kv))
+	for _, item := range kv {
+		if item.Value == nil {
+			continue
+		}
+		switch v := item.Value.GetValue().(type) {
```
ycombinator: We're not handling cases where the value is an array or map.  I think this is fine, since this function is only being used for identifying and non-identifying attributes which should not contain such values.  Maybe consider adding a case for these types and error out or at least add a comment explaining why we're not handling them?  Or, if you want to go for completeness anyway, handle these types of values recursively?

internal/pkg/api/handleOpAMP.go
```diff
@@ -261,15 +275,32 @@ func (oa *OpAMPT) enrollAgent(zlog zerolog.Logger, agentID string, aToS *protobu
 		return nil, fmt.Errorf("failed to marshal local metadata: %w", err)
 	}
 
+	identifyingAttributes, err := ProtobufKVToRawMessage(aToS.AgentDescription.IdentifyingAttributes)
```
ycombinator: According to https://opentelemetry.io/docs/specs/opamp/#agenttoserveragent_description, `aToS.AgentDescription` could be `nil`, so we'll get a nil-pointer panic here.  Maybe we should add a guard here before trying to extract identifying and non-identifying attributes?

internal/pkg/api/handleOpAMP.go
```diff
@@ -291,14 +322,42 @@ func (oa *OpAMPT) updateAgent(zlog zerolog.Logger, agent *model.Agent, aToS *pro
...
+		}
+		healthBytes, err := json.Marshal(aToS.Health)
+		if err != nil {
+			return fmt.Errorf("failed to marshal health: %w", err)
+		}
+		initialOpts = append(initialOpts, checkin.WithHealth(healthBytes))
+	}
+
+	initialOpts = append(initialOpts, checkin.WithStatus(status))
+	initialOpts = append(initialOpts, checkin.WithSequenceNum(aToS.SequenceNum))
```
ycombinator, line 359: Reading through https://opentelemetry.io/docs/specs/opamp/#agent-status-compression, if the Collector restarts, `aToS.SequenceNum` could be 0.  Do we want to write it to ES as 0 in that scenario, potentially overwriting what's already there, e.g. 17?

internal/pkg/api/handleOpAMP.go
```diff
@@ -261,15 +275,32 @@ func (oa *OpAMPT) enrollAgent(zlog zerolog.Logger, agentID string, aToS *protobu
...
 		PolicyID:   rec.PolicyID,
 		Agent: &model.AgentMetadata{
-			ID: agentID,
+			ID:      agentID,
+			Version: meta.Elastic.Agent.Version,
+			Type:    agentType,
 		},
-		LocalMetadata: data,
+		LocalMetadata:            data,
+		PolicyRevisionIdx:        1,
```
ycombinator, line 309: Maybe add a comment here why this can be hardcoded as 1?

model/schema.json
```diff
@@ -744,7 +748,34 @@
...
+        },
+        "non_identifying_attributes": {
+          "description": "Non-identifying attributes of the collector",
+          "format": "raw"
+        },
+        "sequence_num": {
+          "description": "The sequence number of the last collector message",
+          "type": "integer"
+        },
+         "health": {
```
ycombinator, line 764: Extra space in indentation?

```suggestion
        "health": {
```

model/schema.json
```diff
@@ -744,7 +748,34 @@
...
+          "description": "Health information of the collector",
+          "format": "raw"
+        },
+        "capabilities": {
+           "description": "List of capabilities of the collector",
```
ycombinator: Extra space in indentation?

```suggestion
          "description": "List of capabilities of the collector",
```

model/schema.json
```diff
@@ -744,7 +748,34 @@
...
+          "type": "array",
+          "items": {
+            "type": "string"
+          }
+        },
+          "effective_config": {
```
ycombinator: Two extra spaces in indentation?

```suggestion
        "effective_config": {
```

internal/pkg/api/handleOpAMP.go
juliaElastic: So does it mean in case of a restart, sequence_num is restarted from 0?
If so, I think it's still good to save the latest sequence_num, so we can detect if the next message is not incremented by 1. 

Tested and indeed the sequence_num restarts in case of a collector restart.

internal/pkg/api/handleOpAMP.go
juliaElastic: good point, removed the Contains condition

internal/pkg/api/handleOpAMP.go
juliaElastic: added a default case to error out on other types

internal/pkg/api/handleOpAMP.go
juliaElastic: moved attributes handling inside the `AgentDescription != nil` check

internal/pkg/api/handleOpAMP.go
juliaElastic: added a comment

model/schema.json
juliaElastic: fixed

model/schema.json
juliaElastic: fixed

internal/pkg/api/handleOpAMP.go
cmacknz: You could try calling https://pkg.go.dev/google.golang.org/protobuf/encoding/protojson#Marshal on each value as a way to generalize. We definitely want to handle all the types in AnyValue.

Probably you could wrap this in a [KeyValueList](https://github.com/open-telemetry/opamp-spec/blob/7667c08a93a8f0df671495e37fc64ae8ae212b72/proto/anyvalue.proto#L56) which has the exact same wire format and pass that to protojson.Marshal if the resulting format works.

internal/pkg/api/handleOpAMP.go
```diff
@@ -310,10 +378,145 @@ type localMetadata struct {
...
+			redactSensitive(val)
+		}
+	case []interface{}:
+		for i := range typed {
+			redactSensitive(typed[i])
+		}
+	}
+}
+
+func isSensitiveKey(key string) bool {
```
cmacknz: There should be unit tests of this, I don't see any.

internal/pkg/api/handleOpAMP.go
```diff
@@ -310,10 +378,145 @@ type localMetadata struct {
...
+	key = strings.ToLower(strings.TrimSpace(key))
+	if key == "" {
+		return false
+	}
+	for _, token := range []string{
```
cmacknz: https://github.com/elastic/elastic-agent/blob/1c3fb4b4c8989cd2cfb692780debd7619820ae72/internal/pkg/diagnostics/diagnostics.go#L454-L468

That's the list we use in agent diagnostics, probably they should be sync, and probably be in some shared implementation (elastic-agent-libs?).

internal/pkg/api/handleOpAMP.go
juliaElastic: This test covers redaction of `password` key: https://github.com/elastic/fleet-server/pull/6400/changes#diff-c07c5170813ea245fe626640afeaaca3307b0fdd96df454058b27814704afa60R240

internal/pkg/api/handleOpAMP.go
juliaElastic: thanks, I'll copy the same function for now and create a follow up issue to move to elastic-agent-libs

created: https://github.com/elastic/fleet-server/issues/6414

internal/pkg/api/handleOpAMP.go
juliaElastic: added handling of arrays and keyvalue list recursively

ycombinator (dismissed):
LGTM!

internal/pkg/api/handleOpAMP.go
```diff
@@ -200,13 +207,20 @@ func (oa *OpAMPT) handleMessage(zlog zerolog.Logger, apiKey *apikey.APIKey) func
...
 
-func (oa *OpAMPT) findEnrolledAgent(ctx context.Context, _ zerolog.Logger, agentID string) (*model.Agent, error) {
+func (oa *OpAMPT) findEnrolledAgent(ctx context.Context, zlog zerolog.Logger, agentID string) (*model.Agent, error) {
 	agent, err := dl.FindAgent(ctx, oa.bulk, dl.QueryAgentByID, dl.FieldID, agentID)
 	if errors.Is(err, dl.ErrNotFound) {
 		return nil, nil
 	}
 
+	// if agents index doesn't exist yet, it will be created when the first agent document is indexed
+	if err != nil && strings.Contains(err.Error(), "index_not_found_exception") {
```
michel-laterman: Can we compare to `es.ErrIndexNotFound` here instead?

internal/pkg/api/handleOpAMP.go
```diff
@@ -267,9 +299,17 @@ func (oa *OpAMPT) enrollAgent(zlog zerolog.Logger, agentID string, aToS *protobu
 		EnrolledAt: now.UTC().Format(time.RFC3339),
 		PolicyID:   rec.PolicyID,
 		Agent: &model.AgentMetadata{
-			ID: agentID,
+			ID:      agentID,
+			Version: meta.Elastic.Agent.Version,
+			Type:    agentType,
 		},
 		LocalMetadata: data,
+		// Setting revision to 1, the collector won't receive policy changes and 0 would keep the collector in updating state
```
michel-laterman, line 308: qq: the policy on fleet will never have a policy id > 1, if it can then agents can appear as outdated

juliaElastic: The UI hides the policy name and revision for collectors, as policies are not relevant for them (other than needing an enrollment token to authenticate with).

internal/pkg/api/handleOpAMP.go
juliaElastic: done

testing/e2e/stand_alone_test.go
```diff
@@ -697,14 +698,12 @@ func (suite *StandAloneSuite) TestOpAMP() {
...
-	// TODO: uncomment once https://github.com/elastic/fleet-server/pull/6400 is merged
-	// versionOut, err := exec.Command(otelBinaryPath, "--version").Output()
-	// suite.Require().NoError(err)
-	// otelVersion := strings.TrimPrefix(strings.TrimSpace(string(versionOut)), "otelcontribcol version ")
-	// suite.Equal("OPAMP", agentDoc.Type, "expected type to be OPAMP")
-	// suite.Equal("otelcontribcol", agentDoc.Agent.Type, "expected agent.type to be otelcontribcol")
-	// suite.Equal(otelVersion, agentDoc.Agent.Version, "expected agent.version to match otelcol-contrib binary version")
-	// suite.Equal(1, agentDoc.Revision, "expected policy_revision_idx to be 1")
-	// suite.Contains(agentDoc.Tags, "otelcontribcol", "expected tags to contain otelcontribcol")
-	// suite.Equal("online", agentDoc.Status, "expected status to be online")
```
juliaElastic, line 709: removed the status field check from the agent doc, it's calculated by a runtime field by Fleet UI

ycombinator (approved):

michel-laterman (approved):

ycombinator:
Backporting to `9.3` so I can proceed with merging https://github.com/elastic/fleet-server/pull/7024.

PR #6491, opened 2026-03-03
Fix checkin endpoint gzip handling

michel-laterman:

Checkin requests did not support "Content-Encoding: gzip" correctly.

Detect `Content-Encoding: gzip` header when validating checkin bodies and uncompress the body if detected.

## Design Checklist

## Checklist

### DIFF
--- a/changelog/fragments/1772577135-Fix-checkin-compression-support.yaml
+++ b/changelog/fragments/1772577135-Fix-checkin-compression-support.yaml
@@ -18,7 +18,6 @@ summary: Fix checkin endpoint compression support
 # NOTE: This field will be rendered only for breaking-change and known-issue kinds at the moment.
 description: |
   Adds support for gzip compressed requests to the checkin endpoint.
-  Properly detect accept-encoding headers for gzip responses on the checkin endpoint.
 
 # Affected component; usually one of "elastic-agent", "fleet-server", "filebeat", "metricbeat", "auditbeat", "all", etc.
 component: fleet-server


internal/pkg/api/handleCheckin.go
```diff
@@ -188,9 +188,20 @@ func (ct *CheckinT) validateRequest(zlog zerolog.Logger, w http.ResponseWriter,
 	}
 	readCounter := datacounter.NewReaderCounter(body)
 
+	// Decompress the body when the client signals Content-Encoding: gzip.
+	var bodyReader io.Reader = readCounter
+	if r.Header.Get("Content-Encoding") == kEncodingGzip {
+		gr, err := gzip.NewReader(readCounter)
```
cmacknz, line 194: Why does the stdlib not do this for us? Doing a quick search I see a block like the following in Go 1.26 in the read path:

```go
		resp.Body = body
		if rc.addedGzip && ascii.EqualFold(resp.Header.Get("Content-Encoding"), "gzip") {
			resp.Body = &gzipReader{body: body}
			resp.Header.Del("Content-Encoding")
			resp.Header.Del("Content-Length")
			resp.ContentLength = -1
			resp.Uncompressed = true
		}
```

internal/pkg/api/handleCheckin.go
```diff
@@ -677,10 +688,35 @@ func (ct *CheckinT) writeResponse(zlog zerolog.Logger, w http.ResponseWriter, r
...
+// RFC 7231 §5.3.4: multiple encodings may appear as comma-separated tokens
+// within a single header value, each optionally followed by a quality value
+// (e.g. "gzip;q=1.0, deflate;q=0.5").  A bare "*" token is treated as
+// matching any encoding.  A quality value of 0 ("gzip;q=0") means the
+// encoding is explicitly unwanted and is treated as a non-match.
 func acceptsEncoding(r *http.Request, encoding string) bool {
-	for _, v := range r.Header.Values("Accept-Encoding") {
-		if v == encoding {
-			return true
+	for _, headerVal := range r.Header.Values("Accept-Encoding") {
```
cmacknz: Do we need all this? The versions of this in the Go stdlib are a lot simpler, also we know agent doesn't use any of these technically allowed more complex formats.

internal/pkg/api/handleCheckin.go
michel-laterman: I thought that this only applies to HTTP clients
EDIT: It looks like it's linked to the http.Transport: https://pkg.go.dev/net/http#Transport which is only used by the client as far as I can tell

internal/pkg/api/handleCheckin.go
michel-laterman: good point

michel-laterman:
Some findings for validation of this:
- stdlib clients automatically add an `Accept-Encoding: gzip` header by default
- stdlib clients automatically handle responses that have a `Content-Encoding: gzip`
- the current behaviour of the server will server compress the checkin response if they are over 1kb uncompressed

Gathering data from the first couple of checkin an agent with the default policy (monitoring + systems integration) makes with a custom fleet-server build we can see the compressed resp savings, and **estimate** the request size savings to be:

| checkin number | req body size | resp body size | req compressed | resp compressed | notes |
|----------------|---------------|----------------|----------------|-----------------|-|
| 1 | 1409 | 14783 | 780 (-44.6%) | 3571 (-75.8%) | contains POLICY_CHANGE action |
| 2 | 4588 | 35 | 1155 (-74.8%) | 35 (no compression) | no returned actions |

Note that these numbers are for agents running today, without `effective_config` as a part of the checkin request body.

PR #6509, opened 2026-03-04
run go fix ./..., add as step in bk

michel-laterman:

Modernizes code by running `go fix ./...` with go v1.26
Add a `check:fix` target in the magefile, and add that as a target in `check:ci` that runs as a part of the buildkite pipeline


blakerouse (approved):
Looks good. Thanks for cleaning up the code to use the latest syntax.

PR #6517, opened 2026-03-06
Fix differences between openapi spec and implementation.

michel-laterman:

open api spec is missing some error responses.

Fix spec so that it more accurately describes the implantation. Fixes include  using status code 429 for too many requests (instead of 428) and status code 413 for request size too large (instead of 400).

Note that the only effective code change included is fixing a typo in error.go.

## Design Checklist

## Checklist

### DIFF
--- a/changelog/fragments/1772755112-Fix-innacuracies-with-openapi-spec.yaml
+++ b/changelog/fragments/1772755112-Fix-innacuracies-with-openapi-spec.yaml
@@ -11,7 +11,7 @@
 kind: bug-fix
 
 # Change summary; a 80ish characters long description of the change.
-summary: Fix innacuracies with openapi spec
+summary: Fix inaccuracies with openapi spec
 
 # Long description; in case the summary is not enough to describe the change
 # this field accommodate a description without length limits.


changelog/fragments/1772755112-Fix-innacuracies-with-openapi-spec.yaml
```diff
@@ -0,0 +1,32 @@
...
+# - enhancement: extends functionality but does not break or fix existing behavior
+# - feature: new functionality
+# - known-issue: problems that we are aware of in a given version
+# - security: impacts on the security of a product or a user’s deployment.
+# - upgrade: important information for someone upgrading from a prior version
+# - other: does not fit into any of the other categories
+kind: bug-fix
+
+# Change summary; a 80ish characters long description of the change.
+summary: Fix innacuracies with openapi spec
```
ycombinator:
```suggestion
summary: Fix inaccuracies with openapi spec
```

ycombinator: The same typo is in the changelog fragment filename too, but that's not as big of a deal since these get merged away eventually anyway.

ycombinator (dismissed):
Just a typo in the CHANGELOG.  Otherwise, LGTM.  Thanks for addressing these inconsistencies!

PR #6751, opened 2026-04-03
feat: add configurable limit on concurrent bulk dispatch goroutines

ycombinator:

During a 30k Serverless scale test, 22 of 39 fleet-server pods were OOMKilled. Analysis of the captured pod logs showed:

- An upgrade + policy reassignment storm caused checkin durations to spike from normal (~ms) to 5-45 seconds average.
- This created ~479 concurrent checkins per pod, each blocked in `dispatch()` waiting to enqueue onto the bulk engine's channel (capacity 32).
- Elasticsearch was **not** the bottleneck — ES had ~88% free heap after GC with sub-30ms pause times.
- Each blocked goroutine holds its stack (~4-8KB) plus the `bulkT` object. With no upper bound on concurrent dispatches, goroutines piled up until pods hit their memory limit (~154 Mi) and were killed.

This adds an optional cap on concurrent dispatch goroutines to bound memory usage.

The limit check runs at the top of `dispatch()`, before blocking on the channel: https://github.com/elastic/fleet-server/blob/1f456fb664ec564e2066ba0d0f5a28e0c552471c/internal/pkg/bulk/engine.go#L608-L622

When the limit is reached, the dispatch is rejected immediately with `ErrTooManyDispatches`, which maps to HTTP 429 so agents retry on their next checkin interval: https://github.com/elastic/fleet-server/blob/1f456fb664ec564e2066ba0d0f5a28e0c552471c/internal/pkg/api/error.go#L181-L189

The limit is configurable via `max_pending_dispatches` in the bulk config: https://github.com/elastic/fleet-server/blob/1f456fb664ec564e2066ba0d0f5a28e0c552471c/internal/pkg/config/input.go#L52

The default is 0 (no limit) so existing deployments are unaffected. Operators opt in by setting a value appropriate for their deployment size.

```
go test -race ./internal/pkg/bulk/ -run TestDispatch -v
```

## Design Checklist

## Checklist

## Related issues

- Relates #6747

### DIFF
--- a/internal/pkg/bulk/dispatch_limit_test.go
+++ b/internal/pkg/bulk/dispatch_limit_test.go
@@ -8,6 +8,8 @@ import (
 	"context"
 	"sync"
 	"testing"
+
+	"github.com/stretchr/testify/require"
 )
 
 func TestDispatchRejectsWhenLimitReached(t *testing.T) {
@@ -43,9 +45,7 @@ func TestDispatchRejectsWhenLimitReached(t *testing.T) {
 	blk.buf.WriteString(`{"index":"test"}`)
 	resp := b.dispatch(context.Background(), blk)
 
-	if resp.err != ErrTooManyBulkDispatches {
-		t.Fatalf("expected ErrTooManyBulkDispatches, got: %v", resp.err)
-	}
+	require.ErrorIs(t, resp.err, ErrTooManyBulkDispatches)
 
 	// Clean up: drain the channel to unblock the goroutines.
 	for i := 0; i < limit; i++ {
@@ -71,14 +71,10 @@ func TestDispatchAllowsWhenUnderLimit(t *testing.T) {
 	}()
 
 	resp := b.dispatch(context.Background(), blk)
-	if resp.err != nil {
-		t.Fatalf("expected no error, got: %v", resp.err)
-	}
+	require.NoError(t, resp.err)
 
 	// Counter should be back to 0 after dispatch completes.
-	if pending := b.pendingBulkDispatches.Load(); pending != 0 {
-		t.Fatalf("expected 0 pending bulk dispatches, got: %d", pending)
-	}
+	require.Equal(t, int64(0), b.pendingBulkDispatches.Load())
 }
 
 func TestDispatchNoLimitWhenZero(t *testing.T) {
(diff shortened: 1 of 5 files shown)


internal/pkg/bulk/dispatch_limit_test.go
```diff
@@ -0,0 +1,105 @@
...
+	}
+
+	// The next dispatch should be rejected immediately.
+	blk := b.newBlk(ActionSearch, optionsT{})
+	blk.buf.WriteString(`{"index":"test"}`)
+	resp := b.dispatch(context.Background(), blk)
+
+	if resp.err != ErrTooManyBulkDispatches {
+		t.Fatalf("expected ErrTooManyBulkDispatches, got: %v", resp.err)
+	}
```
michel-laterman: As the linter says, `errors.Is` should be used.
Or the `require` package for the error check

internal/pkg/bulk/dispatch_limit_test.go
```diff
@@ -0,0 +1,105 @@
...
+	// Simulate the Run loop responding.
+	go func() {
+		item := <-b.ch
+		item.ch <- respT{}
+	}()
+
+	resp := b.dispatch(context.Background(), blk)
+	if resp.err != nil {
+		t.Fatalf("expected no error, got: %v", resp.err)
+	}
```
michel-laterman: We should add something to `AGENTS.md` that gets claude to use the `require` package

internal/pkg/bulk/opt.go
```diff
@@ -81,15 +81,16 @@ func withAPMLinkedContext(ctx context.Context) Opt {
...
-	policyTokens      []config.PolicyToken
-	bi                build.Info
+	flushInterval            time.Duration
+	flushThresholdCnt        int
+	flushThresholdSz         int
+	maxPending               int
+	blockQueueSz             int
+	apikeyMaxParallel        int
+	apikeyMaxReqSize         int
+	maxPendingBulkDispatches int
```
michel-laterman: Shouldn't this be int64?

michel-laterman (changes requested):
We really should be consistent with the type of `maxPendingBulkDispatches`; either have it as an int64 in the implementation + config or just an int

Also the new tests don't follow our existing test structures with the use of the `require` test package

internal/pkg/bulk/dispatch_limit_test.go
ycombinator: Done in a separate PR: https://github.com/elastic/fleet-server/pull/6756

internal/pkg/bulk/opt.go
ycombinator: Fixed in b0f71eba.

internal/pkg/bulk/dispatch_limit_test.go
ycombinator: Fixed in 9548f868.

michel-laterman (dismissed):
lgtm, should we fix the linter warnings?

michel-laterman (approved):

swiatekm (approved):

PR #6829, opened 2026-04-13
fix: report server capabilities in OpAMP responses

michel-laterman:

The OpAMP spec requires the server to report its capabilities bitmask in the first `ServerToAgent` response. Fleet-server was not setting this field, so agents could not discover what the server supports.

Send the `serverCapabilities` bitmask (`AcceptsStatus`, `AcceptsEffectiveConfig`) in the `ServerToAgent` response when an agent enrolls or reconnects from a non-active status (offline, disconnected). Active agents (online, error, degraded) do not receive capabilities on every message to reduce response size. An `isActiveStatus` helper determines which statuses are considered active.

## Design Checklist

## Checklist

## Related issues

- Closes #6785

### DIFF
--- a/internal/pkg/api/handleOpAMP.go
+++ b/internal/pkg/api/handleOpAMP.go
@@ -36,7 +36,6 @@ import (
 
 const (
 	kOpAMPMod          = "opAMP"
-	statusDisconnected = "disconnected"
 	serverCapabilities = uint64(protobufs.ServerCapabilities_ServerCapabilities_AcceptsStatus |
 		protobufs.ServerCapabilities_ServerCapabilities_AcceptsEffectiveConfig)
 )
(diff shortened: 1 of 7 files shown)


internal/pkg/api/handleOpAMP.go
```diff
@@ -532,6 +539,12 @@ func ProtobufKVToRawMessage(zlog zerolog.Logger, kv []*protobufs.KeyValue) (json
 	return json.RawMessage(b), nil
 }
 
+func isActiveStatus(status string) bool {
+	return status == string(CheckinRequestStatusOnline) ||
+		status == string(CheckinRequestStatusError) ||
+		status == string(CheckinRequestStatusDegraded)
```
ycombinator, line 545: Nit: might be a bit more readable to use [`slices.Contains`](https://pkg.go.dev/slices#Contains), not 100% sure though.

ycombinator (approved):

internal/pkg/api/handleOpAMP.go
michel-laterman: we'll see if the go fix pr catches it: https://github.com/elastic/fleet-server/pull/6509

PR #6830, opened 2026-04-13
Decode all opamp agent capabilites

michel-laterman:

Decode all capabilites an opamp agent can send.

Use the upstream `protobufs.AgentCapabilities_name` as the source of truth to map a capability to a human-readable name.

## Design Checklist

## Checklist

## Related issues

- Closes https://github.com/elastic/fleet-server/issues/6790


ycombinator (dismissed):
LGTM, thanks for fixing.  Would be nice to have a unit test for `decodeCapabilities` but it's not super critical as the function is pretty straightforward.

ycombinator (approved):

PR #6831, opened 2026-04-13
fix: always set ReportFullState flag in OpAMP responses

### LINKED ISSUE #6783: [OpAMP] fleet-server should request a status report if drift is detected
The opamp spec defines a `ServerToAgent.flags.ReportFullState` message that a server can use to request that an agent responds with a full status report on the next message.

Fleet-server does not use this ability at the moment.
We should follow the spec and use this flag to get a full status report if the server detects drift.

This should occur if a sequence number is delivered out of order.

michel-laterman:

Fleet-server does not use the OpAMP spec's `ServerToAgent.flags.ReportFullState` flag. Fleet-server will now always set the flag.

Flag is set for every response, seq no drift is detected and logged as a warning.

## Design Checklist

## Checklist

## Related issues

- Closes #6783

### DIFF
--- a/docs/opamp.md
+++ b/docs/opamp.md
@@ -40,7 +40,7 @@ The following fields are ignored:
 - **Auto-enrollment.** The spec does not define enrollment. Fleet-server auto-enrolls unknown agents on first message using the enrollment API key's associated policy, creating a document in the `.fleet-agents` index with type `OPAMP`.
 - **Health-to-status mapping.** Fleet-server maps `ComponentHealth` to simplified statuses (`online`, `error`, `degraded`). The spec's nested `component_health_map` is not traversed; only the top-level health is used.
 - **Sensitive value redaction.** Fleet-server redacts keys containing `password`, `token`, `key`, `secret`, `auth`, `certificate`, or `passphrase` from the effective config before persisting.
-- **Always requests full state.** Fleet-server sets the `ReportFullState` flag in every `ServerToAgent` response, requesting the agent to include all status fields on every message.
+- **Always requests full state.** Fleet-server sets the `ReportFullState` flag in every `ServerToAgent` response, requesting the agent to include all status fields on every message. This is an intentional change intended to improve reliability by having constant load in order avoid irapid changes in workload.
 
 ### Throttling
 
--- a/docs/opamp.md
+++ b/docs/opamp.md
@@ -40,7 +40,7 @@ The following fields are ignored:
 - **Auto-enrollment.** The spec does not define enrollment. Fleet-server auto-enrolls unknown agents on first message using the enrollment API key's associated policy, creating a document in the `.fleet-agents` index with type `OPAMP`.
 - **Health-to-status mapping.** Fleet-server maps `ComponentHealth` to simplified statuses (`online`, `error`, `degraded`). The spec's nested `component_health_map` is not traversed; only the top-level health is used.
 - **Sensitive value redaction.** Fleet-server redacts keys containing `password`, `token`, `key`, `secret`, `auth`, `certificate`, or `passphrase` from the effective config before persisting.
-- **Always requests full state.** Fleet-server sets the `ReportFullState` flag in every `ServerToAgent` response, requesting the agent to include all status fields on every message. This is an intentional change intended to improve reliability by having constant load in order avoid irapid changes in workload.
+- **Always requests full state.** Fleet-server sets the `ReportFullState` flag in every `ServerToAgent` response, requesting the agent to include all status fields on every message. This is an intentional change intended to improve reliability by having constant load in order avoid rapid changes in workload.
 
 ### Throttling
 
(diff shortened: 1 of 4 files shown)


internal/pkg/api/handleOpAMP.go
```diff
@@ -547,6 +561,58 @@ func isActiveStatus(status string) bool {
...
+		if caps&c.cap != 0 && !c.field {
+			return false
+		}
+	}
+	return true
+}
+
+// shouldRequestFullState determines whether the server should set the
+// ReportFullState flag in its response. False positives are harmless (the agent
+// resends full state) and expected when the stored sequence number lags due to
```
cmacknz: >  False positives are harmless (the agent
// resends full state) and expected when the stored sequence number lags

What is the harm in always requesting the full state? This is pretty much what we do with the existing agent checkin. The system has to be able to tolerate getting the full state of every agent at once so why no always do that?

michel-laterman: The harm would just be extra network traffic; the spec assumes that we want to avoid being too noisy.

cmacknz: Yes traded off against the payload sizes being more dynamic. I mostly worry about scale and reliability with this, where we can work fine unless every agent restarts at once or a configuration change or something requires us to accept the full state.

AWS has a reliability principle specifically about this that I have always agreed with https://docs.aws.amazon.com/wellarchitected/latest/framework/rel_prevent_interaction_failure_constant_work.html

> Systems can fail when there are large, rapid changes in load. For example, if your workload is doing a health check that monitors the health of thousands of servers, it should send the same size payload (a full snapshot of the current state) each time. Whether no servers are failing, or all of them, the health check system is doing constant work with no large, rapid changes.
> 
> For example, if the health check system is monitoring 100,000 servers, the load on it is nominal under the normally light server failure rate. However, if a major event makes half of those servers unhealthy, then the health check system would be overwhelmed trying to update notification systems and communicate state to its clients. So instead the health check system should send the full snapshot of the current state each time. 100,000 server health states, each represented by a bit, would only be a 12.5-KB payload. Whether no servers are failing, or all of them are, the health check system is doing constant work, and large, rapid changes are not a threat to the system stability. This is actually how Amazon Route 53 handles health checks for endpoints (such as IP addresses) to determine how end users are routed to them.

There is an argument that the extra network traffic is an additional cost but I don't think it's ever come up as a problem for Fleet today. The payload body size limit being too low has. Any problem where agents fail to checkin reliably is always a high severity support case. The cons outweight the pros IMO for our circumstances right now.

michel-laterman: 👍 
I'll make the changes

internal/pkg/api/handleOpAMP.go
```diff
@@ -215,6 +217,14 @@ func (oa *OpAMPT) handleMessage(zlog zerolog.Logger, apiKey *apikey.APIKey) func
 			sendCapabilities = true
 		}
 
+		if !newlyEnrolled && message.SequenceNum != uint64(agent.SequenceNum)+1 {
+			zlog.Warn().
```
cmacknz: Can you make this a debug log? This isn't a user actionable log message AFAIK.

cmacknz (dismissed):

internal/pkg/api/handleOpAMP.go
cmacknz: LGTM, CC @ycombinator for the rationale here in case you have a different opinion.

IMO not reporting the full state in our architecture is more valuable between the opamp extension and whatever the supervisor is, we have a version of the same thing there with the config index in the control protocol.

ycombinator: No, I'm good with reporting full state always for now.  If this becomes a problem, we can always re-evaluate at the time, as the spec gives us that flexibility.

docs/opamp.md
```diff
@@ -41,6 +40,7 @@ The following fields are ignored:
 - **Auto-enrollment.** The spec does not define enrollment. Fleet-server auto-enrolls unknown agents on first message using the enrollment API key's associated policy, creating a document in the `.fleet-agents` index with type `OPAMP`.
 - **Health-to-status mapping.** Fleet-server maps `ComponentHealth` to simplified statuses (`online`, `error`, `degraded`). The spec's nested `component_health_map` is not traversed; only the top-level health is used.
 - **Sensitive value redaction.** Fleet-server redacts keys containing `password`, `token`, `key`, `secret`, `auth`, `certificate`, or `passphrase` from the effective config before persisting.
+- **Always requests full state.** Fleet-server sets the `ReportFullState` flag in every `ServerToAgent` response, requesting the agent to include all status fields on every message.
```
ycombinator: Might be good to include the rationale here, as discussed in https://github.com/elastic/fleet-server/pull/6831#discussion_r3081936491.  I think it would be good to do that, in general, for any areas where we _intentionally_ deviate from the spec.

docs/opamp.md
```diff
@@ -41,6 +40,7 @@ The following fields are ignored:
 - **Auto-enrollment.** The spec does not define enrollment. Fleet-server auto-enrolls unknown agents on first message using the enrollment API key's associated policy, creating a document in the `.fleet-agents` index with type `OPAMP`.
 - **Health-to-status mapping.** Fleet-server maps `ComponentHealth` to simplified statuses (`online`, `error`, `degraded`). The spec's nested `component_health_map` is not traversed; only the top-level health is used.
 - **Sensitive value redaction.** Fleet-server redacts keys containing `password`, `token`, `key`, `secret`, `auth`, `certificate`, or `passphrase` from the effective config before persisting.
+- **Always requests full state.** Fleet-server sets the `ReportFullState` flag in every `ServerToAgent` response, requesting the agent to include all status fields on every message. This is an intentional change intended to improve reliability by having constant load in order avoid irapid changes in workload.
```
ycombinator:
```suggestion
- **Always requests full state.** Fleet-server sets the `ReportFullState` flag in every `ServerToAgent` response, requesting the agent to include all status fields on every message. This is an intentional change intended to improve reliability by having constant load in order avoid rapid changes in workload.
```

ycombinator (approved):

swiatekm (approved):

cmacknz (approved):

PR #6834, opened 2026-04-14
feat: respond to OpAMP RequestInstanceUid flag with agent_identification for newly enrolled agents

michel-laterman:

Fleet-server ignores the `AgentToServer.flags.RequestInstanceUid` flag and does not respond with `ServerToAgent.agent_identification`, which is required by the [OpAMP spec](https://github.com/open-telemetry/opamp-spec/blob/main/specification.md).

When an agent sets the `RequestInstanceUid` flag, fleet-server treats the message as a new enrollment regardless of whether an agent with the incoming `instance_uid` already exists:

1. `handleMessage` checks the flag early and, when set, skips the `findEnrolledAgent` lookup entirely.
2. `enrollAgent` generates a fresh UUID v7 and uses it as the agent's ID; the new agent document is stored in `.fleet-agents` under that UID.
3. The response includes `AgentIdentification.NewInstanceUid` with the generated UID. The agent is expected to adopt the new UID for subsequent messages per the spec.
4. The response's `InstanceUid` still echoes the incoming `instance_uid` per the spec.

If `agent_disconnect` is set together with `RequestInstanceUid` (or for an unenrolled agent), fleet-server returns a `BadRequest` error response.

Reassigning an already-enrolled agent's instance UID without re-enrollment is intentionally out of scope for this change. Doing so would require decoupling the `.fleet-agents` document `_id` from `instance_uid`, which today are the same value (consistent with the rest of fleet-server).

1. Configure an OpenTelemetry Collector to connect to fleet-server over OpAMP (see `docs/opamp.md`).
2. Send an `AgentToServer` message with `flags` set to `1` (`RequestInstanceUid`). The response should include `agent_identification.new_instance_uid` with a valid UUID v7 distinct from the incoming `instance_uid`, and a new document in `.fleet-agents` should exist under the generated UID.
3. Confirm a message without the flag continues to behave as before (existing agents check in normally; unknown agents auto-enroll under the incoming UID).

## Design Checklist

## Checklist

## Related issues

- Closes #6789


internal/pkg/api/handleOpAMP.go
```diff
@@ -260,6 +295,55 @@ func (oa *OpAMPT) findEnrolledAgent(ctx context.Context, zlog zerolog.Logger, ag
...
+
+	if _, err := oa.bulk.Create(ctx, dl.FleetAgents, newID, data, bulk.WithRefresh()); err != nil {
+		return fmt.Errorf("failed to create agent with new ID: %w", err)
+	}
+
+	// Delete the old agent document. Log a warning on failure since the new
+	// document was already created successfully.
+	if err := oa.bulk.Delete(ctx, dl.FleetAgents, oldID); err != nil {
+		zlog.Warn().Err(err).Str("old_id", oldID).Str("new_id", newID).Msg("failed to delete old agent document after ID reassignment")
+	}
```
michel-laterman, line 384: I'm not sure this is the approach we should take when an existing agent requests a new UUID.

Right now our ES docs associate the opamp instance_uid attribute as `_id`, which is the same approach we take for notmal fleet instances.
the instance UID is also repeated in local_metadata, and as agent.id.
As an alternative we could seperate the doc ID value from the instance_uid value for OpAMP

cmacknz: > As an alternative we could seperate the doc ID value from the instance_uid value for OpAMP

This feels more correct, but if instance UID changes because the end system has no persistent state does it make any functional difference? Monitored collectors can change instance UID at each restart so this could be a source of churn in the system, but we'd need a way to associate collectors doing this to more stable identities reliably.

michel-laterman: >  but if instance UID changes because the end system has no persistent state does it make any functional difference?

For the current monitoring use case we're implementing, not really. This only becomes an issue when agents have persistent storage/we want to support managing agents.

cmacknz: Yeah I think separating instance_uid from agent ID (or .fleet-agents doc ID) feels nicer but I can't think of a way to do it.

If/when we start provisioning per agent API keys this would become more important, but we don't do that yet.

michel-laterman: I've limited this PR to only generate a new instance_uid if the flag is set for new enrollments

internal/pkg/api/handleOpAMP.go
```diff
@@ -200,12 +201,32 @@ func (oa *OpAMPT) handleMessage(zlog zerolog.Logger, apiKey *apikey.APIKey) func
...
+				zlog.Error().Err(err).Msg("failed to generate new instance UID")
+				return &protobufs.ServerToAgent{
+					InstanceUid: instanceUID.Bytes(),
+					ErrorResponse: &protobufs.ServerErrorResponse{
+						Type:         protobufs.ServerErrorResponseType_ServerErrorResponseType_Unavailable,
+						ErrorMessage: "failed to generate new instance UID",
+					},
+				}
+			}
+			newInstanceUID = &uid
```
ycombinator: Could we set `instanceUID` here to `&uid`? I think it might simplify some of the code later.

internal/pkg/api/handleOpAMP.go
```diff
@@ -272,6 +314,17 @@ func (oa *OpAMPT) findEnrolledAgent(ctx context.Context, zlog zerolog.Logger, ag
 	return &agent, nil
 }
 
+// reassignAgentID functions as a special checkin that associates an existing agent with a new ID.
+//
+// Currently will return an error as reassiging an enrolled agent is not supported.
```
ycombinator:
```suggestion
// Currently will return an error as reassigning an enrolled agent is not supported.
```

internal/pkg/api/handleOpAMP.go
michel-laterman: As per the spec, the response's instance_uid value must match the request's even if the request has the new instance_uid flag set

internal/pkg/api/handleOpAMP.go
```diff
@@ -149,102 +150,106 @@ func (oa *OpAMPT) Enabled() bool {
...
-		zlog.Debug().
-			Msg("received AgentToServer message from agent")
-
-		// Check if Agent is "enrolled"; if it is, update it; otherwise, enroll it.
-		agent, err := oa.findEnrolledAgent(ctx, zlog, instanceUID.String())
-		if err != nil {
-			return &protobufs.ServerToAgent{
-				InstanceUid: instanceUID.Bytes(),
-				ErrorResponse: &protobufs.ServerErrorResponse{
+		enrollRequest := message.Flags&uint64(protobufs.AgentToServerFlags_AgentToServerFlags_RequestInstanceUid) != 0
```
cmacknz, line 174: Do we need to rate limit this separate from the limiter that is already on the /opamp endpoint? There is now the possibility that an agent continuously requests that it be re-enrolled.

Perhaps you could look at the timestamp of the last enrollment in the agent document? You wouldn't be skipping the `findEnrolledAgent` call though, but if it's a document lookup by ID it is reasonably cheap.

cmacknz: I would bet that making the `findEnrollAgent` call has less performance impact on the system than misbehaving collector(s) continuously creating new agent documents unnecessarily.

cmacknz: If we are making the `findEnrollAgent` call anyway, we could additionally delete the old agent document to help this. Then maybe you don't need additional rate limiting because you aren't creating useless agents anymore.

ycombinator (dismissed):

internal/pkg/api/handleOpAMP.go
michel-laterman: We could try to rate limit opamp enrolls; I don't think it makes too much sense to tie it to existing agent IDs, just have a general enrollment rate limit instead. Currently the benefit would be to reduce the amount of new agent docs that are created by a misbehaving collector, in the future it would also limit the collector specific API keys that are issued

I'm hesitant to delete existing agents automatically as it would allow anyone with the enrollment key to delete agent docs. We could use the `audit_unenrolled_*` attributes (and add a new enum to descibe the state) to annotate existing documents instead.

cmacknz: >  I don't think it makes too much sense to tie it to existing agent IDs, just have a general enrollment rate limit instead

👍 

> I'm hesitant to delete existing agents automatically as it would allow anyone with the enrollment key to delete agent docs. We could use the audit_unenrolled_* attributes (and add a new enum to descibe the state) to annotate existing documents instead

👍

internal/pkg/api/handleOpAMP.go
```diff
@@ -584,3 +633,18 @@ func decodeCapabilities(caps uint64) []string {
...
 	return result
 }
+
+func (oa *OpAMPT) unenrollAgent(ctx context.Context, zlog zerolog.Logger, agentID string) error {
+	ts := time.Now().UTC().Format(time.RFC3339)
+	zlog.Debug().Msg("Marking agent as re-enrolled")
+	doc := bulk.UpdateFields{
+		dl.FieldUpdatedAt:             ts,
+		dl.FieldAuditUnenrolledTime:   ts,
+		dl.FieldAuditUnenrolledReason: reenrolled,
```
cmacknz, line 640: Do you need the logic to clear these fields if the same instance UID connects again (without requesting a new instance ID when it does) that we have in the checkin endpoint?

internal/pkg/api/handleOpAMP_test.go
```diff
@@ -220,6 +223,48 @@ func TestEnrollAgentWithAgentToServerMessage(t *testing.T) {
 	bulker.AssertExpectations(t)
 }
 
+// TestEnrollAgentGeneratesNewUIDWhenFlagSet verifies enrollAgent generates a
+// fresh UUID when the AgentToServer message sets the RequestInstanceUid flag,
+// ignoring the incoming uid argument.
+func TestEnrollAgentGeneratesNewUIDWhenFlagSet(t *testing.T) {
```
cmacknz, line 277: Is it worth testing the rate limiter? You'd probably have to configure one that was easier to test e.g. allowing only one request or similar.

cmacknz (approved):

PR #6835, opened 2026-04-14
fix: correct typos in default rate limit YAML keys

michel-laterman:

All 6 embedded default YAML files under `internal/pkg/config/defaults/` used `pgp_retieval_limit` (missing 'r') instead of `pgp_retrieval_limit`. Because the Go struct tag is `config:"pgp_retrieval_limit"`, the YAML values never deserialized — PGP retrieval rate limits silently fell back to hardcoded defaults regardless of agent-count tier.

Additionally, `lte10000_limits.yml` used `policy_interval` instead of `policy_limit`, causing the same silent fallback for the policy rate limit at the 5001-10000 agent tier.

- Fixes `pgp_retieval_limit` → `pgp_retrieval_limit` in all 6 default YAML files
- Fixes `policy_interval` → `policy_limit` in `lte10000_limits.yml`
- Renames `lte2500_limite.yml` → `lte2500_limits.yml` for consistency
- Adds `TestDefaultLimitsYAMLKeys` regression test that verifies all embedded YAML files populate every limit field with non-zero values, catching future key/struct-tag mismatches

```
go test ./internal/pkg/config/ -run "TestDefaultLimitsYAMLKeys" -v
```

## Design Checklist

## Checklist


ebeahan (commented):
Are these config values only exposed internally?

I agree that these changes are a bugfix, but could any of these changes be "breaking" in the sense they are changing the current default behavior?

michel-laterman:
@ebeahan, our defaults are **much** lower than what the fix would load

ycombinator (commented):
Thanks for catching and fixing both typos.  Config keys are now properly aligned with the struct tags in `config.ServerLimits`.

Thanks for adding the test too.  It would be nice to make this test more generic, such that it iterates over all the `server_limits` keys in each YAML file and, for each key, requires that there is a corresponding struct field with that struct tag.  I'll push up a commit for that.

ycombinator:
> It would be nice to make this test more generic, such that it iterates over all the `server_limits` keys in each YAML file and, for each key, requires that there is a corresponding struct field with that struct tag. I'll push up a commit for that.

I made the test more generic using reflection, in 5ad7478759892566598b968449cc4a2a35fec3b5.  This way we don't have to enumerate all the struct fields in the assertions.  But if you prefer the original implementation, feel free to revert/remove my commit.

ycombinator (dismissed):

internal/pkg/config/env_defaults_test.go
```diff
@@ -39,3 +43,46 @@ func TestLoadLimits(t *testing.T) {
...
+		t.Run(name, func(t *testing.T) {
+			f, err := defaultsFS.Open(path)
+			require.NoError(t, err)
+			data, err := io.ReadAll(f)
+			require.NoError(t, err)
+
+			var raw map[string]any
+			require.NoError(t, goyaml.Unmarshal(data, &raw))
+
+			require.Containsf(raw, "server_limits", "%s does not contain server_limits attribute", path)
```
ycombinator:
```suggestion
			require.Containsf(t, raw, "server_limits", "%s does not contain server_limits attribute", path)
```

ycombinator (approved):

PR #6838, opened 2026-04-15
Allow Fleet Server to reload TLS certificates without restarting

### LINKED ISSUE #6433: Allow Fleet Server to reload TLS certificates without restarting
**Describe the enhancement:**

Fleet Server requires TLS certificates to serve HTTPS requests from Elastic Agent. These certificates are placed in external files rather than embedded in the configuration, so if they change, Fleet Server doesn't reload the configuration. Rotating certificates requires an explicit restart, as per the [documentation](https://www.elastic.co/docs/reference/fleet/certificates-rotation).

We should add an option allowing the certificate files to be reloaded without the need to restart Fleet Server. It should start as disabled by default to avoid user surprise, but we should work towards enabling it by default eventually.

In many environments, synchronizing certificate renewal and Fleet Server restarts is quite onerous. In our own Serverless environments, certificates are managed through lightweight-cert-manager, and each project has its own certificates with their own expiration date. The operational toil of this problem would be completely removed by just letting the K8s Secret containing the certificate be remounted in the Pod and Fleet Server reloading it automatically.

In general, this feature would reduce operational burden in most environments, removing the need for external orchestration of certificate rotation.

ycombinator:

Fleet Server currently requires an explicit restart to reload TLS certificates used for serving HTTPS requests to Elastic Agents. In environments with frequent certificate rotation — Kubernetes (where Secrets are remounted into pods), serverless (where cert-manager manages per-project certificates), and general ops — this creates operational burden and requires external orchestration to coordinate cert rotation with server restarts.

When TLS is enabled, Fleet Server now periodically re-reads the configured cert/key files from disk and hot-swaps them into the running TLS listener without restart. This is enabled by default and can be disabled by setting `ssl.certificate_reload.enabled: false`.

This follows the [OTel Collector's `configtls` approach](https://github.com/open-telemetry/opentelemetry-collector/blob/7201768d04fb628e65cb4e3d5ead69d17c43d59d/config/configtls/configtls.go#L152): on each TLS handshake, if the configured reload interval has elapsed, the cert/key files are re-read from disk and validated before being swapped in. No file watchers, no extra goroutines.

The core `CertReloader` implementation and the `CertificateReload` configuration both live in [`elastic-agent-libs` v0.37.0](https://github.com/elastic/elastic-agent-libs) (in the `tlscommon` package), as part of `ServerConfig`, so any component using `tlscommon` gets cert reload support. Fleet Server simply wires it up.

Key design decisions:
- **Polling instead of file watching**: Inspired by OTel Collector's approach. Simpler, no `fsnotify` dependency, no goroutines, no debounce logic. On each `GetCertificate` call, checks if the reload interval has elapsed; if so, re-reads files from disk
- **Enabled by default**: Certificate reload is on unless explicitly disabled, matching the OTel Collector pattern
- **Reload interval**: Defaults to 5 seconds, configurable via `ssl.certificate_reload.reload_interval`. Automated tooling (cert-manager, K8s secret mounts) can use the default or a shorter value, while manual rotation workflows may benefit from a longer window
- **Validation**: Validates the new cert/key pair with `tls.LoadX509KeyPair` before swapping; invalid pairs keep the old cert active and log an error
- **Serving**: Uses `tls.Config.GetCertificate` callback with `sync.RWMutex` and double-checked locking for concurrent reads on every TLS handshake
- **Existing connections**: Unaffected — only new TLS handshakes see the new cert

Modified files:
- `internal/pkg/api/server.go` — wires `CertReloader` into TLS setup when feature is enabled
- `internal/pkg/api/server_test.go` — adds end-to-end cert reload test
- `fleet-server.reference.yml` — documents the new config option

**1. Build Fleet Server**

```bash
mage build:local
```

**2. Generate a CA and server cert/key pair**

```bash
mkdir -p /tmp/tls-reload-test
cd /tmp/tls-reload-test

# Generate CA
openssl req -x509 -newkey rsa:2048 -keyout ca-key.pem -out ca.pem -days 1 -nodes -subj "/CN=Test CA"

# Generate server cert signed by the CA
openssl req -newkey rsa:2048 -keyout server-key.pem -out server.csr -nodes \
  -subj "/CN=localhost" -addext "subjectAltName=DNS:localhost,IP:127.0.0.1"
openssl x509 -req -in server.csr -CA ca.pem -CAkey ca-key.pem -CAcreateserial \
  -out server-cert.pem -days 1 -copy_extensions copyall
```

**3. Create a config file**

```bash
cat > /tmp/tls-reload-test/fleet-server.yml << 'CONF'
output:
  elasticsearch:
    hosts: ['localhost:9200']
    service_token: 'fake-token'

fleet:
  agent:
    id: test-agent-id

inputs:
  - type: fleet-server
    server:
      host: localhost
      port: 8220
      ssl:
        enabled: true
        certificate: /tmp/tls-reload-test/server-cert.pem
        key: /tmp/tls-reload-test/server-key.pem
CONF
```

**4. Start Fleet Server**

```bash
bin/fleet-server -c /tmp/tls-reload-test/fleet-server.yml \
  -E logging.to_files=true \
  -E logging.to_stderr=false \
  -E logging.files.path=/tmp/tls-reload-test \
  -E logging.files.name=fleet-server.log &
```

Wait for the `"Listening on localhost:8220"` log line:

```bash
grep "Listening" /tmp/tls-reload-test/fleet-server.log-*.ndjson
```

**5. Make a TLS connection and note the server cert fingerprint**

```bash
openssl s_client -connect localhost:8220 -CAfile /tmp/tls-reload-test/ca.pem \
  -servername localhost </dev/null 2>&1 | openssl x509 -noout -serial -fingerprint
```

**6. Replace the cert/key files with a new pair (same CA)**

```bash
# Generate the new cert/key outside the watched directory
openssl req -newkey rsa:2048 -keyout /tmp/server-key2.pem -out /tmp/server2.csr -nodes \
  -subj "/CN=localhost" -addext "subjectAltName=DNS:localhost,IP:127.0.0.1"
openssl x509 -req -in /tmp/server2.csr -CA /tmp/tls-reload-test/ca.pem \
  -CAkey /tmp/tls-reload-test/ca-key.pem -CAcreateserial \
  -out /tmp/server-cert2.pem -days 1 -copy_extensions copyall

# Move the new files into the watched directory
mv /tmp/server-cert2.pem /tmp/tls-reload-test/server-cert.pem
mv /tmp/server-key2.pem /tmp/tls-reload-test/server-key.pem
```

**7. Wait ~5 seconds for the reload interval, then make another TLS connection**

```bash
openssl s_client -connect localhost:8220 -CAfile /tmp/tls-reload-test/ca.pem \
  -servername localhost </dev/null 2>&1 | openssl x509 -noout -serial -fingerprint
```

The fingerprint should differ from step 5.

**8. Stop Fleet Server**

```bash
kill %1
```

---

Automated tests:
- `go test -run Test_server_TLSCertReload ./internal/pkg/api/...` — end-to-end integration test

## Design Checklist

## Checklist

## Related issues

- Closes #6433


blakerouse (commented):
How would you feel about changing this implementation? We talked about this at a team sync meeting and I was supposed to make an issue for it, but honestly I forgot until I was reading this PR.

https://github.com/open-telemetry/opentelemetry-collector/blob/7201768d04fb628e65cb4e3d5ead69d17c43d59d/config/configtls/configtls.go#L152

That design of config reload is something we discussed in the call. We overall liked the simple design no need for fsnotify or anything. We discussed about adding it to elastic-agent-libs so it could be used in multiple places.

Maybe we could do the implementation here first in Fleet Server and then split it out later.

ycombinator:
@blakerouse Sure, I like the simplicity of the OTel implementation.  It means having to wait up to `ReloadInterval` for the new cert to be picked up, but I think that's an acceptable tradeoff for the simpler implementation.  I'll implement it in `elastic-agent-libs` and then update this PR here to use that implementation.

go.mod
```diff
@@ -103,3 +103,5 @@ require (
 	gopkg.in/yaml.v2 v2.4.0 // indirect
 	howett.net/plist v1.0.1 // indirect
 )
+
+replace github.com/elastic/elastic-agent-libs => github.com/ycombinator/elastic-agent-libs v0.0.0-20260417224603-74c467d3bcab
```
ycombinator, line 107: Will need to be removed once https://github.com/elastic/elastic-agent-libs/pull/404 is merged and a new version of that module is released.

ycombinator:
@blakerouse Check out https://github.com/elastic/elastic-agent-libs/pull/404.  Also, I've updated this PR here to use `CertReloader` from that PR.

internal/pkg/config/input.go
```diff
@@ -59,26 +60,55 @@ func (c *ServerBulk) InitDefaults() {
 	c.FlushMaxPending = 8
 }
 
+// CertificateReload is the configuration for hot-reloading TLS certificates.
+type CertificateReload struct {
+	Enabled        bool          `config:"enabled"`
```
blakerouse, line 65: Should this just not be default? I feel like this just working out of the box, is better than requiring to turn it on. Feel like its more likely that you want it on than off. Looking at configtls from OTEL doesn't look like it can be turned off, just the interval adjusted.

internal/pkg/config/input.go
```diff
@@ -59,26 +60,55 @@ func (c *ServerBulk) InitDefaults() {
...
+	ReloadInterval time.Duration `config:"reload_interval"`
+}
+
+// ServerTLSConfig wraps tlscommon.ServerConfig with additional Fleet Server
+// specific TLS settings like certificate hot-reload.
+type ServerTLSConfig struct {
+	tlscommon.ServerConfig
+	CertificateReload CertificateReload `config:"certificate_reload"`
```
blakerouse, line 73: Should this just be moved into `tlscommon.ServerConfig`? Being that its now in elastic-agent-libs it would be great to just have this always present.

internal/pkg/config/input.go
ycombinator: I wasn't sure if we wanted to change the existing behavior or not.  If we're OK with that, we should enable reload by default.  WDYT?

internal/pkg/config/input.go
ycombinator: Yup, done: https://github.com/elastic/elastic-agent-libs/pull/405

blakerouse (dismissed):
Nice! Looks good.

internal/pkg/api/server_test.go
```diff
@@ -396,3 +400,148 @@ key: %s`,
...
+	case <-started:
+	case <-time.After(500 * time.Millisecond):
+		require.Fail(t, "timed out waiting for server to start")
+	}
+	select {
+	case err := <-errCh:
+		require.NoError(t, err, "error during startup")
+	case <-time.After(500 * time.Millisecond):
+		break
+	}
```
swiatekm: Is there a way for us to avoid always waiting for 500ms here?

swiatekm (dismissed):
LGTM, one non-blocking question.

internal/pkg/api/server_test.go
ycombinator: Yeah, we could instead use a `require.Eventually` that calls the `/api/status` endpoint on the server to see if it's actually alive.  Done in a6c9f083e928022de37d11842c98068edbbd81d9.

swiatekm (dismissed):

ycombinator:
The E2E test failures (`TestStaticTokenAuthentication`, `TestWithSecretFiles`, `TestStandAloneRunningSuite`) are caused by `CertReloader` using `tls.LoadX509KeyPair`, which cannot handle passphrase-encrypted private keys. These tests configure fleet-server with `key_passphrase_path`, so the server fails at startup with `tls: failed to parse private key`.

Fix: https://github.com/elastic/elastic-agent-libs/pull/412 adds `WithPassphrase` support to `CertReloader`. This PR now passes the resolved passphrase through and temporarily uses a `go.mod` replace pointing to that fork commit.

go.mod
```diff
@@ -103,3 +103,5 @@ require (
 	gopkg.in/yaml.v2 v2.4.0 // indirect
 	howett.net/plist v1.0.1 // indirect
 )
+
+replace github.com/elastic/elastic-agent-libs => github.com/ycombinator/elastic-agent-libs v0.0.0-20260427164627-e9c901a6c5a7
```
ycombinator: Remove this `replace` directive once https://github.com/elastic/elastic-agent-libs/pull/412 is merged and released.

swiatekm (approved):

PR #6864, opened 2026-04-20
Updated validate to fall back to Elastic-Agent-Version header'

michalpristas:
related to https://github.com/elastic/elastic-agent/pull/13695

when uncompatible user agent comes it tries to fallback to version from `Elastic-Agent-Version` header if present.

### DIFF
--- a/internal/pkg/api/handleCheckin.go
+++ b/internal/pkg/api/handleCheckin.go
@@ -144,7 +144,7 @@ func (ct *CheckinT) handleCheckin(zlog zerolog.Logger, w http.ResponseWriter, r 
 	ctx := zlog.WithContext(r.Context())
 	r = r.WithContext(ctx)
 
-	ver, err := validateUserAgent(r.Context(), zlog, userAgent, ct.verCon)
+	ver, err := validateUserAgent(r.Context(), zlog, userAgent, r.Header.Get("Elastic-Agent-Version"), ct.verCon)
 	if err != nil {
 		return err
 	}
--- a/internal/pkg/api/userAgent.go
+++ b/internal/pkg/api/userAgent.go
@@ -47,7 +47,7 @@ func maximizePatch(ver *version.Version) string {
 
 // validateUserAgent validates that the User-Agent of the connecting Elastic Agent is valid and that the version is
 // supported for this Fleet Server.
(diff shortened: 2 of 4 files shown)


blakerouse (approved):
lgtm!

cmacknz:
If this is for agentless, you will need the ECH integrations servers updated before you can rely on this being available which might taken longer than you want.

What problem is this solving? Why is the user agent not compatible?

cmacknz:
Discussion happening in https://github.com/elastic/elastic-agent/pull/13695

internal/pkg/api/userAgent.go
```diff
@@ -77,22 +78,37 @@ func validateUserAgent(ctx context.Context, zlog zerolog.Logger, userAgent strin
...
+	if err == nil {
+		return ver, nil
+	}
+
+	ver, err = version.NewVersion(elasticAgentVersion)
+	if err == nil {
+		return ver, nil
+	}
+
+	return nil, ErrInvalidUserAgent
```
ycombinator, line 113: `ErrInvalidUserAgent` seems overly broad for what this function is checking.  Also, that error is being returned from the error checking being done at this function's call site.  So maybe this function should just return the raw error at the end?
```suggestion
    // Fallback
	return version.NewVersion(elasticAgentVersion)
```

internal/pkg/api/handleCheckin.go
```diff
@@ -144,7 +144,7 @@ func (ct *CheckinT) handleCheckin(zlog zerolog.Logger, w http.ResponseWriter, r
 	ctx := zlog.WithContext(r.Context())
 	r = r.WithContext(ctx)
 
-	ver, err := validateUserAgent(r.Context(), zlog, userAgent, ct.verCon)
+	ver, err := validateUserAgent(r.Context(), zlog, userAgent, r.Header.Get("Elastic-Agent-Version"), ct.verCon)
```
michel-laterman, line 147: Can we add this into the openapi spec?

ebeahan:
@michalpristas checking in on this PR - are you waiting on additional reviews? Or are we waiting on another dependency to move this forward?

michalpristas:
no i'm thinking whether to ship this or not, the underlying issue was really about integrations, but this may be handy but is not required

swiatekm (approved):

PR #6872, opened 2026-04-21
feat(opamp): compute and store effective_config_hash for collector pipelines

juliaElastic:
## Summary

- Computes a SHA-256 hash of the OTel collector's  effective config and stores it as `effective_config_hash` in the `.fleet-agents` index.
- Keys are canonicalized via `yaml.v3` Marshal (sorts alphabetically) before hashing, so key order never affects the output.
- Hash is computed from the raw YAML body (before the sensitive-value redaction pass that produces `effective_config`).

Closes: https://github.com/elastic/ingest-dev/issues/7064

To verify:
- Start ES, Kibana, fleet-server locally https://github.com/elastic/fleet-server/blob/main/docs/opamp.md#setup
- Go to Fleet UI, click on Add collector
- Copy or download OTel config with OpAMP
- Start EDOT or otel-contrib collector with the generated config
- Verify that the `effective_config_hash` is stored in `.fleet-agents`.

```
GET .fleet-agents/_search
{
    "_source": ["effective_config_hash", "effective_config"]
}

    "hits": [
      {
        "_index": ".fleet-agents-7",
        "_id": "95d87b10-299e-45ce-a5dd-72e500451e48",
        "_score": 1,
        "_source": {
          "effective_config": {
           ...
          },
          "effective_config_hash": "4e13e7566f2c31e563e235dd8e1a7ae8998a6677fae6124806ca842904f6aa0e"
        }
      },
```

🤖 Generated with [Claude Code](https://claude.ai/claude-code)


juliaElastic:
CI fails due to an environmental issue, not related to changes in this PR:

The job, [Package x86_64](https://buildkite.com/elastic/fleet-server/builds/14254#019db4d0-b1a3-40e3-b60f-12c6dd24bdb8), has been canceled as it failed to get an agent after 5 tries.

The job, [Package x86_64 FIPS](https://buildkite.com/elastic/fleet-server/builds/14254#019db4d0-b1a3-453e-a366-9ecfa63831c8), has been canceled as it failed to get an agent after 5 tries.

ebeahan:
> CI fails due to an environmental issue, not related to changes in this PR

https://github.com/elastic/fleet-server/pull/6881 should address the CI issues.

internal/pkg/api/configHash.go
```diff
@@ -0,0 +1,65 @@
...
+// OpAMP effective config. Only receivers, processors, exporters, connectors,
+// service.pipelines, and service.extensions are included. Keys are sorted
+// deterministically (yaml.v3 sorts map keys on Marshal) so identical topologies
+// always produce the same hash regardless of key order or non-topology fields.
+// Returns "" with no error when the config is nil or the body is empty.
+func HashEffectiveConfig(effectiveConfig *protobufs.EffectiveConfig) (string, error) {
+	if effectiveConfig.ConfigMap == nil || effectiveConfig.ConfigMap.ConfigMap[""] == nil {
+		return "", nil
+	}
+	body := effectiveConfig.ConfigMap.ConfigMap[""].Body
```
michel-laterman: In theory we should hash everything in the configmap, not just the default config

internal/pkg/api/configHash.go
```diff
@@ -0,0 +1,65 @@
...
+}
+
+func hashConfigBody(body []byte) (string, error) {
+	var full map[string]any
+	if err := yaml.Unmarshal(body, &full); err != nil {
+		return "", fmt.Errorf("unmarshal config for hashing: %w", err)
+	}
+
+	topology := make(map[string]any)
+	for _, k := range []string{"receivers", "processors", "exporters", "connectors"} {
```
michel-laterman: Why do we need to only use these keys when computing the sha?
How do we ensure that a change to another key is emitted to a collector?

internal/pkg/api/configLabel_test.go
```diff
@@ -0,0 +1,80 @@
...
+
+	parts := strings.SplitN(label, "-", 2)
+	require.Len(t, parts, 2, "label must be adjective-noun separated by a hyphen")
+	assert.NotEmpty(t, parts[0], "adjective must not be empty")
+	assert.NotEmpty(t, parts[1], "noun must not be empty")
+}
+
+func TestLabelFromHash_Deterministic(t *testing.T) {
+	hash := "aabbccdd" + strings.Repeat("00", 28) // 64-char hex
+	assert.Equal(t, LabelFromHash(hash), LabelFromHash(hash))
```
michel-laterman: What's the point of this test?

internal/pkg/api/configLabel_test.go
```diff
@@ -0,0 +1,80 @@
...
+	t.Run("nouns", func(t *testing.T) {
+		assert.Len(t, labelNouns, 256)
+		seen := make(map[string]int, 256)
+		for i, w := range labelNouns {
+			require.NotEmpty(t, w, "noun at index %d is empty", i)
+			prev, dup := seen[w]
+			assert.False(t, dup, "noun %q at index %d duplicates index %d", w, i, prev)
+			seen[w] = i
+		}
+	})
```
michel-laterman:
```suggestion
	assert.Len(t, slices.Compact(slices.Sort(labelAdjectives)), 256, "labelAdjectives contains duplicate entry")
	assert.Len(t, slices.Compact(slices.Sort(labelNounts)), 256, "labelNouns contains duplicate entry")
```

internal/pkg/api/handleOpAMP.go
```diff
@@ -424,6 +424,14 @@ func (oa *OpAMPT) updateAgent(zlog zerolog.Logger, agent *model.Agent, aToS *pro
 		if effectiveConfigBytes != nil {
 			initialOpts = append(initialOpts, checkin.WithEffectiveConfig(effectiveConfigBytes))
 		}
+
+		configHash, err := HashEffectiveConfig(aToS.EffectiveConfig)
+		if err != nil {
+			zlog.Warn().Err(err).Msg("failed to compute effective config hash")
+		} else if configHash != "" {
+			initialOpts = append(initialOpts, checkin.WithEffectiveConfigHash(configHash))
+			initialOpts = append(initialOpts, checkin.WithEffectiveConfigLabel(LabelFromHash(configHash)))
```
michel-laterman:
```suggestion
			initialOpts = append(initialOpts, checkin.WithEffectiveConfigHash(configHash), checkin.WithEffectiveConfigLabel(LabelFromHash(configHash)))
```

internal/pkg/api/configLabel_test.go
juliaElastic: it tests that calling the `LabelFromHash` is deterministic, gives the same response when called twice

internal/pkg/api/configHash.go
juliaElastic: > Why do we need to only use these keys when computing the sha?

The issue scope is deliberate: the hash is meant to answer "are two collectors running the same pipeline wiring?", not "are all their configs byte-for-byte identical." The topology keys describe what data flows where. Non-topology fields like `extensions.*` configs or `service.telemetry` describe how the pipeline operates (scrape intervals, log levels, endpoints) but don't change the pipeline graph.

The use case is: two collectors that differ only in their `health_check` endpoint or telemetry verbosity should be considered "the same topology" and produce the same hash, so you can cluster/query them together.

> How do we ensure that a change to another key is emitted to a collector?

Changes to other keys are not surfaced via the hash, but they are captured in full in `effective_config` (the redacted JSON blob already stored in `.fleet-agents`). If a non-topology field changes, `effective_config` changes, but `effective_config_hash` stays the same.

internal/pkg/api/configHash.go
juliaElastic: Added support for multiple config files: https://github.com/elastic/fleet-server/pull/6872/changes/9ac32c8b8f1a973b0f58c54c9a86bba19dcdaa0b

Though I'm not sure if two separate files should hash differently from the same content in one file.

internal/pkg/api/configHash.go
```diff
@@ -0,0 +1,90 @@
...
+func HashEffectiveConfig(effectiveConfig *protobufs.EffectiveConfig) (string, error) {
+	if effectiveConfig.ConfigMap == nil || len(effectiveConfig.ConfigMap.ConfigMap) == 0 {
+		return "", nil
+	}
+
+	keys := make([]string, 0, len(effectiveConfig.ConfigMap.ConfigMap))
+	for k := range effectiveConfig.ConfigMap.ConfigMap {
+		keys = append(keys, k)
+	}
+	sort.Strings(keys)
```
michel-laterman:
```suggestion
	keys := slices.Sorted(maps.Keys(effectiveConfig.ConfigMap.ConfigMap))
```

internal/pkg/api/configHash.go
michel-laterman: Can you add that to the doc string or in schema.json? It's important to note how this hash should differ from the `RemoteConfig` hash

michel-laterman (dismissed):

internal/pkg/api/configHash.go
```diff
@@ -0,0 +1,90 @@
...
+
+// HashEffectiveConfig computes a SHA-256 hash of the pipeline topology fields
+// across all config files in the OpAMP ConfigMap. Each file is processed in
+// sorted key order so the hash is deterministic regardless of map iteration
+// order. Only receivers, processors, exporters, connectors, service.pipelines,
+// and service.extensions are included from each file. Keys within each file are
+// sorted deterministically by yaml.v3 Marshal. Returns "" with no error when
+// the config is nil or all files are empty.
+func HashEffectiveConfig(effectiveConfig *protobufs.EffectiveConfig) (string, error) {
+	if effectiveConfig.ConfigMap == nil || len(effectiveConfig.ConfigMap.ConfigMap) == 0 {
```
ycombinator: Should probably add a `nil` guard here for `effectiveConfig` as well?

model/schema.json
```diff
@@ -775,6 +775,14 @@
...
           "description": "The effective config that the collector is running",
           "format": "raw"
+        },
+        "effective_config_hash": {
+          "description": "SHA-256 hex hash of collector pipeline topology (receivers, processors, exporters, connectors, service.pipelines, service.extensions)",
+          "type": "string"
+        },
+        "effective_config_label": {
+          "description": "Human-readable adjective-noun label derived from the first two bytes of effective_config_hash",
+          "type": "string"
```
ycombinator, line 781: This feels like a presentation concern.  Should it be generated from the `effective_config_hash` in the UI code?

cmacknz: Agree this feels out of place in fleet-server an definitely feels like a presentation concern. This could be calculated in the UI.

internal/pkg/api/configHash.go
cmacknz: Yes, this needs to be differentiated from [AgentRemoteConfig.config_hash](https://github.com/open-telemetry/opamp-spec/blob/main/proto/opamp.proto#L1020-L1029) in the opamp spec.

Reading https://github.com/elastic/ingest-dev/issues/7064 it's not clear why we need this new concept of a topology hash, vs just the hash of the configuration. 

The problem this seems like this topology hash would solve is if you had a lot of collector configurations that had superflous components that did not modify the function of the pipeline. This feels like an optimization that can be done later once we confirm we have this problem.

Is there a reason we can't start with just the config_hash from the spec that includes everything, and only introduce an optimized topology hash as an optimization on that once that's proven to be necessary?

model/schema.json
juliaElastic: If we move it to the UI, it either has to be calculated dynamically or the `.fleet-agents` docs has to be updated from the UI to store it.

If we calculate it dynamically, it should be a runtime field to support filtering on it.

Removed the `effective_config_label` changes from here and added to kibana: https://github.com/elastic/kibana/pull/265005

internal/pkg/api/configHash.go
juliaElastic: I think the reason for the topology hash is to group collectors on the UI that run the same pipelines, @andrewvc might have more context.

Here is Claude's take on it:

## Why not just hash the full config?

The OTel collector expands environment variables before it reports `EffectiveConfig` back to
fleet-server. So a collector config with:

```yaml
service:
  telemetry:
    resource:
      service.instance.id: "${env:HOSTNAME}"
```

arrives at fleet-server as:

```yaml
service:
  telemetry:
    resource:
      service.instance.id: "Julias-MacBook-Pro.local"
```

Every collector in the same group — running identical pipelines — produces a unique full-config
hash just because of its hostname. Grouping collectors by hash becomes impossible.

That is the concrete reason the topology normalization exists: it strips per-instance runtime
values that vary across collectors in the same group. It is not an optimization for superfluous
components.

## When is a full hash sufficient?

If the only requirement is **per-collector change detection** ("has this collector's config
changed since last check-in?"), a full hash is simpler and still useful. The per-instance
expansion issue does not matter when comparing a single collector against its own history.

If the requirement is **topology grouping** ("show N collectors running pipeline X"), the
topology hash is necessary.

cmacknz: Right, the topology hash makes conceptual sense but as I said I would only introduce this when we see it as a need. This would arise when we see many users with functionally equivalent configurations that are not exactly the same. 

I think you should model this in the system where the identity of a configuration can be flexible, allowing us to start with the existing config hash and being able to change to a topology hash or even some other similarity score in the future.

I don't see a need to introduce this topology complexity on day one, let's wait until we see the user need for it to arise but we can anticipate we have to iterate here.

model/schema.json
cmacknz: Is there a large performance concern with this?

I would like to keep fleet-server in the architectural role an action router, this is getting away from that much more than other things we've done. I can also imagine other requirements that would require the UI to write this name.

I could imagine that users might ask for the ability to customize the names of their configurations in the future, like they do for agent policies. The friendly label is just an initial placeholder before we can do that.

juliaElastic: If using a runtime field, it's a small overhead when querying agents, I would start with that.

If kibana wrote the label to `.fleet-agents`, there is a chance of write conflicts with fleet-server, and we would need to poll from kibana to notice changes in the effective config hash to update the label.

internal/pkg/api/configHash.go
juliaElastic: Removed the topology hashing and using the full effective config to calculate the hash.

internal/pkg/api/configHash.go
```diff
@@ -0,0 +1,65 @@
...
+	for _, k := range keys {
+		file := effectiveConfig.ConfigMap.ConfigMap[k]
+		if file == nil || len(file.Body) == 0 {
+			continue
+		}
+		var parsed map[string]any
+		if err := yaml.Unmarshal(file.Body, &parsed); err != nil {
+			return "", fmt.Errorf("unmarshal config %q for hashing: %w", k, err)
+		}
+		canonical, err := yaml.Marshal(parsed)
```
cmacknz: Why do we need to do this? The input to the hash is just bytes, and YAML marshalling and unmarhsalling are not cheap, especially on unknown possibly very large configurations.

cmacknz: We already unmarshal and then marshal to JSON:

https://github.com/elastic/fleet-server/blob/40d76d8de5e40af1b111fbe84438529261b46e6c/internal/pkg/api/handleOpAMP.go#L463-L471

We certainly shouldn't unmarshal twice. We also have the option to hash the JSON document, but you'd still have to compensate for Go maps being unordered which we might want for convenience anyway.

model/schema.json
ycombinator: > we would need to poll from kibana to notice changes in the effective config hash to update the label.

Once a label is set, would we want to keep changing it? As I'm understanding it — and let me know if I'm off here — the hash and the label are answering two different questions for a pipeline:

* hash: has the topology of this pipeline changed?
* label: is this conceptually the same pipeline as before?

juliaElastic: I mean the label has to stay consistent with the effective config hash, and the effective config can change on check-in if the collector config changes.

We are moving away from the topology hash, see https://github.com/elastic/fleet-server/pull/6872#discussion_r3154120046

internal/pkg/api/configHash.go
juliaElastic: good point, updated in https://github.com/elastic/fleet-server/pull/6872/changes/0d15d086b8b1b25c34453c9244ed179eb93b2b17

model/schema.json
cmacknz: The way I view this is that we are trying to build a concept similar to that of an agent policy out of the per agent effective configurations.

That is a configuration shared by multiple agents doing the same job in the same role with a human readable and assignable name.

The problem is the per agent configurations can all vary per agent because parts of the configuration are determined at runtime (just like Elastic Agent) so doing this is not totally trivial.

internal/pkg/api/configHash.go
```diff
@@ -0,0 +1,85 @@
...
+// map[string]any. The returned map is keyed by the ConfigMap entry name.
+// Empty or nil entries are omitted. Call this once and pass the result to both
+// HashEffectiveConfig and the redact/marshal path so the YAML is only parsed once.
+func parseConfigFiles(ec *protobufs.EffectiveConfig) (map[string]map[string]any, error) {
+	out := make(map[string]map[string]any)
+	if ec == nil || ec.ConfigMap == nil {
+		return out, nil
+	}
+	for k, file := range ec.ConfigMap.ConfigMap {
+		if file == nil || len(file.Body) == 0 {
```
cmacknz, line 29: We need to check that the content type is text/yaml as well https://github.com/open-telemetry/opamp-spec/blob/2da595f59a0016abe67b4d44aa52afa3549f8742/proto/opamp.proto#L1046-L1048

internal/pkg/api/handleOpAMP.go
```diff
@@ -417,11 +415,25 @@ func (oa *OpAMPT) updateAgent(zlog zerolog.Logger, agent *model.Agent, aToS *pro
...
+
+		// Hash before redaction: redactSensitive mutates the parsed map in place.
+		configHash, err := HashEffectiveConfig(parsedFiles)
+		if err != nil {
+			zlog.Warn().Err(err).Msg("failed to compute effective config hash")
+		} else if configHash != "" {
+			initialOpts = append(initialOpts, checkin.WithEffectiveConfigHash(configHash))
+		}
+
+		if defaultParsed, ok := parsedFiles[""]; ok {
```
cmacknz: We only include the effective configuration if it uses a single configuration file? If the collector defines multiple configurations this does nothing?

internal/pkg/api/handleOpAMP.go
```diff
@@ -417,11 +415,25 @@ func (oa *OpAMPT) updateAgent(zlog zerolog.Logger, agent *model.Agent, aToS *pro
...
+			redactSensitive(defaultParsed)
+			effectiveConfigBytes, err := json.Marshal(defaultParsed)
```
cmacknz: HashEffectiveConfig marshals each individual configuration JSON already and we throw those bytes away, seems like we should reuse them?

internal/pkg/api/handleOpAMP.go
```diff
@@ -417,11 +415,25 @@ func (oa *OpAMPT) updateAgent(zlog zerolog.Logger, agent *model.Agent, aToS *pro
 	}
 
 	if aToS.EffectiveConfig != nil {
-		effectiveConfigBytes, err := ParseEffectiveConfig(aToS.EffectiveConfig)
+		parsedFiles, err := parseConfigFiles(aToS.EffectiveConfig)
```
cmacknz, line 418: It would help if you had unit tests showing that this handles all the AgentConfigMap arrangements we can have no files, one file, multiple files, non-yaml files, etc

https://github.com/open-telemetry/opamp-spec/blob/2da595f59a0016abe67b4d44aa52afa3549f8742/proto/opamp.proto#L1032-L1039

It's not completely trivial to do this hash so if we don't have a concrete use for it in the UI we could always defer it. I think we have already concluded it's not a great way to group collector configurations together.

internal/pkg/api/configHash.go
juliaElastic: filtered for content type

internal/pkg/api/handleOpAMP.go
juliaElastic: changed to handle multiple files

internal/pkg/api/handleOpAMP.go
juliaElastic: addressed

internal/pkg/api/handleOpAMP.go
juliaElastic: added unit tests

juliaElastic: I moved the PR to draft

internal/pkg/api/configHash.go
andrewvc: @cmacknz with the new opamp metadata around pipeline groups etc. in https://github.com/elastic/fleet-server/pull/6769 I think the hashes will always be unique in the UI so long as well as the case @juliaElastic mentioned. Correct me if I'm wrong but from what I can tell opamp today just uses the bytes of the file right?

If that's the case the flow around "Tell me if the collectors in this metadata group are all running the same config" no longer can be achieved in the UI or API. So, tracking a config rollout would not be possible.

cmacknz: We can still group collectors by the non_identifying_attributes we tell users to populate can't we?

The problem we have with environment variable substitution in those attributes can happen in any part of the collector config, not just service telemetry.

I think this grouping feature is good, I just struggle to see how we can make it work 100% reliably with hashing in Fleet Server. 

We are storing these configurations is a search engine that can compute document similarity in queries based on specific fields so I don't know why we need to try to do something based on exact equivalence of fields in a less flexible part of the system design.

andrewvc: It's a good point @cmacknz , I agree the hashing mechanism is not perfect. I hadn't thought of using relevance to match similar configs, it's a neat idea! I assume we would implement that a bit later?

cmacknz: Julia did a good analysis in https://github.com/elastic/ingest-dev/issues/7064#issuecomment-4358616898 of why using relevance isn't completely straightforward either.

The conclusion on that thread is asking our PMs about how they view the need for this so we can figure out what the right trade off is. This seems like a completely sensible feature to have, but one where a reliable implementation is much harder than we would like it to be.

PR #6878, opened 2026-04-22
Revert "fix: always set ReportFullState flag in OpAMP responses (#6831)"

ycombinator:

Since April 17, the daily 10k OpAMP-on-serverless scale test has failed every run ([details](https://github.com/elastic/observability-perf/issues/869#issuecomment-4300298292)). The failures correlate with the deployment of PR #6831, which set `ServerToAgentFlags_ReportFullState` on every `ServerToAgent` response to every agent.

Reverts #6831 to remove the unconditional `ReportFullState` flag from every OpAMP response.

This is initially a temporary revert to confirm that #6831 is the root cause of the scale test failures. Once confirmed, the `ReportFullState` logic can be re-introduced in a more targeted way (e.g. only on enrollment or drift detection, not on every message).

No special local testing needed — the fix will be validated by the daily 10k OpAMP-on-serverless scale test returning to passing state after deployment.

## Design Checklist

## Checklist

## Related issues

- Reverts #6831
- Relates https://github.com/elastic/observability-perf/issues/869


ebeahan (approved):

swiatekm (approved):

michel-laterman (changes requested):
As mentioned in our standup, even if we revered this change and implemented the spec requirement to collect on seq_no drift; our scale testing _should_ test that a server can gather the full status reports from all agents at once as we would need to ensure we are reliable if an event occurs that causes all agents to drift (restore from snapshot, network connectivity, etc).
I would find out why the test is failing and address the root cause

cmacknz:
> our scale testing should test that a server can gather the full status reports from all agents at once as we would need to ensure we are reliable if an event occurs that causes all agents to drift (restore from snapshot, network connectivity, etc).

Yes this was my motivation for doing it this way, converting the edge cases into always cases so that we didn't have to discover them through incidents or support cases.

I think reverting this to confirm it is the problem is fine, but if it is you still have to fix it.

ycombinator:
Yes, this revert is only to isolate the scale test failures to this specific change, as there have been other changes that went in between the scale tests consistently passing and starting to fail.  Once we've confirmed that this is the specific change that started causing the failures, we'll bring it back in a way that can scale.

michel-laterman (approved):
OK, let's revert and investigate

ycombinator:
I will check the tests on Thursday, which will give us 3 runs between now and then.  If they are all successful, we can isolate the cause to the changes in #6831 and then bring it back in a scalable fashion.

PR #6955, opened 2026-05-04
OpAMP redact slice maps

michel-laterman:

Redact slice maps with the form:
```yaml
- name: Authorization
  value: secretValue
```

Use the shared redact package in elastic-agent-libs

Enroll an opamp agent with sensitive values in the headers slices.

## Design Checklist

## Checklist

## Related issues

- Relates https://github.com/elastic/elastic-agent-libs/issues/415


internal/pkg/api/handleOpAMP_test.go
```diff
@@ -141,6 +143,103 @@ func TestProtobufKVToRawMessage(t *testing.T) {
...
+	}{
+		{
+			name: "sensitive top-level keys are redacted",
+			body: `password: hunter2
+api_key: abcd-1234
+routekey: should-not-be-redacted
+plain: keepme
+`,
+			want: map[string]any{
+				"password": redact.REDACTED,
```
ycombinator, line 160: Probably not a huge deal and also not a concern of this PR, strictly speaking, but `redact.REDACTED` now evaluates to `REDACTED`, whereas previously we were using `[REDACTED]`.

ycombinator (approved):

internal/pkg/api/handleOpAMP_test.go
michel-laterman: Yeah, I thought it would be cleaner to use a single redacted value instead of having a seperate one for URLs and other values

PR #6956, opened 2026-05-04
Add API version constants to pkg/api

ycombinator:

Elastic Agent and Horde both need the Fleet Server API version string (`"2023-06-01"`) and the API version header key (`"Elastic-Api-Version"`) to communicate with Fleet Server. Today each consumer defines these strings locally, which leads to drift and duplication. Fleet Server's `pkg/api` package already exports the generated types (action constants, checkin/enrollment/ack types) but does not export these two version-related constants.

Adds a hand-written `pkg/api/version.go` file exporting two constants:
- `DefaultFleetAPIVersion = "2023-06-01"`
- `ElasticAPIVersionHeaderKey = "Elastic-Api-Version"`

These are part of the API contract but are not expressible in the OpenAPI spec as importable Go constants, so they live in a hand-written file alongside the generated code.

```
go build ./pkg/api/...
go test ./pkg/api/...
```

No behavioral changes — only new exports.

## Design Checklist

N/A — this is a constants-only change with no runtime behavior.

## Checklist

## Related issues

- Relates https://github.com/elastic/ingest-dev/issues/7601

### DIFF
--- a/pkg/api/version.go
+++ b/pkg/api/version.go
@@ -4,10 +4,7 @@
 
 package api
 
-const (
-	// DefaultFleetAPIVersion is the current Fleet Server API version.
-	DefaultFleetAPIVersion = "2023-06-01"
-
-	// ElasticAPIVersionHeaderKey is the HTTP header used to negotiate the API version.
-	ElasticAPIVersionHeaderKey = "Elastic-Api-Version"
-)
+// DefaultFleetAPIVersion is the current Fleet Server API version.
+// The header name is already defined in the OpenAPI spec as a parameter
+// on each endpoint and is handled by the generated client/params structs.
+const DefaultFleetAPIVersion = "2023-06-01"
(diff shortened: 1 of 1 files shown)


michel-laterman:
elastic-api-version is already defined in the openapi.yml spec as a parameter:
https://github.com/elastic/fleet-server/blob/045c0b2bd73e9df2643419cbc98a143c7ecd6880/model/openapi.yml#L1144-L1152

https://github.com/elastic/fleet-server/blob/045c0b2bd73e9df2643419cbc98a143c7ecd6880/model/openapi.yml#L1747

These are meant to be supplied by the params struct that is sent with each request which is seperate from the bodies

ycombinator:
Good point — I've dropped `ElasticAPIVersionHeaderKey` since the header name is already covered by the spec and the generated params structs. The PR now only exports `DefaultFleetAPIVersion` as a canonical default value that consumers can use to populate the `ElasticApiVersion` field on those params structs (or set the header manually until they migrate to the generated client).

michel-laterman (approved):

PR #6988, opened 2026-05-08
Fix effective_config merge bug for OpAMP collectors

### LINKED ISSUE #6877: [OpApmp] When deleting a pipeline from my collector it's not deleted from effective_config
## Description 

In an otel config with multiple pipeline, when I delete a pipeline, it's still present in fleet agent effective_config, should it be deleted?

ycombinator:

When an OpAMP collector removes a pipeline from its configuration, the removed pipeline still appears in the `effective_config` field of the agent document in Elasticsearch. This happens because fleet-server uses Elasticsearch's partial document update API (`{"doc": {...}}`), which recursively merges objects rather than replacing them. So keys present in the old `effective_config` but absent in the new one are preserved.

For example, if the stored config has pipelines `{traces, metrics}` and the collector reports only `{traces}`, the merge would keep `metrics` when it should be removed.

When the checkin includes `effective_config`, the update now uses a new Painless script ([`checkinWithEffectiveConfig.painless`](https://github.com/elastic/fleet-server/blob/main/internal/pkg/checkin/checkinWithEffectiveConfig.painless)) that follows the same pattern as [`deleteAuditFieldsOnCheckin.painless`](https://github.com/elastic/fleet-server/blob/main/internal/pkg/checkin/deleteAuditFieldsOnCheckin.painless), directly assigning fields on `ctx._source` to fully replace `effective_config` instead of merging it. This is a new branch in `flush()` alongside the existing script and doc-merge paths.

```
mage test:integration
```

## Design Checklist

## Checklist

## Related issues

- Closes #6877

### DIFF
--- a/internal/pkg/checkin/bulk.go
+++ b/internal/pkg/checkin/bulk.go
@@ -481,7 +481,6 @@ func encodeParams(now string, data pendingT) (map[string]json.RawMessage, error)
 		isSet              json.RawMessage
 		seqNo              json.RawMessage
 		availableRollbacks json.RawMessage
-		deleteAudit        json.RawMessage
 
 		err error
 	)
@@ -506,8 +505,6 @@ func encodeParams(now string, data pendingT) (map[string]json.RawMessage, error)
 	isSet, err = json.Marshal(data.extra.seqNo.IsSet())
 	Err = errors.Join(Err, err)
 	seqNo, err = json.Marshal(data.extra.seqNo)
-	Err = errors.Join(Err, err)
-	deleteAudit, err = json.Marshal(data.extra.deleteAudit)
 	Err = errors.Join(Err, err)
 	if data.extra.meta != nil {
 		meta = data.extra.meta
@@ -549,6 +546,5 @@ func encodeParams(now string, data pendingT) (map[string]json.RawMessage, error)
 		"SeqNoSet":           isSet,
 		"SeqNo":              seqNo,
 		"AvailableRollbacks": availableRollbacks,
-		"DeleteAudit":        deleteAudit,
 	}, nil
 }
--- a/internal/pkg/checkin/bulk.go
+++ b/internal/pkg/checkin/bulk.go
@@ -469,18 +469,20 @@ func encodeParams(now string, data pendingT) (map[string]json.RawMessage, error)
 		reason  json.RawMessage
 
 		// optional attributes below
-		policyID           json.RawMessage
-		revisionIDX        json.RawMessage
-		sequenceNum        json.RawMessage
-		ver                json.RawMessage
-		meta               json.RawMessage
-		components         json.RawMessage
-		health             json.RawMessage
-		capabilities       json.RawMessage
-		effectiveConfig    json.RawMessage
-		isSet              json.RawMessage
-		seqNo              json.RawMessage
-		availableRollbacks json.RawMessage
+		policyID              json.RawMessage
+		revisionIDX           json.RawMessage
+		ver                   json.RawMessage
+		meta                  json.RawMessage
+		components            json.RawMessage
+		isSet                 json.RawMessage
+		seqNo                 json.RawMessage
+		availableRollbacks    json.RawMessage
+
+		// OpAMP-specific attributes
+		opampSequenceNum      json.RawMessage
+		opampHealth           json.RawMessage
+		opampCapabilities     json.RawMessage
+		opampEffectiveConfig  json.RawMessage
 
 		err error
 	)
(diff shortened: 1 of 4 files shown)


internal/pkg/checkin/bulk.go
```diff
@@ -439,13 +469,19 @@ func encodeParams(now string, data pendingT) (map[string]json.RawMessage, error)
...
+		revisionIDX        json.RawMessage
+		sequenceNum        json.RawMessage
+		ver                json.RawMessage
+		meta               json.RawMessage
+		components         json.RawMessage
+		health             json.RawMessage
+		capabilities       json.RawMessage
+		effectiveConfig    json.RawMessage
+		isSet              json.RawMessage
+		seqNo              json.RawMessage
```
michel-laterman: Having sequenceNum and seqNo makes this really muddy. Can we prefix opamp attributes with OpAMP here?

internal/pkg/checkin/bulk.go
```diff
@@ -463,33 +499,56 @@ func encodeParams(now string, data pendingT) (map[string]json.RawMessage, error)
...
+	if data.extra.capabilities != nil {
+		capabilities, err = json.Marshal(data.extra.capabilities)
+		Err = errors.Join(Err, err)
+	}
+	if data.extra.effectiveConfig != nil {
+		effectiveConfig = data.extra.effectiveConfig
+	}
+	if data.extra.availableRollbacks != nil {
+		availableRollbacks = data.extra.availableRollbacks
+	}
```
michel-laterman: How were we doing this before?

internal/pkg/checkin/bulk_integration_test.go
```diff
@@ -0,0 +1,88 @@
...
+// TestEffectiveConfigReplacesRemovedFields verifies that when a collector
+// removes a pipeline from its configuration, the effective_config stored in
+// Elasticsearch no longer contains the removed pipeline.
+//
+// Reproduces https://github.com/elastic/fleet-server/issues/6877
+func TestEffectiveConfigReplacesRemovedFields(t *testing.T) {
+	ctx := testlog.SetLogger(t).WithContext(t.Context())
+
+	index, bulker := ftesting.SetupCleanIndex(ctx, t, dl.FleetAgents)
+	agentID := uuid.Must(uuid.NewV4()).String()
```
michel-laterman: (nitpick) uuidv7

internal/pkg/checkin/checkinWithEffectiveConfig.painless
```diff
@@ -0,0 +1,42 @@
...
+  if (ctx._source.upgrade == null) {
+    ctx._source.upgrade = new HashMap();
+  }
+  ctx._source.upgrade.rollbacks = params.AvailableRollbacks;
+}
+if (params.DeleteAudit) {
+  ctx._source.remove('audit_unenrolled_reason');
+  ctx._source.remove('audit_unenrolled_time');
+  ctx._source.remove('unenrolled_at');
+}
```
michel-laterman: This will never trigger in this script; the deleteAudit bool is checked in an if statement before the effectiveConfig script is

internal/pkg/checkin/checkinWithEffectiveConfig.painless
```diff
@@ -0,0 +1,42 @@
...
+if (params.Health != null) {
+  ctx._source.health = params.Health;
+}
+if (params.Capabilities != null) {
+  ctx._source.capabilities = params.Capabilities;
+}
+if (params.EffectiveConfig != null) {
+  ctx._source.effective_config = params.EffectiveConfig;
+}
```
michel-laterman: Isn't this always not null?

ycombinator: Yes, good catch.  Fixed in 3e79d60.

internal/pkg/checkin/checkinWithEffectiveConfig.painless
ycombinator: Thanks.  Fixed in 1d4cf8f.

internal/pkg/checkin/bulk_integration_test.go
ycombinator: Fixed in d93208a.

internal/pkg/checkin/bulk.go
ycombinator: Done in 1a3c481.

internal/pkg/checkin/bulk.go
ycombinator: Good catch.  `availableRollbacks` doesn't belong in this PR.  Removed in cfac78c1b300f113c31bef2e332e0013fe5dee9d.

cmacknz:
I don't have a problem with this approach, but one thing we could look at if we get more problems here is moving effective configuration into it's own index that doesn't use the partial document update API.

michel-laterman (approved):
lgtm

PR #7007, opened 2026-05-11
Fix missing `authAgent` in `UploadChunk`

blakerouse:

In the `UploadChunk` the `authAgent` was not being called to validate that the uploading agent of the chunk has the permissions to write that chunk. It validated that there was a valid API key, it just didn't validate that the agent with the valid API key is the agent that should be uploading the chunk.

In `ValidAPIKey` a disabled API key was considered valid even once disabled.

It adds the `authAgent` check to the `UploadChunk`.

It sets `ok = false` when the API key is disabled in `CacheT.ValidAPIKey`.

## Design Checklist

## Checklist

### DIFF
--- a/internal/pkg/cache/cache_test.go
+++ b/internal/pkg/cache/cache_test.go
@@ -8,6 +8,7 @@ package cache
 
 import (
 	"testing"
+	"testing/synctest"
 	"time"
 
 	"github.com/elastic/fleet-server/v7/internal/pkg/config"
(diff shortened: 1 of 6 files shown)


internal/pkg/cache/cache_test.go
```diff
@@ -0,0 +1,75 @@
...
+	}
+	c, err := New(cfg)
+	require.NoError(t, err)
+
+	key := APIKey{ID: "test-id", Key: "test-key"}
+
+	// Cache the key as disabled
+	c.SetAPIKey(key, false)
+	// ristretto is async; wait for value to be available
+	time.Sleep(10 * time.Millisecond)
```
swiatekm: Can we use synctest for these to avoid sleeps?

swiatekm (dismissed):
LGTM, non blocking question.

internal/pkg/cache/cache_test.go
blakerouse: Updated to use `synctest`, if you can give this another look. Thanks!

samuelvl (approved):

swiatekm (approved):

PR #7009, opened 2026-05-11
fix: enforce policy-based access control on artifact downloads

ycombinator:

The artifact download endpoint (`/api/fleet/artifacts/{id}/{sha256}`) only validates the agent's API key but never checks whether the requested artifact belongs to the agent's assigned policy. This means an agent enrolled under one policy can download artifacts belonging to a different policy if it knows the artifact ID and SHA256 hash. For example, an agent enrolled under a policy with no integrations can retrieve Elastic Defend trust lists, exception lists, and other security artifacts from another policy.

Implements the `authorizeArtifact()` function (previously a no-op that returned `nil`) to enforce policy-based access control:

1. Adds a `GetPolicy(ctx, policyID)` method to the `policy.Monitor` interface that returns the cached policy for a given ID (reloads from ES on cache miss).
2. In `authorizeArtifact`, fetches the agent's policy via the monitor using `agent.AgentPolicyID` and verifies that the requested artifact (`identifier` + `decoded_sha256`) appears in the policy's `inputs[].artifact_manifest.artifacts`.
3. Returns **403 Forbidden** (`ErrUnauthorizedArtifact`) if the artifact is not listed in the agent's assigned policy.

1. Set up Fleet Server with Elasticsearch and Kibana
2. Create two agent policies: **Victim-Policy** with Elastic Defend integration (add a trusted application), and **Attacker-Policy** with no integrations
3. Create an enrollment token for Attacker-Policy and enroll an agent
4. Attempt to download an artifact belonging to Victim-Policy using the attacker agent's API key — should now receive **403 Forbidden** instead of the artifact contents
5. Verify that an agent enrolled under Victim-Policy can still download its own artifacts normally (200 OK)

## Design Checklist

## Checklist

### DIFF
--- a/internal/pkg/api/handleArtifacts.go
+++ b/internal/pkg/api/handleArtifacts.go
@@ -178,23 +178,28 @@ func (at ArtifactT) authorizeArtifact(ctx context.Context, agent *model.Agent, i
 
 func policyHasArtifact(pd *model.PolicyData, id, sha2 string) bool {
 	for _, input := range pd.Inputs {
-		manifestRaw, ok := input["artifact_manifest"].(map[string]any)
-		if !ok {
-			continue
-		}
-		artifacts, ok := manifestRaw["artifacts"].(map[string]any)
-		if !ok {
-			continue
-		}
-		entry, ok := artifacts[id].(map[string]any)
-		if !ok {
-			continue
-		}
-		if sha, _ := entry["decoded_sha256"].(string); sha == sha2 {
+		if inputHasArtifact(input, id, sha2) {
 			return true
 		}
 	}
 	return false
+}
+
+func inputHasArtifact(input map[string]any, id, sha2 string) bool {
+	manifestRaw, ok := input["artifact_manifest"].(map[string]any)
+	if !ok {
+		return false
+	}
+	artifacts, ok := manifestRaw["artifacts"].(map[string]any)
+	if !ok {
+		return false
+	}
+	entry, ok := artifacts[id].(map[string]any)
+	if !ok {
+		return false
+	}
+	sha, _ := entry["decoded_sha256"].(string)
+	return sha == sha2
 }
 
 // Return artifact from cache by sha2 or fetch directly from Elastic.
--- a/internal/pkg/api/handleArtifacts.go
+++ b/internal/pkg/api/handleArtifacts.go
@@ -176,41 +176,25 @@ func (at ArtifactT) authorizeArtifact(ctx context.Context, agent *model.Agent, i
 	return ErrUnauthorizedArtifact
 }
 
-type artifactManifest struct {
-	Artifacts map[string]manifestEntry `json:"artifacts"`
-}
-
-type manifestEntry struct {
-	DecodedSha256 string `json:"decoded_sha256"`
-}
-
 func policyHasArtifact(pd *model.PolicyData, id, sha2 string) bool {
 	for _, input := range pd.Inputs {
-		manifest, err := parseArtifactManifest(input)
-		if err != nil || manifest == nil {
+		manifestRaw, ok := input["artifact_manifest"].(map[string]any)
+		if !ok {
 			continue
 		}
-		if entry, ok := manifest.Artifacts[id]; ok && entry.DecodedSha256 == sha2 {
+		artifacts, ok := manifestRaw["artifacts"].(map[string]any)
+		if !ok {
+			continue
+		}
+		entry, ok := artifacts[id].(map[string]any)
+		if !ok {
+			continue
+		}
+		if sha, _ := entry["decoded_sha256"].(string); sha == sha2 {
 			return true
 		}
 	}
 	return false
-}
-
-func parseArtifactManifest(input map[string]any) (*artifactManifest, error) {
-	raw, ok := input["artifact_manifest"]
-	if !ok {
-		return nil, nil
-	}
-	data, err := json.Marshal(raw)
-	if err != nil {
-		return nil, err
-	}
-	var manifest artifactManifest
-	if err := json.Unmarshal(data, &manifest); err != nil {
-		return nil, err
-	}
-	return &manifest, nil
 }
 
 // Return artifact from cache by sha2 or fetch directly from Elastic.
--- a/internal/pkg/api/handleArtifacts.go
+++ b/internal/pkg/api/handleArtifacts.go
@@ -25,6 +25,7 @@ import (
 	"github.com/elastic/fleet-server/v7/internal/pkg/dl"
 	"github.com/elastic/fleet-server/v7/internal/pkg/logger"
 	"github.com/elastic/fleet-server/v7/internal/pkg/model"
+	"github.com/elastic/fleet-server/v7/internal/pkg/policy"
 	"github.com/elastic/fleet-server/v7/internal/pkg/throttle"
 
 	"github.com/rs/zerolog"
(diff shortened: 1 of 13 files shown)


internal/pkg/api/handleArtifacts.go
```diff
@@ -138,19 +143,70 @@ func (at ArtifactT) processRequest(ctx context.Context, zlog zerolog.Logger, age
...
+	return ErrUnauthorizedArtifact
+}
+
+func policyHasArtifact(pd *model.PolicyData, id, sha2 string) bool {
+	for _, input := range pd.Inputs {
+		am, ok := input["artifact_manifest"]
+		if !ok {
+			continue
+		}
+		amMap, ok := am.(map[string]interface{})
```
michel-laterman: go's newer conventions prefers `any` over `interface{}`
this is being enforced by the `go fix` check that can be ran with `mage check:fix`

internal/pkg/api/handleArtifacts.go
```diff
@@ -138,19 +143,70 @@ func (at ArtifactT) processRequest(ctx context.Context, zlog zerolog.Logger, age
...
+	if p.Data != nil && policyHasArtifact(p.Data, id, sha2) {
+		return nil
+	}
+
+	return ErrUnauthorizedArtifact
+}
+
+func policyHasArtifact(pd *model.PolicyData, id, sha2 string) bool {
+	for _, input := range pd.Inputs {
+		am, ok := input["artifact_manifest"]
```
michel-laterman: Should we define this artifact_manifest as a struct somewhere?

ycombinator: Added `model.ArtifactManifest` and `model.ManifestEntry` structs in 3e57a6ed.

internal/pkg/api/handleArtifacts.go
ycombinator: Updated in b00c5df4.

michel-laterman (dismissed):
lgtm

internal/pkg/api/handleArtifacts.go
```diff
@@ -138,19 +143,74 @@ func (at ArtifactT) processRequest(ctx context.Context, zlog zerolog.Logger, age
...
+	}
+	return false
+}
+
+func parseArtifactManifest(input map[string]any) (*artifactManifest, error) {
+	raw, ok := input["artifact_manifest"]
+	if !ok {
+		return nil, nil
+	}
+	data, err := json.Marshal(raw)
```
cmacknz: Marshalling to then unmarshal feels a bit weird, there's no way to just unmarshal this? Maybe an intermediate type that has `artifact_manifest` as `json.RawMessage` instead of an any?

ycombinator: Removed the marshal-then-unmarshal; agree, it was awkward.  Just went with directly accessing map fields, with type assertions for safety: 535b14fa498518d67e17c9ccf16f5550a2c74590

testing/e2e/api_version/client_api_current.go
```diff
@@ -405,15 +405,49 @@ func (tester *ClientAPITester) TestFullFileUpload() {
 }
 
 func (tester *ClientAPITester) TestArtifact() {
-	ctx, cancel := context.WithTimeout(context.Background(), 3*time.Minute)
+	ctx, cancel := context.WithTimeout(context.Background(), 5*time.Minute)
```
michel-laterman: (nitpick) we should use `tester.T().Context()` instead of background

testing/e2e/api_version/client_api_2023_06_01.go
```diff
@@ -305,13 +305,47 @@ func (tester *ClientAPITester20230601) TestFullFileUpload() {
 }
 
 func (tester *ClientAPITester20230601) TestArtifact() {
-	ctx, cancel := context.WithTimeout(context.Background(), 3*time.Minute)
+	ctx, cancel := context.WithTimeout(context.Background(), 5*time.Minute)
```
michel-laterman: (nitpick) we should use `tester.T().Context()` instead of background

michel-laterman (dismissed):
lgtm, just have nitpicks

internal/pkg/policy/monitor.go
```diff
@@ -591,3 +596,32 @@ func (m *monitorT) LatestRev(ctx context.Context, id string) int64 {
...
+	if err := m.loadPolicies(ctx); err != nil {
+		return nil, fmt.Errorf("loading policies: %w", err)
+	}
+
+	m.mut.Lock()
+	p, ok = m.policies[policyID]
+	m.mut.Unlock()
+
+	if !ok {
+		return nil, fmt.Errorf("policy %s not found", policyID)
```
samuelvl: Should we return `ErrUnauthorizedArtifact` here? As it is now, it would return 500 (default in error.go) but I think 403 is more appropriate. Returning 404 would leak whether the policy ID exists.

testing/e2e/scaffold/scaffold.go
```diff
@@ -592,6 +592,78 @@ func (s *Scaffold) FleetHasArtifacts(ctx context.Context) []ArtifactHit {
...
+						return
+					}
+				}
+			}
+			timer.Reset(time.Second)
+		}
+	}
+}
+
+func policyInputHasArtifact(input map[string]any, identifier, decodedSha256 string) bool {
```
samuelvl: nit: this can be re-used from policyHasArtifact in handleArtifacts.go if we define it as:

```golang
func policyHasArtifact(pd *model.PolicyData, id, sha2 string) bool {
    for _, input := range pd.Inputs {
        if inputHasArtifact(input, id, sha2) {
            return true
        }
    }
    return false
}

func inputHasArtifact(input map[string]any, id, sha2 string) bool {
...
}
```

testing/e2e/scaffold/scaffold.go
```diff
@@ -592,6 +592,78 @@ func (s *Scaffold) FleetHasArtifacts(ctx context.Context) []ArtifactHit {
...
+	}
+	artifacts, ok := manifestRaw["artifacts"].(map[string]any)
+	if !ok {
+		return false
+	}
+	entry, ok := artifacts[identifier].(map[string]any)
+	if !ok {
+		return false
+	}
+	sha, _ := entry["decoded_sha256"].(string)
```
samuelvl, line 663: nit: we can read `decoded_sha256` from https://github.com/elastic/fleet-server/blob/main/internal/pkg/dl/constants.go#L68

It can be worth to also add `artifact_manifest` and `artifacts` to this constants.go file.

internal/pkg/policy/monitor.go
ycombinator: Good idea, but I won't have this function itself return `ErrUnauthorizedArtifact` as it's not specific to artifact fetching (that's done by the caller of this function).  So I'll have this function return an error that the caller can look for and return a `ErrUnauthorizedArtifact` from there.

testing/e2e/scaffold/scaffold.go
ycombinator: Addressed via 12faacbe

testing/e2e/scaffold/scaffold.go
ycombinator: Addressed via d347b336

internal/pkg/policy/monitor.go
ycombinator: Addressed via aa125cf2

samuelvl (dismissed):
Thanks for the changes. Some linting issues but LGTM

samuelvl (dismissed):

samuelvl (approved):

PR #7027, opened 2026-05-14
Replace `pkg/api/LICENSE.txt` symlink with actual file

ycombinator:

The `pkg/api/LICENSE.txt` file added in #6982 is a symlink pointing to `../../LICENSE.txt`. The Go module packaging library (`golang.org/x/mod/zip`) excludes symlinks when building module zips — from its [documented constraints](https://pkg.go.dev/golang.org/x/mod/zip):

> Symbolic links and other irregular files are not allowed.

As a result, the LICENSE file is absent from the distributed `github.com/elastic/fleet-server/pkg/api` Go module. Consumers that run NOTICE generation (e.g. elastic-agent using `go-licence-detector`) fail with:

```
no licence file found for github.com/elastic/fleet-server/pkg/api. Add an override entry with licence type to continue.
```

Replaces the symlink with an actual copy of the license text, so the file is present in the distributed Go module zip.

After merging and cutting a new pseudo-version, run `mage notice` in elastic-agent. NOTICE generation should complete without errors.

## Design Checklist

Not applicable — this is a license file fix with no runtime code changes.

## Checklist

## Related issues

- Supersedes #6982
- Unblocks https://github.com/elastic/elastic-agent/pull/14012

### DIFF
--- a/pkg/api/LICENSE.txt
+++ b/pkg/api/LICENSE.txt
@@ -1,1 +1,223 @@
-../../LICENSE.txt
+ELASTIC LICENSE AGREEMENT
+
+PLEASE READ CAREFULLY THIS ELASTIC LICENSE AGREEMENT (THIS "AGREEMENT"), WHICH
+CONSTITUTES A LEGALLY BINDING AGREEMENT AND GOVERNS ALL OF YOUR USE OF ALL OF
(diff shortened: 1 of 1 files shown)


macdewee (approved):
LGTM
I would consider adding some check to keep the files in sync, but on the other hand the LICENSE has not been changed for over 6 years

swiatekm (approved):

ebeahan:
Do we need to backport?

PR #7029, opened 2026-05-14
bk: run faster builds

v1v:
## What is the problem this PR solves?

Checkout faster when running the loading bk pipeline defined in the catalog-info

## How does this PR solve the problem?

Using the spare checkout for the .buildkite folder and the .go-version file

## How to test this PR locally

CI

## Design Checklist

- [ ] I have ensured my design is stateless and will work when multiple fleet-server instances are behind a load balancer.
- [ ] I have or intend to scale test my changes, ensuring it will work reliably with 100K+ agents connected.
- [ ] I have included fail safe mechanisms to limit the load on fleet-server: rate limiting, circuit breakers, caching, load shedding, etc.

## Checklist

- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation
- [ ] I have made corresponding change to the default configuration files
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] I have added an entry in `./changelog/fragments` using the [changelog tool](https://github.com/elastic/fleet-server#changelog)

## Related issues

- see https://github.com/elastic/kibana/pull/265568
- https://github.com/elastic/elastic-agent/pull/12082

PR #7038, opened 2026-05-14
Revert "bk: run faster builds"

v1v:
Reverts elastic/fleet-server#7029

As long as the post-checkout does some logic this cannot be used.

In theory, the BK PR bot could be the solution, and could help to remove all the post-checkout logics required to always test PRs with the merge, but I dont' think it's available yet that feature


ebeahan:
I also reverted https://github.com/elastic/fleet-server/pull/7037, since its purpose was to try to fix the changes from #7029.

ebeahan (approved):

ebeahan (changes requested):
Actually, we may not need to revert. Synthetics are now running again: https://buildkite.com/elastic/gpctl-promote-after-serverless-devenv-synthetics/builds/43741

Let me validate before we merge.

v1v:
I think we need to rebump this. 

The current checkout process with the post-checkout step is not working as expected:

- https://buildkite.com/elastic/fleet-server/builds/14781#019e4149-3bfa-41ce-a41e-1b17fb5a1d4e

```
HEAD is now at ab3b3a1 chore: Update snapshot.yml (#7060)
--
  | Current branch: HEAD
  | Switched to a new branch 'pr_merge_7068'
  | New branch created: pr_merge_7068
  | fatal: refusing to merge unrelated histories
  | Merge failed: 128
  | fatal: There is no merge to abort (MERGE_HEAD missing).
```


I'm not much familiar with the post-checkout script, but it's causing some issues when using the spare-checkout. 

One solution could be to add a condition in the post-checkout so it does not run when running the upload command

ebeahan (approved):

PR #7135, opened 2026-06-01
Add limits.max_agent_doc_size config

michel-laterman:

Prevent fleet-server from going OOM when a compressed checkin request size explodes on decompression.

Add a new config option called limits.max_agent_doc_size that defaults to 20mb. This limit is applied to the gzip.Reader for compressed requests.

## Design Checklist

## Checklist


internal/pkg/config/env_defaults.go
```diff
@@ -26,6 +26,8 @@ const (
 
 	defaultMaxConnections = 0 // no limit
 
+	defaultMaxAgentDocSize = 20 * 1024 * 1024 // 20mb limit, applies to checkin request bodies when decompressing
```
michel-laterman: One thing we can consider is aligning with OpAMP message max default size value of 64mb: https://github.com/open-telemetry/opamp-spec/pull/346
This value applies to messages after decompression

internal/pkg/api/handleCheckin.go
```diff
@@ -196,7 +196,7 @@ func (ct *CheckinT) validateRequest(zlog zerolog.Logger, w http.ResponseWriter,
 			return validatedCheckin{}, &BadRequestErr{msg: "unable to create gzip reader for request body", nextErr: err}
 		}
 		defer gr.Close()
-		bodyReader = gr
+		bodyReader = http.MaxBytesReader(w, gr, ct.cfg.Limits.MaxAgentDocSize)
```
ycombinator: How about `ct.cfg.Limits.CheckinLimit.MaxBodyDecompressed` instead, to parallel the existing `ct.cfg.Limits.CheckinLimit.MaxBody`?  I know that means adding a new `MaxBodyDecompressed` field to the `config.Limit` struct but maybe that could be useful to other  requests besides check-in too in the future, e.g. opamp?

michel-laterman: good idea, i'll add the change

internal/pkg/config/env_defaults.go
michel-laterman: After a brief discussion we're going to start at the smaller 20Mb; we can increase this to match OpAMP later as needed.

internal/pkg/api/handleCheckin_test.go
```diff
@@ -1244,6 +1244,30 @@ func TestValidateCheckinRequest(t *testing.T) {
...
+				_ = gz.Close()
+				return &http.Request{
+					Header: http.Header{"Content-Encoding": []string{"gzip"}},
+					Body:   io.NopCloser(&buf),
+				}
+			}(),
+			cfg: &config.Server{
+				Limits: config.ServerLimits{
+					CheckinLimit: config.Limit{
+						MaxBody:             512, // compressed body is lower then doc size
```
ycombinator: Typo.
```suggestion
						MaxBody:             512, // compressed body is lower than doc size
```

internal/pkg/config/limits.go
```diff
@@ -9,10 +9,11 @@ import (
...
 
 type Limit struct {
-	Interval time.Duration `config:"interval"`
-	Burst    int           `config:"burst"`
-	Max      int64         `config:"max"`
-	MaxBody  int64         `config:"max_body_byte_size"`
+	Interval            time.Duration `config:"interval"`
+	Burst               int           `config:"burst"`
+	Max                 int64         `config:"max"`
+	MaxBody             int64         `config:"max_body_byte_size"`
```
ycombinator, line 15: Should we rename the config setting as well for consistency (with the field name and also to remove units)?
```suggestion
	MaxBody             int64         `config:"max_body_decompressed"`
```

michel-laterman: `MaxBody` is an existing setting, changing the tags would be a breaking change

ycombinator: You're right, duh.

ycombinator (approved):

PR #7202, opened 2026-06-15
Use google/go for FIPS, build windows fips artifacts

michel-laterman:

Use go's FIPS module when compiling artifacts.

Replace all uses of microsoft/go which links to a crypto library on the system (such as OpenSSL) that has a FIPS provider to using go's newly certified FIPS provider that is a part of the standard library.

This is set at compile-time with the env var [`GOFIPS140=certified`](https://go.dev/doc/security/fips140#the-gofips140-environment-variable) variable that enables FIPS mode (but does not force GODEBUG=fips140=only).

Additionally this change builds and releases FIPS windows/amd64 and windows/arm64 artifacts.

`FIPS=true mage docker:release`

## Design Checklist

## Checklist

### DIFF
--- a/.buildkite/scripts/package.sh
+++ b/.buildkite/scripts/package.sh
@@ -4,7 +4,6 @@ set -euo pipefail
 
 source .buildkite/scripts/common.sh
 
-PLATFORM_TYPE=$(uname -m)
 TYPE="$1"
 readonly VERSION_QUALIFIER="${VERSION_QUALIFIER:-""}"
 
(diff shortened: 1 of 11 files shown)


.buildkite/scripts/package.sh
```diff
@@ -13,36 +12,9 @@ if [[ ${BUILDKITE_BRANCH} == "main" && ${TYPE} == "staging" && -z ${VERSION_QUAL
     exit 0
 fi
 
-PLATFORMS=""
-if [[ ${PLATFORM_TYPE} == "arm" || ${PLATFORM_TYPE} == "aarch64" ]]; then
-    PLATFORMS="linux/arm64"
```
michel-laterman, line 18: This was a bug, we didn't export PLATFORMS so all aritfacts were built every time

michel-laterman:
buildkite test this

blakerouse (approved):
I love seeing RED. So much simpler, looks good.

PR #7214, opened 2026-06-17
Improve performance of dl/dsl packages by lowering allocations

### LINKED ISSUE #7257: fleet-server dl/dsl improvements
Improvements to queries and data interactions via `internal/pkg/dl` and `internal/pkg/dsl`

michel-laterman:

Performance improvements to the `dl` and `dsl` packages.

Add benchmarks for all missing test cases, use the `b.Loop()` syntax to make benchmarks follow modern go best practices.

1. Fix bug with pre-allocation in `dls.Template.Resolve` which should reduce the total allocs.
2. ~~Use a fixed array instead of a heap allocted map for rendering tokens in queries. Array size is capped at 8 by default; only 4 tokens have been observed in existing queries.~~
    changes to allocate a slice with 2x len to avoid additional allocations
3. Change return value of `prepareQueryPolicies()` from `*dsl.Tmpl` to `[]byte`.
4. Partially unmarshal policies in order to only process the outputs when `dl.QueryOutputFromPolicy` is called in order to avoid expensive input unmarshal operation. The full policy is only unmarshalled if the output is found in the policy.

`mage test:benchmark`

```
  pkg: internal/pkg/dl
                                      baseline        after           delta
  FindAgentActionsRender  sec/op      642.5n ± 4%     512.7n ± 0%     -20.20% (p=0.000 n=10)
  QueryPoliciesRender     sec/op      48.50n ± 1%     46.20n ± 1%      -4.72% (p=0.000 n=10)
  QueryOutputFromPolicy   sec/op      475.6µ ± 1%     478.7µ ± 1%      +0.65% (p=0.015 n=10)  [noise]
  geomean                             2.456µ          2.247µ           -8.53%

  FindAgentActionsRender  B/op        1064.0          488.0           -54.14% (p=0.000 n=10)
  QueryPoliciesRender     B/op        96.00           96.00            ~
  QueryOutputFromPolicy   B/op        213.1Ki         213.1Ki          ~
  geomean                             2.748Ki         2.119Ki         -22.88%

  FindAgentActionsRender  allocs/op   9               9                ~
  QueryPoliciesRender     allocs/op   1               1                ~
  QueryOutputFromPolicy   allocs/op   6.511k          6.511k           ~
  geomean                             38.84           38.84            0.00%
```
```
  pkg: internal/pkg/dsl
                      baseline        after           delta
  RenderOne  sec/op   161.60n ± 1%    98.59n ± 1%    -38.99% (p=0.000 n=10)
  Render     sec/op   219.7n  ± 1%   151.9n  ± 1%    -30.88% (p=0.000 n=10)
  Render2    sec/op   264.5n  ± 1%   244.1n  ± 0%     -7.71% (p=0.000 n=10)
  geomean             625.8n          532.6n          -14.88%

  RenderOne  B/op     472.0           168.0           -64.41% (p=0.000 n=10)
  Render     B/op     472.0           168.0           -64.41% (p=0.000 n=10)
  Render2    B/op     408.0           344.0           -15.69% (p=0.000 n=10)
  geomean             684.0           471.2           -31.12%

  RenderOne  allocs   4               2               -50.00% (p=0.000 n=10)
  Render     allocs   4               2               -50.00% (p=0.000 n=10)
  Render2    allocs   3               3                ~
  geomean             6.188           4.912           -20.63%
```

Comparison of benchmarks when `-memprofile` was used:

```
  dsl package - net: -77.63 GB across the run (-25.81% of total allocations):

  ┌───────────────────────────┬───────────────────────┐
  │          Symbol           │         Delta         │
  ├───────────────────────────┼───────────────────────┤
  │ bytes.growSlice           │ -62.84 GB             │
  ├───────────────────────────┼───────────────────────┤
  │ bytes.(*Buffer).grow      │ -22.99 GB             │
  ├───────────────────────────┼───────────────────────┤
  │ (*Tmpl).render (removed)  │ -205.74 GB cumulative │
  ├───────────────────────────┼───────────────────────┤
  │ (*Tmpl).renderPairs (new) │ +119.91 GB cumulative │
  └───────────────────────────┴───────────────────────┘

  bytes.growSlice is the direct footprint of the bcnt bug. Before the fix, every call to render() triggered at least one internal buffer reallocation because buf.Grow(sum + t.bcnt) under-allocated. After the fix, t.bcnt correctly reflects the total static byte count, so Grow
  pre-sizes the buffer exactly and growSlice is never called during rendering. The -62 GB is purely that realloc chain disappearing across the benchmark run.
```
```
  dl package - net: -21.80 GB across the run (-15.43% of total allocations):

  ┌───────────────────────────┬───────────────────────┐
  │          Symbol           │         Delta         │
  ├───────────────────────────┼───────────────────────┤
  │ bytes.growSlice           │ -23.71 GB             │
  ├───────────────────────────┼───────────────────────┤
  │ (*Tmpl).render (removed)  │ -116.29 GB cumulative │
  ├───────────────────────────┼───────────────────────┤
  │ (*Tmpl).renderPairs (new) │ +92.58 GB cumulative  │
  ├───────────────────────────┼───────────────────────┤
  │ json.Unmarshal (dl)       │ -363 MB cumulative    │
  └───────────────────────────┴───────────────────────┘

  Same bcnt-fix story on the dl side: FindAgentActionsRender calls QueryAgentActions.Render(params) on every iteration, so the 4-token template was reallocating its buffer every single call. The -23.71 GB in bytes.growSlice is that. The -363 MB in json.Unmarshal reflects the
  policyOutputsOnly probe reducing the total bytes unmarshaled in the QueryOutputFromPolicy scan loop.
```

## Design Checklist

## Checklist


michel-laterman:
buildkite run perf-tests

michel-laterman:
buildkite run perf-tests

internal/pkg/dsl/tmpl.go
```diff
@@ -114,6 +129,9 @@ func (t *Tmpl) Resolve(n *Node) error {
 	if len(matches) != len(rmap) {
 		return ErrTokenUndefined
 	}
+	if len(rmap) > renderPairsCap {
+		return ErrTooManyTokens
```
cmacknz: Can we have a way to catch this at build or test time instead? So that somebody doesn't have to find out that fleet server is failing at runtime because they updated something?

michel-laterman: Resolve is either called directly by `monitor` with a max of 3 tokens (I'll add the missing unit tests to detect this error in case it changes), or it's called at package init time with `dsl.MustResolve` (which panics on an error); unit tests that use these queries, such as the ones in `dl` will panic if it has too many tokens.

In addition to those, I've added a static analyzer with claude to make sure we don't do extra binds

michel-laterman:
daily 30k test succeeded with these changes: https://buildkite.com/elastic/observability-perf/builds/6172
10k custom test also succeeded: https://buildkite.com/elastic/observability-perf/builds/6177

internal/pkg/dsl/tmpl.go
```diff
@@ -128,21 +146,22 @@ func (t *Tmpl) MustResolve(n *Node) *Tmpl {
...
+	var pair [1]namedBytes
+	pair[0] = namedBytes{name, d}
+	return t.renderPairs(pair[:], len(d))
 }
 
 func (t *Tmpl) Render(m map[string]any) ([]byte, error) {
-
-	// Marshal all targets, get byte count
-	marshalMap := make(map[string][]byte, len(m))
+	var fixed [renderPairsCap]namedBytes
```
cmacknz: Instead of maintaining an entire DSL checker and returning an error when exceeding the cap, could you just double the size of this slice when it needs to grow? I suspect 99% of this optimization is from not allocating for every pair anymore, and allocating once every 16 or twice every 32 tokens would still be better and more ergonomic than what was here before.

magefile.go
```diff
@@ -622,9 +622,29 @@ func (Check) Imports() error {
 	return sh.Run("go", "tool", "-modfile", filepath.Join("dev-tools", "go.mod"), "golang.org/x/tools/cmd/goimports", "-w", ".")
 }
 
+// Vet builds and runs the custom dslcheck analyzer with go vet.
+//
+// The dslcheck analayzer ensures that the number of Bind calls does not exceed the pre-allocated array.
```
ycombinator: Typo.

```suggestion
// The dslcheck analyzer ensures that the number of Bind calls does not exceed the pre-allocated array.
```

internal/pkg/dsl/tmpl.go
ycombinator: Could we add a unit test that exercises this error path?

dev-tools/static/dslcheck/testdata/go.mod
```diff
@@ -0,0 +1,3 @@
+module example.com/dslcheck-testdata
+
+go 1.26
```
ycombinator: Could we add this as a target to `.ci/bump-golang.yml`?

internal/pkg/dsl/tmpl.go
michel-laterman: I tried with the change:
```go
pairs := make([]nameBytes, 0, len(m)*2)
```

Results of benchstat:
```

  internal/pkg/dsl - vs 8c3e75b8

                  │     8c3e75b8 (baseline)      │       current (make)         │
                  │           sec/op            │   sec/op     vs base         │
  RenderOne-12             174.7n ± 1%   104.3n ± 0%  -40.27% (p=0.000 n=10)
  Render-12                230.3n ± 0%   176.7n ± 1%  -23.30% (p=0.000 n=10)
  MarshalNode-12           4.414µ ± 0%   4.481µ ± 1%   +1.51% (p=0.000 n=10)
  MarshalNode2-12          8.166µ ± 2%   8.276µ ± 0%   +1.34% (p=0.022 n=10)
  Sprintf-12               180.5n ± 1%   185.6n ± 1%   +2.80% (p=0.000 n=10)
  Render2-12               271.3n ± 1%   284.2n ± 3%   +4.79% (p=0.000 n=10)
  geomean                  643.5n        574.8n       -10.68%

                  │     B/op baseline           │   B/op current               │
  RenderOne-12             472.0          168.0   -64.41%
  Render-12                472.0          248.0   -47.46%
  MarshalNode-12         1.321Ki        1.321Ki       ~
  MarshalNode2-12        2.542Ki        2.542Ki       ~
  Sprintf-12               320.0          320.0       ~
  Render2-12               408.0          504.0   +23.53%

                  │   allocs/op baseline        │ allocs/op current             │
  RenderOne-12              4              2   -50.00%
  Render-12                 4              3   -25.00%
  MarshalNode-12           26             26       ~
  MarshalNode2-12          45             45       ~
  Sprintf-12                1              1       ~
  Render2-12                3              4   +33.33%
  geomean                6.188          5.513   -10.91%

  internal/pkg/dl - vs 8c3e75b8

                                │   8c3e75b8 (baseline)  │   current (make)         │
                                │        sec/op          │  sec/op     vs base      │
  FindAgentActionsRender-12       611.0n ± 1%   569.4n ± 1%   -6.80% (p=0.000 n=10)
  QueryPoliciesRender-12           48.55n ± 0%   42.51n ± 1%  -12.43% (p=0.000 n=10)
  QueryOutputFromPolicyUnmarshal   484.4µ ± 0%   491.0µ ± 0%   +1.36% (p=0.000 n=10)
  geomean                          2.431µ        2.282µ        -6.13%

                                │   B/op baseline        │  B/op current            │
  FindAgentActionsRender-12       1064.0        808.0   -24.06%
  QueryPoliciesRender-12            96.0         96.0       ~
  QueryOutputFromPolicyUnmarshal  213.1Ki      213.1Ki       ~

                                │ allocs/op baseline     │ allocs/op current         │
  FindAgentActionsRender-12          9         10   +11.11%
  QueryPoliciesRender-12             1          1       ~
  QueryOutputFromPolicyUnmarshal  6.511k     6.511k       ~
  geomean                          38.84      40.23    +3.57%

  ---
```

internal/pkg/monitor/monitor_test.go
```diff
@@ -0,0 +1,40 @@
...
+
+	for _, tc := range tests {
+		t.Run(tc.name, func(t *testing.T) {
+			m := &simpleMonitorT{
+				withExpiration: tc.withExpiration,
+				fetchSize:      defaultFetchSize,
+			}
+
+			_, err := m.prepareCheckQuery()
+			require.NoError(t, err, "prepareCheckQuery exceeds dsl.renderPairsCap")
```
cmacknz, line 34: There is no `dsl.renderPairsCap` anymore

internal/pkg/dsl/tmpl.go
```diff
@@ -128,21 +135,21 @@ func (t *Tmpl) MustResolve(n *Node) *Tmpl {
...
+	var pair [1]namedBytes
+	pair[0] = namedBytes{name, d}
+	return t.renderPairs(pair[:], len(d))
 }
 
 func (t *Tmpl) Render(m map[string]any) ([]byte, error) {
-
-	// Marshal all targets, get byte count
-	marshalMap := make(map[string][]byte, len(m))
+	pairs := make([]namedBytes, 0, len(m)*2)
```
cmacknz, line 152: Does the `*2` actually make a difference here? I was thinking of keeping this at 8 and growing x2 from that. I see in the loop below we are iterating over `m` so how would this ever grow past m?

PR #7215, opened 2026-06-17
Increase timeouts in e2e tests to 3m

michel-laterman:

Increase timeouts in e2e tests to 3m and use t.Context() as the root context in test cases. Timeout is increased from 1m to improve reliability in the buildkite test environment.
Add FreePort(8220) check to test tear down in AgentInstallTestSuite to ensure that the test that has not released the port is flagged instead of the next test
Increase timeouts of APM related tests to 10m to improve performance under load.

`mage test:e2e`


blakerouse (approved):
lgtm

PR #7263, opened 2026-06-22
Remove unused arg lookup in package.sh script

michel-laterman:

Failing to package daily release artifacts.

Remove unused arg lookup in package.sh script.

### DIFF
--- a/.buildkite/scripts/package.sh
+++ b/.buildkite/scripts/package.sh
@@ -4,7 +4,6 @@ set -euo pipefail
 
 source .buildkite/scripts/common.sh
 
-TYPE="$1"
 readonly VERSION_QUALIFIER="${VERSION_QUALIFIER:-""}"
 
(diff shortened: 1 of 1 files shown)


ebeahan (approved):

.buildkite/scripts/package.sh
```diff
@@ -4,7 +4,6 @@ set -euo pipefail
 
 source .buildkite/scripts/common.sh
 
-TYPE="$1"
 readonly VERSION_QUALIFIER="${VERSION_QUALIFIER:-""}"
 
 if [[ ${BUILDKITE_BRANCH} == "main" && ${TYPE} == "staging" && -z ${VERSION_QUALIFIER} ]]; then
```
ebeahan, line 9: Now seeing a new error without `TYPE` defined:

```

Running script
--
  | $ ./.buildkite/scripts/package.sh
  | ./.buildkite/scripts/package.sh: line 9: TYPE: unbound variable
```

Issue #7260, opened 2026-06-22
fleet-server policy monitor improvements


michel-laterman:
Performance improvements for fleet-server's `internal/pkg/policy` package

PR #7264, opened 2026-06-22
fleet-server align structs

### LINKED ISSUE #7262: fleet-server struct alignment
Investigate the use of https://github.com/essentialkaos/aligo or https://github.com/dkorunic/betteralign as an automated check to ensure struct alignment.

michel-laterman:

Structs in fleet-server codebase are padded. This results in more memory usage then required.

Use betteralign to automatically align go structs to get rid of padding.
Add check:align target to magefile and as a dependency of the check:ci target so changes to structs can be caught by the CI pipeline.

`mage check:align`

## Design Checklist

## Checklist

## Related issues

- Closes #7262


internal/pkg/file/cbor/chunk.go
```diff
@@ -188,8 +188,8 @@ func writeKey(buf []byte, key string) (int, error) {
 	}
 
 	buf[0] = byte(0x60 + keylen)
-	for i, c := range key {
-		buf[i+1] = byte(c)
+	for i, c := range []byte(key) { // convert string into []byte so any runes are properly handled
+		buf[i+1] = c
```
michel-laterman, line 192: This should prevent issues if a non ascii character is ever used as the key/val
cc: @pzl

michel-laterman (commented):
There are still some linter warnings for the policy and secrets pacakges which will be handled by #7260 and #7265

swiatekm (approved):

PR #7298, opened 2026-06-30
fix: propagate namespaces from enrollment key to OpAMP agent on enroll

juliaElastic:
## Summary

Fixes a bug where OTel (OpAMP) collectors enrolled from a non-default Kibana space always appeared in the **Default** space.

### Root cause

`enrollAgent` in `handleOpAMP.go` fetches the enrollment API key record (`rec`) and correctly sets `PolicyID: rec.PolicyID` on the new agent document — but never set `Namespaces`. The standard enrollment handler (`handleEnroll.go`) passes `enrollAPI.Namespaces` through to `_enroll`, which stamps the agent's `namespaces` field and allows Kibana's space-awareness filtering to work correctly.

### Fix

One line: add `Namespaces: rec.Namespaces` to the `model.Agent` struct in `enrollAgent`, mirroring the standard enroll path.

### Test

Updated `TestEnrollAgentWithAgentToServerMessage` to include `Namespaces: []string{"my-space"}` in the mock enrollment key and assert that both the returned agent and the written ES document have `Namespaces: ["my-space"]`.

Companion Kibana fix: elastic/kibana#275602
Closes: https://github.com/elastic/kibana/issues/275528

### DIFF
--- a/internal/pkg/api/handleOpAMP.go
+++ b/internal/pkg/api/handleOpAMP.go
@@ -386,6 +386,7 @@ func (oa *OpAMPT) enrollAgent(ctx context.Context, zlog zerolog.Logger, uid uuid
 		Active:     true,
 		EnrolledAt: now.UTC().Format(time.RFC3339),
 		PolicyID:   rec.PolicyID,
+		Namespaces: rec.Namespaces,
 		Agent: &model.AgentMetadata{
 			ID:      instanceUID,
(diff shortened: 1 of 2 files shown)


swiatekm (approved):

juliaElastic:
@swiatekm Thanks for the review, could you merge the PR, I don't have access.

PR #7319, opened 2026-07-04
File delivery range support

pzl:

Enables clients (e.g. endpoint) to gracefully resume downloads of (large) files via the HTTP Range header.

This allows for completing large file transfers under unstable or periodic network conditions, as opposed to starting the file download over.

**N.B.**: 
- This only supports single ranges (e.g. `bytes=100-400`) and not multipart range requests (e.g. `bytes=0-500, 1024-2048, -600`).  
- Also HTTP Spec allows for graceful degradation here. For unclear range headers (or for any reason if the server chooses not to fulfill a particular range), either a `416 Range Unsatisfiable` error OR `HTTP 200` with full file contents are both acceptable responses.  This implementation mostly generates 416s for clients to handle malformed byte strings, but in narrow cases where intent is unclear (e.g. the header is present without a value: `Range: `) then we do fall back to HTTP 200 with file contents, as spec defines.

Only fetch chunks that fall inside the requested range

## Design Checklist

## Checklist

### DIFF
--- a/internal/pkg/api/handleFileDelivery.go
+++ b/internal/pkg/api/handleFileDelivery.go
@@ -169,8 +169,8 @@ func (ft *FileDeliveryT) sendFileAsRanges(zlog zerolog.Logger, w http.ResponseWr
 
 	w.WriteHeader(http.StatusPartialContent)
 
-	startOffset := uint64(ra.start % info.File.ChunkSize)
-	endOffset := uint64(((ra.start + ra.length - 1) % info.File.ChunkSize) + 1)
+	startOffset := ra.start % info.File.ChunkSize
+	endOffset := ((ra.start + ra.length - 1) % info.File.ChunkSize) + 1
 	return ft.deliverer.SendChunks(r.Context(), zlog, w, chunks, fileID, &startOffset, &endOffset)
 }
 
--- a/internal/pkg/api/handleFileDelivery.go
+++ b/internal/pkg/api/handleFileDelivery.go
@@ -170,6 +170,9 @@ func (ft *FileDeliveryT) sendFileAsRanges(zlog zerolog.Logger, w http.ResponseWr
 	}
 	w.Header().Set("Content-Range", ra.contentRange(info.File.Size))
 	w.Header().Set("Content-Length", strconv.FormatInt(ra.length, 10))
+	if info.File.Hash != nil && info.File.Hash.SHA2 != "" {
+		w.Header().Set("X-File-SHA2", info.File.Hash.SHA2)
+	}
 
 	startOffset := ra.start % info.File.ChunkSize
 	endOffset := ((ra.start + ra.length - 1) % info.File.ChunkSize) + 1
--- a/internal/pkg/api/handleFileDelivery.go
+++ b/internal/pkg/api/handleFileDelivery.go
@@ -228,7 +228,7 @@ func parseRange(s string, size int64) ([]httpRange, error) {
 				return nil, errors.New("invalid range")
 			}
 			i, err := strconv.ParseInt(end, 10, 64)
-			if i < 0 || err != nil {
+			if i <= 0 || err != nil {
 				return nil, errors.New("invalid range")
 			}
 			if i > size {
--- a/internal/pkg/api/handleFileDelivery.go
+++ b/internal/pkg/api/handleFileDelivery.go
@@ -8,7 +8,9 @@ import (
 	"errors"
 	"fmt"
 	"net/http"
+	"net/textproto"
 	"regexp"
+	"sort"
 	"strconv"
 	"strings"
 
@@ -24,6 +26,7 @@ import (
 )
 
 var (
+	ErrBadRange                = errors.New("range not satisfiable")
 	ErrClientFileForbidden     = errors.New("agent not authorized for library")
 	ErrFileForDeliveryNotFound = errors.New("unable to retrieve file")
 	ErrLibraryFileNotFound     = fmt.Errorf("%w from library", ErrFileForDeliveryNotFound)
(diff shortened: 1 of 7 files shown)


pzl:
@elastic/elastic-agent-control-plane  sorry for the short turnaround, could this be looked over before FF?

It is additional behavior only:  when serving files through the file delivery route, if a `Range` header is present, delivers the subset of bytes instead of the whole file (follows HTTP spec). Oherwise all behavior is the same

internal/pkg/api/handleFileDelivery.go
```diff
@@ -112,10 +135,145 @@ func (ft *FileDeliveryT) handleSendFile(zlog zerolog.Logger, w http.ResponseWrit
...
+
+	// if supporting multiple ranges, this becomes multipart/byteranges !
+	w.Header().Set("Content-Type", "application/octet-stream")
+	if info.File.MimeType != "" {
+		w.Header().Set("Content-Type", info.File.MimeType)
+	}
+	w.Header().Set("Content-Range", ra.contentRange(info.File.Size))
+	w.Header().Set("Content-Length", strconv.FormatInt(ra.length, 10))
+
+	w.WriteHeader(http.StatusPartialContent)
```
cmacknz: Setting this here with WriteHeader means it can't be changed later if `SendChunks` fails. It looks like the implementation of `WriteHeader` only allows it to be called once.

https://cs.opensource.google/go/go/+/refs/tags/go1.26.4:src/net/http/server.go;l=1155-1159

internal/pkg/api/handleFileDelivery.go
```diff
@@ -112,10 +135,145 @@ func (ft *FileDeliveryT) handleSendFile(zlog zerolog.Logger, w http.ResponseWrit
...
-	return ft.deliverer.SendFile(r.Context(), zlog, w, chunks, fileID)
+	return ft.deliverer.SendChunks(r.Context(), zlog, w, chunks, fileID, nil, nil)
 }
 
 func sanitizedIndexInput(s string) string {
 	r := regexp.MustCompile(`[*?"<>\\/,|#]`) // bad characters
 	return r.ReplaceAllString(strings.ToLower(strings.TrimSpace(s)), "")
 }
+
+func (ft *FileDeliveryT) sendFileAsRanges(zlog zerolog.Logger, w http.ResponseWriter, r *http.Request, fileID string, info file.MetaDoc, chunks []file.ChunkInfo, ranges []httpRange) error {
```
cmacknz, line 147: You don't set `X-File-SHA2` on this path, should you?

internal/pkg/api/handleFileDelivery.go
```diff
@@ -112,10 +135,145 @@ func (ft *FileDeliveryT) handleSendFile(zlog zerolog.Logger, w http.ResponseWrit
...
+	if len(ranges) > 1 {
+		return errors.New("multipart ranges not supported")
```
cmacknz: This will end up as a 500 I think, since the table in `NewHTTPErrResp` doesn't recognize this error

https://github.com/elastic/fleet-server/blob/2d294ecb467290346a029c11d1145cf24ebb3f01/internal/pkg/api/error.go#L76

internal/pkg/api/handleFileDelivery.go
```diff
@@ -112,10 +135,145 @@ func (ft *FileDeliveryT) handleSendFile(zlog zerolog.Logger, w http.ResponseWrit
...
+	start, length int64
+}
+
+func (r httpRange) contentRange(size int64) string {
+	return fmt.Sprintf("bytes %d-%d/%d", r.start, r.start+r.length-1, size)
+}
+
+// parseRange parses a Range header string as per RFC 7233.
+// errNoOverlap is returned if none of the ranges overlap.
+func parseRange(s string, size int64) ([]httpRange, error) {
```
cmacknz, line 212: Claude thinks this will `Range: bytes=-0` which will cause us to produce a 206 with a `Content-Range` where start > end. This seems plausible but I can't see it myself just from reading this.

internal/pkg/api/handleFileDelivery.go
```diff
@@ -112,10 +135,145 @@ func (ft *FileDeliveryT) handleSendFile(zlog zerolog.Logger, w http.ResponseWrit
...
+
+	sort.SliceStable(chunks, func(i, j int) bool {
+		return chunks[i].Pos < chunks[j].Pos
+	})
+
+	// reduce to fetch only required chunks.
+	// Int64 Floor division used intentionally for index math
+	chunkIdxStart := ra.start / info.File.ChunkSize
+	chunkIdxStop := (ra.start + ra.length - 1) / info.File.ChunkSize
+	chunks = chunks[chunkIdxStart : chunkIdxStop+1]
```
cmacknz, line 164: I would be defensive and make sure that the indexes you calculate actually fit into `len(chunks)` to avoid a panic here.

internal/pkg/api/handleFileDelivery.go
pzl: ambiguous when sending byte ranges (would it apply to the range or still the hash of the whole file?) so it wasn't used for range requests. It could be, it would just have to be an explicit call if it were needed

internal/pkg/api/handleFileDelivery.go
pzl: `parseRange("bytes=-0", 1024)` gives ~~`nil, nil`~~  `httpRange{length:0}, nil`  which ends up as no ranges specified, so you get `HTTP 200` and full file contents

Will add test cases for this and clear up

internal/pkg/api/handleFileDelivery.go
```diff
@@ -112,10 +136,157 @@ func (ft *FileDeliveryT) handleSendFile(zlog zerolog.Logger, w http.ResponseWrit
...
+	}
+	ra := ranges[0]
+
+	sort.SliceStable(chunks, func(i, j int) bool {
+		return chunks[i].Pos < chunks[j].Pos
+	})
+
+	// reduce to fetch only required chunks.
+	// Int64 Floor division used intentionally for index math
+	chunkIdxStart := ra.start / info.File.ChunkSize
```
cmacknz, line 159: This will panic if `ChunkSize` is ever 0, but ChunkSize doesn't come from user provided input so this is unlikely.

cmacknz (approved):

internal/pkg/api/handleFileDelivery.go
pzl: chunksize is also effectively static at `file.MaxChunkSize` (4MiB). Can see examples of that at `file/uploader/upload.go:107`, where it is not even an input, just fixed at the server-decided size.

Delivery takes a slightly different route for doc population, but is also a fixed value.

This is a good call out, though. I might add a followup fix for the just-in-case, even if not possible during normal usage.

PR #7330, opened 2026-07-07
Add checkin capacity-rejection rate metric to /stats

ycombinator:

The EPA autoscaler for Fleet Server (configured in fleet-controller) currently scales on `http_server.tcp_active` (active TCP connections). This has repeatedly failed to scale up pods fast enough under load in Serverless scale tests — including the 20k → 30k drone check-in test tracked in [elastic/ingest-dev#7719](https://github.com/elastic/ingest-dev/issues/7719), and other runs such as a recent 5k test where the autoscaler did not add pods when it should have.

The root cause: when a pod is saturated, agents being turned away don't count as "active" connections, so `tcp_active` understates real demand and the autoscaler under-reacts.

In [this comment](https://github.com/elastic/ingest-dev/issues/7719#issuecomment-4810371846), @cmacknz suggested scaling on the rate of check-ins rejected because Fleet Server was at capacity instead — a server-side saturation signal that (unlike checkin volume) can't be inflated by a valid-but-misbehaving or malicious client.

Fleet Server already counts these rejections (`http_server.routes.checkin.limit_max`), but it's a monotonically increasing counter, not a rate — not directly usable as an autoscaler trigger (the EPA reads metric values as levels, not deltas).

Adds a new gauge, `http_server.routes.checkin.limit_max_rate`, to the internal `/stats` endpoint. A new background subsystem (`api.RunCheckinRejectionRateSampler`, wired into `Fleet.runSubsystems` alongside the other subsystems) samples the existing `limit_max` counter every 10s and publishes the delta as a rejections/sec rate.

This is intentionally computed **inside Fleet Server**, not left to the consumer (e.g. an autoscaler) to derive from the raw counter, so the rate is available directly via `/stats` in a form ready to use as a metric target.

This PR only adds the metric; it does not change any existing behavior or defaults. A follow-up draft PR to fleet-controller (elastic/fleet-controller#623) wires the EPA to this new metric as an additional (not replacing) autoscaling trigger.

**⚠️ Deployment sequencing note:** the companion fleet-controller change must not be enabled until this change has been rolled out to **all** Fleet Server pods. The EPA has no fallback for a metric that's missing from a pod's `/stats` response — a missing metric aborts that reconcile cycle entirely (including evaluation of the existing `tcp_active` trigger), rather than degrading gracefully. See the fleet-controller PR for how this is gated (metric is opt-in via config, defaulted off).

1. Run fleet-server with the internal HTTP server enabled (`http.enabled: true`, default port `5066`).
2. `curl localhost:5066/stats | jq .http_server.routes.checkin.limit_max_rate` — should read `0` at idle.
3. Drive check-ins past `checkin_limit.max` (e.g. via a load test) and confirm the value rises, then falls back to `0` once rejections stop.
4. Or run `go test ./internal/pkg/api/... -run TestComputeRate` and `-run TestRunCheckinRejectionRateSampler` for the unit-level behavior, and `go test -tags integration ./internal/pkg/api/... -run TestMetricsEndpoints` for the `/stats` JSON shape.

## Design Checklist

- [x] I have ensured my design is stateless and will work when multiple fleet-server instances are behind a load balancer. (Per-pod gauge, sampled from per-pod in-memory counters only — no shared state.)
- [ ] I have or intend to scale test my changes, ensuring it will work reliably with 100K+ agents connected. (Intended as part of the ingest-dev#7719 scale test follow-up; not yet run at 100K+.)
- [x] I have included fail safe mechanisms to limit the load on fleet-server: rate limiting, circuit breakers, caching, load shedding, etc. (Purely observational — a read-only sampler goroutine on a 10s ticker; adds no load to the request path.)

## Checklist

- [x] I have made corresponding change to the default configuration files (none needed — no new config)
- [x] I have added an entry in `./changelog/fragments` using the changelog tool

## Related issues

- Relates elastic/ingest-dev#7719

### DIFF
--- a/changelog/fragments/1783459187-checkin-capacity-rejection-rate-metric.yaml
+++ b/changelog/fragments/1783459187-checkin-capacity-rejection-rate-metric.yaml
@@ -1,0 +1,17 @@
+kind: enhancement
+
+summary: Expose a checkin capacity-rejection rate metric for use as an autoscaling signal
+
+description: |
+  Adds a new http_server.routes.checkin.limit_max_rate gauge to the internal
+  /stats endpoint. It is a rolling, server-computed rate (rejections/sec) of the
+  existing http_server.routes.checkin.limit_max counter, which tracks checkins
+  rejected because Fleet Server was at capacity (the per-route concurrency limit
+  was reached).
+
+  Unlike checkin volume, this counter can only be incremented by Fleet Server
+  itself, so it can't be inflated by a client checking in unusually often. This
+  makes the rate a safer signal than raw checkin throughput for driving
+  autoscaling decisions elsewhere (e.g. fleet-controller's EPA configuration).
+
+component: fleet-server
--- a/internal/pkg/api/metrics.go
+++ b/internal/pkg/api/metrics.go
@@ -9,6 +9,7 @@ import (
 	"errors"
 	"fmt"
 	"sync"
+	"time"
 
 	"github.com/prometheus/client_golang/prometheus"
 	"github.com/prometheus/client_golang/prometheus/promhttp"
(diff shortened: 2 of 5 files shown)


macdewee (commented):
LGTM

ycombinator:
@cavokz I am unable to add you as a reviewer on this PR but would like your input as well.  If not specifically on the implementation then at least on the idea behind this enhancement.  Thank you!

samuelvl (approved):

cmacknz (approved):

cmacknz:
Nice, will be very curious to see how well this works once wired up to the autoscaler.

cavokz (approved):

PR #7332, opened 2026-07-08
Bump elastic-agent-libs and enforce fips140=on at runtime for FIPS builds

macdewee:

Fleet Server's FIPS builds already moved to Go's native FIPS 140-3 support (`GOFIPS140`), but two gaps remained compared to the same migration done in beats/elastic-agent:

- `elastic-agent-libs` was still pinned to v0.44.0, which predates a fix enforcing FIPS 140-3 peer-certificate key-type checks across all TLS verification modes. Fleet Server uses this library's `tlscommon`/`httpcommon` packages for all of its TLS configuration, so FIPS builds could accept non-compliant TLS certificates (e.g. RSA keys below 2048 bits).
- FIPS binaries never actually enforced `fips140=on` at runtime by default — `checkFIPSBinary` only verified the binary was compiled with `GOFIPS140` available, not that it enforces FIPS mode when run.

- Bumped `elastic-agent-libs` to v0.46.0 in `go.mod` and `testing/go.mod`, regenerating `NOTICE.txt`/`NOTICE-fips.txt`.
- Added a `//go:debug fips140=on` directive in a new `requirefips`-gated file (`godebug_fips.go`), so FIPS binaries embed `DefaultGODEBUG=fips140=on` and enforce FIPS mode by default at runtime.
- Extended `checkFIPSBinary` in `magefile.go` to fail if a FIPS binary's build info doesn't show `DefaultGODEBUG` containing `fips140=on`.

```
FIPS=true PLATFORMS=linux/amd64 mage build:local
go version -m bin/fleet-server | grep -E 'GOFIPS140|DefaultGODEBUG|-tags'
```

Confirm `DefaultGODEBUG=fips140=on` is present, and that a non-FIPS build (`mage build:local` without `FIPS=true`) shows none of these markers.

`mage test:unit` and `mage check:imports check:headers check:fix check:notice` all pass.

## Design Checklist

## Checklist

### DIFF
--- a/magefile.go
+++ b/magefile.go
@@ -1669,8 +1669,11 @@ func checkFIPSBinary(path string) error {
 				return fmt.Errorf("GOFIPS140 is empty")
 			}
 		case "DefaultGODEBUG":
-			if strings.Contains(setting.Value, "fips140=on") {
-				foundFIPSDefault = true
+			for _, entry := range strings.Split(setting.Value, ",") {
+				if key, val, ok := strings.Cut(entry, "="); ok && key == "fips140" && val == "on" {
+					foundFIPSDefault = true
+					break
+				}
 			}
 		}
 	}
--- a/magefile.go
+++ b/magefile.go
@@ -1640,7 +1640,7 @@ func checkFIPSBinary(path string) error {
 	if err != nil {
 		return fmt.Errorf("unable to read buildinfo: %w", err)
 	}
-	var foundTags, foundFIPS140 bool
+	var foundTags, foundFIPS140, foundFIPSDefault bool
 
 	for _, setting := range info.Settings {
 		switch setting.Key {
@@ -1668,6 +1668,10 @@ func checkFIPSBinary(path string) error {
 			if setting.Value == "" {
 				return fmt.Errorf("GOFIPS140 is empty")
 			}
+		case "DefaultGODEBUG":
+			if strings.Contains(setting.Value, "fips140=on") {
+				foundFIPSDefault = true
+			}
 		}
 	}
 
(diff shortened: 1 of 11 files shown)


magefile.go
```diff
@@ -1668,6 +1668,10 @@ func checkFIPSBinary(path string) error {
 			if setting.Value == "" {
 				return fmt.Errorf("GOFIPS140 is empty")
 			}
+		case "DefaultGODEBUG":
+			if strings.Contains(setting.Value, "fips140=on") {
```
ycombinator: This would also catch `fips140=only`, which we probably don't want since that could lead to unexpected panics at runtime for FIPS-140 non-compliant uses.  Consider a more structured approach like so:

```go
case "DefaultGODEBUG":
    for _, entry := range strings.Split(setting.Value, ",") {
        if key, val, ok := strings.Cut(entry, "="); ok && key == "fips140" && val == "on" {
            foundFIPSDefault = true
            break
        }
    }
```

ycombinator:
Thanks for enforcing `GODEBUG=fips140=on` at a code/build level.  To clarify, this was not a gap for Docker-based deployments before as we were enforcing this at the packaging level: https://github.com/elastic/fleet-server/blob/7caed5a18b31319b843dee767336d96d06d7eaed/Dockerfile.fips#L59  

This form of enforcement was, of course, only limited to Docker images since our goal at the time was FRH environment deployments in ESS. 

All that said, the change in this PR is better as it enforces this now regardless of the packaging artifact used for deployment.  Maybe we could remove the Dockerfile line now. Or we could leave it as-is for defense in depth.  WDYT?

magefile.go
macdewee: good catch, I will fix it in beats and agent as well

macdewee:
> All that said, the change in this PR is better as it enforces this now regardless of the packaging artifact used for deployment. Maybe we could remove the Dockerfile line now. Or we could leave it as-is for defense in depth. WDYT?

I would leave it. It is not a cost and gives as additional defense in depth

ycombinator (approved):

ycombinator:
I'm going to backport this PR to all active branches, otherwise we cannot bump github.com/elastic/elastic-agent-libs to 0.46.1 in those branches (see backports of https://github.com/elastic/fleet-server/pull/7385).

PR #7352, opened 2026-07-13
Fallback Chunk Size in Range calculations

pzl:

Follow-up to #7319, adds fallback to known default chunk size when chunk size is omitted in document metadata

tests updated to include this case

(reviewers: is changelog needed for this? This is a small addition to #7319 which hasn't been released yet)


cmacknz (approved):

pzl:
@Mergifyio backport 9.5

PR #7416, opened 2026-07-20
Store output API key secrets in .fleet-secrets instead of .fleet-agents

ycombinator:

The `kibana_system` built-in role has broad `read` (actually `all`) access to `.fleet-agents*`. Fleet Server currently writes output API key secrets as plaintext `id:secret` strings into `outputs.{name}.api_key` on agent documents in `.fleet-agents-7`. Any principal with `kibana_system` credentials can therefore extract these secrets and use them to write to data streams that `kibana_system` itself cannot access — a privilege escalation.

Instead of storing the raw `id:secret` string in `.fleet-agents-7`, Fleet Server now:

1. **On new key generation** (`policy_output.go`): writes the key material to `.fleet-secrets` via `POST /_fleet/secret` and stores only a `$co.elastic.secret{<secretId>}` reference in `outputs.{name}.api_key`.
2. **At agent checkin** (`policy_output.go`): detects the reference format and resolves it via `GET /_fleet/secret/{id}` before injecting `api_key` into the policy payload sent to the agent.
3. **On ACK / key rotation** (`handleAck.go`): deletes the retired secret from `.fleet-secrets` via `DELETE /_fleet/secret/{id}`. The secret ID is carried in a new `secret_id` field on `ToRetireAPIKeyIdsItems` (schema change in `model/schema.json`).

**Backward compatibility**: existing plaintext values already stored in `.fleet-agents-7` continue to work unchanged. The check `secret.ParseSecretReference(output.APIKey)` returns `false` for plain `id:secret` strings, so they are used directly. Agents migrate to the new format naturally on the next key rotation.

**Dependency**: requires [elastic/elasticsearch#154498](https://github.com/elastic/elasticsearch/pull/154498) (granting `write_fleet_secrets` cluster privilege to the `elastic/fleet-server` service account) to be merged and deployed first.

Run the integration test suite against a stack with [elastic/elasticsearch#154498](https://github.com/elastic/elasticsearch/pull/154498) applied:

```
mage integration:test
```

Verify that:
- Newly enrolled agents receive a valid `api_key` in checkin responses
- `.fleet-agents-7` documents show `$co.elastic.secret{...}` in `outputs.*.api_key` rather than a plaintext key
- `.fleet-secrets` contains the corresponding entries
- After a policy change that rotates the key, the old secret is deleted from `.fleet-secrets`

## Design Checklist

- [x] I have ensured my design is stateless and will work when multiple fleet-server instances are behind a load balancer. (Secrets are resolved per-request from ES; no local state.)
- [ ] I have or intend to scale test my changes, ensuring it will work reliably with 100K+ agents connected. (Each checkin that needs to rotate a key adds one `POST /_fleet/secret` write; reads are already part of the checkin path via `ReadSecrets`.)
- [x] I have included fail safe mechanisms to limit the load on fleet-server: rate limiting, circuit breakers, caching, load shedding, etc. (Secret writes/deletes are only on key rotation, not every checkin.)

## Checklist

## Related issues

- Requires elastic/elasticsearch#154498


internal/pkg/policy/policy_output.go
lorienhu: I think SecretID needs to be set here too or it won't get picked up by deleteRetiredSecrets.

ycombinator: Added the secret ID in 7c8ef4e9d1c8b2b0ce9b7fa928652c91122079b0.

internal/pkg/bulk/secret.go
```diff
@@ -46,6 +47,38 @@ func (c *ExtendedAPI) Read(ctx context.Context, secretID string) (*SecretRespons
...
+	}
+
+	req, err := http.NewRequestWithContext(ctx, "POST", "/_fleet/secret", bytes.NewReader(body))
+	if err != nil {
+		return "", err
+	}
+	req.Header.Set("Content-Type", "application/json")
+	req.Header.Set("Accept", "application/json")
+
+	res, err := c.Perform(req)
```
samuelvl, line 73: `c.Perform` returns no error if ES returns 4xx/5xx and `resp.ID` will be empty, causing a bad `$co.elastic.secret{}` reference to be silently stored in the agent document.

It would be good to fix/document `c.Perform()` behavior because it's easy to make a mistake.

internal/pkg/policy/policy_output.go
```diff
@@ -307,8 +312,14 @@ func (p *Output) prepareElasticsearch(
 			Str(ecs.DefaultOutputAPIKeyID, outputAPIKey.ID).
 			Msg("Updating agent record to pick up default output key.")
 
+		secretID, err := bulker.WriteSecret(ctx, outputAPIKey.Agent())
```
samuelvl, line 315: For existing deployments, can we ensure ES will be upgraded before fleet-server so it has the write_fleet_secrets privilege when this is reached? Or will it fail if ES is not upgraded yet? Should we fallback to the old behavior if that happens?

internal/pkg/policy/policy_output.go
```diff
@@ -362,7 +377,15 @@ func (p *Output) prepareElasticsearch(
...
 	// agent policy to multiple agents.
 	// See: https://github.com/elastic/fleet-server/issues/1301
-	outputMap[p.Name]["api_key"] = output.APIKey
+	apiKey := output.APIKey
+	if secretID, ok := secret.ParseSecretReference(apiKey); ok {
+		resolved, err := bulker.ReadSecrets(ctx, []string{secretID})
+		if err != nil {
+			return fmt.Errorf("failed resolving output API key secret: %w", err)
+		}
+		apiKey = resolved[secretID]
```
samuelvl: Should we check if secretID exists in resolved to avoid delivering an empty api_key? The secret might have been deleted between write and read

internal/pkg/api/handleAck.go
lorienhu, line 579: Could we retire secrets here? Looks like unenroll keys are missed right now.

lorienhu: Actually, it might be better to do this in ext.go Agent.APIKeyIDs when retirement entries are added.

internal/pkg/policy/policy_output.go
ycombinator: Fleet-server enforces `ES >= fleet_major.fleet_minor.0` at startup via [`ver.CheckCompatibility`](https://github.com/elastic/fleet-server/blob/742984f1d6a795231898938fda4d3fd8d2268a6e/internal/pkg/ver/check.go#L33). Since this PR and elastic/elasticsearch#154498 are targeting the same minor release, ES will already carry the `write_fleet_secrets` privilege on the service account by the time any fleet-server instance with this code is running.

If somehow the privilege is absent (e.g. standalone mode, where the version check is skipped), the `POST /_fleet/secret` call returns a 4xx. There's a separate fix in this PR for `bulk/secret.go` that makes `Write()` treat non-2xx as an error — once that lands, `WriteSecret` will return a real error, checkin will fail, and the agent will retry. That's the right failure mode: I don't think we want to silently fall back to storing plaintext, since that would re-introduce the privilege-escalation window this PR is closing?

internal/pkg/policy/policy_output.go
ycombinator: Fixed in 8ea5c20d.

internal/pkg/bulk/secret.go
ycombinator: Added error handling for status code >= 400 in 0aa3e352.  Let me know what you think.

internal/pkg/api/handleAck.go
ycombinator: Good catch — `handleUnenroll` invalidates the API keys but never deletes their secrets.

On the `ext.go` approach: `APIKeyIDs()` can't call `secret.ParseSecretReference` because the `secret` package imports `model`, so adding that import would be circular. Instead, we could handle it directly in `handleUnenroll`: after `ack.invalidateAPIKeys`, iterate over `agent.Outputs`, parse any secret reference from each `output.APIKey`, and call `deleteRetiredSecrets` with the resulting items. The `ToRetireAPIKeyIds` entries already carry `SecretID`, so those get handled for free — the only gap is the current output key, which is what the extra iteration covers.

WDYT?

internal/pkg/policy/policy_output.go
samuelvl: Thanks for the detailed explanation! I think retrying is correct in this case 👍

samuelvl (dismissed):

internal/pkg/api/handleAck.go
blakerouse: @ycombinator I believe that flow of doing it in `handleUnenroll` seems correct.

ycombinator: Thank you both.  Implemented in 020d99e5.

lorienhu (dismissed):

blakerouse (commented):
Overall the code looks good, but I do have a worry. Correct me if I am off base here.

Being that the API key is now in secrets and this is not running like a transactional database across two tables there is a case that might not be handled correctly.

Write the secret to the secret store, but then writing the reference into `.fleet-agents` fails. I don't see where if that fails the secret is then deleted. I feel like it would need to be deleted then, because it is just left in there with no reference.

Also what happens if writing into the secret store fails? Does that just error back to the caller and they make the API call again? Seems like it, I believe that is correct.

ycombinator:
> Write the secret to the secret store, but then writing the reference into `.fleet-agents` fails. I don't see where if that fails the secret is then deleted. I feel like it would need to be deleted then, because it is just left in there with no reference.

Yep, good catch.  This is a gap.  Fixed now in 0818c0ebf69a2ff613e7f911dca95578b710ddf7.

> Also what happens if writing into the secret store fails? Does that just error back to the caller and they make the API call again? Seems like it, I believe that is correct.

Yes, that's what happens.  This [early return](https://github.com/elastic/fleet-server/pull/7416/changes#diff-85caf77b270865c04855a90fc60468c90130cd83d2218aa66f362b32c178f4e0R317) when `bulker.WriteSecret` fails means no secret is written and the Agent doc is not updated either.  When Agent makes the API call again, the whole flow will be retried.

blakerouse (approved):
Nice! Thanks for the fixes.

PR #7422, opened 2026-07-21
[Fleet] Write policy_base_id field at agent enrollment

juliaElastic:
## Summary

Adds a denormalized `policy_base_id` field written at agent enrollment to enable cheap `term`/`terms` queries in place of expensive prefix/wildcard queries.

### Background

Fleet supports version-specific agent policies where `policy_id` gets a `#major.minor` suffix (e.g. `my-policy#9.2`). Querying agents by policy currently requires:

```json
{ "bool": { "should": [{ "term": { "policy_id": "my-policy" } }, { "prefix": { "policy_id": "my-policy#" } }] } }
```

The `policy_base_id` field stores the canonical base ID (suffix stripped), enabling a single `term: { policy_base_id: "my-policy" }` query instead.

### Changes

- **`internal/pkg/dl/constants.go`**: Add `FieldPolicyBaseID = "policy_base_id"` constant
- **`internal/pkg/model/schema.go`**: Add `PolicyBaseID string json:"policy_base_id,omitempty"` field to the `Agent` model
- **`internal/pkg/api/handleEnroll.go`**: Set `PolicyBaseID: policyBaseID(policyID)` in the agent struct at enrollment; add `policyBaseID` helper that strips the `#version` suffix

### Notes

- Checkin (bulk.go) writes `agent_policy_id` via partial update — no change needed there; `policy_base_id` tracks `policy_id`, not `agent_policy_id`
- Kibana backfills existing documents via `update_by_query` during Fleet setup (companion PR: elastic/kibana#279716)
- Companion ES mapping PR: elastic/elasticsearch#154521

Part of elastic/ingest-dev#8624 (Workstream B2).


ycombinator (commented):
### `internal/pkg/api/handleEnroll_test.go` — `TestEnrollWithAgentIDExistingActive` (lines 382–410)

`TestEnrollWithAgentIDExistingActive` exercises the re-enrollment `UpdateFields` path — the exact path where the original omission was caught by Copilot review. The mock accepts any args on `Update`, so a missing `FieldPolicyBaseID` would pass silently. Suggested change: use a versioned policy ID and assert the field is written.

```go
// before (lines 382–410)
	source := fmt.Sprintf(`{"active":true,"agent":{"id":"1234","version":"8.9.0"},"type":"PERMANENT","policy_id":"1234","replace_token":"%s"}`, replaceHash)
	// ... bulker mock setup ...
	resp, _ := et._enroll(ctx, rb, zlog, req, "1234", []string{}, "8.9.0")
	// only resp.Action and resp.Item.Id are asserted
}
```

```go
// after
	policyID := "my-policy#9.2"
	source := fmt.Sprintf(`{"active":true,"agent":{"id":"1234","version":"8.9.0"},"type":"PERMANENT","policy_id":"%s","replace_token":"%s"}`, policyID, replaceHash)
	// ... bulker mock setup unchanged ...
	resp, _ := et._enroll(ctx, rb, zlog, req, policyID, []string{}, "8.9.0")

	if resp.Action != "created" {
		t.Fatal("enroll failed")
	}
	if resp.Item.Id != agentID {
		t.Fatalf("agent ID should have been %s (not %s)", agentID, resp.Item.Id)
	}

	var updateBody []byte
	for _, c := range bulker.Calls {
		if c.Method == "Update" {
			updateBody = c.Arguments.Get(3).([]byte)
			break
		}
	}
	var updateDoc struct {
		Doc map[string]any `json:"doc"`
	}
	assert.NoError(t, json.Unmarshal(updateBody, &updateDoc))
	assert.Equal(t, "my-policy", updateDoc.Doc[dl.FieldPolicyBaseID])
}
```

ycombinator:
While this PR handles the enrollment path, my concern was around the policy reassignment path. But it looks like the companion Kibana PR takes care of setting both `policy_id` and the new `base_policy_id` fields on the Agent doc during policy reassignment as well. 👍

juliaElastic:
@ycombinator thanks for the review, updated the test

If you ready to approve, can you help with the merge too? I don't have access.

ycombinator (approved):

PR #7446, opened 2026-07-22
fix: treat enrollment CREATE 409 version conflict as success

ycombinator:

During high-load enrollment (e.g. 100k agents), a primary shard relocation on `.fleet-agents` causes in-flight `CREATE` bulk operations to fail with a `409 version_conflict_engine_exception`. Fleet Server previously propagated this as an HTTP 500 to the enrolling agent.

Observed in a 2026-07-22 100k-agent scale test: 661 enrollment requests returned 500 over a ~10-second window following a single `.fleet-agents` shard relocation. The failing requests had been in-flight for 43–57s before receiving the 409. Without a fix, agents receive a 500, retry enrollment, and loop indefinitely against a document that already exists in ES.

A `409 version_conflict_engine_exception` on `op_type: create` is a definitive signal that the document already exists — the write was committed to the old primary before the conflict response was returned. This was confirmed by pulling raw ES and fleet-server logs from the scale test failure window (Buildkite #6617, 2026-07-22): zero occurrences of `if_seq_no` or `if_primary_term` in fleet-server's requests; the 409s came from the document already existing after the shard handoff.

Following the Beats/Logstash convention, `createFleetAgent` now returns `nil` on `ErrElasticVersionConflict`: enrollment completes successfully and the agent gets an HTTP 200 with the already-created access API key.

The `updateFleetAgent` path (re-enrollment of existing agents) already uses `retry_on_conflict: 3` at the Elasticsearch bulk-API level and is unaffected.

```
go test ./internal/pkg/api/... -run TestCreateFleetAgent
```

`TestCreateFleetAgentVersionConflictSucceeds` — bulker returns `ErrElasticVersionConflict` → `createFleetAgent` returns `nil` (enrollment succeeds).

## Design Checklist

## Checklist


internal/pkg/api/handleEnroll.go
```diff
@@ -647,16 +648,44 @@ func createFleetAgent(ctx context.Context, bulker bulk.Bulk, id string, agent mo
...
+			}
+			// The document exists with different data: a concurrent writer created it.
+			return lastErr
+		}
+		if !errors.Is(readErr, dl.ErrNotFound) {
+			// Unexpected error reading the document.
+			return readErr
+		}
+		// Document not found – the Create was not committed. Retry.
 	}
```
blakerouse: This just places this in `handleEnroll`, but can't this happen everywhere in any call? Would it be possible to move this into the bulker directly so these issues are retried there?

If this has to be handled everywhere this is going to get nasty all over the code base.

ycombinator: Good call. Moved the retry into `Bulker.Create` (up to 3 attempts on `409 version_conflict_engine_exception`) so any calls that go through the Bulker will benefit (dce43824).

One caveat: this is a blind retry — we don't read the document back afterward. If the `Create` was actually committed on the old primary before the shard relocation reported the 409, the retry will see the document already exists, get another 409, and surface the error to the enrolling agent as a 500. The agent would then need to retry enrollment. Handling that case correctly requires reading back the document and comparing `access_api_key_id` to distinguish "our write landed" from "concurrent writer" — which is fleet-agent-specific knowledge that cannot belong in the bulker.

In practice the common failure mode during shard relocation is the write not being committed at all, so blind retry should resolve the bulk of the failures we saw during the 100k check-in scale test. The committed-before-409 case is rarer and the agent retry path should handle it acceptably.

ycombinator:
@blakerouse @macdewee I added some scope to this PR.  See 36bb5087, which extends the retry logic to also cover `context.DeadlineExceeded` / `context.Canceled` from the ES client.  This was based on failures observed in a 100k scale test run today where `req.Do` timed out 4000+ bulk writes over a 9-minute window while an ES shard was still bootstrapping.

internal/pkg/bulk/opBulk.go
```diff
@@ -20,12 +20,28 @@ import (
 )
 
 func (b *Bulker) Create(ctx context.Context, index, id string, body []byte, opts ...Opt) (string, error) {
-	item, err := b.waitBulkAction(ctx, ActionCreate, index, id, body, opts...)
-	if err != nil {
+	const maxRetries = 3
+	var lastErr error
+	for range maxRetries {
+		item, err := b.waitBulkAction(ctx, ActionCreate, index, id, body, opts...)
+		if err == nil {
```
blakerouse: It would probably just be simpler here to check `if ctx.Err() != nil`. `return ctx.Err()`. Then the check below for the deadline can be removed. You basically just checking if the error here is from the context or from the actual action on ES side.

internal/pkg/bulk/bulk_test.go
```diff
@@ -13,16 +13,180 @@ import (
...
+func TestCreateRetriesOnVersionConflict(t *testing.T) {
+	mock := &conflictThenSuccessTransport{}
+
+	bulker := NewBulker(mock, nil, WithFlushThresholdCount(1))
+	go func() { _ = bulker.Run(t.Context()) }()
+
+	id, err := bulker.Create(t.Context(), "test-index", testDocID, []byte(`{"field":"value"}`))
+	if err != nil {
+		t.Fatalf("expected Create to succeed after retry, got: %v", err)
+	}
```
macdewee: Why not


```suggestion
	require.NoError(t, err, "expected Create to succeed after retry")
```

macdewee: and analogical for other `Fatalf` and `Errorf`

internal/pkg/bulk/opBulk.go
ycombinator: Done in cadf830d9cb64309634e8751db695a90e4fcd7ca.

internal/pkg/bulk/bulk_test.go
ycombinator: Good catch.  Fixed this and other similar occurrences in https://github.com/elastic/fleet-server/pull/7446/changes/2415726b0c138cef3128a3f23573975ea102939e.

internal/pkg/api/handleEnroll.go
macdewee: shouldn't we document possible issue somewhere? can be gh issue or something easy to find in case it happens

macdewee (dismissed):
LGTM

internal/pkg/api/handleEnroll.go
cmacknz: > we don't read the document back afterward. If the Create was actually committed on the old primary before the shard relocation reported the 409, the retry will see the document already exists, get another 409, and surface the error to the enrolling agent as a 500

Getting a 409 conflict where the document failed to index completely does not feel entirely correct on the ES side to me. Is this definitely what happens? If it is this is something to raise with the ES team.

Both Filebeat and Logstash take document level 409 responses to be ignorable generally, as indicators that the document did index already.

Filebeat having to query out data on 409 to confirm there was no data loss would scale terribly as you could imagine, and also runs into the problem that many writing clients do not have permission to read in the first place.

blakerouse (dismissed):
Thanks for the fixes.

internal/pkg/api/handleEnroll.go
ycombinator: You're right that it doesn't happen — a 409 on `op_type: create` is a definitive signal that the document exists.

I went back and pulled the ES and fleet-server logs from the scale test failure (Buildkite #6617, 2026-07-22). The 661 failures are plain `op_type: create` 409s. The failing requests had been in-flight for ~43–57s before resolving as 409, meaning the writes were committed to the old primary before the conflict response arrived.

So I'm updating the implementation to match the Beats/Logstash convention — treat `ErrElasticVersionConflict` on CREATE as success. Concretely: `createFleetAgent` will return `nil` on `ErrElasticVersionConflict` instead of propagating it, so the enrollment completes successfully rather than returning a 500 to the agent. The retry logic in the bulker for `Create` will also be removed since retrying a 409 is always futile — the document already exists.

ycombinator: Changes made in df626805,.

blakerouse (dismissed):
I see this is much simpler! ;-)

macdewee (dismissed):

internal/pkg/api/handleEnroll.go
```diff
@@ -653,10 +654,12 @@ func createFleetAgent(ctx context.Context, bulker bulk.Bulk, id string, agent mo
...
 
 	_, err = bulker.Create(ctx, dl.FleetAgents, id, data, bulk.WithRefresh())
-	if err != nil {
-		return err
+	// A 409 on op_type:create is a definitive signal the document already exists.
+	// Treat it as success per the Beats/Logstash convention.
+	if errors.Is(err, es.ErrElasticVersionConflict) {
+		return nil
```
cmacknz, line 662: You probably want to log that this happened, but still treat it as success as you do here.

ycombinator: Done in 31ace0a4.

cmacknz (dismissed):

changelog/fragments/1784763000-enrollment-version-conflict-retry.yaml
```diff
@@ -0,0 +1,7 @@
+kind: bug-fix
+
+summary: Retry enrollment agent-document write on ES version conflict and write timeout
```
cmacknz: You have two changelogs now

ycombinator: Ugh, sorry, fixing.

cmacknz (approved):

PR #7468, opened 2026-07-27
Remove `service_token` from `secret_path` for remote ES

### LINKED ISSUE #5664: "Unable to find secret path" warning is displayed while collecting diagnostics when Remote ES is selected.
**Kibana Build details:**
```
VERSION: 9.0.0-SNAPSHOT
BUILD: 84188
COMMIT: 5c231847869504d37f8f4f5ac098a42429ee854c
```
Artifact Link: https://snapshots.elastic.co/9.0.0-528cb7b2/downloads/beats/elastic-agent/elastic-agent-9.0.0-SNAPSHOT-linux-x86_64.tar.gz

**Preconditions:**
1. 9.0.0-SNAPSHOT Kibana cloud environment should be available.
2. Remote Elasticsearch should be configured.

**Steps to reproduce:**
1. Install an agent with agent policy having Remote ES selected.
2. Run `sudo elastic-agent diagnostics collect` command.
3. Observe warning is displayed: "Unable to find secret path".

**Expected Result:**
"Unable to find secret path" warning should not be displayed while collecting diagnostics when Remote ES is selected.

**Screenshot:**

Image

Image

**Logs:**
[elastic-agent-diagnostics-2025-03-25T11-01-11Z-00.zip](https://github.com/user-attachments/files/19446597/elastic-agent-diagnostics-2025-03-25T11-01-11Z-00.zip)

**Agent JSON:**
[ip-172-31-20-104-agent-details.zip](https://github.com/user-attachments/files/19446700/ip-172-31-20-104-agent-details.zip)

### LINKED ISSUE #15798: Agents logging constant error: "failed to add secret markers: secret path outputs.OUTPUT_ID.service_token does not exist"
Hello,

We noticed recently that agents are constantly reporting the following error:

> failed to add secret markers: secret path outputs.OUTPUT_ID.service_token does not exist

Where the `OUTPUT_ID` is the remote elasticsearch output configured in the Elastic Agent policy to send monitoring logs and metrics to our monitoring cluster.

This does not seems to completely impact the delivery of logs and metrics as we are seeing these logs in the monitoring cluster.

Checking the output settings using the request: `GET kbn:/api/fleet/outputs/OUTPUT_ID`, we have this response:

```
{
  "item": {
    "id": "OUTPUT_ID",
    "name": "es-cloud-monitoring",
    "type": "remote_elasticsearch",
    "hosts": [
      "https://ENDPOINT.aws.found.io:443"
    ],
    "is_default": false,
    "is_default_monitoring": true,
    "preset": "balanced",
    "config_yaml": "",
    "secrets": {
      "service_token": {
        "id": "SERVICE_TOKEN_ID"
      }
    },
    "service_token": null,
    "sync_integrations": false,
    "kibana_url": null,
    "sync_uninstalled_integrations": false,
    "write_to_logs_streams": false,
    "kibana_api_key": null,
    "ssl": {
      "certificate": "",
      "certificate_authorities": []
    }
  }
}
```

We can see that the field `service_token` is `null`, but the field `secrets.service_token` is populated.

The configuration of the output was done following the instructions in the Fleet UI.

It seems that this issue was introduced between `8.19.2` and `9.4.4`, I had an agent on `8.19.2` that was not reporting this error and as soon as I upgrade it to `9.4.4` it started reporting those messages.

On version `8.19.2`:
Image

After upgrade to version `9.4.4`:
Image

The integrations that I have in the policies that are sending data to the monitoring cluster are:

- `Elastic Agent`
- `System` (with only the _Collect metrics from System instances_ option enabled)

### To replicate:

I think this can be replicated with the followint steps:

- Create 2 cloud deployments, one will be used for monitoring.
- Add a remote elasticsearch output on the main deployment.
- Create a policy to send metrics and elastic agent logs to the monitoring cluster.
- Add an agent on version 9.4.4 to this policy.
- Validate the error logs on the monitoring cluster.

khushijain21:
We delete the `service_token` from the parsed policy when output is of remoteES type. [See](https://github.com/khushijain21/fleet-server/blob/834152573966d725b6d6e6cf7aacbd889282fd16/internal/pkg/policy/policy_output.go#L348-L354). But we do not remove the reference for it from `secret_path.output.[OuputID].service_token` which causes constant stream of errors on elastic-agent side when attempting to read the nil `service_token`

The error logged is 
```
failed to add secret markers: secret path outputs.OUTPUT_ID.service_token does not exist
```

This PR removes the corresponding `service_token` from `secret_path` before sending the policy to elastic-agent.

## Design Checklist

## Checklist

## Related issues

 
- Closes https://github.com/elastic/elastic-agent/issues/15798
- Closes https://github.com/elastic/fleet-server/issues/5664


samuelvl (dismissed):

internal/pkg/policy/parsed_policy.go
```diff
@@ -77,11 +77,17 @@ func NewParsedPolicy(ctx context.Context, bulker bulk.Bulk, p model.Policy) (*Pa
...
 	for name, policyOutput := range p.Data.Outputs {
+		outputType, _ := policyOutput[FieldOutputType].(string)
 		ks, err := secret.ProcessOutputSecret(policyOutput, secretValues)
 		if err != nil {
 			return nil, fmt.Errorf("failed to replace secrets in output section of policy '%s': %w", name, err)
 		}
 		for _, key := range ks {
+			// Do not advertise remote ES service_token in secret_paths; Prepare(...) deletes it
+			// before the policy is sent to agents
+			if outputType == OutputTypeRemoteElasticsearch && key == FieldOutputServiceToken {
```
cmacknz: This should be conditional on the service_token actually being removed shouldn't it? Otherwise we leak the service_token.

Is there a way to do this such that if Prepare changes to retain the token this behaves properly?

khushijain21: Added the conditional. PTAL when you can

cmacknz (approved):

PR #7493, opened 2026-07-28
test: merge duplicate dispatch limit test functions into table-driven test

ycombinator:

`TestDispatchAllowsWhenUnderLimit` and `TestDispatchNoLimitWhenZero` in `dispatch_limit_test.go` are structurally identical — both dispatch a single operation through a bulker with a goroutine simulating the Run loop, differing only in the `max_pending_dispatches` value (10 vs 0). golangci-lint's `dupl` checker (threshold: 100 tokens) flags these as duplicate code on any PR that touches the `bulk` package, because the linter runs with `--whole-files`.

The issue is latent on `main` and `9.5` today and will surface on the next PR that modifies any file in `internal/pkg/bulk/`. It is already actively blocking backport PRs for #6751 on `9.4`, `9.3`, and `8.19`.

The root cause: golangci-lint was v2.5.0 when these tests were introduced (#6751, April 2026) and is now v2.11.4, which triggers the finding at this token count.

Merges the two functions into a single `TestDispatchAllowsDispatch` table-driven test using a `map[string]int64` of limit values. This is idiomatic Go, eliminates the dupl finding, and keeps the same coverage.

```
go test ./internal/pkg/bulk/ -run TestDispatch -v
```

## Design Checklist

## Checklist

## Related issues

- Relates #6751

### DIFF
--- a/internal/pkg/bulk/dispatch_limit_test.go
+++ b/internal/pkg/bulk/dispatch_limit_test.go
@@ -59,16 +59,13 @@ func TestDispatchRejectsWhenLimitReached(t *testing.T) {
 }
 
 func TestDispatchAllowsDispatch(t *testing.T) {
-	tests := []struct {
-		name  string
-		limit int64
-	}{
-		{name: "under limit", limit: 10},
-		{name: "no limit when zero", limit: 0},
+	tests := map[string]int64{
+		"under limit":      10,
+		"no limit when zero": 0,
 	}
-	for _, tt := range tests {
-		t.Run(tt.name, func(t *testing.T) {
-			b := NewBulker(nil, nil, WithBlockQueueSize(1), WithMaxPendingBulkDispatches(tt.limit))
+	for name, limit := range tests {
+		t.Run(name, func(t *testing.T) {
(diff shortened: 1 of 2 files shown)


internal/pkg/bulk/dispatch_limit_test.go
```diff
@@ -58,42 +58,28 @@ func TestDispatchRejectsWhenLimitReached(t *testing.T) {
...
-		item.ch <- respT{}
-	}()
-
-	resp := b.dispatch(context.Background(), blk)
-	require.NoError(t, resp.err)
-
-	// Counter should be back to 0 after dispatch completes.
-	require.Equal(t, int64(0), b.pendingBulkDispatches.Load())
-}
+func TestDispatchAllowsDispatch(t *testing.T) {
```
lorienhu: Nit but since you have to fix the golangci-lint anyway, could we rename this? 'DispatchAllowsDispatch' is a little confusing.

lorienhu (dismissed):

internal/pkg/bulk/dispatch_limit_test.go
ycombinator: Addressed in 761b8844.

lorienhu (dismissed):

lorienhu (approved):

PR #7499, opened 2026-07-29
feat: add http_server.tcp_rejected_rate metric for connection-cap saturation

ycombinator:

Fleet Server's existing autoscaling metric, `http_server.tcp_active`, is a utilization signal: EPA uses it to maintain headroom by targeting N active connections per pod. This works well at steady state but understates demand when connections are short-lived — specifically during large-scale agent enrollment, where each enrollment connection completes in under a second.

`tcp_active` is a point-in-time gauge: it reflects the instantaneous count of open HTTP connections at the moment EPA polls `/stats`. A burst of enrollment connections that arrive and clear between two EPA polls leaves no trace in the metric. The result is that pods plateau well below the theoretical requirement, agents receive HTTP 429s, and enter exponential backoff — turning a short enrollment phase into a multi-hour convergence tail.

Adds `http_server.tcp_rejected_rate` to `/stats`: a gauge, in rejections/sec, derived from an internal counter that increments whenever the chi `Throttle` middleware returns HTTP 429 due to the per-pod connection cap (`max_connections`). The rate is sampled every 10 s by a new `RunConnRejectionRateSampler` goroutine.

**`tcp_rejected_rate` vs `tcp_active`**

| | `tcp_active` | `tcp_rejected_rate` |
|---|---|---|
| Signal type | Utilization | Saturation |
| When it fires | Active connections approach the per-pod target | Pod is already at its connection ceiling and returning 429s |
| Enrollment bursts | Understates demand — burst clears before next EPA poll | Catches the burst; rejections accumulate in the counter until sampled |
| Steady state | Correct — long-held checkin connections register reliably | Zero — pod is not at ceiling, no 429s |
| EPA behavior | Scale to maintain headroom | Scale the moment a pod is overwhelmed |

Because EPA takes the **maximum** desired replica count across all configured metrics, wiring both allows proactive scaling (`tcp_active`: headroom at steady state) and reactive scaling (`tcp_rejected_rate`: act immediately under a burst). Neither metric suppresses the other.

The raw rejection count is kept as an unexported `atomic.Uint64` — only the derived rate gauge is exposed in `/stats`, since that is the only value EPA consumes. The chi `Throttle` middleware behaviour is unchanged; the wrapper only observes the response status code.

1. Configure Fleet Server with `max_connections: N` and start it.
2. Send more than N concurrent requests (e.g. with `hey` or `wrk`).
3. `curl localhost:5066/stats | python3 -m json.tool | grep tcp_rejected` — confirm `tcp_rejected_rate` becomes non-zero after ~10 s.
4. Let load drop — `tcp_rejected_rate` should return to 0 within the next sample interval.

## Design Checklist

## Checklist

## Related issues

- Relates https://github.com/elastic/fleet-controller/pull/638

### DIFF
--- a/internal/pkg/api/metrics.go
+++ b/internal/pkg/api/metrics.go
@@ -332,13 +332,9 @@ func attachPrometheusEndpoint(router metricsRouter, reg *prometheus.Registry, bi
 	router.AddRoute("/metrics", promhttp.InstrumentMetricHandler(reg, h).ServeHTTP)
 }
 
-// CheckinRateSampleInterval is how often RunCheckinRejectionRateSampler recomputes
-// the checkin capacity-rejection rate.
-const CheckinRateSampleInterval = 10 * time.Second
-
-// ConnRejectionRateSampleInterval is how often RunConnRejectionRateSampler recomputes
-// the connection-cap rejection rate.
-const ConnRejectionRateSampleInterval = CheckinRateSampleInterval
+// RejectionRateSampleInterval is how often the rejection-rate samplers recompute
+// the checkin and connection-cap rejection rates.
+const RejectionRateSampleInterval = 10 * time.Second
 
 // computeRate returns the count of rejections (cur-prev) divided by an elapsed
 // time (dt), rounded to the nearest integer. It returns 0 if dt is non-positive
--- a/internal/pkg/api/metrics.go
+++ b/internal/pkg/api/metrics.go
@@ -9,6 +9,7 @@ import (
 	"errors"
 	"fmt"
 	"sync"
+	"sync/atomic"
 	"time"
 
 	"github.com/prometheus/client_golang/prometheus"
@@ -36,8 +37,10 @@ var (
 	cntHTTPNew          *statsCounter
 	cntHTTPClose        *statsCounter
 	cntHTTPActive       *statsGauge
-	cntHTTPRejected     *statsCounter
 	cntHTTPRejectedRate *statsGauge
+	// cntHTTPRejected is an internal-only accumulator; only the derived rate is
+	// exposed in /stats. Not registered with the metrics registry.
+	cntHTTPRejected atomic.Uint64
 
 	cntCheckin       routeStats
 	cntEnroll        routeStats
@@ -63,7 +66,6 @@ func init() {
 	cntHTTPNew = newCounter(registry, "tcp_open")
 	cntHTTPClose = newCounter(registry, "tcp_close")
 	cntHTTPActive = newGauge(registry, "tcp_active")
-	cntHTTPRejected = newCounter(registry, "tcp_rejected")
 	cntHTTPRejectedRate = newGauge(registry, "tcp_rejected_rate")
 
 	routesRegistry := registry.newRegistry("routes")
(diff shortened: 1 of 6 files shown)


ebeahan (dismissed):

swiatekm:
The logic looks right to me, but calculating the rate of change this way within the application looks like an antipattern. Could we emit it as a `total` counter and calculate the delta downstream, either in a library or the monitoring solution?

ycombinator:
> The logic looks right to me, but calculating the rate of change this way within the application looks like an antipattern. Could we emit it as a `total` counter and calculate the delta downstream, either in a library or the monitoring solution?

I had looked into this previously (for a similar trigger that we may/may not need: https://github.com/elastic/fleet-controller/issues/626).  Unfortunately, Elastic Pod Autoscaler (EPA) doesn't support computing rates from a counter.  We could implement this around fleet-controller, but that would require it to poll the `/stats` API on every Fleet Server pod, retain per-pod counter history, compute the rate, and expose the derived metric for EPA to consume — responsibilities it does not have today.

That would be a larger architectural change, so this PR keeps the rate calculation in Fleet Server for now.

[UPDATE] I've created https://github.com/elastic/ecp-services-team/issues/2139 so the EPA maintainers can discuss and decide if they want to support deriving rate metrics from counter metrics in the future.

internal/pkg/api/router.go
```diff
@@ -74,6 +74,34 @@ func Limiter(cfg *config.ServerLimits) *limiter {
...
+			}
+		})
+	}
+}
+
+// statusRecorder captures the HTTP status code written by a downstream handler.
+type statusRecorder struct {
+	http.ResponseWriter
+	status int
+}
```
macdewee, line 98: We are missing implementation of `Write` function, and if anyone will use it in the future, the status will be always 200

https://pkg.go.dev/net/http#ResponseWriter
```
	// If [ResponseWriter.WriteHeader] has not yet been called, Write calls
	// WriteHeader(http.StatusOK) before writing the data.
```

If it is intended we should document it

internal/pkg/api/router_test.go
```diff
@@ -105,3 +107,42 @@ func TestLimiter(t *testing.T) {
...
+	t.Run("increments counter when connection cap is hit", func(t *testing.T) {
+		before := cntHTTPRejected.Load()
+
+		started := make(chan struct{})
+		h := throttleWithCount(1)(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+			close(started)
+			time.Sleep(100 * time.Millisecond)
+		}))
+
+		go h.ServeHTTP(httptest.NewRecorder(), httptest.NewRequestWithContext(context.Background(), http.MethodGet, "/", nil))
```
macdewee: Possible go routine leak

internal/pkg/api/metrics.go
```diff
@@ -330,6 +336,10 @@ func attachPrometheusEndpoint(router metricsRouter, reg *prometheus.Registry, bi
 // the checkin capacity-rejection rate.
 const CheckinRateSampleInterval = 10 * time.Second
 
+// ConnRejectionRateSampleInterval is how often RunConnRejectionRateSampler recomputes
+// the connection-cap rejection rate.
+const ConnRejectionRateSampleInterval = CheckinRateSampleInterval
```
macdewee: Not sure if binding two consts is a good idea, unless it is intended to have the same interval for both of them, but then we could use one constant

macdewee (commented):
We could also add test for `tcp_rejected_rate should return to 0 within the next sample interval`

internal/pkg/api/router.go
ycombinator: Implemented in ebb79f3a7e084a70af83bfabc49da33140aa0751.

internal/pkg/api/router_test.go
ycombinator: Good catch.  Fixed in d8cef17441d4ca6a7144d0bf2084ec16ecc137cc.

internal/pkg/api/metrics.go
ycombinator: Using a single constant in ce96565dff45ebca9cf3357c5840035b1dc59df8.

ycombinator:
> We could also add test for `tcp_rejected_rate should return to 0 within the next sample interval`

Added in 47aa51d9e10d56f82a5c587d490a1b63766b98f1.

swiatekm (approved):

macdewee (approved):

PR #7528, opened 2026-07-31
Parameterize output names in Elasticsearch update scripts

ycombinator:

Policy output names were rendered directly into Painless source used to update Fleet agent documents. An output name containing script syntax could therefore change the meaning of the update or removal script.

Pass output names through Painless `params` in both the output update and output removal scripts. Output names are now handled only as map keys and are never included in script source.

Regression tests verify the generated requests and exercise a script-shaped output name against Elasticsearch.

```bash
mage test:unit
TEST_RUN=TestRenderUpdatePainlessScript mage test:integration
TEST_RUN=TestRenderRemoveOutputPainlessScript mage test:integration
```

## Design Checklist

- [ ] I have or intend to scale test these changes; the request structure and execution model are unchanged.
- [ ] I have included fail safe mechanisms; no new load-producing behavior is introduced.

## Checklist

- [x] The parameterized script construction makes the intent explicit.
- [x] No documentation changes are required.
- [x] No default configuration changes are required.
- [x] I have added tests that prove the fix is effective.
- [x] I have added a changelog fragment.

### DIFF
--- a/changelog/fragments/1785505460-parameterize-output-painless-scripts.yaml
+++ b/changelog/fragments/1785505460-parameterize-output-painless-scripts.yaml
@@ -1,0 +1,3 @@
+kind: security
+summary: Treat output names as parameters in Elasticsearch update scripts
+component: fleet-server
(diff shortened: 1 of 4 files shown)


blakerouse (approved):
Looks good.

PR #7533, opened 2026-07-31
Retain output secret when agent update fails

ycombinator:

When Fleet Server creates an output API key, it stores the encoded key in `.fleet-secrets` and then updates the agent document in `.fleet-agents` with the secret reference. Elasticsearch can commit that update while the client still receives an error, for example when the request context expires while waiting for the response. This can happen under scale load and was observed during a 100k-agent scale test.

The existing error path immediately deleted the secret. In the ambiguous-commit case, this left the agent document pointing at a missing secret and subsequent check-ins failed while resolving the output API key.

Retain the newly created secret whenever the agent update returns an error. This chooses a possible orphaned secret over damaging an agent document with a dangling reference.

A follow-up PR adds conservative, out-of-band reconciliation for these retained candidates.

## How to test

- `go test ./internal/pkg/policy -run TestPolicyOutputESPrepare -count=1`
- `mage test:unit`

`mage check:all` currently reports 58 pre-existing linter findings in unrelated files under the repository's pinned Go/toolchain configuration. The files changed by this PR are clean.

## Design Checklist

- The solution is stateless and assumes a horizontally scaled Fleet Server deployment.
- The changed path is intended for 100k-agent deployments.
- The error behavior is fail-safe: an uncertain write retains the secret so a committed agent reference remains usable.

## Checklist

- [x] I have added an entry in `CHANGELOG.md`

Documentation and configuration changes are not applicable. The changelog entry is supplied as a fragment.

### DIFF
--- a/internal/pkg/policy/policy_output.go
+++ b/internal/pkg/policy/policy_output.go
@@ -347,9 +347,10 @@ func (p *Output) prepareElasticsearch(
 
 		if err = bulker.Update(ctx, dl.FleetAgents, agent.Id, body, bulk.WithRefresh(), bulk.WithRetryOnConflict(3)); err != nil {
 			zlog.Error().Err(err).Msg("fail update agent record")
-			if delErr := bulker.DeleteSecret(ctx, secretID); delErr != nil {
-				zlog.Warn().Err(delErr).Str("secret_id", secretID).Msg("failed to delete orphaned output API key secret after agent update failure")
-			}
+			// The update may have been committed by Elasticsearch even when the client
+			// returns an error, for example when the request context expires while
+			// waiting for the response. Deleting the secret here can therefore leave
+			// the agent document pointing at a missing secret.
 			return fmt.Errorf("fail update agent record: %w", err)
 		}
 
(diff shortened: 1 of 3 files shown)


internal/pkg/policy/policy_output.go
```diff
@@ -347,9 +347,10 @@ func (p *Output) prepareElasticsearch(
 
 		if err = bulker.Update(ctx, dl.FleetAgents, agent.Id, body, bulk.WithRefresh(), bulk.WithRetryOnConflict(3)); err != nil {
 			zlog.Error().Err(err).Msg("fail update agent record")
-			if delErr := bulker.DeleteSecret(ctx, secretID); delErr != nil {
-				zlog.Warn().Err(delErr).Str("secret_id", secretID).Msg("failed to delete orphaned output API key secret after agent update failure")
-			}
+			// The update may have been committed by Elasticsearch even when the client
```
blakerouse, line 350: Which client are you speaking of here? The users HTTP client (aka the HTTP request being handled) or the client Fleet Server uses to communicate with elasticsearch?

I feel like it should be possible to check the error to get more detail on the error to really determine what should be done here. I feel like leaving these leads to an issue where the secrets index just grows unbounded overtime.

blakerouse (changes requested):

blakerouse (approved):
Actually after reviewing https://github.com/elastic/fleet-server/pull/7534 I am okay with this actually.

PR #7534, opened 2026-07-31
Reconcile orphaned output secrets after agent update failures

ycombinator:

#7533 prevents an ambiguous `.fleet-agents` update failure from deleting a secret that Elasticsearch may already have referenced. The safe failure behavior can leave an orphaned `.fleet-secrets` document when the update truly did not commit.

This failure mode was observed under load during a 100k-agent scale test.

Add an out-of-band, candidate-based reconciler:

- The failed update path only enqueues the newly created secret and returns; it performs no cleanup I/O in the check-in request.
- Candidates remain in memory for a 10-minute grace period.
- The reconciler rereads the agent with a fresh background context and preserves candidates referenced by either the current output or output API-key retirement history.
- A secret must be observed unreferenced twice, at least 5 minutes apart, before deletion.
- Read and delete failures retain the candidate for retry.
- Work is capped at 100 candidate checks per minute per Fleet Server to avoid creating a cleanup load spike.
- A full queue fails safe by leaking a secret rather than blocking a check-in or risking deletion.

Candidates intentionally do not survive a Fleet Server restart. A restart during the reconciliation window can therefore leave an orphaned secret, but cannot damage an agent by deleting a referenced secret.

This PR is stacked on and depends on #7533. Its implementation is the second commit (`b3f11ce9`). Once #7533 merges, GitHub will remove the shared first commit from this PR's diff.

## How to test

- `go test ./internal/pkg/policy ./internal/pkg/gc ./internal/pkg/api ./internal/pkg/server`
- `mage test:unit`

`mage check:all` currently reports the same 58 pre-existing linter findings as #7533 in unrelated files. None are in files changed by this PR.

## Design Checklist

- The reconciler is local to each horizontally scaled Fleet Server and requires no cross-instance coordination because each secret candidate is created by one request on one instance.
- The queue, grace period, confirmation reads, and per-run cap are designed for 100k-agent deployments.
- Every uncertain state fails safe by retaining the secret.

## Checklist

- [x] I have added an entry in `CHANGELOG.md`

Configuration changes are not applicable. Package documentation was updated and the changelog entry is supplied as a fragment.

### DIFF
--- a/changelog/fragments/1785539201-reconcile-orphaned-output-secrets.yaml
+++ b/changelog/fragments/1785539201-reconcile-orphaned-output-secrets.yaml
@@ -1,0 +1,3 @@
+kind: bug-fix
+summary: Reconcile orphaned output secrets after ambiguous agent update failures
+component: fleet-server
--- a/internal/pkg/api/handleCheckin.go
+++ b/internal/pkg/api/handleCheckin.go
@@ -85,8 +85,20 @@ type CheckinT struct {
 	// gwPool is a gzip.Writer pool intended to lower the amount of writers created when responding to checkin requests.
 	// gzip.Writer allocations are expensive (~1.2MB each) and can exhaust an instance's memory if a lot of concurrent responses are sent (this occurs when a mass-action such as an upgrade is detected).
 	// effectiveness of the pool is controlled by rate limiter configured through the limit.action_limit attribute.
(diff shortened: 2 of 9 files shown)


blakerouse (dismissed):
This looks good.

blakerouse (approved):

PR #7535, opened 2026-08-01
fix: Import math and validate request body size

odaysec:
Add math import and check for bulk request body size. fix protect all allocation-size arithmetic by checking for overflow before computing/using capacities. For this specific issue, make `Buf.grow` compute a target capacity without overflowing, and panic with a clear message if growth cannot be represented as `int`.

Best targeted fix (no functional change beyond safety):
- Edit `internal/pkg/danger/buf.go`, function `grow`.
- Replace `2*cap(b.buf)+n` with guarded arithmetic:
  - Validate `n >= 0` (defensive, even though `Grow` checks).
  - Compute required capacity `need := len(b.buf) + n` with overflow check.
  - Compute doubled capacity from current cap with overflow check/saturation.
  - Choose `newCap = max(need, doubledCap)`.
  - Allocate with `make([]byte, len(b.buf), newCap)`.
- Add `math` import for `math.MaxInt`.

## Design Checklist

## Checklist

### DIFF
--- a/internal/pkg/bulk/opBulk.go
+++ b/internal/pkg/bulk/opBulk.go
@@ -9,6 +9,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"math"
 	"time"
 
 	"github.com/elastic/go-elasticsearch/v8/esapi"
@@ -59,6 +60,9 @@ func (b *Bulker) waitBulkAction(ctx context.Context, action actionT, index, id s
 
 	// Serialize request
 	const kSlop = 64
+	if len(body) > math.MaxInt-kSlop {
+		return nil, fmt.Errorf("bulk request body too large")
+	}
 	blk.buf.Grow(len(body) + kSlop)
 
 	if err := b.writeBulkMeta(&blk.buf, action.String(), index, id, opt.RetryOnConflict); err != nil {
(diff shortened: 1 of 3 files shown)


internal/pkg/danger/buf.go
```diff
@@ -33,7 +34,31 @@ func (b *Buf) Reset() {
 }
 
 func (b *Buf) grow(n int) {
-	buf := make([]byte, len(b.buf), 2*cap(b.buf)+n)
```
odaysec, line 36: Performing calculations involving the size of potentially large strings or slices can result in an overflow (for signed integer types) or a wraparound (for unsigned types). An overflow causes the result of the calculation to become negative, while a wraparound results in a small (positive) number. This can cause further issues. If, for example, the result is then used in an allocation, it will cause a runtime panic if it is negative, and allocate an unexpectedly small buffer otherwise.

References
[Integer overflow](https://golang.org/ref/spec#Integer_overflow)
[Making slices, maps and channels](https://golang.org/ref/spec#Making_slices_maps_and_channels)

internal/pkg/bulk/opBulk.go
```diff
@@ -59,6 +60,9 @@ func (b *Bulker) waitBulkAction(ctx context.Context, action actionT, index, id s
 
 	// Serialize request
 	const kSlop = 64
+	if len(body) > math.MaxInt-kSlop {
```
kruskall, line 63: Could you share what system you are running to have a byte slice of len(math.MaxInt) in memory ?

PR #7567, opened 2026-08-06
bulk: add configurable concurrency limit for ReadSecrets

ycombinator:

`ReadSecrets` makes direct HTTP calls to the ES Fleet secrets API (one per secret reference per agent checkin) without going through the bulk dispatch queue. This means it is **not subject to the `max_pending_bulk_dispatches` cap** introduced in #6751. Under high concurrent checkin load — as seen in large serverless projects — each checkin goroutine issues an independent ES connection, causing unbounded concurrent ES connections and additional memory pressure (goroutines, HTTP buffers, response allocations).

This was observed as a contributing factor in an OOM incident for a large production serverless security project. PR #7416 added the `ReadSecrets` per-checkin call ~24h before OOMs began.

Adds a `semaphore.Weighted` (`readSecretsLimit`) to `Bulker`, capping concurrent in-flight secret reads at 32 (matching `apikeyLimit`). `ReadSecrets` acquires a slot before each ES call and releases it immediately after — callers that arrive when all slots are taken block until one is free or their context is cancelled. Setting `readSecretsLimit` to `nil` (by passing `WithMaxConcurrentSecretReads(0)`) disables the cap entirely; the default is always 32.

The limit is intentionally not exposed as a user-facing config option, consistent with how `apikeyMaxParallel` is handled.

## Files changed

- `internal/pkg/bulk/engine.go`: add `readSecretsLimit *semaphore.Weighted` to `Bulker`; initialize in `NewBulker` when limit > 0; nil-check acquire/release in `ReadSecrets`; add `defaultMaxConcurrentSecretReads = 32`
- `internal/pkg/bulk/opt.go`: add `maxConcurrentSecretReads` field, `WithMaxConcurrentSecretReads` BulkOpt, zerolog logging
- `internal/pkg/bulk/secret_limit_test.go`: unit tests covering the concurrency limit, context cancellation while waiting, zero-value (no limit), and default capacity

```
go test -v -count=1 -run=TestReadSecrets ./internal/pkg/bulk
```

## Design Checklist

## Checklist

## Related issues

- Relates elastic/ingest-dev#8991
- Relates #7416
- Relates #6751

### DIFF
--- a/internal/pkg/bulk/secret_limit_test.go
+++ b/internal/pkg/bulk/secret_limit_test.go
@@ -29,10 +29,12 @@ func (m *blockingTransport) RoundTrip(_ *http.Request) (*http.Response, error) {
 	m.inFlight.Add(1)
 	defer m.inFlight.Add(-1)
 	<-m.gate
+	h := http.Header{}
+	h.Set("X-Elastic-Product", "Elasticsearch")
 	body := `{"value":"test"}`
 	return &http.Response{
 		StatusCode: http.StatusOK,
-		Header:     make(http.Header),
+		Header:     h,
 		Body:       io.NopCloser(strings.NewReader(body)),
 	}, nil
 }
--- a/internal/pkg/bulk/secret_limit_test.go
+++ b/internal/pkg/bulk/secret_limit_test.go
@@ -60,15 +60,12 @@ func TestReadSecretsLimitsConcurrency(t *testing.T) {
 	var wg sync.WaitGroup
 	errs := make([]error, 2)
 
-	wg.Add(2)
-	go func() {
-		defer wg.Done()
+	wg.Go(func() {
 		_, errs[0] = b.ReadSecrets(t.Context(), []string{"id1"})
-	}()
-	go func() {
-		defer wg.Done()
+	})
+	wg.Go(func() {
 		_, errs[1] = b.ReadSecrets(t.Context(), []string{"id2"})
-	}()
+	})
 
 	require.Eventually(t, func() bool {
 		return mt.inFlight.Load() >= 1
--- a/internal/pkg/bulk/secret_limit_test.go
+++ b/internal/pkg/bulk/secret_limit_test.go
@@ -13,6 +13,7 @@ import (
 	"sync"
 	"sync/atomic"
 	"testing"
+	"time"
 
 	"github.com/elastic/go-elasticsearch/v8"
 	"github.com/stretchr/testify/require"
@@ -69,10 +70,9 @@ func TestReadSecretsLimitsConcurrency(t *testing.T) {
 		_, errs[1] = b.ReadSecrets(context.Background(), []string{"id2"})
 	}()
 
-	// Spin until at least one request is in the transport.
-	for mt.inFlight.Load() < 1 {
-		// wait for the first goroutine to enter the transport
-	}
+	require.Eventually(t, func() bool {
+		return mt.inFlight.Load() >= 1
+	}, time.Second, time.Millisecond, "wait for the first goroutine to enter the transport")
 
 	// With semaphore capacity 1, the second goroutine is blocked on Acquire
 	// and cannot have entered the transport yet.
@@ -147,10 +147,9 @@ func TestReadSecretsDefaultConcurrency(t *testing.T) {
 		})
 	}
 
-	// Spin until all slots are occupied.
-	for mt.inFlight.Load() < int64(defaultMaxConcurrentSecretReads) {
-		// wait for all goroutines to enter the transport
-	}
+	require.Eventually(t, func() bool {
+		return mt.inFlight.Load() >= int64(defaultMaxConcurrentSecretReads)
+	}, time.Second, time.Millisecond, "wait for all goroutines to enter the transport")
 
 	// The semaphore is now full; a cancelled-context acquire must return immediately.
 	ctx, cancel := context.WithCancel(context.Background())
--- a/internal/pkg/bulk/secret_limit_test.go
+++ b/internal/pkg/bulk/secret_limit_test.go
@@ -23,7 +23,12 @@ import (
 // and tracks how many requests are currently in-flight.
 type blockingTransport struct {
 	gate     chan struct{}
+	gateOnce sync.Once
 	inFlight atomic.Int64
+}
+
+func (m *blockingTransport) unblock() {
+	m.gateOnce.Do(func() { close(m.gate) })
 }
 
 func (m *blockingTransport) RoundTrip(_ *http.Request) (*http.Response, error) {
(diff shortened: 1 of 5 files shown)


belimawr (commented):
I tried to run the tests but they failed... Could you take a look?

Do I need anything aside `go test ...` to run those tests?

Also, could you clarify how to test this locally?

<details><summary>Go test output</summary>
<p>

```
% go test -v -count=1 -run=TestReadSecrets ./internal/pkg/bulk
=== RUN   TestReadSecretsLimitsConcurrency
    secret_limit_test.go:83: 
                Error Trace:    /home/tiago/fleet-server/internal/pkg/bulk/secret_limit_test.go:83
                Error:          Received unexpected error:
                                the client noticed that the server is not Elasticsearch and we do not support this unknown product
                Test:           TestReadSecretsLimitsConcurrency
--- FAIL: TestReadSecretsLimitsConcurrency (0.00s)
=== RUN   TestReadSecretsContextCancelledWhileWaiting
--- PASS: TestReadSecretsContextCancelledWhileWaiting (0.00s)
=== RUN   TestReadSecretsNoLimitWhenZero
    secret_limit_test.go:121: 
                Error Trace:    /home/tiago/fleet-server/internal/pkg/bulk/secret_limit_test.go:121
                Error:          Received unexpected error:
                                the client noticed that the server is not Elasticsearch and we do not support this unknown product
                Test:           TestReadSecretsNoLimitWhenZero
--- FAIL: TestReadSecretsNoLimitWhenZero (0.00s)
=== RUN   TestReadSecretsDefaultConcurrency
--- PASS: TestReadSecretsDefaultConcurrency (0.00s)
FAIL
FAIL    github.com/elastic/fleet-server/v7/internal/pkg/bulk    0.006s
FAIL
```

</p>
</details>

internal/pkg/bulk/secret_limit_test.go
```diff
@@ -0,0 +1,149 @@
...
+	// a single unique secret ID. Each goroutine will acquire one semaphore slot
+	// and block in the transport, filling all capacity.
+	var wg sync.WaitGroup
+	for i := range defaultAPIKeyMaxParallel {
+		wg.Add(1)
+		go func() {
+			defer wg.Done()
+			_, _ = b.ReadSecrets(context.Background(), []string{fmt.Sprintf("id%d", i)})
+		}()
+	}
```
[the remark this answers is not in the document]
ycombinator: Yes, it does.  We are using Go v1.26.5.

ycombinator:
@belimawr Try running the tests again now.  They were missing the `X-Elastic-Product` header.

internal/pkg/bulk/secret_limit_test.go
```diff
@@ -0,0 +1,164 @@
...
+	}()
+
+	// Spin until at least one request is in the transport.
+	for mt.inFlight.Load() < 1 {
+		// wait for the first goroutine to enter the transport
+	}
+
+	// With semaphore capacity 1, the second goroutine is blocked on Acquire
+	// and cannot have entered the transport yet.
+	require.Equal(t, int64(1), mt.inFlight.Load())
```
belimawr: nit: the code can be modernised with `wg.Go` and use `require.Eventually` instead of a busy wait:
```suggestion
	wg.Go(func() { _, errs[0] = b.ReadSecrets(t.Context(), []string{"id1"}) })
	wg.Go(func() { _, errs[0] = b.ReadSecrets(t.Context(), []string{"id2"}) })

	// With semaphore capacity 1, the second goroutine is blocked on Acquire
	// and cannot have entered the transport yet.
	require.Eventually(
		t,
		func() bool { return mt.inFlight.Load() == 1 },
		time.Second,
		10*time.Millisecond,
		"there are one or less requests in flight")
```

internal/pkg/bulk/secret_limit_test.go
```diff
@@ -0,0 +1,164 @@
...
+	for i := range defaultMaxConcurrentSecretReads {
+		wg.Go(func() {
+			_, _ = b.ReadSecrets(context.Background(), []string{fmt.Sprintf("id%d", i)})
+		})
+	}
+
+	// Spin until all slots are occupied.
+	for mt.inFlight.Load() < int64(defaultMaxConcurrentSecretReads) {
+		// wait for all goroutines to enter the transport
+	}
```
belimawr: That can be replaced by a `require.Eventually`
```suggestion
	require.Eventually(
		t,
		func() bool { return mt.inFlight.Load() == defaultMaxConcurrentSecretReads },
		time.Second,
		10*time.Millisecond,
		"expecting %d in flight requests to block",
		defaultMaxConcurrentSecretReads,
	)
```

belimawr (commented):
Only the busy wait are blocking.

One last nit pick (feel free to ignore it): you could use `t.Context()` instead of `context.Background()`.

internal/pkg/bulk/secret_limit_test.go
ycombinator: Done in aaa3d9de and 229dfcef.

internal/pkg/bulk/secret_limit_test.go
ycombinator: Done in https://github.com/elastic/fleet-server/commit/aaa3d9de8eacbe1e5b2167b3b4d39769fa7f5835.

swiatekm:
The logic looks correct to me, but why is this a separate ES connection for each goroutine? Shouldn't it multiplex or go through a pool?

ycombinator:
> The logic looks correct to me, but why is this a separate ES connection for each goroutine? Shouldn't it multiplex or go through a pool?

The requests do go through a connection pool.  The ES client used by the individual requests is shared. That client's HTTP transport [pools and reuses connections](https://github.com/ycombinator/fleet-server/blob/229dfcefb7b24fd756dadd820e02e1d58b3567a4/internal/pkg/config/output.go#L99-L110).  

The requests remain individual, one per secret, because the Fleet secrets ES API has no bulk-read endpoint.  This change limits concurrent in-flight reads to 32.

swiatekm (dismissed):

internal/pkg/bulk/secret_limit_test.go
```diff
@@ -0,0 +1,160 @@
...
+	require.NoError(t, err)
+	return NewBulker(esClient, nil, opts...)
+}
+
+// TestReadSecretsLimitsConcurrency verifies that WithMaxConcurrentSecretReads(1)
+// allows at most one ReadSecret HTTP call in-flight at a time across concurrent
+// ReadSecrets callers.
+func TestReadSecretsLimitsConcurrency(t *testing.T) {
+	mt := &blockingTransport{gate: make(chan struct{})}
+	b := newTestBulkerWithTransport(t, mt, WithMaxConcurrentSecretReads(1))
```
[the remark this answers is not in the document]
ycombinator: Fixed in bd742bad.

belimawr (approved):

blakerouse (approved):
Looks good.

ycombinator:
@Mergifyio backport 9.5 9.4 8.19

PR #7568, opened 2026-08-06
fix: size ristretto cache from GOMEMLIMIT instead of host RAM

ycombinator:

`memEnvLimits()` in `internal/pkg/config/env_defaults.go` calls `memory.TotalMemory()` to select the ristretto cache tier. `memory.TotalMemory()` returns the **host node's total physical RAM**, not the container's cgroup memory limit. On a Kubernetes node with ≥16 GB of RAM, this selects a `MaxCost` of 256–512 MB — 2–4× the pod's GOMEMLIMIT (128 MB at the 256 M default pod limit). The cache is allowed to grow past the GOMEMLIMIT, triggering OOMKills even when live application heap is within budget.

This was the primary driver of the OOMKills observed for a high-volume serverless project in [elastic/ingest-dev#8991](https://github.com/elastic/ingest-dev/issues/8991).

Introduces `containerMemoryMB()`, which reads the current GOMEMLIMIT via `debug.SetMemoryLimit(-1)` and converts it to MiB. When GOMEMLIMIT is set (i.e. ≠ `math.MaxInt64`), that value drives cache tier selection — so a pod with a 256 M limit and GOMEMLIMIT = 128 M will select a cache sized for 128 MB of available memory, not 16+ GB of node RAM.

When GOMEMLIMIT is unset (`math.MaxInt64`), the function falls back to `memory.TotalMemory()`, preserving the existing behaviour for non-containerised deployments.

`memMB` remains a `var` pointing to `containerMemoryMB`, so existing tests that stub `memMB` are unaffected.

Run the config package tests:

```
go test ./internal/pkg/config/... -run TestContainerMemoryMB -v
```

`TestContainerMemoryMB` covers both the GOMEMLIMIT path (sets a 256 MiB limit and asserts the function returns 256) and the fallback path (clears the limit and asserts the function returns host RAM in MiB).

To verify the end-to-end effect: start fleet-server with `GOMEMLIMIT=128MiB` (or via `server.runtime.memory_limit`) and observe that the logged `recommended_mb` value is ≤ 128, not 256–512.

## Design Checklist

## Checklist

## Related issues

- Closes elastic/ingest-dev#8991

### DIFF
--- a/internal/pkg/config/env_defaults.go
+++ b/internal/pkg/config/env_defaults.go
@@ -312,9 +312,9 @@ func loadLimits(log *zerolog.Logger, agentLimit int) *envLimits {
 		// get nearest limits for configured agent numbers
 		if l.Agents.Min <= agentLimit && agentLimit <= l.Agents.Max {
 			log.Info().Msgf("Using system limits for %d to %d agents for a configured value of %d agents", l.Agents.Min, l.Agents.Max, agentLimit)
-			ramSize := memory.TotalMemory() / 1024 / 1024
+			ramSize := memMB()
 			if ramSize < l.RecommendedRAM {
-				log.Warn().Msgf("Detected %d MB of system RAM, which is lower than the recommended amount (%d MB) for the configured agent limit", ramSize, l.RecommendedRAM)
+				log.Warn().Msgf("Detected %d MB of available memory, which is lower than the recommended amount (%d MB) for the configured agent limit", ramSize, l.RecommendedRAM)
 			}
 			return l
 		}
--- a/internal/pkg/config/env_defaults.go
+++ b/internal/pkg/config/env_defaults.go
@@ -11,6 +11,7 @@ import (
 	"io/fs"
 	"math"
 	"runtime"
+	"runtime/debug"
 	"strings"
 	"time"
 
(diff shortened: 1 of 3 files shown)


lorienhu (dismissed):
Just a suggestion but since [GOMEMLIMIT isn't currently automatically set](https://github.com/golang/go/issues/75164), might be worth looking at reading it from cgroup limits similar to https://github.com/KimMachineGun/automemlimit.

internal/pkg/config/env_defaults.go
lorienhu: Do we need to respect GOMEMLIMIT here too?

ycombinator: Yes! Thanks for catching!  Fixed in be781d31.

ycombinator:
> Just a suggestion but since [GOMEMLIMIT isn't currently automatically set](https://github.com/golang/go/issues/75164), might be worth looking at reading it from cgroup limits similar to https://github.com/KimMachineGun/automemlimit.

It's a good idea.  Let's do it in a follow up PR since this PR here addresses an immediate need in Serverless Production.

lorienhu (approved):

ycombinator:
> > Just a suggestion but since [GOMEMLIMIT isn't currently automatically set](https://github.com/golang/go/issues/75164), might be worth looking at reading it from cgroup limits similar to https://github.com/KimMachineGun/automemlimit.
> 
> It's a good idea. Let's do it in a follow up PR since this PR here addresses an immediate need in Serverless Production.

https://github.com/elastic/fleet-server/pull/7573

PR #7571, opened 2026-08-06
fix: gracefully handle missing secrets in policy monitor

### LINKED ISSUE #7536: Add graceful handling for missing secrets in policy monitor
## Summary

Fleet Server currently fails fatally on startup if any secret referenced in its compiled agent policy cannot be fetched from `/_fleet/secret`. A single 404 on any `secret_references` entry causes the policy monitor to abort — there is no partial-load path, no recovery, and no actionable surface in the Fleet UI.

## Current behavior

On startup (and on each policy refresh), Fleet Server fetches all secrets listed in `secret_references` of its compiled policy. If any secret returns 404, Fleet Server logs:

```
Error: failed to get secret values: unexpected status 404 from fleet secret read:
{"error":{"root_cause":[{"type":"resource_not_found_exception","reason":"No secret with id [<id>]"}]}}
Fleet Server failed
```

Fleet Server then enters a crash loop, retrying indefinitely. There is no indication in the Fleet UI that the fleet-server agent is in this state or which secret is missing.

## Why this is a problem

Fleet Server's `secret_references` aggregates secrets from **all** sources in its compiled policy:
- Output SSL keys
- Fleet Server Host SSL keys
- **All package policy credentials** on the same agent policy (integration API keys, etc.)

A missing secret from an integration policy — unrelated to fleet-server's own connectivity — can prevent Fleet Server from starting at all. This is a disproportionate blast radius: one deleted integration credential can take the entire Fleet Server offline.

Additionally, the failure mode is silent from a user perspective: the Fleet UI shows the fleet-server agent as offline with no actionable error message pointing to the missing secret or which integration caused it.

## Proposed behavior

Evaluate one or more of the following:

1. **Confirm if a guard already exists** — if fleet-server already has a path to skip or warn on individual secret fetch failures, document it; the behavior observed may be a configuration or version-specific issue.

2. **Log and skip non-critical missing secrets** — if the missing secret is not required for fleet-server's own connectivity (i.e., it belongs to a package policy credential, not an SSL key or output credential), log a warning and continue loading the policy without that secret. The affected integration would degrade gracefully rather than blocking fleet-server.

3. **Surface the error in the Fleet UI** — expose the missing secret ID (or the integration it belongs to) as a health check failure on the fleet-server agent, so operators can identify and remediate without grepping logs.

4. **Categorize secrets by criticality** — distinguish between secrets that are required for fleet-server operation (SSL keys for the Fleet Server Host and ES output) and secrets that belong to package policies (integration credentials). Only the former should block startup.

## Steps to reproduce

1. Fleet Server agent policy includes at least one integration policy with a secret (e.g., an API key stored via `/_fleet/secret`)
2. The secret document is deleted from Elasticsearch (e.g., via accidental policy update, ES-level operation, or snapshot restore)
3. Fleet Server restarts — it cannot load its policy and enters a crash loop
4. No recovery is possible without re-entering the integration credentials in the Fleet UI to generate a new secret

## Environment

Observed on Elastic Cloud Serverless (Security project) but the code path is general to all Fleet Server deployments. Affects any version where fleet-server fetches secrets atomically on policy load.

## Related

- Kibana `full_agent_policy.ts` — `secret_references` aggregation: [source](https://github.com/elastic/kibana/blob/main/x-pack/platform/plugins/shared/fleet/server/services/agent_policies/full_agent_policy.ts)
- Kibana #273040: Fleet KQL injection in secret cleanup (separate but related to secret lifecycle)

ycombinator:

Fleet Server crash-loops when any secret referenced in an agent policy's `secret_references` cannot be found in the Fleet secrets store. A single missing secret — from a deleted or orphaned integration credential — causes the entire policy monitor to abort with:

```
Error: failed to get secret values: unexpected status 404 from fleet secret read: ...
Fleet Server failed
```

Fleet Server then restarts, re-reads the same stale policy document from the policy monitor (the checkpoint hasn't advanced), encounters the same 404, and crashes again — indefinitely. Every agent on that fleet-server instance goes offline for the duration.

This is a disproportionate blast radius: one deleted integration credential (unrelated to fleet-server's own connectivity) can take the entire Fleet Server offline.

The root causes that produce missing secrets are tracked in Kibana (elastic/kibana#282911 and elastic/kibana#282912) and in fleet-server (#7533, #7534). This PR is a **defense-in-depth fix**: even after those root causes are resolved, fleet-server should not treat a single missing package-policy credential as a reason to crash-loop and take all agents offline.

Two changes:

1. **`bulk/secret.go` — `ExtendedAPI.Read`**: Returns a typed `ErrSecretNotFound` sentinel when the Fleet secrets API responds with 404. This lets callers distinguish "this secret document doesn't exist" from network/auth failures (403, 500, etc.), which should still be fatal.

2. **`bulk/engine.go` — `ReadSecrets`**: When `ReadSecret` returns `ErrSecretNotFound`, logs a warning with the secret ID and skips it (continues the loop) rather than aborting. The returned map contains only the secrets that were successfully fetched.

With this change, a missing secret causes the policy to load without that secret value — the raw `$co.elastic.secret{id}` placeholder remains unsubstituted in the affected integration's config. That integration's credentials won't resolve (it will fail to authenticate), but fleet-server itself stays up and continues serving all other agents.

Callers that need hard failure for missing secrets (e.g. output API key resolution, where an empty key would silently break agent→ES connectivity) can check whether the expected ID is present in the returned map — as the second commit in #7416 already does.

1. Start a local stack with fleet-server
2. Create an integration that uses secrets (e.g. any OAuth-based integration)
3. Directly delete the secret document from Elasticsearch: `DELETE /_fleet/secret/<id>`
4. Restart fleet-server and observe that it:
   - Logs `WARN secret not found; policy will load without it secret_id=<id>`
   - Continues to start successfully
   - Serves other agents normally
5. Verify (via agent diagnostics) that the affected integration's credential field contains the raw `$co.elastic.secret{id}` placeholder rather than the secret value

## Design Checklist

- [x] I have ensured my design is stateless and will work when multiple fleet-server instances are behind a load balancer. (Each instance evaluates `ReadSecrets` independently per policy load; skipping is per-request with no shared state.)
- [ ] I have or intend to scale test my changes, ensuring it will work reliably with 100K+ agents connected. (The change reduces work on failure — skipping a secret is cheaper than returning an error — so scale impact is neutral or positive.)
- [x] I have included fail safe mechanisms to limit the load on fleet-server: rate limiting, circuit breakers, caching, load shedding, etc. (No new load introduced; 404s are now fast-path skips.)

## Checklist

## Related issues

- Closes #7536
- Root cause (how secrets become dangling in the first place):
  - elastic/kibana#282911 — `agentPolicyService.delete()` can leave orphaned `.fleet-policies` documents when the cleanup call after package-policy deletion fails silently
  - elastic/kibana#282912 — `deleteSecretsIfNotReferenced` only checks live package-policy saved objects, not `.fleet-policies`, so secrets are deleted even when an orphaned policy document still references them
- Mitigation for the agent-update-failure source of dangling secrets: #7533, #7534
- Context: the `>= 400` check in `ExtendedAPI.Read` was introduced in #7416. That PR's own second commit described the *original* tolerant design — `ReadSecrets` was not expected to fail on a missing secret; hard failure for the output API key was added at the caller level. This PR restores that layered design by making 404 non-fatal in `ReadSecrets` while preserving fatal behavior for other 4xx/5xx responses.

### DIFF
--- a/changelog/fragments/1786053515-graceful-missing-secrets.yaml
+++ b/changelog/fragments/1786053515-graceful-missing-secrets.yaml
@@ -1,0 +1,20 @@
+kind: bug-fix
+
+summary: Gracefully handle missing secrets in policy monitor instead of crash-looping
+
+description: |
+  Fleet Server no longer crash-loops when a secret referenced in an agent
+  policy's `secret_references` cannot be found in the Fleet secrets store.
+  Previously, any 404 from `GET /_fleet/secret/{id}` was treated as a fatal
+  error, causing the entire policy monitor to abort and fleet-server to
+  restart indefinitely — taking all agents offline.
+
+  Now, a missing secret is logged as a warning and skipped. The policy loads
+  without that secret value; affected integration credentials will not be
+  substituted (the raw `$co.elastic.secret{id}` placeholder remains), causing
+  only that integration to degrade, while fleet-server itself continues
+  serving all other agents normally.
+
+component: fleet-server
+
+pr: https://github.com/elastic/fleet-server/pull/7566
--- a/changelog/fragments/1786053515-graceful-missing-secrets.yaml
+++ b/changelog/fragments/1786053515-graceful-missing-secrets.yaml
@@ -17,4 +17,4 @@ description: |
 
 component: fleet-server
 
-pr: https://github.com/elastic/fleet-server/pull/7566
+pr: https://github.com/elastic/fleet-server/pull/7571
--- a/internal/pkg/bulk/engine.go
+++ b/internal/pkg/bulk/engine.go
@@ -330,6 +330,10 @@ func (b *Bulker) ReadSecrets(ctx context.Context, secretIds []string) (map[strin
 	for _, id := range secretIds {
 		val, err := ReadSecret(ctx, esClient, id)
 		if err != nil {
+			if errors.Is(err, ErrSecretNotFound) {
+				zerolog.Ctx(ctx).Warn().Str("secret_id", id).Msg("secret not found; policy will load without it")
+				continue
+			}
 			return nil, err
 		}
 		result[id] = val
--- a/internal/pkg/bulk/secret.go
+++ b/internal/pkg/bulk/secret.go
@@ -8,6 +8,7 @@ import (
 	"bytes"
 	"context"
 	"encoding/json"
+	"errors"
 	"fmt"
 	"io"
 	"net/http"
@@ -15,6 +16,9 @@ import (
 	"github.com/elastic/go-elasticsearch/v8"
 	"go.elastic.co/apm/v2"
 )
+
+// ErrSecretNotFound is returned when a secret document does not exist in the Fleet secrets store.
+var ErrSecretNotFound = errors.New("secret not found")
 
 type ExtendedClient struct {
 	*elasticsearch.Client
@@ -40,6 +44,9 @@ func (c *ExtendedAPI) Read(ctx context.Context, secretID string) (*SecretRespons
 		return nil, err
 	}
 	defer res.Body.Close()
+	if res.StatusCode == http.StatusNotFound {
+		return nil, fmt.Errorf("%q: %w", secretID, ErrSecretNotFound)
+	}
 	if res.StatusCode >= 400 {
 		body, _ := io.ReadAll(res.Body)
 		return nil, fmt.Errorf("unexpected status %d from fleet secret read: %s", res.StatusCode, body)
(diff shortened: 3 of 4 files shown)


changelog/fragments/1786053515-graceful-missing-secrets.yaml
```diff
@@ -0,0 +1,20 @@
...
+
+  Now, a missing secret is logged as a warning and skipped. The policy loads
+  without that secret value; affected integration credentials will not be
+  substituted (the raw `$co.elastic.secret{id}` placeholder remains), causing
+  only that integration to degrade, while fleet-server itself continues
+  serving all other agents normally.
+
+component: fleet-server
+
+pr: https://github.com/elastic/fleet-server/pull/7566
```
MichelLosier:
```suggestion
pr: https://github.com/elastic/fleet-server/pull/7571
```

MichelLosier (approved):
LGTM!

blakerouse (approved):

PR #7572, opened 2026-08-06
fix: add revision guard to updatePolicy to reject stale policy revisions

### LINKED ISSUE #7570: policy/monitor: updatePolicy() has no revision guard — change-feed can overwrite cached policy with stale revisions
- **Version:** serverless production (confirmed via `bf10b309` policy across `#8.9`, `#9.3`, `#9.5`, `#9.6` index generations)
- **Operating System:** Linux (Kubernetes)
- **Discuss Forum URL:** N/A — internal bug, discovered during customer incident investigation (sdh-beats#7452)
- **Steps to Reproduce:**

  1. Have a `.fleet-policies` index that has gone through one or more Kibana index migrations (e.g. 8.9 → 9.3 → 9.5 → 9.6). Each migration reindexes all documents into a new backing index, assigning new `_seq_no` values.
  2. If during a migration, an older revision (e.g. `revision_idx=7`) is written to the new backing index *after* a newer revision (`revision_idx=8`), the older revision ends up with a higher `_seq_no` within that index.
  3. Fleet-server starts. The startup path (`QueryLatestPolicies` — terms aggregation + top_hits) correctly loads only `revision_idx=8`.
  4. The change-feed (`simpleMonitorT`) begins polling `.fleet-policies` by `_seq_no` range. It delivers `revision_idx=8` in one batch, advancing the checkpoint.
  5. In a subsequent batch it delivers `revision_idx=7` (which has a higher `_seq_no` from the migration). `groupByLatest()` cannot deduplicate across batches.
  6. `updatePolicy()` unconditionally overwrites the correctly-cached `revision_idx=8` with the stale `revision_idx=7`.
  7. If `revision_idx=7` references secrets that have since been deleted, `NewParsedPolicy` → `secret.GetSecretValues` fails and fleet-server crashes.
  8. On restart, the checkpoint rolls back to its pre-crash state, so `revision_idx=7` (with its migration-assigned `_seq_no`) is above the checkpoint again. The cycle repeats indefinitely.

## Root cause

`updatePolicy()` in `internal/pkg/policy/monitor.go` has no revision guard:

```go
// current — unconditional overwrite:
p.pp = *pp
m.policies[newPolicy.PolicyID] = p
```

The fix is to reject incoming documents with a `RevisionIdx` ≤ the currently cached revision:

```go
if newPolicy.RevisionIdx <= p.pp.Policy.RevisionIdx {
    zlog.Warn().
        Str("policy_id", newPolicy.PolicyID).
        Int64("incoming_rev", newPolicy.RevisionIdx).
        Int64("cached_rev", p.pp.Policy.RevisionIdx).
        Msg("ignoring stale revision from change-feed")
    return false
}
p.pp = *pp
```

## Related issues

- #7536 — fleet-server treats any missing secret as fatal (no graceful degradation); this is the reason a stale revision causes a crash rather than a warning
- elastic/kibana#282911 — agent policy not cleaned up on integration deletion (produces orphaned `.fleet-policies` documents)
- elastic/kibana#282912 — secrets deleted without checking `.fleet-policies` for active references

ycombinator:
## Summary

- `updatePolicy()` in `internal/pkg/policy/monitor.go` unconditionally overwrote the in-memory policy cache with any document the policy monitor delivered, regardless of its `revision_idx`.
- After a Kibana index migration (e.g. 8.9→9.3→9.5→9.6), older revision documents can be reindexed with a higher `_seq_no` than the latest revision. The policy monitor then delivers the stale revision in a later batch (after `groupByLatest()` has already passed the clean revision through), overwriting the cache.
- If the stale revision references deleted secrets, `NewParsedPolicy` fails and fleet-server crash-loops indefinitely (the checkpoint never advances past the stale doc).
- This adds a strict `revision_idx` guard: incoming documents whose `revision_idx` is not greater than the cached revision are rejected with a warning log. Unit tests cover the equal, older, and newer revision cases.

## References

Closes #7570
Related: #7536 (missing-secrets crash that this guard also mitigates)

### DIFF
--- a/internal/pkg/policy/monitor.go
+++ b/internal/pkg/policy/monitor.go
@@ -451,16 +451,6 @@ func (m *monitorT) updatePolicy(ctx context.Context, pp *ParsedPolicy) bool {
 		return false
 	}
 
-	// Secondary stale-revision guard (primary check is in processPolicies before parsing).
-	// Rejects any revision whose revision_idx is not newer than the cached revision,
-	// guarding against races or direct callers that bypass the pre-parse check.
-	if newPolicy.RevisionIdx <= p.pp.Policy.RevisionIdx {
-		zlog.Warn().
-			Int64("cached_revision_idx", p.pp.Policy.RevisionIdx).
-			Msg("ignoring stale policy revision from policy monitor")
-		return false
-	}
-
 	// Cache the old stored policy for logging
 	oldPolicy := p.pp.Policy
 
--- a/internal/pkg/policy/monitor.go
+++ b/internal/pkg/policy/monitor.go
@@ -434,6 +434,17 @@ func (m *monitorT) updatePolicy(ctx context.Context, pp *ParsedPolicy) bool {
 		return false
 	}
 
+	// Reject stale revisions — the change-feed can deliver old documents (e.g. after a
+	// Kibana index migration that assigns higher _seq_no values to older revisions) after
+	// a newer revision has already been cached. Without this guard, an old revision with
+	// dangling secrets would overwrite the good cache entry and crash-loop fleet-server.
+	if newPolicy.RevisionIdx <= p.pp.Policy.RevisionIdx {
+		zlog.Warn().
+			Int64("cached_revision_idx", p.pp.Policy.RevisionIdx).
+			Msg("ignoring stale policy revision from change-feed")
+		return false
+	}
+
 	// Cache the old stored policy for logging
 	oldPolicy := p.pp.Policy
 
--- a/internal/pkg/policy/monitor.go
+++ b/internal/pkg/policy/monitor.go
@@ -378,6 +378,13 @@ func (m *monitorT) processPolicies(ctx context.Context, policies []model.Policy)
 
 	latest := m.groupByLatest(policies)
 	for _, policy := range latest {
+		if m.isStaleRevision(policy.PolicyID, policy.RevisionIdx) {
+			m.log.Warn().
+				Str(ecs.PolicyID, policy.PolicyID).
+				Int64(ecs.RevisionIdx, policy.RevisionIdx).
+				Msg("skipping stale policy revision; secret resolution and policy parse skipped")
+			continue
+		}
 		pp, err := NewParsedPolicy(ctx, m.bulker, policy)
 		if err != nil {
 			return err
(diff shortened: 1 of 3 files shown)


internal/pkg/policy/monitor.go
```diff
@@ -434,6 +434,17 @@ func (m *monitorT) updatePolicy(ctx context.Context, pp *ParsedPolicy) bool {
 		return false
 	}
 
+	// Reject stale revisions — the change-feed can deliver old documents (e.g. after a
+	// Kibana index migration that assigns higher _seq_no values to older revisions) after
+	// a newer revision has already been cached. Without this guard, an old revision with
+	// dangling secrets would overwrite the good cache entry and crash-loop fleet-server.
```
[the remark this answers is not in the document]
ycombinator: Good catch — reworked the comment to focus on cache correctness and outdated policy delivery rather than crash-loop, since #7571 already makes missing secrets non-fatal.

internal/pkg/policy/monitor.go
lorienhu, line 388: Shouldn't we do revision rejection here if NewParsedPolicy is referencing stale secrets?

internal/pkg/policy/monitor.go
```diff
@@ -434,6 +434,17 @@ func (m *monitorT) updatePolicy(ctx context.Context, pp *ParsedPolicy) bool {
...
+	if newPolicy.RevisionIdx <= p.pp.Policy.RevisionIdx {
+		zlog.Warn().
+			Int64("cached_revision_idx", p.pp.Policy.RevisionIdx).
+			Msg("ignoring stale policy revision from change-feed")
+		return false
+	}
```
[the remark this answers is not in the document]
ycombinator: Addressed in 6ed24f08.

internal/pkg/policy/monitor.go
ycombinator: Yes, same as https://github.com/elastic/fleet-server/pull/7572#discussion_r3732235259; addressed in 6ed24f08.

internal/pkg/policy/monitor.go
```diff
@@ -434,6 +451,16 @@ func (m *monitorT) updatePolicy(ctx context.Context, pp *ParsedPolicy) bool {
 		return false
 	}
 
+	// Secondary stale-revision guard (primary check is in processPolicies before parsing).
+	// Rejects any revision whose revision_idx is not newer than the cached revision,
+	// guarding against races or direct callers that bypass the pre-parse check.
```
macdewee: > guarding against races or direct callers that bypass the pre-parse check.

Do we have any situations like that, or is it strictly just in case?

macdewee (dismissed):
LGTM, just one question to get better understanding

internal/pkg/policy/monitor.go
ycombinator: It was strictly "just in case", a defensive guard.  We don't actually need it.  The primary guard in `processPolicies` is sufficient.

I've removed this guard to simplify the code in bce98b3c.

macdewee (approved):
Thanks!

PR #7573, opened 2026-08-06
fix: read cgroup memory limit for cache sizing when GOMEMLIMIT is not set

ycombinator:

When `GOMEMLIMIT` is not explicitly set, `containerMemoryMB()` (introduced in #7568) falls back directly to host RAM. A container with a cgroup memory limit but no explicit `GOMEMLIMIT` would use the node's full RAM for memory-based sizing decisions in fleet-server.

Extends `containerMemoryMB()` with a cgroup tier between GOMEMLIMIT and host RAM:

1. **GOMEMLIMIT** — used when explicitly set
2. **cgroup v2** — reads `/sys/fs/cgroup/memory.max`; skips if value is `"max"` (unlimited)
3. **cgroup v1** — reads `/sys/fs/cgroup/memory/memory.limit_in_bytes`; skips if value ≥ `MaxInt64` (unlimited sentinel)
4. **host total RAM** — final fallback for non-containerised deployments

On non-Linux hosts the cgroup files simply don't exist, so the reads fail silently and the fallback chain continues — no platform-specific build tags needed.

```
go test ./internal/pkg/config/... -run "TestContainerMemoryMB|TestReadCgroupMemoryFile" -v
```

`TestReadCgroupMemoryFile` covers: valid byte-limit, `"max"` unlimited sentinel, `MaxInt64` unlimited sentinel (cgroup v1), missing file, and invalid content.

## Design Checklist

## Checklist

## Related issues

- Follow-up to #7568 (suggested by @lorienhu in https://github.com/elastic/fleet-server/pull/7568#pullrequestreview-4878590311)
- Closes elastic/ingest-dev#8991


internal/pkg/config/env_defaults.go
```diff
@@ -323,17 +325,55 @@ func loadLimits(log *zerolog.Logger, agentLimit int) *envLimits {
...
+	data, err := os.ReadFile(path)
+	if err != nil {
+		return 0, false
+	}
+	s := strings.TrimSpace(string(data))
+	if s == "max" {
+		return 0, false
+	}
+	n, err := strconv.ParseUint(s, 10, 64)
+	if err != nil || n >= math.MaxInt64 {
```
lorienhu: According to [this](https://unix.stackexchange.com/questions/420906/what-is-the-value-for-the-cgroups-limit-in-bytes-if-the-memory-is-not-restricted) the sentinel value for cgroup v1 is not quite MaxInt64 because it's rounded. Double checked [here](https://github.com/torvalds/linux/blob/3f008280327ba5ad132965abab0c7846283cef0c/mm/memcontrol-v1.c#L1623).

internal/pkg/config/env_defaults.go
```diff
@@ -323,17 +325,55 @@ func loadLimits(log *zerolog.Logger, agentLimit int) *envLimits {
...
+	if mb, ok := cgroupMemMB(); ok {
+		return mb
+	}
 	return memory.TotalMemory() / 1024 / 1024
 }
 
+// cgroupMemoryLimitMB reads the container memory limit from cgroup files.
+// It tries cgroup v2 first, then cgroup v1. Returns (0, false) when no
+// applicable limit is found (unlimited, missing file, or parse error).
+func cgroupMemoryLimitMB() (uint64, bool) {
```
lorienhu, line 352: Might want to add a comment here that this doesn't work with nested cgroups.

internal/pkg/config/env_defaults.go
ycombinator: Nice catch. Fixed in 94964ece.

internal/pkg/config/env_defaults.go
ycombinator: Done in e0789559.

lorienhu (approved):

PR #7593, opened 2026-08-07
Add option to enable graceful unenroll to invalid API key agents.

blakerouse:

Once an Elastic Agent is force unenrolled the Elastic Agents will continue to communicate to Fleet Server. There are cases where it would be best to just have the Elastic Agent stop all of its components, unenroll if it can (those with tamper protection on will not be able to), and those that cannot will continue to receive 401's for a full hour, until the cycle starts again.

This changes the behavior of invalid API keys from being a 401 error that just gets retried non-stop to a 200 with a policy change action that is an empty policy, then a unenroll action, and then back to the 401 error.

1. The empty policy will cause the Elastic Agent to stop all of its running components.
2. The unenroll action will cause the Elastic Agent to stop communicating with Fleet (as long as there is not tamper protection enabled, if enabled this will not stop the Elastic Agent).
3. 401 original error returns for a full hour.

1. Enroll an Elastic Agent.
4. Perform force unenroll in Fleet.
5. Wait for next check-in and see that it stops all of its running components.
6. Wait for follow-up check-in and see that it unenrolls.

## Design Checklist

## Checklist


internal/pkg/api/handleCheckin.go
```diff
@@ -182,6 +187,49 @@ func invalidateAPIKeysOfInactiveAgent(ctx context.Context, zlog zerolog.Logger,
...
+func isInvalidAPIKeyErr(err error) bool {
+	return errors.Is(err, apikey.ErrAPIKeyNotFound) ||
+		errors.Is(err, apikey.ErrUnauthorized) ||
+		errors.Is(err, ErrAPIKeyNotEnabled) ||
+		errors.Is(err, ErrAgentInactive)
+}
+
+// writeUnenrollResponse writes a 200 check-in response containing a single UNENROLL action.
+// It is used when UnenrollOnInvalidAPIKey is enabled and the agent's API key is invalid.
+func (ct *CheckinT) writeUnenrollResponse(zlog zerolog.Logger, w http.ResponseWriter, agentID string) error {
```
cmacknz, line 318: For users with tamper protected Elastic Defend, will this work? The action would need to be signed by Kibana for endpoint to accept it

https://github.com/elastic/elastic-agent/blob/07b2c23a7e7b09abe1611cbd7be9bfe030eb0e93/internal/pkg/agent/application/actions/handlers/handler_action_unenroll.go#L73-L88

```go
func (h *Unenroll) handle(ctx context.Context, a fleetapi.Action, acker acker.Acker, timeout time.Duration) error {
	h.log.Debugf("handlerUnenroll: action '%+v' received", a)
	action, ok := a.(*fleetapi.ActionUnenroll)
	if !ok {
		return fmt.Errorf("invalid type, expected ActionUnenroll and received %T", a)
	}

	if h.tamperProtectionFn() {
		// Find inputs that want to receive UNENROLL action
		// Endpoint needs to receive a signed UNENROLL action in order to be able to uncontain itself
		state := h.coord.State()
		ucs := findMatchingUnitsByActionType(state, a.Type())
		if len(ucs) > 0 {
			err := notifyUnitsOfProxiedAction(ctx, h.log, action, ucs, h.coord.PerformAction)
			if err != nil {
				return err
			}
		} else {
			// Log and continue
			h.log.Debugf("No components running for %v action type", a.Type())
		}
	}
```

cmacknz: I think sending an empty policy change might work though

blakerouse: I have updated this PR to switch from sending the UNENROLL action to instead have it send the POLICY_CHANGE action with an empty policy.

blakerouse:
@cmacknz Changing to `POLICY_CHANGE` does have a side-effect. The `UNENROLL` action stops the Elastic Agent from communicating with Fleet. `POLICY_CHANGE` will just result in it returning 200, and then check back-in and then get another 200. That will just keep happening.

At the moment I have changed this to wait check-in long poll time before responding. That will prevent the storm of requests. But that does mean we are holding a connection open for a long period of time for a 401 invalid credential.

Another option is to still send 401's and track how long the client has been tried to connect for. Then after a full hour of 401's from the client, only then send the 200 POLICY_CHANGE. We could even get a little more advanced. We could have it send a `UNENROLL` and then if it comes back (because `UNENROLL` doesn't work because of tamper protection) then we send a `POLICY_CHANGE`.

The only weird part of the tracking is that when there are multiple fleet-server's how do we tracking this information cross process. We could add it to a new elasticsearch index, or we could just keep it in-memory per each and then it really just depend on how the routing occurs, after enough time the right thing would happen on one of them.

cmacknz:
> At the moment I have changed this to wait check-in long poll time before responding. That will prevent the storm of requests. But that does mean we are holding a connection open for a long period of time for a 401 invalid credential.

This is what the agent would have been doing when it was enrolled, so in theory the system will tolerate this.

We need to rework the way unenroll works to something like what is described in https://github.com/elastic/kibana/issues/196621#issuecomment-4263211158 so I think we'll get another iteration to polish this behaviour.

> Another option is to still send 401's and track how long the client has been tried to connect for. Then after a full hour of 401's from the client, only then send the 200 POLICY_CHANGE. We could even get a little more advanced. We could have it send a UNENROLL and then if it comes back (because UNENROLL doesn't work because of tamper protection) then we send a POLICY_CHANGE.
> 
> The only weird part of the tracking is that when there are multiple fleet-server's how do we tracking this information cross process. We could add it to a new elasticsearch index, or we could just keep it in-memory per each and then it really just depend on how the routing occurs, after enough time the right thing would happen on one of them.

Doing this correctly sounds too complex for what we need this to do, and as mentioned above I think we will change the way unenroll works and add an uninstall action as a proper prevention of this type of problem.

blakerouse:
> > At the moment I have changed this to wait check-in long poll time before responding. That will prevent the storm of requests. But that does mean we are holding a connection open for a long period of time for a 401 invalid credential.
> 
> This is what the agent would have been doing when it was enrolled, so in theory the system will tolerate this.

Except in this case its a client with a bad API key. So anyone can hold that slot. The comment from the link below doesn't address that either. This is a hard force unenroll path and that comment provides not cleanup or help for that situation. We need a solution for customers on current versions that have already done force unenroll.

> 
> We need to rework the way unenroll works to something like what is described in [elastic/kibana#196621 (comment)](https://github.com/elastic/kibana/issues/196621#issuecomment-4263211158) so I think we'll get another iteration to polish this behaviour.

This is all great but doesn't solve the issues we have with customers now. Where they have done the force unenroll and are stuck in this situation. I agree with all points in that issue, but it doesn't solve the problem at hand.

> 
> > Another option is to still send 401's and track how long the client has been tried to connect for. Then after a full hour of 401's from the client, only then send the 200 POLICY_CHANGE. We could even get a little more advanced. We could have it send a UNENROLL and then if it comes back (because UNENROLL doesn't work because of tamper protection) then we send a POLICY_CHANGE.
> > The only weird part of the tracking is that when there are multiple fleet-server's how do we tracking this information cross process. We could add it to a new elasticsearch index, or we could just keep it in-memory per each and then it really just depend on how the routing occurs, after enough time the right thing would happen on one of them.
> 
> Doing this correctly sounds too complex for what we need this to do, and as mentioned above I think we will change the way unenroll works and add an uninstall action as a proper prevention of this type of problem.

This doesn't solve the problem people have today.

cmacknz:
I think we should focus on providing a correction for the ongoing incident we know about with this, and change the design of our unenroll mechanism later to remove this possibility.

I don't want us to try to generalize this too much, especially because the most general solution (unenroll action) doesn't work in the specific situation we are trying to help with this change. 

Separately I saw the suggestion of having the agent run an empty policy when in the long scheduler mode, I like that more than making Fleet Server more complex. We could also have agent start respecting the Retry-After header and give fleet server more control over the interval (though this won't protect against malicious clients).

I think we need to be careful with allowing Fleet Server to do work for unauthenticated clients, and keep that behavior opt in and off by default, and change the design of the involved systems so that we don't hit failure modes that require it in the future.

blakerouse:
> I think we should focus on providing a correction for the ongoing incident we know about with this, and change the design of our unenroll mechanism later to remove this possibility.

Thats for sure, but this is not the only customer that is having this issue. I have another.

> 
> I don't want us to try to generalize this too much, especially because the most general solution (unenroll action) doesn't work in the specific situation we are trying to help with this change.
> 
> Separately I saw the suggestion of having the agent run an empty policy when in the long scheduler mode, I like that more than making Fleet Server more complex. We could also have agent start respecting the Retry-After header and give fleet server more control over the interval (though this won't protect against malicious clients).

Yes I agree that is the best option, but not something that helps customers today.

> 
> I think we need to be careful with allowing Fleet Server to do work for unauthenticated clients, and keep that behavior opt in and off by default, and change the design of the involved systems so that we don't hit failure modes that require it in the future.

This is definitely an opt-in never on by default, worst-case scenario enablement option. That is why I preferred the UNENROLL as it just stops the Elastic Agent's. That really is the best option, sadly the signing does prevent that from working. I think we need this option, we just need to determine if its okay for a 401 response actually holding a long poll connection. Without it, Elastic Agent will just spam fleet-server non-stop.

blakerouse:
I don't think we have the option of not doing something to solve this issue for existing customers having this problem for older versions of the Elastic Agent.

testing/e2e/stand_alone_test.go
```diff
@@ -732,6 +734,187 @@ func (suite *StandAloneSuite) TestOpAMPWithUpstreamCollector() {
...
+// fleet-server returns an UNENROLL action due to an invalidated API key.
+//
+// Fleet-server runs with:
+//   - unenroll_on_invalid_api_key: true
+//   - ttl_api_key: 2s  (cache expires quickly after key invalidation)
+//   - checkin_long_poll: 5s  (agent retries within seconds)
+//
+// The test observes only the elastic-agent's own log output — not fleet-server's API response —
+// to confirm the agent processes the UNENROLL action and stops checking in.
+func (suite *StandAloneSuite) TestAgentUnenrollsOnInvalidAPIKey() {
```
cmacknz: The test name still references `Unenroll` which is not what we do anymore.

fleet-server.reference.yml
```diff
@@ -286,6 +286,12 @@ fleet:
 #        // enable OpAMP endpoint (enabled by default)
 #        enable_opamp: true
 #
+#        // When true, a checkin with an invalid or disabled API key returns HTTP 200 with a
+#        // POLICY_CHANGE action carrying an empty policy instead of HTTP 401. This causes the
+#        // agent to stop all inputs.
```
cmacknz: We should document the explicit risk here that we will hold connections for unauthenticated agents open for the long poll interval, and that this is a setting to be used in emergency recovery situations and should otherwise be disabled.

This will help us avoid the LLM generated security report that will probably follow after we merge this.

internal/pkg/api/handleCheckin.go
```diff
@@ -182,6 +187,72 @@ func invalidateAPIKeysOfInactiveAgent(ctx context.Context, zlog zerolog.Logger,
...
+	}
+
+	zlog.Info().
+		Str(ecs.AgentID, agentID).
+		Str(ecs.ActionID, action.Id).
+		Msg("Returning empty POLICY_CHANGE for agent with invalid API key")
+
+	resp := CheckinResponse{
+		Action:  "checkin",
+		Actions: []Action{action},
```
cmacknz, line 305: We could include an UNENROLL action in this list as the second action to make this slightly safer in the non-tamper protected case?

We'd have to test this on a real agent though.

cmacknz: Though UNENROLL does come with the potential for more abuse, in that there is no recovery from unenroll right now so maybe let's avoid this.

internal/pkg/api/handleCheckin.go
```diff
@@ -153,6 +155,9 @@ func (ct *CheckinT) handleCheckin(zlog zerolog.Logger, w http.ResponseWriter, r
 			ctx := zlog.WithContext(r.Context())
 			invalidateAPIKeysOfInactiveAgent(ctx, zlog, ct.bulker, agent)
 		}
+		if ct.cfg.Features.EmptyPolicyOnInvalidAPIKey && isInvalidAPIKeyErr(err) {
+			return ct.writeEmptyPolicyResponse(zlog, w, r, id)
+		}
```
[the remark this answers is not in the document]
cmacknz: This is valid but there's not a simple way around it, perhaps we should have a separate configuration for how long to hold connections for unauthenticated agents open (that defaults to the checkin duration). This would let us tune this value to lower but still helpful values (30s) without affecting still enrolled agents.

cmacknz:
Is there any way to incorporate a real agent with endpoint into the automated tests here? Or do we have to do that in the agent repository? I think it's important we have that test case captured somewhere since it's the entire point of this feature.

internal/pkg/api/handleCheckin.go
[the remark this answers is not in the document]
blakerouse: I have changed the design of this, so this is no longer a valid point. Instead of holding a connection open I have Fleet Server do the following:

1. Return empty policy on first 401.
2. Return unenroll action on second 401.
3. Return 401 on all remaining 401's (for an hour).

This completely allows no need for state between Fleet Servers as no matter the round robin of Fleet Servers it doesn't matter. POLICY_CHANGE will always happen first, it could happen 3 times in a row (each a different Fleet Server, that is fine). Then finally it hits the same server gets a UNENROLL (that might fail, it might get that 3 times). Then finally it hits the server again and gets the 401.

At no point does a 401 elastic-agent cause a connection to be held any longer than it currently does for standard 401's. It does use a `sync.Map` to track this per Fleet Server, but this is very small and has a clear that removes elastic-agent's that stop communicating after a period of time.

fleet-server.reference.yml
blakerouse: See my comment above on the AI suggestion about holding connections open. I didn't like this, so I have changed the design. See what you think.

testing/e2e/stand_alone_test.go
blakerouse: Fixed.

blakerouse:
> Is there any way to incorporate a real agent with endpoint into the automated tests here? Or do we have to do that in the agent repository? I think it's important we have that test case captured somewhere since it's the entire point of this feature.

Yes that is being done in this PR already with the e2e tests here - https://github.com/elastic/fleet-server/pull/7593/changes#diff-a9b2609c608e655f4f1e170ea17f0b582801d8ccc6705f570cd27cec17025789R750

internal/pkg/api/handleCheckin.go
```diff
@@ -88,6 +101,10 @@ type CheckinT struct {
 	gwPool                         sync.Pool
 	bulker                         bulk.Bulk
 	outputSecretCandidateCollector policy.OutputSecretCandidateCollector
+
+	// invalidKeyStates tracks per-agent invalid-API-key state for the EmptyPolicyOnInvalidAPIKey
+	// feature. Entries expire and are removed after invalidKeyStateReset.
+	invalidKeyStates sync.Map
```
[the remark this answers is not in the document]
blakerouse: Yes this is expected. I am working on adding a memory cap.

internal/pkg/api/handleCheckin.go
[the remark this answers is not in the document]
cmacknz: Nice, I like this design it fixes every concern with the previous one.

internal/pkg/api/handleCheckin.go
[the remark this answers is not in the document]
blakerouse: LRU is added to cap the memory to ensure this doesn't cause a DDOS attack vector when this feature is enabled.

cmacknz (dismissed):
LGTM, thanks!

I think we should write an integration test in the elastic-agent repository that configures Fleet Server with this setting and confirms it correctly stops the OTel collector in an agent with endpoint security installed. We need to test this before releasing so better to just automate it rather than rely on manual testing.

swiatekm:
I have two questions:

1. If we tell the agent to unenroll on a 401, and the root cause in ES is fixable (by revalidating the key, or maybe the ES failure is itself infrastructural and temporary), then we're forcing the user to manually re-enroll with a new key later, right? At the very least, there should be a delay on this, I think.

2. Is there a reason we're not actually persisting the state to Fleet rather than storing it in the cache. Sounds like it'd be both simpler and more robust, even if it might require new fields in the index.

cmacknz:
> If we tell the agent to unenroll on a 401, and the root cause in ES is fixable (by revalidating the key, or maybe the ES failure is itself infrastructural and temporary), then we're forcing the user to manually re-enroll with a new key later, right? At the very least, there should be a delay on this, I think.

I am personally a bit hesitant to include UNENROLL in this for this reason, any kind of automatic unenroll we've ever introduced has lead to an incident of this sort. That said in this situation, the agents are already likely unmanaged and a users has opted into this, so it isn't automatic. If this weren't behind a feature flag it would be much more dangerous, though at the same time we could have people forget to turn it off and then get hit by this.  We will eventually make unenroll continue to contact Fleet Server as part of the uninstall action anyway so maybe the unenroll has less value considering this, though until we do this work there is no other way to get agents to stop contacting fleet server.

> Is there a reason we're not actually persisting the state to Fleet rather than storing it in the cache. Sounds like it'd be both simpler and more robust, even if it might require new fields in the index.

It leaves a permanent record of this feature in Elasticsearch, there is a chance we can remove this feature, or at least need for it should reduce greatly once we rework how unenroll works as part of the uninstall work. I'm also in favor of trying to keep this simple because there is some urgency to produce it for the affected user.

blakerouse:
> > If we tell the agent to unenroll on a 401, and the root cause in ES is fixable (by revalidating the key, or maybe the ES failure is itself infrastructural and temporary), then we're forcing the user to manually re-enroll with a new key later, right? At the very least, there should be a delay on this, I think.
> 
> I am personally a bit hesitant to include UNENROLL in this for this reason, any kind of automatic unenroll we've ever introduced has lead to an incident of this sort. That said in this situation, the agents are already likely unmanaged and a users has opted into this, so it isn't automatic. If this weren't behind a feature flag it would be much more dangerous, though at the same time we could have people forget to turn it off and then get hit by this. We will eventually make unenroll continue to contact Fleet Server as part of the uninstall action anyway so maybe the unenroll has less value considering this, though until we do this work there is no other way to get agents to stop contacting fleet server.

If we want to cause this to be a little safer we could send 401 for the first hour, then only after a full hour of 401 then start the process of sending the empty policy, unenroll. That would allow the case for elasticsearch having interment issues for less than an hour to not be an issue, but if it went over an hour it would be the same result (which begs wether its worth doing at all then).

Do me this is a feature you turn on, leave on for 2 hours and then turn off. It should not be something that remains on, this is only an escape out of the mess and not the keep on all the time flow.

cmacknz:
> Do me this is a feature you turn on, leave on for 2 hours and then turn off. It should not be something that remains on, this is only an escape out of the mess and not the keep on all the time flow.

This is also how I am thinking about it.

cmacknz (approved):
Latest changes from my last review are just to the test, LGTM.

I still think we need an automated test of this with endpoint installed, probably on the agent side. That will likely be the most common use for this. I am particularly interested in making sure the unenroll action behaves reasonably since we know endpoint will reject it.

PR #7638, opened 2026-08-13
fix: add memory tiers for 256 MB and 512 MB pods

ycombinator:
Related: elastic/fleet-controller#645

## What

Adds two new memory tiers below the existing 1024 MB floor:

| File | RAM | Agents | Cache |
|---|---|---|---|
| `lte0500_limits.yml` | 256 MB | 0–500 | 50 MiB |
| `lte1000_limits.yml` | 512 MB | 501–1000 | 64 MiB |

All other limits (checkin burst/max, enroll, ack, connections, etc.) are derived from `lte2500` with minimum floors applied where strict proportional scaling would produce impractically low values. Also bumps `lte2500`'s `min` from 0 to 1001 to make all agent count ranges non-overlapping.

## Why

Previously, the lowest tier required `recommended_min_ram: 1024`. Pods with less memory — common in containerised deployments such as Elastic Cloud Serverless — fell through to `defaultEnvLimits`, receiving a 50 MB cache regardless of their actual agent load. After elastic/fleet-server#7573 correctly began reading the pod's cgroup memory limit instead of host RAM, this became a problem in practice: Staging fleet-server pods with a 256 MB cgroup limit started getting bare-default configs, causing cache misses and ES overload at scale.

The proper fix is for fleet-controller to set cache limits explicitly (elastic/fleet-controller#645), but adding these tiers means fleet-server degrades gracefully for any small-memory deployment not managed by fleet-controller.

### DIFF
--- a/internal/pkg/config/defaults/lte1000_limits.yml
+++ b/internal/pkg/config/defaults/lte1000_limits.yml
@@ -1,0 +1,59 @@
+num_agents:
+  min: 501
+  max: 1000
+recommended_min_ram: 512
+cache_limits:
+  num_counters: 10000
+  max_cost: 67108864 # 64MiB
+server_limits:
+  max_connections: 3000
+  action_limit:
+    interval: 5ms
+    burst: 1
+  policy_limit:
+    interval: 5ms
+    burst: 1
+  checkin_limit:
+    interval: 5ms
+    burst: 200
+    max: 1000
+  artifact_limit:
+    interval: 5ms
+    burst: 200
+    max: 400
+  enroll_limit:
+    interval: 10ms
+    burst: 20
+    max: 40
+  ack_limit:
+    interval: 4ms
+    burst: 200
+    max: 400
+  status_limit:
+    interval: 5ms
+    burst: 20
+    max: 40
+  upload_start_limit:
+    interval: 2s
+    burst: 5
+    max: 10
+  upload_end_limit:
+    interval: 2s
+    burst: 5
+    max: 10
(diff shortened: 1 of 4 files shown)


internal/pkg/config/env_defaults_test.go
```diff
@@ -31,8 +31,9 @@ func TestLoadLimits(t *testing.T) {
 		ExpectedAgentLimit   int
 	}{
 		{"default", -1, int(getMaxInt())},
-		{"few agents", 5, 2500},
-		{"512", 512, 2500},
+		{"few agents", 5, 500},
+		{"512", 512, 1000},
+		{"lte2500 lower bound", 1001, 2500},
 		{"lesser bound", 5001, 10000},
```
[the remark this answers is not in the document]
ycombinator: Addressed in 1f6268a.

internal/pkg/config/defaults/lte1000_limits.yml
```diff
@@ -0,0 +1,59 @@
+num_agents:
+  min: 501
+  max: 1000
+recommended_min_ram: 512
+cache_limits:
+  num_counters: 10000
+  max_cost: 67108864 # 64MiB
```
lorienhu: This is higher than the other ones for lte2500+. We should either raise those or lower this one.

ycombinator: Thanks. Done in 7a72299.

lorienhu (approved):

PR #7668, opened 2026-08-20
Add workflow to sync PR branches

nuugen:

PR branches need a manual way to record and push a sync update.

Adds a dispatchable workflow that appends a UTC timestamp and pushes it to the selected branch. It also carries over the API-key authentication text changes from #7667.

`git diff --check` passes. The workflow was not dispatched.

## Design Checklist

- [x] Not applicable; this is a GitHub workflow change.

## Checklist

- [x] No documentation, configuration, test, or changelog changes are required.

## Related issues

- Relates #7667

Made with [Cursor](https://cursor.com)


swiatekm:
Why do we need to do this? What is the purpose of the sync file?

nuugen:
> Why do we need to do this? What is the purpose of the sync file?

Sorry for the noise - this is a test related to an infosec indent (260819 on Slack). I will close the PR as the testing is done.

PR #7682, opened 2026-08-22
Add support for multiple upgrade artifact source URIs

### LINKED ISSUE #2586: [Fleet Server] Allow multiple artifact URLs to be configured for a single policy
**Describe the enhancement:**
Today you can configure one or more artifact repositories for air gapped networks to serve downloads but you can only assign one download URI per policy. The enhancement would be like that of the output configurations in Kibana -> Fleet -> Settings where you can then set many URIs for the Elastic agent to download updates from. 

The use case is that we prefer agents to use our hosted downloads but if they are unavailable to from 1 source they can try another source such as another server in the network or from the default Elastic artifacts repo. I understand a load balancer could handle this but something built in would reduce complexity.

lorienhu:

Currently, only one source for upgrade artifacts can be specified, meaning upgrades fail if that source is down.

Adds support for sending multiple upgrade artifact source URIs in an upgrade action with a new sources that deprecates source_uri. Backwards compatibility is maintained.

## Design Checklist

- [ ] ~~I have ensured my design is stateless and will work when multiple fleet-server instances are behind a load balancer.~~
- [ ] ~~I have or intend to scale test my changes, ensuring it will work reliably with 100K+ agents connected.~~
- [ ] ~~I have included fail safe mechanisms to limit the load on fleet-server: rate limiting, circuit breakers, caching, load shedding, etc.~~

## Checklist

- [ ] ~~I have commented my code, particularly in hard-to-understand areas~~
- [ ] ~~I have made corresponding changes to the documentation~~
- [ ] ~~I have made corresponding change to the default configuration files~~

## Related issues

- Closes https://github.com/elastic/fleet-server/issues/2586
- Relates https://github.com/elastic/elastic-agent/issues/4752
- Relates https://github.com/elastic/elastic-agent/pull/16097

### DIFF
--- a/internal/pkg/api/handleCheckin.go
+++ b/internal/pkg/api/handleCheckin.go
@@ -1023,9 +1023,13 @@ func convertActionData(aType ActionType, raw json.RawMessage) (ad Action_Data, e
 		if err != nil {
 			return
 		}
-		if d.Sources != nil && len(*d.Sources) > 0 {
-			sourceURI := (*d.Sources)[0]
-			d.SourceUri = &sourceURI
+		if d.Sources != nil {
+			if len(*d.Sources) > 0 {
+				sourceURI := (*d.Sources)[0]
+				d.SourceUri = &sourceURI
+			} else {
+				d.SourceUri = nil
+			}
 		} else if d.SourceUri != nil {
 			sources := []string{*d.SourceUri}
 			d.Sources = &sources
--- a/internal/pkg/api/handleCheckin.go
+++ b/internal/pkg/api/handleCheckin.go
@@ -1031,7 +1031,10 @@ func convertActionData(aType ActionType, raw json.RawMessage) (ad Action_Data, e
 				d.SourceUri = nil
 			}
 		} else if d.SourceUri != nil {
-			sources := []string{*d.SourceUri}
+			sources := []string{}
+			if *d.SourceUri != "" {
+				sources = append(sources, *d.SourceUri)
+			}
 			d.Sources = &sources
 		}
 		err = ad.FromActionUpgrade(d)
--- a/internal/pkg/api/handleCheckin.go
+++ b/internal/pkg/api/handleCheckin.go
@@ -1023,6 +1023,13 @@ func convertActionData(aType ActionType, raw json.RawMessage) (ad Action_Data, e
 		if err != nil {
 			return
 		}
+		if d.Sources != nil && len(*d.Sources) > 0 {
+			sourceURI := (*d.Sources)[0]
+			d.SourceUri = &sourceURI
+		} else if d.SourceUri != nil {
+			sources := []string{*d.SourceUri}
+			d.Sources = &sources
+		}
 		err = ad.FromActionUpgrade(d)
 		return
 	case REQUESTDIAGNOSTICS:
--- a/internal/pkg/api/handleCheckin.go
+++ b/internal/pkg/api/handleCheckin.go
@@ -52,6 +52,7 @@ var (
 	ErrNoPolicyOutput         = errors.New("output section not found")
 	ErrFailInjectAPIKey       = errors.New("failure to inject api key")
 	ErrInvalidUpgradeMetadata = errors.New("invalid upgrade metadata")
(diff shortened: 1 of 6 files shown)


internal/pkg/api/handleCheckin.go
```diff
@@ -971,6 +971,13 @@ func convertActionData(aType ActionType, raw json.RawMessage) (ad Action_Data, e
 		if err != nil {
 			return
 		}
+		if d.Sources != nil && len(*d.Sources) > 0 {
+			sourceURI := (*d.Sources)[0]
+			d.SourceUri = &sourceURI
+		} else if d.SourceUri != nil {
+			sources := []string{*d.SourceUri}
+			d.Sources = &sources
```
ycombinator, line 993: Will this work correctly if `sources` is `[]`?

internal/pkg/api/handleCheckin_test.go
```diff
@@ -125,8 +125,20 @@ func TestConvertActionData(t *testing.T) {
...
+		raw:    json.RawMessage(`{"sources":["https://first.example.com","https://second.example.com"],"version":"1.2.3"}`),
+		expect: Action_Data{json.RawMessage(`{"source_uri":"https://first.example.com","sources":["https://first.example.com","https://second.example.com"],"version":"1.2.3"}`)},
+		hasErr: false,
+	}, {
+		name:   "upgrade action populates sources from source uri",
+		aType:  UPGRADE,
+		raw:    json.RawMessage(`{"source_uri":"https://legacy.example.com","version":"1.2.3"}`),
+		expect: Action_Data{json.RawMessage(`{"source_uri":"https://legacy.example.com","sources":["https://legacy.example.com"],"version":"1.2.3"}`)},
 		hasErr: false,
 	}, {
```
ycombinator, line 173: Could we add a test case where `raw` contains both `source_uri` and `sources`?

lorienhu: Added in [163a660](https://github.com/elastic/fleet-server/pull/7682/commits/163a6607d927b90100c0ef3befe49bf05e014c6b) and [4963720](https://github.com/elastic/fleet-server/pull/7682/commits/4963720a478a4d050ed56d725d40ef52fa9a2e20).

internal/pkg/api/handleCheckin.go
lorienhu: I think this was correct, but looking at this I noticed some inconsistencies with the null/empty case. Updated to always choose sources > sourceURI if sources is non-nil.

model/openapi.yml
```diff
@@ -743,6 +743,13 @@ components:
         source_uri:
           description: The source of the upgrade artifact.
           type: string
+          deprecated: true
+          x-deprecated-reason: Replaced by sources.
+        sources:
+          description: An ordered list of sources for the upgrade artifact.
+          type: array
+          items:
+            type: string
```
ycombinator, line 752: Currently, this array is unbounded.  A client could send thousands of `sources` and eat up memory in Fleet Server unnecessarily. Consider setting a practical limit like, IDK, 20?

ycombinator (commented):
Could we add a test case for where neither `sources` nor `source_uri` is set?

ycombinator (commented):
Also, this is more of a naming nit, but would `source_uris` be more consistent than `sources`?

model/openapi.yml
lorienhu: Added a check in [2eeb96d](https://github.com/elastic/fleet-server/pull/7682/commits/2eeb96d5eed3c1a0623e129b3ab1b794d8d2dc2b).

lorienhu:
> Also, this is more of a naming nit, but would `source_uris` be more consistent than `sources`?

I picked sources because the naming for sourceURI vs actual sourceURI was colliding in downloadArtifact before with the old name, since it's technically the base path/base URI.

lorienhu:
> Could we add a test case for where neither `sources` nor `source_uri` is set?

Added anyway for clarity, but this just means neither is in the action, following the previously existing behaviour.

ycombinator (approved):

PR #7689, opened 2026-08-26
Compare base policy IDs in replace-enrollment check

seanrathier:

An agent that has been reassigned to a version-specific policy variant (e.g. `my-policy#9.6`) by Kibana's version-specific policy assignment task can never re-enroll via the enroll-with-replace flow. Enrollment API keys are only ever bound to the **base** policy, so the enrolling agent presents the base policy ID while the stored agent document carries the versioned one. The replace check compares the two raw IDs, sees a mismatch, and returns `AgentNotReplaceable` (403) on every attempt.

Containerized (agentless) agents hit this whenever their pod is recreated after such a reassignment — e.g. after any policy edit rolls the pod. The agent stays offline permanently and the pod crash-loops on enrollment. Observed in production (elastic/sdh-beats#7493) and reproduced in elastic/ingest-dev#9238: all 8 offline agents in the repro environment had `#9.6`-suffixed `policy_id`s with still-active agent docs, while 0 of 15 enrollment keys were bound to a versioned policy.

The replace check now compares **base** policy IDs using the existing `policyBaseID()` helper (already used in the same file when writing `policy_base_id` on agent documents):

```go
if policyBaseID(agent.PolicyID) != policyBaseID(policyID) {
```

A version-suffixed assignment no longer blocks replacement, while enrolling into a genuinely different policy is still rejected. The replace path already leaves the agent document's `policy_id` untouched (it only resets `policy_revision_idx` and updates `policy_base_id`), so the agent keeps its version-specific assignment and receives the latest policy on next checkin.

Unit test `TestEnrollWithAgentIDExistingActive_VersionedPolicy` covers the exact production scenario: an existing active agent with `policy_id: "my-policy#9.6"` and a valid replace token, re-enrolling with an enrollment key bound to `"my-policy"`. It fails with `AgentNotReplaceable` before this change and passes after. `TestEnrollWithAgentIDExistingActive_WrongPolicy` verifies genuinely different policies are still rejected.

Manual reproduction (before the fix): create an agentless/containerized agent on a policy whose package declares `conditions.agent.version` (e.g. aws ≥ 7.0.0), wait for the version-specific policy assignment task to reassign the agent to `<policy>#<minor>`, then recreate the container so it re-enrolls with its replace token — enrollment loops on `AgentNotReplaceable`.

## Design Checklist

- [x] I have included fail safe mechanisms to limit the load on fleet-server: rate limiting, circuit breakers, caching, load shedding, etc. (no new load paths; one string comparison changed)

## Checklist

## Related issues

- Closes elastic/ingest-dev#9238
- Relates elastic/sdh-beats#7493
- Relates elastic/kibana#281074

### DIFF
--- a/changelog/fragments/1787691869-replace-enroll-versioned-policy.yaml
+++ b/changelog/fragments/1787691869-replace-enroll-versioned-policy.yaml
@@ -1,0 +1,17 @@
+kind: bug-fix
+
+summary: Allow replace-enrollment of agents assigned to version-specific policy variants
+
+description: |
+  The enroll-with-replace path rejected any agent whose document had been
+  reassigned to a version-specific policy variant (e.g. "policy#9.6") by
+  Kibana's version-specific policy assignment task, because it compared the
+  raw stored policy ID against the enrollment key's policy ID, and enrollment
+  API keys are only ever bound to the base policy. Containerized (agentless)
+  agents hit this whenever their pod was recreated after such a reassignment:
+  every re-enrollment attempt failed with AgentNotReplaceable and the agent
+  stayed offline permanently. The check now compares base policy IDs, so a
+  version-suffixed assignment no longer blocks replacement while enrolling
+  into a genuinely different policy is still rejected.
+
+component: fleet-server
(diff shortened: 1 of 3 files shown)


juliaElastic (approved):

MichelLosier:
Superceded by: https://github.com/elastic/fleet-server/pull/7719

PR #7711, opened 2026-08-28
fix: prevent ghost agents with synchronous enrollment writes

ycombinator:

When an enrollment write is committed to fleet-server's in-memory bulk queue but the HTTP response is lost before it reaches the agent (network drop, EOF), the agent retries enrollment. If the retry lands on a different pod, that pod's `FindAgent` search finds nothing — the write hasn't been flushed to Elasticsearch yet — and creates a second agent document. These duplicate documents ("ghost agents") accumulate and do not resolve on their own.

The batched pre-refresh approach introduced in #7662 attempted to fix this by refreshing the index and deduplicating concurrent requests by `enrollment_id` within a single flush batch. It cannot help when the retry reaches a different pod, because cross-pod retries are never in the same batch. Evidence from the 30k serverless checkin scale test: 35 ghost agents with `enrolled_at` timestamps clustered in a 21-second window (11:53:42–11:54:03 UTC), zero `ErrEnrollDuplicate` 429s — confirming the in-batch dedup never fired for any of the 35 retries.

This PR removes the `kQueueEnrollSearch` batching machinery from #7662 and #7688 and replaces it with a direct synchronous write using `op_type=create&refresh=wait_for`. ES blocks until the document is committed and visible to search on all shards before returning. The enrollment response is not sent until this write completes, so any retry on any pod finds the existing document immediately (returning a 409 treated as success).

The new write path is gated behind `inputs[].server.feature_flags.sync_enrollment_write` (default: `false`). When `false`, the existing async bulk queue path is used (current production behaviour). When `true`, the synchronous path is active. The flag lets us enable and validate the change in Staging before it reaches Serverless Production, and provides an emergency fallback if needed. Once the 30k checkin scale test validates the sync path, the flag and the async fallback will be removed.

To enable in serverless-gitops:

```yaml
inputs:
  - type: fleet-server
    server:
      feature_flags:
        sync_enrollment_write: true
```

Enable the flag and run the 30k Fleet Server checkin scale test against Staging. Confirm ghost agent count is 0 after enrollment completes.

## Design Checklist

## Checklist

## Related issues

- Supersedes #7662
- Supersedes #7688
- Relates elastic/ingest-dev#9167


ycombinator:
Superseded by a cleaner PR that keeps the existing async path intact and only adds the sync path + feature flag.

PR #7719, opened 2026-08-28
Fix agentless replacement enrollments when policies are versioned

MichelLosier:

Agentless agents use a replace-enrollment flow when their pod restarts. The replace check in fleet-server compared the full stored `policy_id` against the enrollment key's policy ID. Kibana's version-specific policy assignment task rewrites an agent doc's `policy_id` to a versioned variant (e.g. `my-policy#9.6`). Enrollment keys are always bound to the base policy ID. After reassignment, the stored ID and the enrollment key's ID never match. Fleet-server returns `AgentNotReplaceable` on every attempt. The pod crash-loops and the agent stays offline permanently.

A second problem exists after the crash-loop is fixed. Agentless agents upgrade by replacing the pod with a new image. They do not go through a Fleet upgrade action, so `upgraded_at` is never set. Kibana's version-specific policy assignment task uses `upgraded_at` to find agents that need to move to a new policy variant. Without it, the agent stays pinned to the variant for its previous minor version. If a package moves its `conditions.agent.version` constraint to a newer minor, the agent silently stops receiving the affected inputs.

**Fix 1** (cherry-picked from elastic/fleet-server#7689, authored by @seanrathier): The replace check now compares base policy IDs using the existing `policyBaseID` helper. A version-suffixed assignment no longer blocks replacement. Enrolling into a genuinely different policy is still rejected.

**Fix 2**: The replace-enrollment path stamps `upgraded_at` when the incoming agent version differs from the version in the existing agent doc. Kibana's version-specific policy assignment task picks up the agent within its 30-minute window and moves it to the correct variant for the new version. The comparison is direction-agnostic, matching the existing behavior of `handleUpgrade`, which already stamps `upgraded_at` on Kibana-initiated rollbacks.

Unit tests cover the fix directly. Run:

```
TEST_RUN='TestEnrollWithAgentIDExisting' mage test:unit
```

`TestEnrollWithAgentIDExistingActive_VersionedPolicy` fails before Fix 1 and passes after. `TestEnrollWithAgentIDExistingActive_UpgradedAtOnVersionChange` and `_UpgradedAtOnDowngrade` fail before Fix 2 and pass after. `TestEnrollWithAgentIDExistingActive_NoUpgradedAtOnSameVersion` verifies the field is absent on a same-version re-enrollment.

## Design Checklist

- [x] I have included fail safe mechanisms to limit the load on fleet-server: rate limiting, circuit breakers, caching, load shedding, etc. _(no new load paths; one string comparison and one conditional field added to an existing update)_

## Checklist

## Related issues

- Closes elastic/ingest-dev#9238
- Relates elastic/kibana#281074
- Relates elastic/ingest-dev#9341
- Supersedes elastic/fleet-server#7689


internal/pkg/api/handleEnroll.go
```diff
@@ -391,6 +399,10 @@ func (et *EnrollerT) _enroll(
 			dl.FieldUnenrolledReason:      nil,
 			dl.FieldUpdatedAt:             now.UTC().Format(time.RFC3339),
 		}
+		// stamp upgraded_at so Kibana knows it should evaluate which version-specific policy it should receive
```
ycombinator: Nit: this reads a bit awkwardly. 


```suggestion
		// stamp upgraded_at so Kibana's version-specific policy assignment task will re-evaluate this agent
```

internal/pkg/api/handleEnroll_test.go
```diff
@@ -413,18 +431,213 @@ func TestEnrollWithAgentIDExistingActive(t *testing.T) {
...
+	rb := &rollback.Rollback{}
+	zlog := zerolog.Logger{}
+	agentID := "1234"
+	replaceToken := "replace_token"
+	var pbkdf2Cfg config.PBKDF2
+	pbkdf2Cfg.InitDefaults()
+	replaceHash, err := hashReplaceToken(replaceToken, pbkdf2Cfg)
+	if err != nil {
+		t.Fatalf("error generating bcrypt hash: %v", err)
 	}
```
ycombinator: Any reason not to use `require.NoError` here like in the other tests added in this PR?

internal/pkg/api/handleEnroll_test.go
```diff
@@ -413,18 +431,213 @@ func TestEnrollWithAgentIDExistingActive(t *testing.T) {
...
+		ReplaceToken: &replaceToken,
 	}
-	assert.NoError(t, json.Unmarshal(updateBody, &updateDoc))
-	assert.Equal(t, "my-policy", updateDoc.Doc[dl.FieldPolicyBaseID])
+	verCon := mustBuildConstraints("8.9.0")
+	cfg := &config.Server{}
+	cfg.InitDefaults()
+	c, _ := cache.New(config.Cache{NumCounters: 100, MaxCost: 100000})
+	bulker := ftesting.NewMockBulk()
+	et, _ := NewEnrollerT(verCon, cfg, bulker, c)
```
ycombinator: This block of code, with some small variations, is repeated across the four tests added in this PR.  Consider either extracting it into a helper function or consolidating the four tests into one table-driven test.

internal/pkg/api/handleEnroll.go
```diff
@@ -391,6 +399,10 @@ func (et *EnrollerT) _enroll(
 			dl.FieldUnenrolledReason:      nil,
 			dl.FieldUpdatedAt:             now.UTC().Format(time.RFC3339),
 		}
+		// stamp upgraded_at so Kibana's version-specific policy assignment task will re-evaluate this agent
+		if prevVer != "" && prevVer != ver {
+			doc[dl.FieldUpgradedAt] = now.UTC().Format(time.RFC3339)
```
cmacknz, line 404: Will this have any consequence for non-containerized agents going through the actual Fleet upgrade process?

cmacknz: Agents with a version difference outside containers can be rolled back, just seeing the new version isn't the completion of the upgrade process necessarily.

MichelLosier: Yeah I thought the guards above it would be exclusive to replacements, but I think I'm wrong there, going to add an extra guard to make sure its for replacements.

> Agents with a version difference outside containers can be rolled back, just seeing the new version isn't the completion of the upgrade process necessarily.

Yeah I had considered this initially, but it looks like (keep me honest) that downgrades already are dispatched as an UPGRADE from kibana through `handleAck.handleUpgrade` path, and results in an upgraded_at on any version change. So current convention already overloads the meaning here.

cmacknz: There are intentional downgrades and rollbacks, and an automated rollback (e.g. agent upgrades and defend crashes so we roll back to auto fix).

The replace token guard is good, another effective guard would probably be to look at the update state but I'm good with this initially.

ycombinator (approved):

cmacknz (approved):

harshitgupta-qasource:
Hi Team,

As requested, we have validated the issue on the latest 9.5.3-SNAPSHOT Kibana Cloud environment, and the functionality is **working as expected**.

We have also validated the relevant scenarios, including the restart and re-enrollment flows.

**Build details:**
```
VERSION: 9.5.3-SNAPSHOT
BUILD: 107530
COMMIT: 2169f1de4c917fce03905cc3110f3f523e2184d2
 ```

 **Observations:**

- Added the AWS integration with Elastic-managed integration using the Identity Federation setup method.

https://github.com/user-attachments/assets/64f6ed48-d635-4923-907b-a901e2fa4eee

Image
- Validated the re-enrollment and restart scenarios.

https://github.com/user-attachments/assets/6f30ea34-a24a-4cb2-9be4-aaaece4452ad

Image

- Validated that after updating the Agentless integration, the Agentless agent does not go offline and continues to work as expected

 https://github.com/user-attachments/assets/5086e3a8-a845-4aa1-b42c-71c014b22012
 
image


Please let us know if we missed any specific step, scenario, or configuration that needs to be validated from the our side.
 
 Thank you

PR #7729, opened 2026-09-01
fix: retry enrollment write on transient refresh-wait failure to prevent ghost agents

ycombinator:
⚠️ _I would like someone from the Elasticsearch team to confirm the root cause mentioned below before putting this PR into review._

In stateless Elasticsearch, the primary (index) shard is not searchable — search queries are handled by separate `SEARCH_ONLY` shard nodes. In stateless, the primary has role `INDEX_ONLY` (`promotable=true, searchable=false`) and search replicas have role `SEARCH_ONLY` (`promotable=false, searchable=true`) ([`ShardRouting.java:937`](https://github.com/elastic/elasticsearch/blob/71e50fce48adb6239e2b22f2aa8897f8c6b426f6/server/src/main/java/org/elasticsearch/cluster/routing/ShardRouting.java#L937-L939), [`StatelessShardRoutingRoleStrategy.java:17`](https://github.com/elastic/elasticsearch/blob/71e50fce48adb6239e2b22f2aa8897f8c6b426f6/x-pack/plugin/stateless/src/main/java/org/elasticsearch/xpack/stateless/allocation/StatelessShardRoutingRoleStrategy.java#L17)). Fleet-server uses `refresh=wait_for` so enrollment blocks until those `SEARCH_ONLY` shards have loaded the new agent document, at which point credentials are returned and Horde can immediately verify them.

When `refresh=wait_for` is used on a non-searchable (`INDEX_ONLY`) primary, `PostWriteRefresh.refreshUnpromotables` ([`PostWriteRefresh.java:104`](https://github.com/elastic/elasticsearch/blob/71e50fce48adb6239e2b22f2aa8897f8c6b426f6/server/src/main/java/org/elasticsearch/action/support/replication/PostWriteRefresh.java#L104)) registers a flush listener that, once the Lucene commit completes, sends an `UnpromotableShardRefreshRequest` to each `SEARCH_ONLY` shard node. Each `SEARCH_ONLY` shard must call `waitForPrimaryTermAndGeneration` ([`TransportUnpromotableShardRefreshAction.java:181`](https://github.com/elastic/elasticsearch/blob/71e50fce48adb6239e2b22f2aa8897f8c6b426f6/server/src/main/java/org/elasticsearch/action/admin/indices/refresh/TransportUnpromotableShardRefreshAction.java#L181)) to confirm it has loaded the new generation from blob storage before the write response is returned to the caller.

A `SEARCH_ONLY` shard that is bootstrapping onto a new node cannot respond to that notification. This happens in (at least) two observed scenarios:

1. **ES cluster scale-out** — the autoscaler adds nodes and shard rebalancing migrates `.fleet-agents` to a new node. Observed in [build #6895](https://buildkite.com/elastic/observability-perf/builds/6895): EPA over-provisioned fleet-server to 113 pods, causing ES to autoscale 3 new nodes (00:17:53–00:18:11 UTC). Shard bootstrapping windows at 00:22:25, 00:22:39, 00:24:55, 00:26:44 UTC produced 232 ghost agents.

2. **ES index-node rolling update** — a Kubernetes rolling restart of the `es-es-index` deployment terminates old pods one by one, each handoff migrating the `.fleet-agents` `INDEX_ONLY` shard via peer recovery to a new pod, triggering a fresh `SEARCH_ONLY` bootstrapping window on each migration. Observed in [build #6897](https://buildkite.com/elastic/observability-perf/builds/6897): 10+ ES index pods received SIGTERM between 06:48–06:56 UTC; the `.fleet-agents-7[0]` shard migrated across 4 distinct node replicasets; 42 `"failed to notify search shards after create commit"` events ([`StatelessCommitService.java:2370`](https://github.com/elastic/elasticsearch/blob/71e50fce48adb6239e2b22f2aa8897f8c6b426f6/x-pack/plugin/stateless/src/main/java/org/elasticsearch/xpack/stateless/commits/StatelessCommitService.java#L2370)) in two windows (06:52:27 and 06:55:50 UTC) produced 13 ghost agents. This run used only ~52 fleet-server pods — the same count as the passing run — confirming the failure is not bounded by EPA tuning.

In both cases the notification timeout causes fleet-server to return HTTP 500 — but the underlying Lucene commit is not rolled back. As `InternalEngine.addFlushListener` documents ([`InternalEngine.java:4163`](https://github.com/elastic/elasticsearch/blob/71e50fce48adb6239e2b22f2aa8897f8c6b426f6/server/src/main/java/org/elasticsearch/index/engine/InternalEngine.java#L4163)): *"flush does not imply commit durability"* — durability (blob storage upload) is handled separately by `IndexEngine.waitForCommitDurability` ([`IndexEngine.java:855`](https://github.com/elastic/elasticsearch/blob/71e50fce48adb6239e2b22f2aa8897f8c6b426f6/x-pack/plugin/stateless/src/main/java/org/elasticsearch/xpack/stateless/engine/IndexEngine.java#L855)) and proceeds independently of the refresh notification. The document will eventually become visible once the `SEARCH_ONLY` shard finishes bootstrapping.

Fleet-server returned 500 to Horde. Horde retried enrollment with the same enrollment ID. At retry time, `FindAgent` searched the still-bootstrapping shard, got "not found", and fleet-server created a second agent document under a new UUID. When the original document later became visible, it appeared as a "ghost" agent in Fleet — counted in the total agent count but never having checked in.

Fleet-server retries the `op_type=create` enrollment write once before surfacing the error. Because `op_type=create` is idempotent on the document ID, a retry that finds the original write already committed receives `409 Conflict`, which is treated as success — the caller gets a valid enrollment response and no ghost is created. The retry is skipped when the outer HTTP request context has expired.

Run the `api` package tests:
```
go test ./internal/pkg/api/...
```

`TestCreateFleetAgentSyncWriteRetry` covers four cases: transient error → 409 success, transient error → 201 success, persistent error on both attempts, and context cancellation before retry.

## Design Checklist

## Checklist

### DIFF
--- a/changelog/fragments/1788232734-enrollment-create-retry.yaml
+++ b/changelog/fragments/1788232734-enrollment-create-retry.yaml
@@ -1,0 +1,19 @@
+kind: bug-fix
+
+summary: Retry enrollment write on transient refresh-wait failure to prevent ghost agents
+
+description: |
+  In stateless Elasticsearch, a synchronous enrollment write (op_type=create +
+  refresh=wait_for) can commit the agent document to blob storage and then time out
+  while waiting for a search shard to acknowledge the refresh — for example when the
+  search shard is bootstrapping onto a new node after cluster scale-out. Fleet-server
+  received an error and returned 500, causing the agent to retry enrollment and create
+  a second document for the same agent. The first, now-orphaned document became a
+  "ghost" agent visible in Fleet.
+
+  fleet-server now retries the same op_type=create write once before surfacing the
+  error. If the original write committed, the retry gets 409 Conflict, which is treated
+  as success — the caller receives a valid enrollment response and no ghost is created.
+  The retry is skipped when the outer HTTP request context has expired.
+
+component: fleet-server
--- a/internal/pkg/api/handleEnroll.go
+++ b/internal/pkg/api/handleEnroll.go
@@ -663,28 +663,7 @@ func createFleetAgent(ctx context.Context, bulker bulk.Bulk, id string, agent mo
 	}
 
 	if syncWrite {
-		// Sync path: write directly with refresh=wait_for so the document is immediately
-		// visible to any retry pod's FindAgent search, preventing ghost agents.
-		req := esapi.IndexRequest{
-			Index:      dl.FleetAgents,
-			DocumentID: id,
-			Body:       bytes.NewReader(data),
-			OpType:     "create",
-			Refresh:    "wait_for",
-		}
-		res, err := req.Do(ctx, bulker.Client())
-		if err != nil {
-			return err
-		}
-		defer res.Body.Close()
-		if res.StatusCode == http.StatusConflict {
-			zerolog.Ctx(ctx).Debug().Str("agent_id", id).Msg("agent document already exists on enrollment create, treating as success")
-			return nil
-		}
-		if res.IsError() {
-			return fmt.Errorf("createFleetAgent: %s", res.String())
-		}
-		return nil
+		return createFleetAgentSync(ctx, bulker, id, data)
 	}
 
 	_, err = bulker.Create(ctx, dl.FleetAgents, id, data, bulk.WithRefresh())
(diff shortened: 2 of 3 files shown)


ycombinator:
Closing this PR after further investigation revealed that the underlying hypothesis was incorrect.

**Original hypothesis:** When `refresh=wait_for` is used and the SEARCH_ONLY shard notification fails during bootstrapping, ES returns a 5xx to fleet-server, which then returns an error to the agent, causing the agent to retry enrollment with a new ID and creating a ghost doc alongside the original.

**What the logs actually show:**
- The SEARCH_ONLY bootstrapping notification in stateless ES is a background async operation (via `StatelessCommitService`). When it fails, ES logs a warning and swallows the error — the enrollment write HTTP response is unaffected and returns 200.
- Ghost agents in build #6895 (232 ghosts) coincided with an ES autoscale event that caused `context deadline exceeded` / `net/http: timeout awaiting response headers` errors on fleet-server's bulk write operations, triggering 9,900+ enrollment rollbacks — not related to `refresh=wait_for` SEARCH_ONLY notify failures.
- The `op_type=create` + 409-as-success approach only prevents duplicates when the agent retries with the **same** document ID. Log analysis confirms the retrying agents used new IDs, so 409 would never fire.

The ghost agent problem requires a different approach.

PR #7740, opened 2026-09-01
Fix processPolicy using shared secret keys between agents

### LINKED ISSUE #7739: Checkin processPolicy corrupts shared ParsedPolicy.SecretKeys
Dispatch of a new policy revision sends a pointer to a single ParsedPolicy to every agent's checkin goroutine. Each processPolicy then mutates this ParsedPolicy's SecretKeys with no synchronization.

This previously didn't result in any noticeable problems until DeleteFunc was added in https://github.com/elastic/fleet-server/pull/7468, which could result in SecretKeys being corrupted to [""]. This will cause any agent receiving a policy update after this corruption has occurred to crashloop in `addSecretMarkers`. Editing the policy and creating a new revision will clear this but potentially corrupt again. If the corruption occurs early in rollout, the impact could be severe.

This is the root cause of https://github.com/elastic/sdh-beats/issues/7546.

lorienhu:

Dispatch of a new policy revision sends a pointer to a single ParsedPolicy to every agent's checkin goroutine. Each processPolicy then mutates this ParsedPolicy's SecretKeys with no synchronization. This results in agents potentially receiving a [""] in secret paths, which causes them to crashloop.

Fix processPolicy so that processPolicy for each agent doesn't mutate shared ParsedPolicy.SecretKeys.

`go test -race -count=1 -run TestProcessPolicySecretPathsConcurrentDispatch ./internal/pkg/api/`

## Design Checklist

## Checklist

## Related issues

- Closes #7739
- Relates https://github.com/elastic/sdh-beats/issues/7546

### DIFF
--- a/internal/pkg/api/handleCheckin_test.go
+++ b/internal/pkg/api/handleCheckin_test.go
@@ -18,7 +18,9 @@ import (
 	"io"
 	"net/http"
 	"net/http/httptest"
+	"slices"
 	"strings"
+	"sync"
 	"testing"
 	"time"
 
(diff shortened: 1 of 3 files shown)


internal/pkg/api/handleCheckin_test.go
```diff
@@ -1859,3 +1861,87 @@ func TestProcessPolicyRemoteESServiceTokenSecretPaths(t *testing.T) {
...
+				return
+			}
+			secretPaths[a] = pc.Policy.SecretPaths
+		})
+	}
+	wg.Wait()
+
+	for a := range agents {
+		require.NoError(t, errs[a])
+		assert.NotContains(t, secretPaths[a], "", "agent %d received a corrupt empty secret path", a)
```
ycombinator: Would be nice to also assert that the expected secret path (not just a non-empty one) is present.  That way we know that the correct secret paths are being used.

ebeahan (approved):

ycombinator (approved):

lorienhu:
@ebeahan Can we get this forced merged? Failures aren't related to the change.

Issue #7741, opened 2026-09-01
Ghost agent documents created when ES is disrupted during enrollment


ycombinator:
## Summary

Under ES load, fleet-server can leave orphaned `.fleet-agents` documents — agents that enrolled but never checked in. These "ghost" agents inflate the `active` count in the Fleet agent status API above the true `online` count.

## What is a ghost agent

A ghost agent is a `.fleet-agents` document that:
- Has an `enrolled_at` timestamp (fleet-server wrote it successfully)
- Has **no `last_checkin` field** (the agent never contacted Fleet after enrollment)
- Is counted as `active` but not `online` in the Fleet agent status API

## Evidence

Observed in consecutive daily 30k checkin scale test runs:

| Build | Ghosts (`all - online`) | Notes |
|-------|------------------------|-------|
| [#6895](https://buildkite.com/elastic/observability-perf/builds/6895) | 232 | ES autoscale event; bulk write timeouts during enrollment window 00:26–00:37 UTC |
| [#6899](https://buildkite.com/elastic/observability-perf/builds/6899) | 79 | Shorter disruption |
| [#6901](https://buildkite.com/elastic/observability-perf/builds/6901) | 0 | No ES disruption; clean run |

Ghost `enrolled_at` timestamps span the ES disruption window exactly. Total agent documents created exceeded the requested count by the ghost count (e.g. 30,232 documents for 30,000 requested agents), indicating re-enrollments occurred.

## Related

- observability-perf test workaround: elastic/observability-perf#937

ycombinator:
Another instance observed in build [elastic/observability-perf#6929](https://buildkite.com/elastic/observability-perf/builds/6929) (2026-09-02, still running at time of writing):

- **295 ghost documents** at 23:06:19 UTC (`all=30,295` for 30,000 requested agents); grew to **464** by 23:13:21 UTC (`all=30,464`)
- EPA scaled fleet-server to **171 pods** (out of `maxReplicas=180`) — highest pod count seen across all runs

**fleet-server pod logs during enrollment (23:11:13–23:13:05 UTC):**

- 178 `"perform rollback on enrollment failure"` events — ES write timed out (`context deadline exceeded`: 103, `net/http: timeout awaiting response headers`: 75)
- 165 `"agent record failed to delete"` / `"Error when trying to delete old agent with enrollment id"` — rollback deletions also timed out, leaving ghost docs in place
[...]

PR #7770, opened 2026-09-04
feat: log enrollment_id on successful enrollment

ycombinator:

When ghost agents appear in `.fleet-agents` (documents with `enrolled_at` but no `last_checkin`), we cannot determine from fleet-server logs whether the ghost write was one that fleet-server considered successful. The "Elastic Agent successfully enrolled" log entry records `agent_id` but not `enrollment_id`, so there is no way to correlate a ghost document (identified by `enrollment_id` via aggregation) back to a fleet-server log entry.

This gap was discovered during ghost agent investigation ([elastic/fleet-server#7741](https://github.com/elastic/fleet-server/issues/7741)): in [observability-perf build #6943](https://buildkite.com/elastic/observability-perf/builds/6943), 3 ghost documents had enrollment_ids that appeared in neither fleet-server write-error logs nor ES error logs, making the root cause unclear.

Adds `enrollment_id` to the "Elastic Agent successfully enrolled" log entry. With this field present, the next run with ghost agents can confirm whether ghost enrollment_ids appear in the success log, distinguishing between:
- Ghost write succeeded from FS perspective (enrollment_id IS in success log)
- Ghost write failed silently without reaching the success log path

The change threads `enrollmentID` from `processRequest` (where the request body is parsed) through `handleEnroll` to `writeResponse`. `EnrollResponse` is auto-generated from OpenAPI spec and is not modified.

```bash
# Start fleet-server and enroll an agent. Verify the log entry:
# {"message":"Elastic Agent successfully enrolled","fleet.agent.id":"...","enrollment_id":"..."}
```

## Design Checklist

- [x] I have included fail safe mechanisms to limit the load on fleet-server: rate limiting, circuit breakers, caching, load shedding, etc. *(logging-only change; no load impact)*

## Checklist

## Related issues

- Relates #7741

### DIFF
--- a/internal/pkg/api/handleEnroll.go
+++ b/internal/pkg/api/handleEnroll.go
@@ -597,7 +597,7 @@ func writeResponse(ctx context.Context, zlog zerolog.Logger, w http.ResponseWrit
 		Str(LogPolicyID, resp.Item.PolicyId).
 		Str(LogAccessAPIKeyID, resp.Item.AccessApiKeyId)
 	if enrollmentID != "" {
-		logEvent = logEvent.Str("enrollment_id", enrollmentID)
+		logEvent = logEvent.Str(dl.FieldEnrollmentID, enrollmentID)
 	}
 	logEvent.
 		Int(ECSHTTPResponseBodyBytes, numWritten).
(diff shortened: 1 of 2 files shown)


samuelvl (dismissed):
I think this suggestion makes sense https://github.com/elastic/fleet-server/pull/7770/changes#r3931453220 but probably it's not a blocker

ycombinator (commented):
Addressed in b6ff1a6e: `enrollment_id` is now only logged when non-empty, so clients that omit the field won't produce an empty-string log entry.

samuelvl (dismissed):

samuelvl (approved):

PR #7781, opened 2026-09-08
[8.19] fix: prevent shared ParsedPolicy.SecretKeys mutation in processPolicy

### LINKED ISSUE #7739: Checkin processPolicy corrupts shared ParsedPolicy.SecretKeys

ycombinator:

`processPolicy` is called concurrently — one goroutine per agent checkin — and all goroutines share the same `*ParsedPolicy` pointer for a given policy revision. The function mutated `pp.SecretKeys` directly via `append`, `slices.Sort`, and `slices.Compact`, all of which race on the shared backing array. Under concurrent execution this corrupts the `SecretPaths` slice delivered to agents (e.g. empty strings appear due to torn string reads during concurrent sort), causing agents to crash-loop in `addSecretMarkers` with:

```
panic: runtime error: index out of range [-1]
goroutine N [running]:
github.com/elastic/go-ucfg.idxField.GetValue(...)
...
github.com/elastic/elastic-agent/internal/pkg/diagnostics.addSecretMarkers(...)
```

The corrupted `SecretPaths` are persisted to the agent's local state, so the crash-loop continues across restarts. Only a full uninstall/reinstall (clearing local state) recovers the agent.

This is a manual backport of #7740 to the 8.19 branch. The Mergify auto-backport #7744 was closed without merging because the cherry-pick produced unresolvable conflicts: main had diverged from 8.19 via a refactor of `ProcessOutputSecret` (different package and signature) and the addition of `slices.DeleteFunc` (from #7468), neither of which landed in 8.19. The conflict markers were committed as-is, breaking the build, and the PR was abandoned.

This PR re-applies the same logical fix — clone `pp.SecretKeys` into a local variable at the top of `processPolicy` so all subsequent appends, sort, and compact operate on a per-goroutine copy, leaving the shared `ParsedPolicy` field untouched — adapted cleanly to the 8.19 code.

```
go test -race -count=1 -run TestProcessPolicySecretPathsConcurrentDispatch ./internal/pkg/api/
```

## Design Checklist

## Checklist

## Related issues

- Closes #7739
- Supersedes #7744

### DIFF
--- a/changelog/fragments/1788278234-fix-secret-keys-corruption.yaml
+++ b/changelog/fragments/1788278234-fix-secret-keys-corruption.yaml
@@ -1,0 +1,7 @@
+kind: bug-fix
+
+summary: Fix secret keys corruption in policy processing
+
+component: fleet-server
+
+# pr: filled automatically by tooling
--- a/internal/pkg/api/handleCheckin.go
+++ b/internal/pkg/api/handleCheckin.go
@@ -854,6 +854,9 @@ func processPolicy(ctx context.Context, zlog zerolog.Logger, bulker bulk.Bulk, a
 	}
 
 	data := model.ClonePolicyData(pp.Policy.Data)
+	// Clone pp.SecretKeys so concurrent processPolicy calls for the same policy
+	// revision do not race on the shared ParsedPolicy field.
+	secretKeys := slices.Clone(pp.SecretKeys)
 	for policyName, policyOutput := range data.Outputs {
 		// NOTE: Not sure if output secret keys collected here include new entries, but they are collected for completeness
 		ks, err := policy.ProcessOutputSecret(ctx, policyOutput, bulker) // makes a bulk request to get secret values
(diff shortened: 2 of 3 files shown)


ycombinator:
buildkite build this

ycombinator:
> Both failing Buildkite package jobs are blocked by an **infrastructure issue** in the base image apt metadata, not by this PR’s code changes. `apt-get update` in `Dockerfile.build` fails because Debian bullseye-security `InRelease` is expired; rerunning once the mirror metadata refreshes (or using a non-expired base) should unblock.

Temporary workaround PR to `main` that'll need to be backported to other branches, including `8.19`: https://github.com/elastic/fleet-server/pull/7782

lorienhu (approved):
LGTM, but I'm surprised this is causing noticeable issues without the DeleteFunc change that triggered the problem in the other branches.

PR #7784, opened 2026-09-09
fix: clear upgrade_started_at when agent upgrades faster than one checkin interval

ycombinator:
## What is the problem?

When an Elastic Agent upgrades faster than one checkin interval (~60 seconds), Fleet Server never observes any intermediate `upgrade_details` states (downloading, extracting, watching). The agent completes its upgrade and restarts at the new version within a single checkin gap.  In practice, this is unlikely to happen with an actual Elastic Agent.  But it is possible with Horde drones and, in theory, it could happen with Elastic Agents too.

In this case:
- `upgrade_started_at` is set on the agent document (written by Kibana's `bulk_upgrade` API) when the upgrade action is dispatched;
- `upgrade_details` is `null` on the agent document;
- The agent checks in with `upgrade_details: null` in the checkin body.

`processUpgradeDetails` calls `markUpgradeComplete` when the checkin body has `upgrade_details: null`. However, `markUpgradeComplete` had an early-return guard:

```go
if agent.UpgradeDetails == nil {
    return nil
}
```

Since `agent.UpgradeDetails` is also `nil` (no intermediate states were ever stored), this was a no-op. `upgrade_started_at` was never cleared, leaving the agent permanently in the "updating" state in Fleet UI — even though it had successfully upgraded to the target version.

This was observed in a 30,000-agent scale test (build [#6981](https://buildkite.com/elastic/observability-perf/builds/6981)) where one agent (`eh-ThoseChillwave-LHTb`) remained stuck in "updating" state, blocking Step10 from converging to `online == 30,000`.

The NOP guard is extended to also require `upgrade_started_at` to be empty:

```go
if agent.UpgradeDetails == nil && agent.UpgradeStartedAt == "" {
    return nil
}
```

If `upgrade_started_at` is set but `upgrade_details` is nil, an upgrade was dispatched (by Kibana) but no intermediate states were ever recorded by Fleet Server. When the agent then checks in with `upgrade_details: null`, Fleet Server now correctly clears `upgrade_started_at`, sets `upgraded_at`, and marks the upgrade complete — regardless of whether any intermediate `upgrade_details` states were observed.

## How to test locally

1. Start a Fleet Server with an enrolled agent.
2. Trigger an upgrade via Kibana's `POST /api/fleet/agents/bulk_upgrade`. This sets `upgrade_started_at` on the agent document.
3. Before the agent sends any checkin with `upgrade_details`, manually set `upgrade_details: null` and the new version on the agent document (simulating a fast upgrade).
4. Send a checkin with `upgrade_details: null` (the default for a running, non-upgrading agent).
5. Verify that Fleet Server clears `upgrade_started_at` and sets `upgraded_at` on the agent document.

Alternatively, run the unit tests:
```
go test ./internal/pkg/api/ -run TestProcessUpgradeDetails -v
```

The new test case `"agent has upgrade_started_at but no upgrade_details, checkin details are nil (fast upgrade race)"` covers this scenario.

## Design Checklist

- [ ] The code is stateless (no local caching that could cause inconsistency between instances)
- [ ] The code has been tested with more than 100K agents
- [ ] There are fail safes in the code (e.g. timeouts, max values, graceful degradation)

## Checklist

- [x] I have made corresponding changes to the tests
- [x] I have added an entry in `./changelog/fragments` using the changelog tool

### DIFF
--- a/changelog/fragments/1788912764-fix-upgrade-started-at-stuck-fast-upgrade.yaml
+++ b/changelog/fragments/1788912764-fix-upgrade-started-at-stuck-fast-upgrade.yaml
@@ -1,0 +1,25 @@
+kind: bug-fix
+
+summary: Fix agent permanently stuck in "updating" state after fast upgrade
+
+# When an agent upgrades faster than one checkin interval, Fleet Server never
+# observes any upgrade_details states. markUpgradeComplete was a NOP when
+# upgrade_details was nil on both the checkin body and the agent doc, leaving
+# upgrade_started_at set indefinitely and the agent permanently "updating".
+description: |
+  When an Elastic Agent completes an upgrade faster than one checkin interval
+  (~60 s), Fleet Server never observes any intermediate upgrade_details states
+  (downloading, extracting, watching). In this case the agent document has
+  upgrade_details: null while upgrade_started_at remains set from the original
+  bulk upgrade request. The markUpgradeComplete function previously returned
+  early (NOP) whenever upgrade_details was nil on the agent document, so
+  upgrade_started_at was never cleared. This left the agent permanently in the
+  "updating" state in Fleet UI, even though it had successfully upgraded.
+
+  The fix extends the NOP guard to also check upgrade_started_at: if an upgrade
+  was dispatched (upgrade_started_at is set) but no upgrade_details were ever
+  recorded (fast upgrade race), Fleet Server now correctly clears
+  upgrade_started_at on the next checkin with nil upgrade_details and marks the
+  upgrade complete.
+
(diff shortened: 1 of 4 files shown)


blakerouse (approved):
Looks good.

cmacknz:
> upgrade_details is null on the agent document;
> The agent checks in with upgrade_details: null in the checkin body.

The agent is supposed to check in with upgrade details set to upgrade complete isn't it? The main case where you'd have `upgrade_details: null` would be an agent upgrading before upgrade details was supported, in which case the upgrade action ack is the signal the upgrade is completed.

PR #7827, opened 2026-09-15
feat: handle shard_restoring_exception from _fleet/global_checkpoints

ycombinator:
## Summary

- Elasticsearch is changing `_fleet/global_checkpoints` to return HTTP 409 (`shard_restoring_exception`) instead of HTTP 503 when a request targets a shard being restored from a snapshot (see [elastic/dev#3625](https://github.com/elastic/dev/issues/3625), planned for ES 9.6.0).
- Today the go-elasticsearch client silently retries 503s (up to 5×), so the monitor never sees them. With 409 those implicit retries disappear, making polling more frequent and logging the error as an unexpected failure.
- This PR adds `shard_restoring_exception` as a recognised error type (`ErrShardRestoring` sentinel) and handles it explicitly in the index monitor so it logs a clear warning and retries at the normal cadence rather than treating it as an unexpected error.

## Details

**`internal/pkg/es/error.go`**
- New `shardRestoringErrorType` constant and `ErrShardRestoring` sentinel error.
- `ErrElastic.Unwrap()` now returns `ErrShardRestoring` when `Type == "shard_restoring_exception"`, enabling `errors.Is` checks up the call stack.
- Added to `knownErrorTypes` so string-based error matching also recognises it.

**`internal/pkg/monitor/monitor.go`**
- Handles `ErrShardRestoring` explicitly in the `WaitAdvance` error path: logs at `WARN` level with a clear message that the shard is being restored, then retries after `retryDelay` — same as the `ErrIndexNotFound` path.

## Backwards compatibility

This change is purely additive. `ErrShardRestoring` is never returned by the current Elasticsearch stack (which still returns 503), so behaviour is unchanged until the ES 9.6.0 change ships. It is safe to merge and release on any 9.x line ahead of the ES change.

## Testing

- Unit tests added for `shard_restoring_exception` in both string and structured (`ErrorT`) form, including an `errors.Is(err, ErrShardRestoring)` assertion.
- `mage check:fix` and `mage check:imports` clean.


blakerouse (dismissed):
Mostly mechanical, looks good.

blakerouse (approved):
Still very mechanical, looks good.

PR #7844, opened 2026-09-17
pull-requests.json: update the skip_ci if no product or BK specific

v1v:
## What is the problem this PR solves?

Run faster buidls if no product changes

## Design Checklist

- [ ] I have ensured my design is stateless and will work when multiple fleet-server instances are behind a load balancer.
- [ ] I have or intend to scale test my changes, ensuring it will work reliably with 100K+ agents connected.
- [ ] I have included fail safe mechanisms to limit the load on fleet-server: rate limiting, circuit breakers, caching, load shedding, etc.

## Checklist

- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation
- [ ] I have made corresponding change to the default configuration files
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] I have added an entry in `./changelog/fragments` using the [changelog tool](https://github.com/elastic/fleet-server#changelog)

PR #7849, opened 2026-09-18
fix: use https_endpoint for fleet_url Terraform output

ycombinator:
## Summary

- The `elastic/ec` Terraform provider no longer populates `integrations_server.endpoints.fleet` for `9.6.0-SNAPSHOT` deployments — that nested block is `null`.
- This caused `fleet_url` to output `""`, which propagated as `FLEET_SERVER_URL is empty` in the cloud E2E test step, aborting all cloud E2E runs.
- Fix: switch to `integrations_server.https_endpoint`, which is always populated and is consistent with how `elasticsearch_url` and `kibana_url` are already defined.

Affects PRs #7827 and #7844 (and any other PR currently running cloud E2E tests).


samuelvl (approved):

ycombinator:
Closing this — the fix is incorrect. `integrations_server.https_endpoint` is the APM endpoint, not the fleet-specific URL (fleet uses `*.fleet.*` hostnames). The `endpoints.fleet` attribute is genuinely null for 9.6.0-SNAPSHOT deployments in the ec provider v0.12.2,. The problem is underlying ESS provisioning failures due to Kibana crash on startup in the current 9.6.0-SNAPSHOT used by Fleet Server for E2E tests.

PR #7854, opened 2026-09-20
updatecli: refactor (use oblt-actions, update scm github)

v1v:

use the centralised github actions we own
update scm spec to avoid redundant configuration and verify commits

REduce the maintenance of the github actions by delegating that to the oblt-actions

## Design Checklist

## Checklist


fr4nc1sc0-r4m0n (approved):
LGTM

PR #7860, opened 2026-09-21
Fix flaky TestOpAMPWithUpstreamCollector

swiatekm:
`TestOpAMPWithUpstreamCollector` burned its entire 5 minute context on every run, pass or fail - CI consistently reports 300.3xs for it. Root cause: the test did `defer cancel()` for the test context and, further down, `defer otelCmd.Wait()`. Defers run LIFO, so `Wait()` ran before `cancel()`. The collector is started with `exec.CommandContext` and only receives SIGTERM once that context is done, so `Wait()` blocked until the 5 minute deadline expired on its own.

Stop the collector explicitly as soon as the assertions are done: `Wait()` now has a single call site in a goroutine, and a deferred stop sends SIGTERM, waits up to 30s for the process to be reaped, and falls back to Kill. Being a defer, it also runs when an assertion fails, so the failure path no longer costs five minutes either.

The shallow clone of opentelemetry-collector-contrib and `make otelcontribcol` shared that same 5 minute budget and were killed on slow CI agents (#6590). They now get their own budget of 15 minutes, comfortably above the build time on a healthy agent. The OpAMP assertions keep a tight 2 minute budget, matching the sibling `TestOpAMPWithEDOTCollector`; enrollment is a matter of seconds. The outer context spans both phases so fleet-server and the collector are still torn down on every exit path. This mitigates #6590 until we decide how to handle it better.


testing/e2e/stand_alone_test.go
```diff
@@ -650,7 +650,20 @@ func (suite *StandAloneSuite) writeOpAMPCollectorConfig(configFilePath, instance
...
+		// of killing the build mid-compile (see elastic/fleet-server#6590).
+		buildTimeout = 15 * time.Minute
+		// opampTimeout covers the OpAMP assertions once the collector is built. Enrollment
+		// is a matter of seconds; the sibling EDOT test uses the same budget.
+		opampTimeout = 2 * time.Minute
+	)
+
+	// ctx bounds the whole test so the fleet-server and collector processes started with
+	// it are torn down no matter which phase fails.
+	ctx, cancel := context.WithTimeout(suite.T().Context(), buildTimeout+opampTimeout)
```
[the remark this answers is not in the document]
swiatekm: I don't think this is necessary, the build timeout is very generous to start with and should easily absorb this. Plus, we'll probably fix this in a different way in the near future.

testing/e2e/stand_alone_test.go
```diff
@@ -715,12 +733,32 @@ func (suite *StandAloneSuite) TestOpAMPWithUpstreamCollector() {
...
 
-	defer otelCmd.Wait()
+	// The standard library forbids calling cmd.Wait() more than once (it races and
+	// returns "no child processes" on the second call). The goroutine below is the
+	// single Wait() call site; it forwards the exit status over a buffered channel so
+	// the deferred stop below can observe process exit without calling Wait() itself.
+	otelExited := make(chan error, 1)
+	go func() { otelExited <- otelCmd.Wait() }()
+
+	defer func() {
```
ycombinator, line 743: Now that cleanup is owned by this defer function, I'm not sure `otelCmd` should also take the `ctx`?  Otherwise two `SIGTERM`s could be sent, one from `ctx` cancellation and one from the `defer`.

swiatekm: It might, but then we'll just log that it happened and cleanup will otherwise proceed normally.

ycombinator (approved):

Issue #7832, opened 2026-09-16
Flaky TestAgentGracefulForceUnenroll


macdewee:
## Flaky Test

* **Test Name:** TestAgentGracefulForceUnenroll
* **Link:** https://github.com/elastic/fleet-server/blob/98c1938ef8071984dd5c91a602e76e6e80715351/testing/e2e/stand_alone_test.go#L750
* **Branch:** [main](https://github.com/elastic/fleet-server/tree/98c1938ef8071984dd5c91a602e76e6e80715351)
* **Artifact Link:** https://buildkite.com/elastic/fleet-server/builds/16929/list?sid=01a0a943-d1d1-40d8-8389-d5a9f9bcaa81&tab=artifacts
* **Notes:** In progress

### Stack Trace

```
<testcase name="TestStandAloneRunningSuite/TestAgentGracefulForceUnenroll" classname="github.com/elastic/fleet-server/testing/e2e" time="221.050">
<failure message="Failed">
<![CDATA[     stand_alone_test.go:753: Using version 9.6.0-SNAPSHOT for agent download
    stand_alone_test.go:753: Using cached elastic-agent from /tmp/fleet-server-e2e/elastic-agent-9.6.0-SNAPSHOT-linux-x86_64.tar.gz
    stand_alone_test.go:820: enroll output:
        {"log.level":"info","@timestamp":"2026-09-16T09:20:50.500Z","log.origin":{"function":"github.com/elastic/elastic-agent/internal/pkg/agent/application/enroll.EnrollWithBackoff","file.name":"enroll/enroll.go","file.line":87},"message":"Starting enrollment to URL: https://localhost:8220/","ecs.version":"1.6.0"}
        {"log.level":"info","@timestamp":"2026-09-16T09:20:51.570Z","log.origin":{"function":"github.com/elastic/elastic-agent/internal/pkg/agent/cmd.(*enrollCmd).Execute","file.name":"cmd/enroll_cmd.go","file.line":210},"message":"Elastic Agent has been enrolled; start Elastic Agent","ecs.version":"1.6.0"}
        Successfully enrolled the Elastic Agent.
    stand_alone_test.go:860: Waiting for enrolled agent to appear online in Kibana...
    stand_alone_test.go:876: agent 886df709-ed73-4780-85eb-1eaa4119638a status=updating
    stand_alone_test.go:885: Agent 886df709-ed73-4780-85eb-1eaa4119638a is active and checking in
[...]

PR #7866, opened 2026-09-21
fix(apikey): log warning when Invalidate finds key not present in ES

ycombinator:

`apikey.Invalidate` (and by extension `Bulker.APIKeyInvalidate`) never reads the Elasticsearch response body. `DELETE /_security/api_key` always returns HTTP 200 — even when zero keys were actually invalidated — so silent failures where the key ID was not found in ES were completely invisible. This affected two callers:

- **`handleEnroll.go`**: invalidating the old API key when an agent re-enrolls
- **`handleAck.go`**: retiring old output API keys when a policy is acked

If the key was not found (already expired, already deleted, or wrong ID), the call succeeded silently and the stale key remained active.

The response body is now decoded. Any requested key ID absent from both `invalidated_api_keys` and `previously_invalidated_api_keys` is logged at warn level. A non-zero `error_count` from ES is surfaced as an error. Keys in `previously_invalidated_api_keys` are treated as already-invalidated (not a warning).

Unit tests: `go test ./internal/pkg/apikey/...`

Integration test with a real ES: invalidate a non-existent key ID and observe the warning log.

## Design Checklist

## Checklist

## Related issues

- Relates #7832


internal/pkg/apikey/invalidate.go
```diff
@@ -49,5 +54,32 @@ func Invalidate(ctx context.Context, client *elasticsearch.Client, ids ...string
...
+		actioned[id] = struct{}{}
+	}
+	for _, id := range ids {
+		if _, ok := actioned[id]; !ok {
+			zerolog.Ctx(ctx).Warn().Str("api_key.id", id).Msg("API key not found during invalidation; it may have already expired")
+		}
+	}
+
+	if result.ErrorCount > 0 {
+		return fmt.Errorf("InvalidateAPIKey: ES reported %d error(s) invalidating keys", result.ErrorCount)
```
swiatekm, line 81: We're changing behavior here, though I'm not sure to what extent this is user-visible. But we should have a changelog entry for this PR.

PR #7874, opened 2026-09-22
fix(terraform): set config to null when no docker image is provided

ycombinator:

When a docker image variable is empty (the common case when no custom image is needed), the `elasticsearch`, `kibana`, and `integrations_server` resource `config` blocks evaluated to an object with a `null` `docker_image` field. The `elastic/ec` Terraform provider returns `null` for the entire config block in that case, causing a **"Provider produced inconsistent result after apply"** error on every `terraform apply` run where no custom docker images are provided.

Make the entire `config` block `null` when no docker image variable is set, by changing:
```hcl
config = {
  docker_image = var.X != "" ? local.docker_image_X : null
}
```
to:
```hcl
config = var.X != "" ? {
  docker_image = local.docker_image_X
} : null
```

This matches what the provider actually returns, eliminating the inconsistency. All three affected resources (`elasticsearch`, `kibana`, `integrations_server`) are fixed.

Run `terraform apply` in `dev-tools/cloud/terraform/` without setting any docker image variables (i.e. leaving `elastic_agent_docker_image`, `kibana_docker_image`, and `elasticsearch_docker_image` empty). The apply should complete without the "provider produced inconsistent result" error.

## Design Checklist

## Checklist

## Related issues


changelog/fragments/1790080744-fix(terraform)-set-config-to-null-when-no-docker-image-is-provided.yaml
```diff
@@ -0,0 +1,32 @@
+# Kind can be one of:
```
swiatekm, line 1: Why do we need a changelog entry? We're just fixing our e2e test setup.

swiatekm (approved):

